PR #351, opened 2014-10-28
Add tests for managedsave command.

rbian:
Test bypass_cache option
Test bypass_cache and start_delay option in libvirt-guests configuration file.


rbian:
@yumingfei , all failures should be time dependent, we need to consider more factors for this test. I can't reproduce your problem, but I will try to fix it.

rbian:
@yumingfei , could you please help check if this version works? Thanks :)

yumingfei:
Good enough now.
Thanks @rbian, @will-Do , applied.

PR #379, opened 2014-11-24
Improve code implementation.

rbian:
wwn may contain 0x prefix in qemu-kvm command line.
set serail console to kernel boot parameters for serial tests.
update disk format probe mothed.


rbian:
Thanks @will-Do , has updated the code.

will-Do:
Thanks, cases running smoothly, merged.

PR #394, opened 2014-12-17
blockpull: Add various disk types for testing

waynesun09:
Add network disk with gluster and iscsi type, block disk use iscsi
backend, netfs disk mounted on local for testing.

Also added test case for option --async and --keep-relative.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockpull.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockpull.cfg
@@ -4,8 +4,13 @@
     variants:
         - normal_test:
             status_error = "no"
+            needs_agent = "yes"
             variants:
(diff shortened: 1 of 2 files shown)


yumingfei:
LGTM.
But why not run complete blockpull cases in a single round?

waynesun09:
@yumingfei I also like to collect result in one run, but my machine could fail specific type cases some time and succeed other time (mostly iscsi and gluster related), so I need to split runs and solve all problems, the run result could also be collected from different machines and os.

yumingfei:
OK, I just want to confirm that failed testcases will not effect others.

waynesun09:
@yumingfei the most disturbing problem is gluster bug 1143800 and it did affect the env, which I've solved with commit 789a623. So, it's all good. :)

PR #402, opened 2014-12-24
Remove checking present cpu num with /sys/devices/system/cpu/present

kylazhang:
As it is not reliable


yumingfei:
Can you find some documents about the cpu parameters(present, possible, online, offline)?
As I know, possible cpus mean the count of internal cpus and external cpus,
while present cpus mean internal cpus.

Since we need to get present cpus here, maybe we should get the root reason
why present count is not accurate.

kylazhang:
@yumingfei https://www.kernel.org/doc/Documentation/cputopology.txt I have no idea why it is wrong and maybe it is kernel bug. If you want more info, I think you can find kernel patches and analyse

will-Do:
Okay, merge this PR

PR #405, opened 2014-12-29
virsh.vol_create: Update and extend cases

will-Do:
1. Test create 128 partition volume on disk pool(gpt disk)
2. Create thin lvm pool/volume in libvirt logical pool
3. Create snapshopt volume in libvirt logical pool
4. Deactivate volume in libvirt logical pool
5. Use invalid target path('/dev/') to create pool/volume, which should pass
6. Use 'qemu-img' to cretae vmdk v4/v5 images in libvirt dir pool


yumingfei:
It's OK, thanks.

PR #407, opened 2015-01-06
virsh.iface: Free mac address if domain interface changed

will-Do:
In the case, domain interface changed by detach and attach interface
operation, then vm.wait_for_login() always failed as vm.virtnet not
update with the new interface, free the original mac addresses can
fix this issue.


PandaWei:
you spelt one word of commit title wrong, "chnaged" should be "changed"

yumingfei:
Thanks, @will-Do

PR #410, opened 2015-01-07
Add test for different init scripts of libvirt

Hao-Liu:
libvirt using different init scripts for different init systems.
This test mainly check the availability of libvirt init scripts
and start, restart, reload and stop libvirt daemon with all
available init systems.


yumingfei:
OK, it runs well.
Thanks~

PR #411, opened 2015-01-07
Add test for killing qemu process

Hao-Liu:
Kill started qemu VM process with different signals and check
the status of the VM changes accordingly.

### DIFF
--- a/libvirt/tests/src/daemon/kill_qemu.py
+++ b/libvirt/tests/src/daemon/kill_qemu.py
@@ -1,0 +1,38 @@
+import os
+import signal
+from autotest.client import utils
+from autotest.client.shared import error
+from virttest.libvirt_xml import vm_xml
+
+
+def run(test, params, env):
+    """
+    Kill started qemu VM process with different signals and check
+    the status of the VM changes accordingly.
+    """
+    vm_name = params.get('main_vm')
+    sig_name = params.get('signal', 'SIGSTOP')
+    vm_state = params.get('vm_state', 'running')
+    expect_stop = params.get('expect_stop', 'yes') == 'yes'
+    vm = env.get_vm(vm_name)
+
+    xml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    try:
+        if vm_state == 'running':
+            pass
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/daemon/kill_qemu.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    vm_name = params.get('main_vm')
+    sig_name = params.get('signal', 'SIGSTOP')
+    expect_stop = params.get('expect_stop', 'yes') == 'yes'
+    vm = env.get_vm(vm_name)
+
+    pid = vm.get_pid()
+    os.kill(pid, getattr(signal, sig_name))
+
+    stopped = bool(utils.wait_for(lambda: vm.state() == 'shut off', 10))
```
yumingfei: Do we need to deal with abnormal states?

Hao-Liu: Do you mean "abnormal states" is for the state of VM before or after sending the signal?
If for after, I think the state "shut off" should be an only expected result.
If for before, I totally agree with changing the VM's state before sending the signal to cover a larger test dimension.

yumingfei: I mean after, maybe I'm to sensitive :P, what if vm.state() is in other state such as running, paused?

Hao-Liu: No, It will timeout after 10 seconds and raise a TestFail for any states except "shut off".

yumingfei: Hmm, it seems that I can't find TestFail in wait_for(...).
BTW, why not use utils_misc.wait_for? They are same.

Ha, I got it.
You mean the result checking but I mean the vm recovering...

Hao-Liu:
@yumingfei, I added two states (paused, pmsuspended) before kill qemu.
Add for the environment recovering, I tested leaving vm in different states (running, pause, pmsuspend) without kill the process, our test framework will always destroy the vm. Could you please help check this again. Thanks.

yumingfei:
Not bad, applied.
Thanks~

PR #412, opened 2015-01-07
Add test for killing daemon when starting or started

Hao-Liu:
This PR depends on https://github.com/autotest/virt-test/pull/1947.

These test focus on testing whether libvirtd works properly
when recieved different kill signals. It includes two parts:

kill_started:
    Kill libvirt daemon with different signals and check
    whether daemon restart properly and leaving no pid file
    if stopped.

kill_starting:
    Start libvirt daemon with break point inserted.
    And then kill daemon ensure no sigsegv happends.


yumingfei:
Well, it's good.
Thanks.

PR #414, opened 2015-01-07
conf_file..unix_sock: Change test command to net-define

Hao-Liu:
To test whether readonly socket is working, this commit change the virsh
test command from `list` to `net-define`.

### DIFF
--- a/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
+++ b/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
@@ -1,9 +1,9 @@
 import grp
-import time
 import stat
 import logging
 import os.path
 from virttest import virsh
+from virttest import data_dir
 from virttest import utils_config
 from virttest import utils_libvirtd
 from autotest.client.shared import error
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
```diff
@@ -75,11 +74,15 @@ def check_unix_sock(group, perms, path):
 
         # Check virsh connection.
         uri = 'qemu+unix:///system?socket=%s' % path
-        result = virsh.dom_list('--all', uri=uri)
+        result = virsh.net_autostart('default', uri=uri)
```
will-Do: Then we need make sure 'default' does exist on the host

yumingfei: Yes, you have to deal with it, even it runs well currently.
And, why should you change it to net-autostart?

Hao-Liu: Well, I just want to change the virsh command to another one which need authorization. Thus I can test the difference between ro and rw sockets. But it seems very difficult to find one that is both easy the implement and reliable. Do you have any suggestions?

yumingfei: Maybe you can try to define your own network in this module.
It's easy to be implemented by providing a fixed network xml.

Hao-Liu: Sure, I will change to net-define, thanks for the advice.

Hao-Liu:
Changed test command from net-autostart and net-define to prevent potential issue. Thanks @will-Do, @yumingfei.

yumingfei:
Good, thanks. Applied.

PR #416, opened 2015-01-07
Fix the sysinfo problem when dmidecode can not give the version

kylazhang:
Remove the status check for dmidecode commands output as with pipe it can not show the right commands status, and use output check instead

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_sysinfo.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_sysinfo.py
@@ -15,9 +15,8 @@ def get_processor_version():
     processor_version = []
     cmd = "dmidecode -t processor | awk -F: '/Version/ {print $2}'"
     cmd_result = utils.run(cmd, ignore_status=True)
-    status = cmd_result.exit_status
     output = cmd_result.stdout.strip()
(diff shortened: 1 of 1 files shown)


yumingfei:
Reasonable, but do we need to check the output format first?

kylazhang:
@yumingfei In my opinion only 2 types of result we can get, the version string or NULL. I think we do not need to check more.

yumingfei:
Oh, I meant the output data:
[root@domain virt-test]# dmidecode -t processor | awk -F: '/Version/ {print $2}'
 Intel(R) Core(TM)2 Duo CPU     E7500  @ 2.93GHz     

Since you simply compare command output and xml output, I'm afraid that little difference may cause unnecessary failure, such as one more space for command output.
So I think that splitting the elements(Intel, 2 Duo CPU, E7500...) in output would be more reasonable.

kylazhang:
@yumingfei This patch is only for correcting the wrong judge situation which cause my test fail, I think the problem you think about should be another issues. Thanks

kylazhang:
@waynesun09 how do you think about the issue that yumingfei said?

liyangfnst:
good enough, merge it now.

PR #432, opened 2015-01-17
make usb-hotplug cases remain consistent style with other files

PandaWei:
just rename the file about usb hotplug at the directory libvirt_bench,
and modify the first line of cfg file.


yumingfei:
OK.

PR #433, opened 2015-01-19
virsh_iface_bridge: bugfix and improvement for iface-bridge test

PandaWei:
1) bugfix for recovering iface script
   the iface configuration script maybe changed and it will be backed up
   as iface_script_bk. finally, all procedure of test ends up,
   the existence of 'iface_script_bk' should be checked, not iface_script.

2) imporve the procedure of reloading script
   'ifdown' should be executed before 'ifup'.


yumingfei:
Yes, it should be.

PR #437, opened 2015-01-21
add "utils" commands

zcyes:
as described in comments commits.

### DIFF
--- a/libguestfs/tests/cfg/guestfish_utils.cfg
+++ b/libguestfs/tests/cfg/guestfish_utils.cfg
@@ -81,59 +81,59 @@
                 - delete_event:
                     guestfish_function = event
                 - set_append:
-                    guestfish_function = set_append
+                    guestfish_function = set_get_append
                 - get_append:
-                    guestfish_function = get_append
+                    guestfish_function = set_get_append
                 - set_smp:
-                    guestfish_function = set_smp
+                    guestfish_function = set_get_smp
                 - get_smp:
-                    guestfish_function = get_smp
+                    guestfish_function = set_get_smp
                 - set_pgroup:
-                    guestfish_function = set_pgroup
+                    guestfish_function = set_get_pgroup
                 - get_pgroup:
-                    guestfish_function = get_pgroup
+                    guestfish_function = set_get_pgroup
                 - set_attach_method:
-                    guestfish_function = set_attach_method
+                    guestfish_function = set_get_attach_method
                 - get_attach_method:
-                    guestfish_function = get_attach_method
+                    guestfish_function = set_get_attach_method
                 - set_autosync:
-                    guestfish_function = set_autosync
+                    guestfish_function = set_get_autosync
                 - get_autosync:
-                    guestfish_function = get_autosync
+                    guestfish_function = set_get_autosync
                 - set_direct:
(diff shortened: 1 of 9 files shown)


zcyes:
add test case: sleep, quit, man, launch

libguestfs/tests/guestfish_utils.py
```diff
@@ -0,0 +1,1570 @@
...
+          <target dev='hda' bus='ide'/>\n\
+        </disk>\n\
+      </devices>\n\
+    </domain>\n\
+    " % (test_domain_name, image_path)
+    f = open(test_xml, "w")
+    f.write(xml_content)
+    f.close()
+
+    os.system("virsh define %s > /dev/null" % test_xml)
```
shi2wei3: We'd better do not call virsh command directly, we can try to reuse the libvirt_vm, you can reference to guestfs_add.py

libguestfs/tests/guestfish_utils.py
```diff
@@ -0,0 +1,1570 @@
...
+    Test command add_drive:
+    """
+    add_ref = params.get("gf_add_ref", "disk")
+    readonly = params.get("gf_add_readonly", "no")
+
+    gf = utils_test.libguestfs.GuestfishTools(params)
+
+    image_path = params.get("image_path")
+    pv_name = params.get("pv_name")
+    alloc_file = '~/../../home/test_add_drive.img'
```
shi2wei3: It is a bad idea to use hardcode path especially with '~', small temp images could save to /tmp or params.get("image_path")

libguestfs/tests/cfg/guestfish_utils.cfg
```diff
@@ -0,0 +1,158 @@
...
+    lv_name = vol_file
+    image_formats = "raw"
+    image_name = utils
+    tarball_file = utils.tgz
+    image_size = "1G"
+
+    variants:
+        - utils_cmds:
+            variants:
+                - add_domain:
```
yumingfei: Shouldn't these add-operations be put into guestfs_add or some other module?

yumingfei: BTW, what's the principle to collect these testcases?

libguestfs/tests/guestfish_block_dev.py
```diff
@@ -34,12 +34,9 @@ def test_blockdev_flushbufs(vm, params):
     readonly = params.get("gf_add_readonly", "no")
 
     gf = utils_test.libguestfs.GuestfishTools(params)
-    if add_ref == "disk":
-        image_path = params.get("image_path")
-        gf.add_drive_opts(image_path, readonly=readonly)
-    elif add_ref == "domain":
-        vm_name = params.get("main_vm")
-        gf.add_domain(vm_name, readonly=readonly)
+
```
yumingfei: Do you mean these test modules should only test with images but domains?

yumingfei:
Another problem about these patches is that you'd better rename your patches' subject to
make it more clear and useful.
We need to check them and know what we have done in future.

Generally ,we have following format:
module: the changes your have done for this module.
(less than 70 character as possible) :)

waynesun09:
Close as no update for long time

PR #441, opened 2015-01-23
virsh.pool_edit: Fixup pool cleanup usage

will-Do:
The cleanup_pool() method has been update to use kwargs for these
optional paramters, so update the caller here.
BTW, remove an useless variable here.


will-Do:
Tiny fix, merged.

PR #442, opened 2015-01-23
Add tests for attach-interface with noexisted script.

rbian:


yumingfei:
It's good. And we can add some positive cases for script in future~

PR #444, opened 2015-01-26
Fix var readonly assignment to boolean type instead of string

shi2wei3:
var readonly is a boolean type in virttest/utils_libguestfs.py under virt-test

if readonly:
    cmd += " readonly:true"
else:
    cmd += " readonly:false"

Change the assignment statement for this variable for all libguestfs cases


shi2wei3:
Reference to "Bug fix for add-drive-opts readonly #1976"

PR #446, opened 2015-01-27
Modify case for the change of apis' location

zcyes:
Apis are moved to another place, GuestfishPersistent contains no api any more

### DIFF
--- a/libguestfs/tests/guestfs_add.py
+++ b/libguestfs/tests/guestfs_add.py
@@ -2,8 +2,7 @@ import logging
 import re
 import commands
 from autotest.client.shared import error, utils
-from virttest import utils_libguestfs as lgf
-from virttest import aexpect
+from virttest import aexpect, utils_test
 
 
 def primary_disk_virtio(vm):
@@ -121,7 +120,7 @@ def run(test, params, env):
         disk_or_domain = add_ref
         add_ref = "disk"
 
-    guestfs = lgf.GuestfishPersistent()
+    guestfs = utils_test.libguestfs.GuestfishTools(params)
     set_guestfs_args(guestfs)
 
     add_error = params.get("guestfs_add_error", "no")


zcyes:
@yumingfei 
Hi, I modify the file "guestfs_add.py" for the change of apis' location
Cause, as we discussed in the virt-test project, I move the libguestfs apis from GuestfishPersistent(in utils_libguestfs.py) to LibguestfsAPIs(in libguestfs_api.py), in consideration of the scale of utils_libguestfs.py.
What you do thinks? if there's any question, please let me know. Thank you. :)

shi2wei3:
acked, pending on "Move libguestfs apis to libguestfs_api.py #1983"

yumingfei:
We can add a link to:
https://github.com/autotest/virt-test/pull/1983/files

libguestfs/tests/guestfs_add.py
```diff
@@ -121,7 +120,7 @@ def run(test, params, env):
         disk_or_domain = add_ref
         add_ref = "disk"
 
-    guestfs = lgf.GuestfishPersistent()
+    guestfs = utils_test.libguestfs.GuestfishTools(params)
```
yumingfei: Do you have other purpose by using GuestfishTools and why not use GuestfishAPIs?

shi2wei3:
Actually, i am trying to understanding your design methodology for GuestfishTools and GuestfishPersistent, and i think virttest/utils_test/libguestfs.py would be a best entrance for all libguestfs callings, GuestfishTools handle APIs calling, VirtTools handle virt-tools calling.
So structure adjustment won't just involve virt-test updates, tp-libvirt won't have and changes. Because we only expose virttest/utils_test/libguestfs.py to the outside.

If we both using GuestfishPersistent, GuestfishTools and GuestfishAPIs, there would be a mass and not easy to maintain, and i saw most of your script using GuestfishTools, so i choose the same class.

Do i understand this correctly?

yumingfei:
Yes, utils_test.libguestfs should be an entrance for libguestfs callings.
But My thought is that it's open for test module like libguestfs/tests/guestfs_add,
while for those modules in virttest, we should use utils_libguestfs/xxx.

The original idea I worked GuestfishTools out is because I wrote some complex functions
by using apis of guestfish.But your description about setting it as the entrance is right.
You could and should use it in guestfs_add, so please ignore my comment above. :)
I had that question above because I felt that we have to pass a dict to GuestfishTools may
cause some confusion to tester.

While for GuestfishPersistent, we should only use it in virttest modules, so your choice is right.

waynesun09:
close as no more update

PR #447, opened 2015-01-28
virtual_disks_multidisks: Modify code of getting guest console output.

rbian:
Only get stripped output of bios,
get rid of kernel serial output can't be decoded.


yumingfei:
Acked.

yumingfei:
Thanks, @rbian

PR #452, opened 2015-02-04
libvirt.test: Raise TestNAError for all qed type disk for snapshot.

yumingfei:
As bugzilla reported, all qed type disk should be skipped:
qemu-kvm component, BZ#731570
The QEMU Enhanced Disk format (QED) for KVM guest virtual machines
is not supported in Red Hat Enterprise Linux 7. Use the qcow2 image
format instead.

Signed-off-by: Yu Mingfei <yumingfei@cn.fujitsu.com>


waynesun09:
ACK
Current no_pool qed test will skip at attach-disk, this fix is better. 
Nice catch, Thanks.

yumingfei:
Thanks @waynesun09

PR #453, opened 2015-02-04
hotplug_serial: improve the way of getting temp directory

PandaWei:
* use test.tmpdir to get the path
  * remove the entire temporary directory at last

### DIFF
--- a/libvirt/tests/src/hotplug_serial.py
+++ b/libvirt/tests/src/hotplug_serial.py
@@ -4,6 +4,7 @@ import stat
 import subprocess
 import time
 import socket
+import shutil
 
 from autotest.client import utils
 from autotest.client.shared import error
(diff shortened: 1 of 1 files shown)


PandaWei:
add a new patch that make hotplug_serial using common function to create channel xmlfile
    
The following have been changed,
        * call virt-test/utils_test.libvirt.create_channel_xml
        * remove the unused configuration item 'xml_file'

libvirt/tests/src/hotplug_serial.py
```diff
@@ -153,7 +142,7 @@ def dup_hotplug(type, char_dev, id, dup_charid=False, dup_devid=False, diff_devi
         return result
 
     def confirm_hotplug_result(char_dev, id=0):
-        tmp_file = "/tmp/%s" % char_dev
+        tmp_file = "%s/%s" % (tmp_dir, char_dev)
```
yumingfei: Using os.path.join will be more suitable.

libvirt/tests/src/hotplug_serial.py
```diff
@@ -49,50 +55,36 @@ def run(test, params, env):
...
-        xmlf = open(channelxml.xml)
-        try:
-            xml_lines = xmlf.read()
-        finally:
-            xmlf.close()
-        return xml_lines
+        params = {'channel_type_name': channel_type,
+                  'source_path': channel_path,
+                  'target_type': 'virtio',
+                  'target_name': "%s" % char_type}
```
will-Do: Why not use the variable directly?

will-Do:
Good patches.

libvirt/tests/src/hotplug_serial.py
PandaWei: @will-Do 
the funciton create_channel_xml() has been added at virttest/utils_test/libvirt.py.
In order to make it be consistent with the old function, such as create_nwfilter_xml.
just adopt dict params[] as input parameter, which
can reduce the number of input parameters.
The style is marked by simplicity, clarity.
so I think it's a good method.

PandaWei:
update..

libvirt/tests/src/hotplug_serial.py
will-Do: No, I mean this line:
```
'target_name': "%s" % char_type}
```
Is 'char_type' not a string, so you need use this way to pass it?

yumingfei:
OK.

PR #455, opened 2015-02-05
numatune: Fix problems and update get numa parameter method

waynesun09:

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_numatune.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_numatune.cfg
@@ -13,6 +13,7 @@
                             start_vm = "no"
                             variants:
                                 - change_mode:
+                                    numa_nodeset = 0
                                     variants:
                                         - preferred:
                                             numa_mode = 'preferred'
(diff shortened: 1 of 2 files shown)


waynesun09:
This depends on
https://github.com/autotest/virt-test/pull/1999

libvirt/tests/cfg/virsh_cmd/domain/virsh_numatune.cfg
```diff
@@ -127,6 +128,7 @@
                                         - interleave:
                                             numa_mode = 'interleave'
                                         - strict:
+                                            numa_nodeset = 0
```
yumingfei: I have one question that whether parameters in test module should be changed in same format(Integer)?
Such as following code:
```
nodeset = params.get("numa_nodeset", "")
```

waynesun09: @yumingfei as I remember the parameters parsed from cfg file are string type, so the integer you see in cfg will be string value in params dict.
I just tested, it did is string value.
As whether to change to same format, it depends on the how it used in the module, nodeset in virsh numatune command is string format.

yumingfei: OK, it's acceptable.

PR #457, opened 2015-02-06
Wait for vm is started before save/pmsuspend operation.

rbian:
Fix virt_clone_cmd parameter problem.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
@@ -176,10 +176,14 @@ def run(test, params, env):
                 # Run a series of operations to check libvirtd status.
                 ret = virsh.dompmwakeup(vm_name, **virsh_dargs)
                 libvirt.check_exit_status(ret)
+                # Wait for vm is started
+                vm.wait_for_login()
                 ret = virsh.save(vm_name, savefile, **virsh_dargs)
                 libvirt.check_exit_status(ret)
                 ret = virsh.restore(savefile, **virsh_dargs)
                 libvirt.check_exit_status(ret)
+                # Wait for vm is started
+                vm.wait_for_login()
                 # run pmsuspend again
                 ret = virsh.dompmsuspend(vm_name, suspend_target, **virsh_dargs)
                 libvirt.check_exit_status(ret)


libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
```diff
@@ -176,10 +176,14 @@ def check_vm_guestagent(session):
                 # Run a series of operations to check libvirtd status.
                 ret = virsh.dompmwakeup(vm_name, **virsh_dargs)
                 libvirt.check_exit_status(ret)
+                # Wait for vm is started
+                vm.wait_for_login()
```
yumingfei: Why wait_for_login is necessary here? It seems that save and managedsave do not need this step.

rbian: Yes, save and restore don't need this, but if guestagent in vm isn't running before SAVE operation, second dompmsuspend command will fail.

yumingfei: OK, got it. Have you tested it?

rbian: Yes, it has passed my test.

libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -141,7 +141,7 @@ def check_multi_guests(guests, start_delay, libvirt_guests):
         for i in range(int(guests)):
             dst_vm = "%s_%s" % (vm_name, i)
             utils_libguestfs.virt_clone_cmd(vm_name, dst_vm,
-                                            True, timeout)
+                                            True, timeout=float(timeout))
```
yumingfei: Maybe we should convert it to integer or float in utils_libguestfs.virt_clone_cmd.
Otherwise we have to check it in test module every time it is called.

rbian: Yes, that should be better.

will-Do:
ACKed this PR

Hao-Liu:
ACK and merging, Thanks @rbian @yumingfei @will-Do.

PR #460, opened 2015-02-09
Remove --unsafe option

kylazhang:
1. we do not need it for migrate command to do migration
2. other commands do not support this option and it may cause error


yumingfei:
But it often incurs error when disk is not flagged with cache='none'.

Antique:
Yes, I agree that we can use "--unsafe" for migration part of the domjobabort test.

https://github.com/autotest/tp-libvirt/pull/464

yumingfei:
Close this one since we have adopt @Antique 's idea.

PR #464, opened 2015-02-13
virsh_domjobabort: fix the wrong usage of "--unsafe"

Antique:
Only command "virsh migrate" has an option called unsafe therefor we
can use it only for migrate action.

Signed-off-by: Pavel Hrdina <phrdina@redhat.com>


yumingfei:
Acked.

will-Do:
ACK

PR #465, opened 2015-02-15
libvirt: Add test case for domstats command

will-Do:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
@@ -1,0 +1,209 @@
+import logging
+from autotest.client.shared import error
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices.panic import Panic
+from virttest.aexpect import ShellTimeoutError
+from virttest.aexpect import ShellProcessTerminatedError
+
+
+def prepare_vm_state(vm, vm_state):
+    """
+    Prepare the given state of the domain.
+
+    :param vm: Libvirt VM instance
+    :param vm_state: Domain state to set
+    """
+    if vm_state == "active":
+        if not vm.is_alive():
+            vm.start()
+    elif vm_state == "inactive":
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
```diff
@@ -0,0 +1,210 @@
...
+                panic_dev.addr_iobase = "0x505"
+                vmxml.add_device(panic_dev)
+                vmxml.sync()
+                virsh.start(vm.name, ignore_status=False)
+                vmxml_new = vm_xml.VMXML.new_from_dumpxml(vm.name)
+                # Skip this test if no panic device find
+                if not vmxml_new.xmltreefile.find('devices').findall('panic'):
+                    raise error.TestNAError("No 'panic' device in the guest,"
+                                            " maybe your libvirt version"
+                                            " doesn't support it.")
```
rbian: Need to restore vm xml if there is any error here.

waynesun09: @will-Do as @rbian commented, maybe you could just move this part into try..finally for this.

will-Do: @rbian  @waynesun09 Thanks
In fact, if error occurred here, it will not affect the domain xml. Anyway, I can move it to ```try.. finally``` if you insistent about it

waynesun09: @will-Do add panic device could fail, but vmxml.on_crash = "preserve" could be updated, so better restore it after.

will-Do: Okay, will update it later

waynesun09:
@will-Do it'll fail on rhel6, better check libvirt version and skip it for new commands.

waynesun09:
The test is good on rhel7 and this is new command since rebase on rhel7, not supported in rhel6, so merging this now.

PR #471, opened 2015-02-22
libvirt: Fix PEP8 W503

Hao-Liu:

### DIFF
--- a/libvirt/tests/src/guest_kernel_debugging/nmi_test.py
+++ b/libvirt/tests/src/guest_kernel_debugging/nmi_test.py
@@ -46,7 +46,7 @@ def run(test, params, env):
     if not libvirt_version.version_compare(1, 1, 1):
         if params.get('setup_libvirt_polkit') == 'yes':
             raise error.TestNAError("API acl test not supported in current"
-                                    + " libvirt version.")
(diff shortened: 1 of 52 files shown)


libvirt/tests/src/guest_kernel_debugging/nmi_test.py
```diff
@@ -45,8 +45,8 @@ def run(test, params, env):
 
     if not libvirt_version.version_compare(1, 1, 1):
         if params.get('setup_libvirt_polkit') == 'yes':
-            raise error.TestNAError("API acl test not supported in current"
-                                    + " libvirt version.")
+            raise error.TestNAError("API acl test not supported in current" +
```
waynesun09: The operator '+' is not necessary, implicit line joining inside parentheses could be all right.
@Hao-Liu what do you think?

waynesun09:
Lot of the W503 problems found out here are caused by me, sigh...

waynesun09:
ACK

waynesun09:
@yumingfei @will-Do Can you guys help check this, as all PR will fail without W503 fix here.

will-Do:
ACK and merge this fix, thanks

PR #475, opened 2015-02-26
virsh_domif_setlink_getlink: Use 'ip link' to set link state in guest.

rbian:


yumingfei:
Do they have discrepancy may trouble the test?

rbian:
Ifup script also call "ip link" to set link state, but ifup use the interface name which should have a configuration file, otherwise it will report error.
For example:
if we use "ifup eth2", but there is no ifcfg-eth2 on /etc/sysconfig/network-scripts/, we can't get the right result.

yumingfei:
OK.

chuanchang:
ACKed.

will-Do:
Patch is good, I guess you can amend your patch description to add the detail reason just like you said in comment 2

rbian:
Thanks @will-Do , updated the patch.

waynesun09:
ACK

will-Do:
Thx all, merge this PR

PR #473, opened 2015-02-25
nodeinfo: Fix a typo

waynesun09:
s/topolopy/topology/g

Signed-off-by: Wayne Sun <gsun@redhat.com>

PR #478, opened 2015-02-27
typo fix

will-Do:
Commit d8c0eb04 find this typo but fix is incomplete.


will-Do:
This's complement for autotest/tp-libvirt#473

PR #481, opened 2015-03-03
blockcommit: Fix check loop and destroy problem

waynesun09:
As shutdown might fail due to acpi setting, use destroy to avoid fail.
Update destroy with gracefully as False in setup and cleanup.

Fix the loop problem when without pivot and timeout, blockcommit could
abort succeed without mirror phase.


yumingfei:
LGTM.

will-Do:
ACK

PR #426, opened 2015-01-15
Update with guest agent related changes

Hao-Liu:
There patches update the guest agent related tests to follow the new
method for preparation of guest agent. And some small fixes included.

This PR depends on https://github.com/autotest/virt-test/pull/1961.

Signed-off-by: Hao Liu <hliu@redhat.com>

PR #484, opened 2015-03-04
libvirt_vcpu_plug_unplug: fix bug for qemu-agent

PandaWei:
After suspending the process of qemu-agent by kill,
it will start itself again on RHEL7.1,
which could lead to the failure of
libvirt_vcpu_plug_unplug.negative_test.no_start_qemuga.

so, it's more better that stop qemu-guest-agent service.


PandaWei:
#426 was merged just now.
the problem has been fixed.
close this PR

PR #487, opened 2015-03-05
virsh_cmd: update cases as virt-test API refactorized

PandaWei:
The function set_agent_channel/remove_agent_channels of
class VMXML had refactorized by commit 002d51f,
which leads to so many testcases failed.
so all of caller need to update.


waynesun09:
This is fixed by #426.

PandaWei:
@waynesun09  
OK, thanks for your reminder.

PR #489, opened 2015-03-06
virsh_undefine: Fixed a bug related to test pool creation and deletion

svirt:
The unnecessary checks are removed from the '--remove-all-storage'
option processing block which allows the test flow to continue.
It doesn't look dangerous for the domain.

Fixed the 'emulated_size' parameter bug in test pool creation function
pre_pool() call. The 'emulated_size=emulated_size' is the correct way
to pass it.

Now the test pool is being created and deleted (wiped) correctly.

Minor: text message fixed


waynesun09:
ACK, merging now. Thanks!

PR #493, opened 2015-03-12
Add more time wait function during guest operation for rhel6 guest.

rbian:


Hao-Liu:
This PR LGTM if CI passed and dependence merged.

will-Do:
Patch is OK. Just a suggestion for the concise of the code.
1. The ```if``` sentence is a bit long, it's not necessary to use ```re``` for this str judgement, and use ```any``` maybe a good choice.
2. For the wait func, you can define a tiny function for calling, such as:
``` def f(): return vm.state == True```

rbian:
Thanks @will-Do , updated the code.

will-Do:
Merge this PR now

PR #496, opened 2015-03-16
virtual_disks/virtual_disks_multidisks: close session

PandaWei:
It's better that close the session between Host and Guest
before shutdowning VM.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -445,7 +445,7 @@ def run(test, params, env):
     if test_boot_console:
         if vm.is_dead():
             vm.start()
-        session = vm.wait_for_login()
+        vm.wait_for_login().close()
         # Setting console to kernel parameters
         vm.set_kernel_console("ttyS0", "115200")
         vm.shutdown()


Hao-Liu:
Is there any fail case that indicate session should be closed? I thought it is expected to be closed automatically after vm shutdown, or there will might be a bug.
P.S.: I don't think capture and re-emit exceptions is required. Any exceptions will be captured by virt_test and treated as fail. @PandaWei Do you think so?

PandaWei:
@Hao-Liu 
Thanks for your response.

it won't introduce any fail that the close operation for session disappear.
My intention is that make it more safely and be consistent with the other similar caller.
session = vm.wait_for_login() and session.close() always come in pairs, *excludes negative case*.
obviously, The metioned is not a negative case.

and vm.set_kernel_console() will create a new session, 
so，The session created by vm.wait_for_login() turns unuseful no longer .

for your " P.S",
I'm agree with you.
I will update and remove the try/catch.
is it acceptable?
I will change TestError to TestFail

Hao-Liu:
Sure, it's nice to keep test safe. ACK.

liyangfnst:
how about `vm.wait_for_login().close()` since variable `session` won't be used..

Hao-Liu:
@liyangfnst , That's terse and good.

PandaWei:
Thanks @Hao-Liu  and @liyangfnst.
Update.

Hao-Liu:
Thanks @PandaWei @liyangfnst. Merging.

PR #502, opened 2015-03-23
virsh_managedsave: Fixed reading logs in check_multi_guests()

svirt:
Fixed argument passing for function virt_clone_cmd().

Reading the logs with the help of "systemctl" instead of "service",
as it's the correct way to obtain logs from systemd.

TODO: The function check_multi_guests() must also support the
      (older) systems which don't have systemd

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
@@ -141,7 +141,7 @@ def run(test, params, env):
         for i in range(int(guests)):
             dst_vm = "%s_%s" % (vm_name, i)
             utils_libguestfs.virt_clone_cmd(vm_name, dst_vm,
-                                            True, timeout)
+                                            True, timeout=float(timeout))
             virsh.start(dst_vm)
 
         # Wait 10 seconds for vm to start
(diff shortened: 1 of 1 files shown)


svirt:
Hello, could this be merged please? thanks

libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -153,12 +153,12 @@ def check_multi_guests(guests, start_delay, libvirt_guests):
         # add a time wait here.
 
         def wait_func():
-            return not utils.run("service libvirt-guests status"
+            return not utils.run("systemctl status libvirt-guests"
```
waynesun09: @svirt for backward compatibility on rhel6, systemctl command will fail, so can you stick on service?

will-Do: ```systemctl``` only work for systemd, so if the OS which not using systemd(RHEL6. etc.), this will fail

waynesun09:
@Hao-Liu @will-Do there is service module under staging, any idea ever it's used for the systemctl or service situation here?

svirt:
Hello, I will try to add a check to see if systemctl or service should be used on particular system

Hao-Liu:
Hi @svirt @waynesun09. I believe we should always use module `virttest/staging/service.py` when dealing with something related to init services to let virt-test handle different service systems.
If this module cannot solve your problem, it's welcome to change this module to make it better.

svirt:
Hi **@Hao-Liu**, I will check to see if it fits this case, if needed I will try to provide the needed functionality there

PR #504, opened 2015-03-24
Fix issues for rhel6 test.

rbian:
Attach-interface operation don't support "--live" option,
Error messages differ from rhel7 to rhel6.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_hotplug.py
+++ b/libvirt/tests/src/virtual_network/iface_hotplug.py
@@ -63,6 +63,8 @@ def run(test, params, env):
         try:
             # Attach an interface when vm is running
             iface_list = []
+            err_msgs = ("No more available PCI slots",
+                        "No more available PCI addresses")
             if attach_device:
                 for i in range(int(iface_num)):
                     logging.info("Try to attach interface loop %s" % i)
@@ -76,8 +78,13 @@ def run(test, params, env):
                                               flagstr=attach_option,
                                               ignore_status=True)
                     if ret.exit_status:
-                        if ret.stderr.count("No more available PCI slots"):
+                        if any([msg in ret.stderr for msg in err_msgs]):
+                            logging.debug("No more pci slots, can't"
+                                          " attach more devices")
                             break
+                        elif (ret.stderr.count("doesn't support option %s"
+                                               % attach_option)):
+                            raise error.TestNAError(ret.stderr)
                         elif status_error:
                             continue
                         else:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -76,8 +76,14 @@ def create_iface_xml(mac):
                                               flagstr=attach_option,
                                               ignore_status=True)
                     if ret.exit_status:
-                        if ret.stderr.count("No more available PCI slots"):
+                        if (ret.stderr.count("No more available PCI slots") or
+                                ret.stderr.count("No more available"
+                                                " PCI addresses")):
```
will-Do: I suggest that use a err_msg list here, so we can append the list if any more error need append.

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -76,8 +76,14 @@ def create_iface_xml(mac):
...
                             break
+                        elif (ret.stderr.count("doesn't support option %s"
+                                               % attach_option)):
+                            raise errorTestNAError("Not supported option %s"
+                                                   % attach_option)
```
will-Do: Just need raise ```ret.stderr`` here

will-Do:
BTW, pls fix the Travis CI complain.

rbian:
Thanks @will-Do , fixed.

Hao-Liu:
@rbian, It seems the current `break` can only break out the internal for-loop, which is not expected result.   Please help check this again. Thanks.

rbian:
@Hao-Liu , Yes, that's a problem.

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -76,8 +78,16 @@ def create_iface_xml(mac):
                                               flagstr=attach_option,
                                               ignore_status=True)
                     if ret.exit_status:
-                        if ret.stderr.count("No more available PCI slots"):
+                        find_msg = False
+                        for msg in err_msgs:
+                            if ret.stderr.count(msg):
+                                find_msg = True
+                                break
+                        if find_msg:
```
Hao-Liu: @rbian, How about use `any` to make it cleaner. like:
```
if any([msg in ret.stderr for msg in err_msgs]):
    log something
    break
```

rbian: That's great! I will modify the code.

Hao-Liu:
@rbian, Now this PR LGTM. please test it and we can merge this.

will-Do:
ACK and merge this PR

PR #511, opened 2015-04-01
virsh_cpu_compare: Add a new cpu element if it not exist

will-Do:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_cpu_compare.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_cpu_compare.py
@@ -89,6 +89,10 @@ def run(test, params, env):
     try:
         # Prepare VM cpu feature if necessary
         if modify_target in ['feature_name', 'feature_policy', 'delete']:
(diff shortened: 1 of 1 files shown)


PandaWei:
Look good, ACK

libvirt/tests/src/virsh_cmd/domain/virsh_cpu_compare.py
```diff
@@ -89,6 +89,12 @@ def get_cpu_xml(target, mode):
     try:
         # Prepare VM cpu feature if necessary
         if modify_target in ['feature_name', 'feature_policy', 'delete']:
+            try:
```
chuanchang: The exception process may be not necessary in here, it should be enough to use 'if not vmxml.get('cpu'):' as a condition.

will-Do: Oh, yes, I will update it, thanks.

waynesun09:
ACK and merging now, thx!

PR #512, opened 2015-04-01
virsh_cpu_baseline: Remove stderr check for negative test

will-Do:
The stderr may be empty if command failed, so remove the check
condition.


PandaWei:
ACK.

chuanchang:
ACKed.

waynesun09:
Thanks for the fix, merging now.

PR #517, opened 2015-04-08
libvirt_bench_ttcp_from_guest_to_host: adjust the timeout

PandaWei:
The timeout used by utils_misc.wait_for is set to 5, there's
not enough time to start ttcp server, so provide more time to it.
On the other hand,
The maximum duration time between server and client is too long,
so diminish it to half of the original.


will-Do:
ACK

Hao-Liu:
LGTM. Merging this. Thanks @PandaWei @will-Do

PR #520, opened 2015-04-13
virsh_snapshot_disk: disable selinux for start glusterd

PandaWei:
Encounter the following errors while start glusterd,
  ERROR: failed to create logfile
         "/var/log/glusterfs/etc-glusterfs-glusterd.vol.l...enied
  ERROR: failed to open logfile
         /var/log/glusterfs/etc-glusterfs-glusterd.vol.log
so disbale selinux for associated cases.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_disk.py
+++ b/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_disk.py
@@ -8,6 +8,7 @@ from virttest import virsh, qemu_storage, data_dir
 from virttest import libvirt_xml
 from virttest import libvirt_storage
 from virttest import utils_libvirtd
+from virttest import utils_selinux
 from virttest.utils_test import libvirt as utlv
 from provider import libvirt_version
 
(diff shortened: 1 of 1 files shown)


yangdongsheng:
Sounds good

waynesun09:
@PandaWei we did not met the selinux problem, and 'enforcing' is the requirement for our testing. So in test case module we did not suggest set selinux permissive, and if selinux deny then there might be a selinux-policy bug or on other component.
The error in your description is from start glusterd, the service is config and managed in virt-test, if really want disable selinux to let the service work, then it should be in virt-test gluster module, and restore env to enforce after service is working.

yangdongsheng:
@waynesun09 Oh, my bad!! Sorry for the rough review. You are right this should be intentional case.

PandaWei:
@waynesun09 @yangdongsheng 
Thanks for your comments.

I'm still confused...
If 'enforcing' is the requirement for redhat tests and it's intentional as yangdongsheng mentioned,
It sounds more reasonable that set selinux enforce explicitly.
but i haven't any statements to do that,
so i think it's not intentional.

could you give me more explanation? thanks in advance.

will-Do:
By default, selinux should be 'Enforcing', which means we run all test under this condition without such specification, then if encounter any permission deny, we should any analyse if it's a selinux policy bug or other problems.

PandaWei:
OK,
Thanks for @waynesun09 @yangdongsheng @will-Do 
close this PR now

PR #524, opened 2015-04-14
Storage discard

PandaWei:


PandaWei:
@Hao-Liu 
I rebase it and push a new PR.
but it still appear.
I'm not sure, there's a another vg_create which has two paramters
/virttest/staging/lv_utils.py:193:def vg_create(vg_name, pv_list):

does it make a difference in CI?

Hao-Liu:
I think this pylint fail has nothing to do with virt-test. I guess the cause of this failure is that this `force` kwarg was introduced lately but the autotest package using in Travis CI is too old to include this change.
@lmr, I saw the current autotest package for trusty in your ppa is 0.16.1, but for utopic there is a more modern one 0.16.2-4-g5042f. Is it safe to update travis autotest package from trusty to utopic in virt-test related repos? Thanks.

lmr:
@Hao-Liu, I see, thanks for bringing this to my attention. Let me see what I can do.

libvirt/tests/src/storage_discard.py
```diff
@@ -74,7 +75,7 @@ def create_volume(device, vgname="vgthin", lvname="lvthin"):
     iscsi service if it is None.
     """
     # Create volume group
-    lv_utils.vg_create(vgname, device)
+    lv_utils.vg_create(vgname, device, force=True)
```
liyangfnst: The `force` wasn't used.

libvirt/tests/src/storage_discard.py
```diff
@@ -124,6 +125,19 @@ def get_vm_disks(vm):
...
 
+def _sync_finished():
+    """
+    check volume's capacity once every 60 seconds
+    """
+    lv_cap_before = volumes_capacity("lvthin")
+    time.sleep(60)
+    lv_cap_later = volumes_capacity("lvthin")
+    logging.debug("Synchronizing: %s -> %s", lv_cap_before, lv_cap_later)
+    if lv_cap_before != lv_cap_later:
```
liyangfnst: `return (lv_cap_before == lv_cap_later)` is fine.

libvirt/tests/src/storage_discard.py
```diff
@@ -124,6 +125,19 @@ def get_vm_disks(vm):
     return disks
 
 
+def _sync_finished():
+    """
+    check volume's capacity once every 60 seconds
+    """
+    lv_cap_before = volumes_capacity("lvthin")
+    time.sleep(60)
```
liyangfnst: why `60s`, I just think it will take too long time, we can check them every 2s in a circle, and set timeout=60(return False).

PandaWei: No,
The change process is very slow in practice
2s is too short to detect any difference.

libvirt/tests/src/storage_discard.py
PandaWei: Good idea, i will adopt it.

libvirt/tests/src/storage_discard.py
PandaWei: can you explain it more clear?

liyangfnst: sorry, I missed, I thought it's in the definition of the method :(

libvirt/tests/src/storage_discard.py
liyangfnst: I meant this:
```
lv_cap_before = volumes_capacity("lvthin")
timeout=60
cost_time=0
while cost_time < timeout:
    time.sleep(2)
    lv_cap_later = volumes_capacity("lvthin")
    if lv_cap_later == lv_cap_before:
        return True
    cost_time = cost_time + 2
return False
```

PandaWei: I have got what you want to say.
it's not unworkable.
the reason is that the change process is very *slow* in practice.
if just sleep two seconds.
it will return true in the first loop. but the synchronization operation doesn't finish yet.
That's why i set the interval to 60 seconds

PandaWei: Just for supplement,
utils_misc.wait_for() will call _sync_finished in loop.

liyangfnst: Thanks for you explanation. it's fine.

PandaWei:
@Hao-Liu  Is there any news about CI failure?

Hao-Liu:
Hi @lmr, I did an experiment to use autotest-0.16.2-4-g5042f in https://github.com/autotest/tp-libvirt/pull/540. It looks fine. Can we introduce this to tp-libvirt Travis CI? This is really needed to get new PRs relying on newer autotest pass.

lmr:
@Hao-Liu, thanks, that patch looks fine, I'm going to submit it and merge it.

Hao-Liu:
It seems the fix on Travis CI works. Thank @lmr for fixing this.
This PR looks good. I'm going to merge it. Thanks @PandaWei @liyangfnst.

PR #525, opened 2015-04-15
virsh_domblkstat.cfg: Space extra option cfg fix

svirt:
domblkstat_extra field should contain a space (space_extra_option).
Currently there is only '' (empty) which means that in fact no extra
option is passed to virsh domblkstat command, thus the error case
is not properly formed and some of the virsh versions return success
because of omitting the empty '' paramater


svirt:
Hopefully this can be merged easily

waynesun09:
ACK and merging now, @svirt thanks.

PR #530, opened 2015-04-17
virsh_cpu_baseline.py: Unsupported CPU/flag (AMD) fail case fix

svirt:
This particular test and its config by default are designed to be
run on Intel CPUs. I was running it on AMD-Opteron based systems and
the test virsh.cpu_baseline.positive_test.config_guest was raising
an exception (due to unnsupported CPU/flags, e.g. on AMD /proc/cpuinfo
has no flag "acpi") resulting in test failure. Additional exception
handling is added to avoid this situation.
TODO: In the future might be a good idea to adjust the test for AMD
      CPUs too.


svirt:
Tested locally both on Intel and AMD

waynesun09:
Test good on rhel6 and fedora, ACK and merging now.

PR #532, opened 2015-04-21
virsh_snapshot: repalce variable snaps with snap

PandaWei:
The snapshot will be removed one by one.
When we get failure, the whole snapshot list will be logged,
not a specific snapshot. so,
It's hard to find out which snapshot is the cause of failure.


waynesun09:
ACK and merging the trivial fix, Thanks @PandaWei

PR #533, opened 2015-04-21
perf_kvm: remove the unused function run_dd_on_guest

PandaWei:
run_dd_on_guest never be used, so delete it

### DIFF
--- a/libvirt/tests/src/perf_kvm.py
+++ b/libvirt/tests/src/perf_kvm.py
@@ -134,14 +134,6 @@ def get_kernel_file(vm):
     return (guestkallsyms, guestmodules)
 
 
-def run_dd_on_guest(vm):
-    """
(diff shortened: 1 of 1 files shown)


will-Do:
Yes, the function is useless, many cases which need do 'dd' in guest just execute the command manually. I'm okay with this remove.
BTW,  we'd better not touch such cases(no used functions) usually, as some classes/functions are created along with(or without) the test framework for future intent.

PandaWei:
Thanks for your Okay and comment.
yes , i agree that some classes/functions are created for future intent and don't touch them.
In fact, I just think dd command is performed usually on guest for many cases.
and i want to move it to virt-test.

Hao-Liu:
This seems like an orphan function will wasn't used at all. I believe we should remove this. ACK this. Thanks @PandaWei @will-Do.

PR #536, opened 2015-04-24
nwfilter_vm_start: Fix clean console serial problem

waynesun09:
The read only tmp dir will cause default temp session file failure, so
skip serial login for noexec_tmp_dir case.


will-Do:
ACK

Hao-Liu:
ACK and merging. Thanks @waynesun09 @will-Do.

PR #537, opened 2015-04-24
Fix set_agent_channel problem.

rbian:


waynesun09:
ACK, thx for the quick fix

PR #538, opened 2015-04-24
virsh_vcpupin.py: Fix for cpu_list_x-y test case

svirt:
The domain should have "shut off" status before restarting with newly
set pinned vcpu options

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_vcpupin.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_vcpupin.py
@@ -6,6 +6,7 @@ from autotest.client.shared import error
 from virttest import virsh, utils_test
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
+from virttest import utils_misc
 
 
 def run(test, params, env):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_vcpupin.py
```diff
@@ -118,6 +119,7 @@ def offline_pin_and_check(vm, vcpu, cpu_list):
         vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm.name)
         vmxml.cputune = cputune
         vmxml.sync()
+        utils_misc.wait_for(lambda: vm.state() == "shut off", 10)
```
will-Do: So it just wait 10 sec if vm.state() is not 'shut off', it is what you want? 
Maybe the problem you mentioned is: 
```
103                 if options == "--config":
104                     virsh.destroy(vm.name)


148     if offline_pin:
149         vcpucount_option = "--config --active"

svirt:
Hi,
thanks for the review.
yes, just need to ensure the domain is shut off from the previous actions, otherwise the next command virsh.start will return an error. So I think this wait_for() is enough for this case.

will-Do:
I'm not run this case, just want to confirm that for offiline_pin, vm will be destroyed as expected. If you run the cases and no error/fail occurred, we can merge this PR

svirt:
I ran all the 24 tests from the set virsh.vcpupin , and there is no error/fail

will-Do:
Greate, now we can merge this PR, thx

PR #539, opened 2015-04-27
snapshot_create_as: Fix snapshot name exist problem for multi snapshots

waynesun09:
The multi snapshots test could fail with snapshot name exist, which
by default using incremental time string when snapshot name not
specified. Fix this by sleep 1 second if doing multiple snapshots.


Hao-Liu:
ACK this. BTW I think 1.1s will be safer.

will-Do:
ACK and merge

PR #543, opened 2015-04-28
multiqueue: remove impossible except error.TestFail

PandaWei:
The wrapped statements would never throw error.TestFail,
so discard it.


Hao-Liu:
Sure it should be removed. Thanks @PandaWei

PR #545, opened 2015-04-30
libvirt/tests/src/macvtap: fix nested try-except bug

PandaWei:
def a():
    Maybe raise ShellCmdError

def call_a():
    try:
        a()
    except ShellCmdError, e:
        pass

def call_call_a():
    try:
        call_a()
    except ShellCmdError, e:
        # The following would never be run,
        # even if ShellCmdError is throwed by a()
        TestFail()

so it's need to remove the try-except of call_a().

### DIFF
--- a/libvirt/tests/src/macvtap.py
+++ b/libvirt/tests/src/macvtap.py
@@ -206,12 +206,12 @@ def run(test, params, env):
     try:
         try:
             session = guest_config(vm1, vm1_ip)
-        except (remote.LoginTimeoutError, aexpect.ShellCmdError), fail:
+        except remote.LoginTimeoutError, fail:
             raise error.TestFail(str(fail))
         if vm2:
             try:
                 guest_config(vm2, vm2_ip)
-            except (remote.LoginTimeoutError, aexpect.ShellCmdError), fail:
+            except remote.LoginTimeoutError, fail:
                 raise error.TestFail(str(fail))
 
         # Four mode test


libvirt/tests/src/macvtap.py
```diff
@@ -94,16 +94,9 @@ def guest_config(vm, ip_addr):
...
-        except aexpect.ShellCmdError:
-            pass
-
-        try:
-            session.cmd("ifup %s" % eth_name)
-        except aexpect.ShellCmdError:
-            pass
+        # Attached interface maybe already active
+        session.cmd("ifdown %s" % eth_name)
+        session.cmd("ifup %s" % eth_name)
```
yangdongsheng: WHY?

libvirt/tests/src/macvtap.py
```diff
@@ -94,16 +94,9 @@ def guest_config(vm, ip_addr):
                                         eth_config_file)
         remote_file.truncate()
         remote_file.add(eth_config_detail_list)
-        try:
-            # Attached interface maybe already active
-            session.cmd("ifdown %s" % eth_name)
-        except aexpect.ShellCmdError:
-            pass
```
waynesun09: The pass here have a assumption that: if the interface already down and the ifdown command will fail then let the test ongoing.

And this also works for ifup command later.

So caught ShellCmdError here is not for raise TestFail as you explain and changed here.

waynesun09:
@PandaWei follow my last comment, maybe the focus will be shift to whether need to check ShellCmdError when use the guest_config function:

```
try:
    guest_config()
except aexpect.ShellCmdError:
    raise error.TestFail()
```

PandaWei:
@waynesun09 
Yes, you are right.
We should focus on the try-catch for guest_config().
I have updated and pushed  it again.
Looking forward your feedback.

PandaWei:
libvirt/tests/src/macvtap: remove unused nested-try-catch
    
    Explanation as following,
    
    def a():
        Maybe raise ShellCmdError
    
    def call_a():
        try:
            a()
        except ShellCmdError, e:
            pass
    
    Obviously,
    The ShellCmdError has been caught and ignored by itself.
    No need to catch it by the caller of call_a().
    
    def call_call_a():
        try:
            call_a()
        except ShellCmdError, e:
            # The following would never be run,
            # even if ShellCmdError is throwed by a().
            # The try-catch for ShellCmdError is unuseful.
    
    This patch aims to remove unuseful catch for exception.

    def call_call_a():
        try:
            call_a()
        except ShellCmdError, e:
            # The following would never be run,
            # even if ShellCmdError is throwed by a().
            # The try-catch for ShellCmdError is unuseful.
    
    This patch aims to remove unuseful catch for exception.

libvirt/tests/src/macvtap.py
PandaWei: Because aexpect.ShellCmdError will be caught at the caller of guest_config.
and I have updated it.
Please take a look at the following conversation.

waynesun09:
Just checked the function, as get_linux_ifname handled ShellCmdError and I don't think remote module raise ShellCmdError, so only session.cmd() will raise the error in guest_config, then the change here will not get the error missed.
ACK.

Hao-Liu:
Thanks @PandaWei @yangdongsheng @waynesun09. Now merging this.

PR #549, opened 2015-05-06
virsh_net_dhcp_leases: fix pep8 E701

PandaWei:
E701 means multiple statements on one line (colon)


Hao-Liu:
ACK and merging for trivial fix. Thanks @PandaWei

PR #553, opened 2015-05-13
virsh_migrate_stress: fix bug for option --compressed

PandaWei:
The variable 'options' that may be set with '--compressed'
in order to compresss repeated pages during live migration.

The fed9e53 declares a new variable 'migrate_options' and
replace 'options' with it for do_migration(), which leads
to all of cases that used to test compresss repeated pages
during live migration don't do what they ought to do.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
@@ -59,9 +59,9 @@ def do_stress_migration(vms, srcuri, desturi, stress_type,
     fail_info = utils_test.load_stress(stress_type, vms, params)
 
     migtest = utlv.MigrationTest()
-    options = None
+    options = ''
     if migration_type == "compressed":
-        options = "--live --compressed --timeout 60"
+        options = "--compressed"
         migration_type = "orderly"
         shared_dir = os.path.dirname(data_dir.get_data_dir())
         src_file = os.path.join(shared_dir, "scripts", "duplicate_pages.py")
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
```diff
@@ -77,8 +77,8 @@ def do_stress_migration(vms, srcuri, desturi, stress_type,
         logging.warning("Add stress for migration failed:%s", fail_info)
 
     logging.debug("Starting migration...")
-    migrate_options = ("--live --unsafe --timeout %s"
-                       % params.get("virsh_migrate_timeout", 60))
+    migrate_options = ("--live --unsafe %s --timeout %s"
+                       % (options, params.get("virsh_migrate_timeout", 60)))
```
waynesun09: if migration_type not as "compressed", here options will be default to None, won't it cause problem?

PandaWei: It's my fault, i didn't say it clearly at commit log.
I have updated the commit log.

The key point is that '--live --timeout %s'  will be set twice for compress case.

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
```diff
@@ -61,7 +61,7 @@ def do_stress_migration(vms, srcuri, desturi, stress_type,
     migtest = utlv.MigrationTest()
     options = None
```
waynesun09: Ok, I should directly comment here as:
s/options = None/options = ''/

my concern is on "--live --unsafe None --timeout 60" as option for none compressed cases.

PandaWei:
@waynesun09 
Yes, you are right.
Thanks for your comments.
Update it

waynesun09:
ACK and merge now, thx

PR #554, opened 2015-05-13
virsh_list.py: Bigger timeout is needed on slower connection/systems

svirt:
Bigger maximum timeout is needed on slower connection/systems

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_list.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_list.py
@@ -35,7 +35,7 @@ def run(test, params, env):
         command_on_remote = ("virsh -c %s list %s"
                              % (complete_uri, options_ref))
         session = remote.remote_login(
-            "ssh", remote_ip, "22", "root", remote_passwd, "#")
+            "ssh", remote_ip, "22", "root", remote_passwd, "#", timeout=60)
         time.sleep(5)
         status, output = session.cmd_status_output(
             command_on_remote, internal_timeout=5)


libvirt/tests/src/virsh_cmd/monitor/virsh_list.py
```diff
@@ -35,7 +35,7 @@ def list_local_domains_on_remote(options_ref, remote_ip,
         command_on_remote = ("virsh -c %s list %s"
                              % (complete_uri, options_ref))
         session = remote.remote_login(
-            "ssh", remote_ip, "22", "root", remote_passwd, "#")
+            "ssh", remote_ip, "22", "root", remote_passwd, "#", timeout=60)
```
yumingfei: Can you provide this timeout in virsh_list.cfg? I prefer to resize this timeout by testers.

svirt:
Hi, thanks for the review.
I think probably so far we can delay with those timeout changes application, as seemed there was a network/ssh related problem causing timeouts and even on slower system the current settings are fine after fixing those problems. I will see how are going the tests in the future and maybe if needed will come with those timeout changes.

PandaWei:
Hi @svirt
I agree with @yumingfei 's comment, 
It's better that adjust timeout by testers

Of course, your consideration seems reasonable to me.
so I recommend you that close this PR firstly.
and Maybe you can push another new PR to modify the default of timeout used by remote.remote_login().
which can refer to the implementation of BaseVM().remote_login()

def remote_login(self, nic_index=0, timeout=LOGIN_TIMEOUT,
                     username=None, password=None):

Use the same default value of timeout looks more better.

That's just reminder.

svirt:
@PandaWei
Hi, thanks for your note, yes I agree,  it will be better i think. I will close this PR, will review the tests' behavior and in case if needed will open another PR

PR #556, opened 2015-05-15
virsh_qemu_monitor_command: remove unuesed 'start_vm'

PandaWei:
The variable 'start_vm' is assigned to but never used,
and The virt-test framework is responsible for checking
parameters and starting a VM or not.

So remove it and it's assignment statement.


yumingfei:
OK.

PR #559, opened 2015-05-22
libvirt/tests/src/virt_cmd: don't get VM object

PandaWei:
It's meaningless to get VM object, VM name is enough.


yumingfei:
Yes, it should be.

PR #427, opened 2015-01-15
Add test for virsh command domfsfreeze and domfsthaw

Hao-Liu:
This test virsh domfsfreeze and domfsthaw commands and their options.

1) Start a guest with/without guest agent configured;
2) Freeze the guest file systems with domfsfreeze;
3) Create a file on guest to see command hang;
4) Thaw the guest file systems with domfsthaw;
5) Check the file is already created;
6) Retouch the file the ensure guest file system are not frozen;
7) Cleanup test environment.

This PR depends on https://github.com/autotest/virt-test/pull/1961, https://github.com/autotest/virt-test/pull/1962, and https://github.com/autotest/virt-test/pull/1964

Signed-off-by: Hao Liu <hliu@redhat.com>

PR #567, opened 2015-06-03
Add test case for virsh domfsfreeze

liuzzfnst:


PandaWei:
Hi,
This PR has the same problem as #568 .
Please fix it and attach your test result

PandaWei:
I have no other suggestions about this PR.
but show us the result.
Thanks

PandaWei:
@liuzzfnst 
Hi, 
there're some testcases about domfsfreeze. #427 
So, Please take a look at it, and found if it has any difference。
if not, i suggest you close this PR.

Maybe we should check the PR list before pushing next time.

liuzzfnst:
@PandaWei 
I think a acl test case should be added into #427, I will fix my patches according to #427.

PandaWei:
OK,
Let's wait for updating #427 by @Hao-Liu .

libvirt/tests/src/virsh_cmd/domain/virsh_domfsfreeze.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if mountpoint:
+            cmd_result = virsh.domfsfreeze(vm_ref, mountpoint=mountpoint,
+                                           options=options,
+                                           unprivileged_user=unprivileged_user,
+                                           uri=uri, debug=True)
+        else:
+            cmd_result = virsh.domfsfreeze(vm_ref, options=options,
+                                           unprivileged_user=unprivileged_user,
+                                           uri=uri, debug=True)
```
will-Do: The ```if...else``` block is unnecessary.
Just let ```mountpoint = params.get("domfsfreeze_mnt")```.

will-Do:
And, just check the command return value may be not enough for our test, expected more detail checking added.

libvirt/tests/src/virsh_cmd/domain/virsh_domfsfreeze.py
```diff
@@ -0,0 +1,93 @@
...
+        if not status_error:
+            if cmd_result.exit_status != 0:
+                raise error.TestFail("Fail to do virsh domfsfreeze, error %s" %
+                                     cmd_result.stderr)
+        else:
+            if cmd_result.exit_status == 0:
+                raise error.TestFail("Command 'virsh domfsfreeze' failed ")
+
```
will-Do: BTW, there's a function ```check_exit_status``` in virttest/utils_test/libvirt.py, you can use it

PandaWei:
@liuzzfnst 
Long time no response.
what's up? do you have time to make some modification as comments?

waynesun09:
The PR have been picked up with new PR:
https://github.com/autotest/tp-libvirt/pull/1202

close this one now, thx

PR #569, opened 2015-06-04
fix delimiter problem about iSCSi

PandaWei:
The targetcli doesn't support '_' like below,
    /> /iscsi/ create iqn.2015-06.com:emulated_iscsi.target
    WWN not valid as: iqn, naa, eui
And, There're many string(s) of 'emulated_image',
which selects '-' as the delimiter.

So, replace '_' with '-' to keep consistence.

Signed-off-by: Wei,Jiangang <weijg.fnst@cn.fujitsu.com>

### DIFF
--- a/libvirt/tests/cfg/resource_abnormal.cfg
+++ b/libvirt/tests/cfg/resource_abnormal.cfg
@@ -6,7 +6,7 @@
     pool_name = "test_pool"
     volume_name = "test_vol"
     volume_size = "2G"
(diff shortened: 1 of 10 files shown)


PandaWei:
This PR depends on virt-test#2202

waynesun09:
virt-test depend PR merged now, ACK and merging this to avoid testing problem, Thanks @PandaWei

PR #570, opened 2015-06-08
Add memory device tests.

rbian:


libvirt/tests/src/libvirt_mem.py
```diff
@@ -0,0 +1,438 @@
...
+    def get_numastat():
+        """
+        Return total memory usage from numastat output
+        """
+        pid = vm.get_pid()
+        cmd = "numastat %s" % pid
+        numa_mem = utils.run(cmd).stdout.strip()
+        logging.debug("Numastat output: %s", numa_mem)
+        total_line = re.findall(r"^Total.*", numa_mem, re.M)[0]
+        return re.findall(r"(\d+.\d+)", total_line)
```
will-Do: This's more like a common function, maybe we can move it into virt-test

libvirt/tests/src/libvirt_mem.py
```diff
@@ -0,0 +1,438 @@
...
+        session = vm.wait_for_login()
+        # Hot-plugged memory should be online by udev rules
+        udev_file = "/lib/udev/rules.d/80-hotplug-cpu-mem.rules"
+        udev_rules = ('SUBSYSTEM=="memory", ACTION=="add", TEST=="state",'
+                      ' ATTR{state}=="offline", ATTR{state}="online"')
+        cmd = ("grep memory %s || echo '%s' >> %s"
+               % (udev_file, udev_rules, udev_file))
+        session.cmd(cmd)
+        # Wait a while for new memory to be detected.
+        time.sleep(3)
```
will-Do: Is sleep 3 sec enough or we need a better solution to detected it?

libvirt/tests/src/libvirt_mem.py
```diff
@@ -0,0 +1,438 @@
...
+        global new_cur_mem
+        if attach_option.count("config"):
+            dom_xml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+        else:
+            dom_xml = vm_xml.VMXML.new_from_dumpxml(vm_name)
+        xml_tree = dom_xml.xmltreefile
+        try:
+            xml_max_mem_rt = int(xml_tree.find("maxMemory").text)
+            xml_max_mem = int(xml_tree.find("memory").text)
+            xml_cur_mem = int(xml_tree.find("currentMemory").text)
```
will-Do: VMXML should deal with these properties, so you can use them directly

libvirt/tests/src/libvirt_mem.py
```diff
@@ -0,0 +1,438 @@
...
+            mem_xml.address = mem_xml.new_mem_address(
+                **{"attrs": mem_addr})
+
+        logging.debug("Memory device xml: %s", mem_xml)
+        return mem_xml.copy()
+
+    def add_device(dev_xml, at_error=False):
+        """
+        Add memory device by attachment or modify domain xml.
+        """
```
will-Do: Dose this func has a similar purpose  with ```modify_domain_xml``` below? So I think you can merge them together

waynesun09:
This depends on:
https://github.com/autotest/virt-test/pull/2206

libvirt/tests/src/libvirt_mem.py
rbian: This func is for memory device, it can be used for attach/detach device. It's better to keep it.

waynesun09:
ACK

waynesun09:
Thx and merge now.

PR #575, opened 2015-06-12
Fix a bug for managedsave test.

rbian:


PandaWei:
Good Catch!
ACK.

But it's better to add some commit message.
Just like,
"Only if both ‘Suspending VM’ and ‘stopped, with saved guests’ aren't found, it will raise a Testfail"

rbian:
Thanks @PandaWei . Updated the patch.

PandaWei:
Thanks @rbian
Merge now

PR #576, opened 2015-06-12
virsh_iface_edit: fix bug about catching exception

PandaWei:
If catching more than two exceptions without brackets,
for an example,
    try:
        ...
    except error.CmdError, IndexError:
        pass
The 'IndexError' will be regard as local variable,
not a exception type.

The standing procedure like below,
    try:
        ...
    except (RuntimeError, TypeError):
        pass or handle

This patch is used to solve the above problem.


waynesun09:
nice catch with good commit explanation, thx @PandaWei , merge now.

PR #571, opened 2015-06-10
find_storage_pool_sources_as: delete check iscsi command

waynesun09:
The iscsi support will be checked in iscsi module, no need check
in test module.

Signed-off-by: Wayne Sun <gsun@redhat.com>

PR #577, opened 2015-06-15
find_storage_pool_sources: Replace localhost to ipv4 address

waynesun09:
Replace localhost to 127.0.0.1 to avoid ipv6 resovle problem.


waynesun09:
Merge this as a follow up with https://github.com/autotest/tp-libvirt/pull/571

PR #580, opened 2015-06-17
Add test case about combint '--live' and '--config' for

liuzzfnst:
virsh blkdeviotune.

No matter getting or setting blkdeviotune value for device,
'--live' and '--config' can be used at the same time.

This PR depends on autotest/virt-test#2221.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_blkdeviotune.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_blkdeviotune.cfg
@@ -16,9 +16,9 @@
                                         - none:
                                             blkdevio_options =
                                         - config:
-                                            blkdevio_options = "config"
+                                            blkdevio_options = "--config"
                                         - current:
-                                            blkdevio_options = "current"
+                                            blkdevio_options = "--current"
                         - running_guest:
                             start_vm = "yes"
                             variants:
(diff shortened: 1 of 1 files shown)


liuzzfnst:
Actually, the new added case fails.
```
23:55:48 ERROR| FAIL type_specific.io-github-autotest-libvirt.virsh.blkdeviotune.positive_testing.get_blkdevio_parameter.running_guest.options.live_and_config -> TestFail: error: Unable to get number of block I/O throttle parameters
error: flags 'affect live' and 'affect config' in virDomainGetBlockIoTune are mutually exclusive
```
But I think it should be libvirt's bug, and the manpage about virsh blkdeviotune says:
```
Both --live and --current flags may be given, but --current is exclusive.
```
It should be 
```
Both --live and --config flags may be given, but --current is exclusive.
```

PandaWei:
Hi,
I don't think it's a bug and no clue supports your idea.
As the log you mentioned, I guesss ....,
Maybe "--config" is just effective on un-running VM.
Do you have verify it?
If you have another new information supports your idea,
Please attach them.

and I have a question,
you said "Both --live and --config flags may be given, but --current is exclusive",
why '--current' is exclusive  when just specify '--live' and '--config'?

waynesun09:
@liuzzfnst 
--live with --config is working fine

```
# virsh list
 Id    Name                           State
----------------------------------------------------
 56    virt-tests-vm1                 running

# virsh blkdeviotune virt-tests-vm1 --config --live --device vda --total-bytes-sec 9223372036854775807

# echo $?
0
```

as https://github.com/autotest/virt-test/pull/2221 merged, the case start to fail now without this fix, I raised a new PR to fix this then find out this one not resolved yet.

As updated blkdevio_options in cfg, you need also update check in test module, as I put it in PR:
https://github.com/autotest/tp-libvirt/pull/620

after you updated the code here, I'll close mine.

waynesun09:
PR: #620 merged, this PR could be close now.

PR #581, opened 2015-06-17
Fix pool problem as after 1.2.15 dir pool will not be autostarted

kylazhang:


waynesun09:
ACK and merge as trivial fix.

PR #583, opened 2015-06-19
Modify domfstrim cases

kylazhang:
1. Change the disk_path as /dev/disk/by-path can not be find anymore
2. Add judgement for scsi disk exist for current guest and then do different operation
3. Add vm.shutdown as for guest agent vm destroy will crash the guest agent in guest


waynesun09:
ACK and merging now

PR #588, opened 2015-07-08
Add 16M hugepage size support.

rbian:
16M hugepage size is used on PPC machine, verify supported hugepage size
before the test, only continue the test if hugepage size is in suppported
size list.


waynesun09:
ACK and merge now, thx @rbian

PR #589, opened 2015-07-08
Remove all files in dir pool before pool-delete

kylazhang:


will-Do:
Hi kyla, I think you can rewrite the patch summary

kylazhang:
@will-Do yes, I will update, thanks

waynesun09:
ACK and merge now

PR #600, opened 2015-07-15
attach_device: Attach scsi controller before scsi disk

will-Do:
Attach scsi disk required scsi controller, so try to attach scsi
controller first in case it not exist.


waynesun09:
ACK

PR #603, opened 2015-07-20
Test iothread option for gluster disk.

rbian:


libvirt/tests/src/virtual_disks/virtual_disks_gluster.py
```diff
@@ -91,6 +93,9 @@ def build_disk_xml(disk_img, disk_format, host_ip):
                 host_dict.update({"transport": transport})
             disk_xml.source = disk_xml.new_disk_source(
                 **{"attrs": source_dict, "hosts": [host_dict]})
+        # set domain options
+        if dom_iothreads:
+            vmxml.iothreads = int(dom_iothreads)
```
will-Do: It's better to make sure this will not fail for ValueError

will-Do:
ACK

PR #608, opened 2015-07-21
virsh_freepages: bugfix

Chenditang:
If freepages_cellno is not set，cellno will be None, and then the 
for loops will never be processed
113     # Run test
114     for cell in cellno_list:
115         for page in pagesize_list:
116             result = virsh.freepages(cellno=cell,
some test cases will always return PASS.
pagesize has the same problem.


will-Do:
ACK

PR #615, opened 2015-07-29
libvirt: USB device passthrough test case using hostdev clause

bssrikanth:
Libvirt test cases added to test USB passthrough. This test has two variants 
which offers users to select a specific USB device for passthrough or going 
ahead with all available USB device for passthrough test.


bssrikanth:
sorry.. will fix pep8 errors and then create a pull request

PR #621, opened 2015-08-06
blockcopy: update options for reuse external

waynesun09:
Update options for first blockcopy try on reuse external cases.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -457,6 +457,10 @@ def run(test, params, env):
         # Run blockcopy command
         if rerun_flag == 1:
             options1 = "--wait %s --finish --verbose" % dest_format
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -457,6 +457,10 @@ def _make_snapshot():
         # Run blockcopy command
         if rerun_flag == 1:
             options1 = "--wait %s --finish --verbose" % dest_format
+            if with_blockdev:
+                options1 += " --blockdev"
```
PandaWei: It seems that the old version of virsh doesn't support '--blockdev',
Do we need to consider it?

waynesun09: yeah, I'll update with check version

will-Do:
ACK

waynesun09:
thx, merge now

PR #623, opened 2015-08-06
remote_access.py: Fix skip message

lmr:
The exception message has to use interpolation if you
want it to be correctly displayed to the user.

Signed-off-by: Lucas Meneghel Rodrigues <lmr@redhat.com>


PandaWei:
ACK and Merge Now.  Thanks @lmr 
I will push a new PR to fix the similar problems.

PR #624, opened 2015-08-07
fix exception message

PandaWei:
1. Fix skip message for TestNAError() and TestFail()
   please refer to the commit beacb9d for more details.

2. remote_access.py: fix %s in message
  The '%s' which in exception string should be replaced
  with variable 'pattern'.


waynesun09:
ACK and merging now, thx

PR #625, opened 2015-08-11
virsh: CPU related cases update to adapt PPC64 arch

will-Do:
1. On ppc64/ppc64le, processor ID is discontinuous, so the range notaion
is not support(so as the exclusive naotaion).
2. As the processor count may not be a valid processor id, so we fetch
all the valid numbers from /proc/cpuinfo and then select from them
randomly.
3. In case virsh dumpxml, it hard code cpu feature names to update and
check, and the features are not support on PPC64, so disable them too.


waynesun09:
ACK and merge now, thx

PR #628, opened 2015-08-17
virt_top: Add some negative case

Chenditang:
test some invalid option: "virt-top -0" and "virt-topi --xyz"

### DIFF
--- a/libvirt/tests/src/virt_cmd/virt_top.py
+++ b/libvirt/tests/src/virt_cmd/virt_top.py
@@ -20,30 +20,41 @@ def run(test, params, env):
     vm_name = params.get("main_vm", "virt-tests-vm1")
     output = params.get("output_file", "output")
     output_path = os.path.join(data_dir.get_tmp_dir(), output)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virt_cmd/virt_top.py
```diff
@@ -20,30 +20,41 @@ def run(test, params, env):
...
-        else:
-            continue
-    if not success:
-        raise error.TestFail("Command virt-top exit successfully, but "
-                             "domid is expected")
+        if not success:
+            raise error.TestFail("Command virt-top exit successfully, but "
+                                 "domid is expected")
+    else:
+        if cmd_result.exit_status != 2:
```
will-Do: What't the meaning of exit code equal to 2?

Chenditang: According to my test, command with invalid options will return 2.

will-Do:
OK, ACK this update

PR #631, opened 2015-08-21
Do not add panic device if already exist

waynesun09:
If test domain already have panic device, no need to add new one in
xml.


will-Do:
ACK

PandaWei:
LGTM. 
Thanks @waynesun09 @will-Do

PR #634, opened 2015-08-26
virsh_domif_setlink_getlink: Fix a code bug.

rbian:
Since login by ssh may timeout, remove the operation before setlink on guest.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domif_setlink_getlink.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domif_setlink_getlink.py
@@ -177,8 +177,9 @@ def run(test, params, env):
 
             # Set the link up make host connect with vm
             domif_setlink(vm_name, device, "up", "")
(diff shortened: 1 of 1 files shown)


PandaWei:
We can increase the waiting time to avoid timeout.
Why remove it directly?

rbian:
If login timeout, wait_for_login will login by serial console, this will disrupt the serial sesion which defined previously in the funciton. I thinks jsut checking 'ip link' on the guest is okay.

PandaWei:
Indeed, we can check it's status with 'ip link'.
But .no codes to do it.
if you remove wait_for_login(), 
the following code is 'cmd_status = session.cmd_status('ip link set dev %s down'',
which will set it down directly without checking if it's up.

rbian:
Okay, modified the code to check link state after domif_setlink. Thanks @PandaWei

PandaWei:
ACK.

waynesun09:
ACK and merge now, thx.

PR #610, opened 2015-07-22
domdisplay: update cfg to avoid cartesian keywords

waynesun09:
Update 'vnc' and 'spice' in cfg to avoid cartersian keywords problem
on some env.

Signed-off-by: Wayne Sun <gsun@redhat.com>

PR #637, opened 2015-08-27
Update 'bridge' in cfg to avoid cartersian keywords problem on some env.

rbian:
Fix same issue as autotest/tp-libvirt#610


waynesun09:
ACK and merge as trivial fix, Thx!

PR #552, opened 2015-05-12
Bugfix: Fix the cases NG when the domain has cdrom in virsh-domblklist

liuzzfnst:
When the domain has a none source cdrom driver, new_disk.source would
point to no object. And sometimes the output from xml and command output's
order is different, so we should order them and then compare them.

PR #640, opened 2015-08-31
Bugfix: Fix the cases NG when the domain has cdrom in virsh-domblklist

liuzzfnst:
Change the autotest/tp-libvirt#552 to this branch.

When the domain has a none source cdrom driver, new_disk.source would
point to no object. And sometimes the output from xml and command output's
order is different, so we should order them and then compare them.


PandaWei:
You'd better to clean up your own branch.

Chenditang:
@PandaWei 
Updated, Thanks~~

will-Do:
ACK

waynesun09:
ACK and merging now, thx

PR #642, opened 2015-09-04
Blockpull test config fix

svirt:
Now blockpull commands in virsh in some cases return different values
for success and error cases. The corresponding changes were done in
libvirt before. So in virt-test the blockpull tests must process
the return values (success/fail) correctly

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockpull.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockpull.cfg
@@ -85,3 +85,8 @@
                     snap_in_mirror = "yes"
                     status_error = "no"
                     snap_in_mirror_err = "yes"
+    # Now blockpull commands in virsh in some cases return different values for success and error cases. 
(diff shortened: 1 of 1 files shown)


libvirt/tests/cfg/virsh_cmd/domain/virsh_blockpull.cfg
```diff
@@ -85,3 +85,5 @@
                     snap_in_mirror = "yes"
                     status_error = "no"
                     snap_in_mirror_err = "yes"
+    normal_test.file_disk.local.timeout.nobase.async: status_error = "yes"
+    normal_test.file_disk.local.timeout.nobase.no_async: status_error = "yes"
```
will-Do: Maybe you can add a comment in the cfg, let others know why theses two cases will fail

svirt:
**@will-Do:**  Comment added. Thanks for the review

will-Do:
@svirt 
Thx, I'm OK with this patch, so ACKed.

PR #643, opened 2015-09-08
dompmsuspend: Add new fail pattern.

Hao-Liu:
Fail message changed when doing `dompmsuspend` on a paused vm.
Add a new fail pattern to fix this.

Wait for state change to `pmsuspended` before continuing additional
testing.


will-Do:
ACK

PR #645, opened 2015-09-14
v2v: Add new cases

will-Do:
1. Fix the vm_type to 'libvirt' in this test as the VMs which need
to be handled are in local host(libvirt).
2. Expend rhev related cases, now we can login the VM and do checks.
3. Expend input disks, so the related cases matrix could be:
    in :
        format: raw/qcow2
        allocation: sparse/preallocated
    out:
        format: raw/qcow2
        allocation: sparse/preallocated
4. Add cases for specific disk types: virtio scis, sata.
5. Add cases for multiple disks VMs(including Linux and Windows VM).


waynesun09:
ACK

PR #646, opened 2015-09-16
iscsi: Refresh the pool after pool creation.

rbian:
In some cases, automation tests is running so fast that vol-list
command couldn't get volume info after pool creation, manual test
haven't his problem. Run pool-reresh command after pool creation
to solve this problem.


PandaWei:
Look good.

will-Do:
ACK

PR #649, opened 2015-09-18
libvirt_hugepage: Always cleanup and trivial bugs fix

will-Do:
1. If any exception raise during the test, the cleanup steps will not
execute, so move the tests steps into try.finally block to aviod the
problem.
2. If the test requirements are not met(multiple Vms), skip them asap.
3. unixbench test need 'patch' and 'perl' commands installed.
4. Remove two invalid unixbench related cases, as if nr_huagepageas is 0,
VM can not started with memoryBackging.
4. Format/typo fix.


will-Do:
@lmr 
Thanks, just update these two messages according to your comments.

waynesun09:
ACK

PR #651, opened 2015-09-30
graphics: Fix case for recent libvirt change

Hao-Liu:
1) Renamed some options in config files to avoid name collision.
2) Some negative cases became positive after qemu change.
    http://git.qemu.org/?p=qemu.git;a=commitdiff;h=cf7856ad
3) Add option negative_test to ensure all negative cases are really
    negative.
4) Message changed for port range checking.
5) VNC unix socket file location changed.
6) Some logic and logging tweaks.


will-Do:
ACK

waynesun09:
ACK and merging this now, thx

PR #654, opened 2015-10-13
use API function to check the result of libvirt command

PandaWei:
The function check_exit_status() used to check libvirt command
has been declared at virttest/utils_test/libvirt.py.
so it's better to use it and no need to re-write it respectively.


will-Do:
Yes, I wrote the helper func after wrote the testcase.

ACK

waynesun09:
ACK and merging now, thx.

PR #655, opened 2015-10-13
virttest: Fix skip message for TestNAError()

PandaWei:
In order to display message to users correctly,
The exception message has to use interpolation.


PandaWei:
trivial fix, merge now.

PR #656, opened 2015-10-15
virsh_migrate_multi_vms: Fix wrong params for multi_migration

PandaWei:
The value of simultaneous and jobabort should have been obtained
from cfg file. But they're all fixed to False.


will-Do:
ACK

will-Do:
Merge this trivial fix

PR #658, opened 2015-10-20
libvirt: update cases for avocado execution

will-Do:
1. Replace the default VM name as avocado using 'avocado-vt-vm1'.
2. Function utils_misc.find_command is deprecated, and which will not
raise ValueError when command not find, so replace it by the right function.


PandaWei:
ACK

waynesun09:
ACK

waynesun09:
Please on hold of merging this, as different group haven't fully migrated to avocado-vt yet, this will possibly fail testing for users still on virt-test. You could use the PR to run tests on avocado-vt for now.

will-Do:
Now our job already migrate to avocado, so let's merge this PR

PR #659, opened 2015-10-20
Add tests for scsi disk with ioeventfd option.

rbian:


will-Do:
ACK

waynesun09:
ACK and merging now, thx

PR #660, opened 2015-10-21
virt_sysprep: Add CmdError for attach_disk

PandaWei:
If the exit status of attach_disk is non-zero and
ignore_status=False, CmdError will be raised.
so catch it.


will-Do:
ACK

waynesun09:
ACK and merging now, thx

PR #661, opened 2015-10-30
virsh_hostname.py: hostname test fix for Avocado usage

svirt:
The exception handler is not necessary here while being used
with Avocado. In Virt-test it was working properly in any case,
while Avocado reports test failures which are related to extra
exception handlers

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
@@ -26,13 +26,12 @@ def run(test, params, env):
 
     # Run test case
     option = params.get("virsh_hostname_options")
-    try:
-        hostname_test = virsh.hostname(option,
-                                       ignore_status=False,
-                                       debug=True)
-        status = 0  # good
-    except error.CmdError:
-        status = 1  # bad
+    hostname_test = virsh.hostname(option,
+                                   ignore_status=True,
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
```diff
@@ -26,13 +26,12 @@ def run(test, params, env):
...
-                                       ignore_status=False,
-                                       debug=True)
-        status = 0  # good
-    except error.CmdError:
-        status = 1  # bad
+    hostname_test = virsh.hostname(option,
+                                   ignore_status=True,
+                                   debug=True)
+    status = 0
+    if hostname_test == '':
```
PandaWei: I'm OK for removing try-except.
But the check way of hostname_test is unreasonable.
Maybe you should check hostname_test.exit_status

svirt:
@PandaWei 
Hi, I agree with your point but unfortunately the hostname() function in virsh.py returns text. 

return command("hostname %s" % option, **dargs).stdout.strip()

So probably it's better for this test to check the hostname_test, rather than to change hostname() function ?

PandaWei:
Yes, 
I ignore the special return of hostname().
and No need to change the hostname().
I'm OK for this PR.

will-Do:
LGTM

PR #662, opened 2015-11-02
libvirt_hooks: Modify interface type before test.

rbian:
Network hook only for interface with network type, reset the domain
xml before the test.

### DIFF
--- a/libvirt/tests/src/libvirt_hooks.py
+++ b/libvirt/tests/src/libvirt_hooks.py
@@ -73,6 +73,7 @@ def run(test, params, env):
         prepare_hook_file(hook_script %
                           (vm_name, hook_log))
         vm.start()
+        vm.wait_for_login().close()
         try:
             hook_str = hook_para + " prepare begin -"
             assert check_hooks(hook_str)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/libvirt_hooks.py
```diff
@@ -255,10 +255,25 @@ def attach_hook():
...
+        vmxml.add_device(iface_xml)
+        vmxml.sync()
+
     def network_hook():
         """
         Check network hooks.
         """
+        # Set interface to use default network
+        edit_iface()
         net_name = "default"
```
will-Do: Maybe it's better to make the net_name configurable, then pass it to the edit func, what do u think?

rbian: Currently we always use 'default' network for testing. It's okay to make it configurable. Thanks.

will-Do:
It's better to add 'net_name' to cfg too, that's the reason we want it configurable. Thanks.

rbian:
Thanks @will-Do , updated the code.

will-Do:
ACK and merge it, thanks

PR #663, opened 2015-11-02
aexpect module import fix

svirt:
Now aexpect is not anymore part of virttest or avocado but is
installed as a separate package. Thus the remaining imports
should be corrected


ldoktor:
Yep, those are the only occurrences in the latest 57841c1d0712093a9acaea9d3a95b6e698baafef master.

Acked-by: Lukáš Doktor <ldoktor@redhat.com>

will-Do:
ACK and merge this PR

PR #664, opened 2015-11-04
Tests for attach/detach operation with option matrix.

rbian:


waynesun09:
saw lots of fail with change media matrix, please check the internal CI jobs for this.

libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface_matrix.py
```diff
@@ -0,0 +1,305 @@
+import logging
+import time
+from autotest.client.shared import error
+from autotest.client.shared import utils
```
waynesun09: new cases have been switch to avocado utils:
from avocado.utils import process

libvirt/tests/src/virsh_cmd/domain/virsh_update_device_matrix.py
```diff
@@ -0,0 +1,432 @@
+import os
+import time
+import shutil
+import logging
+from autotest.client.shared import error
+from autotest.client.shared import utils
```
waynesun09: also here

libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface_matrix.py
rbian: Thanks, I will update it.

waynesun09:
ACK
The matrix is huge and testing result on x86 CI is great, so I'm gonna merge this now, thanks.

PR #666, opened 2015-11-06
virsh_uri.py: uri test fix for Avocado usage

svirt:
The exception handler is not necessary here while being used
with Avocado. In Virt-test it was working properly in any case,
while Avocado reports test failures which are related to extra
exception handlers


will-Do:
ACK

will-Do:
So the real problem is described in https://github.com/autotest/tp-libvirt/pull/668#issuecomment-168594365, could you please update your patch, thanks.

will-Do:
A new https://github.com/autotest/tp-libvirt/pull/685 fix all this kind of problem, so close this PR.

svirt:
Thanks for the info

PR #668, opened 2015-11-11
virsh_net_create.py: net_create test fix for Avocado usage

svirt:
Now when ignore_status is True the result of net_create()
is determined by the exit status. In Virt-test it was working
properly in any case, while Avocado reports test failures
which are related to the different way of exception handling
in Avocado.


will-Do:
No, the root problem is that virsh moduel using avocado.utils.process to run command, so it command fail with ignore_status=False, it will raise process.CmdError instead of error.CmdError(which is come from autotest).
This problem is fixed by @rbian in #683, so close this one, thanks.

PR #671, opened 2015-11-30
virsh_attach_device.py: added correct tmp directory usage

svirt:
Now get_tmp_dir() is used to get the correct tmp directory of Avocado,
so the tests will be using it.


svirt:
strange failure... but seems it is not related with my changes....

lmr:
@svirt yes, it's not your fault. It's a new pylint that came out and is more picky. I've fixed the problem.

svirt:
**@lmr**, thanks a lot

PR #678, opened 2015-12-15
virsh_dump: Rename parameter to avoid collision

Hao-Liu:
It seems `timeout` parameter is already taken for avocado. Reusing of
this parameter lead this case occasionally fail if the duration is
longer than 5s.


will-Do:
ACK and merge

PR #679, opened 2015-12-15
virsh_event: Extend event_timeout to reduce failure

Hao-Liu:
Setting `event_timeout` to 30s might be not enough since this process
includes guest booting up, which is definitely possible for some guests.

Extend this timeout will greatly reduce failure rate.
Also added a hint for better troubleshooting if the issue happens again.


will-Do:
ACK

PR #683, opened 2015-12-28
virsh_net_create: Replace error.CmdError with process.CmdError.

rbian:


will-Do:
Yes, there are lots of cases which using the error module of autotest, we need fix them.
ACK and merge this trivial fix, thanks.

PR #684, opened 2016-01-07
virtual_disk_gluster: check mount point before umount

will-Do:
If case fail before mount operation, then the check step is necessary
to avoid umount fail.


waynesun09:
ACK

rbian:
ACK

PR #685, opened 2016-01-07
Catch process.CmdError instead of error.CmdError

will-Do:
Now we using process module from avocado to run commands
and raise exceptiones, so adapt tp-libvirt cases to it.


rbian:
ACK.

waynesun09:
ACK and merge now, thx.

PR #688, opened 2016-01-11
Using the right tmpdir

will-Do:
If the base dir of save file is not exist, virsh save will fail. And we
already have a function to get public tmpdir, so use it here.


waynesun09:
ACK

rbian:
ACK

PR #689, opened 2016-01-11
virsh_setmem: Fix the expected outside memory value

will-Do:
Commit 15526941e is not a complete fix for expected memory value.
For expected memory outside the VM, if memory not change, it should be
the original outside memory, if memory changed, which should be the
value given by virsh command.


waynesun09:
ACK

PandaWei:
Looks  Good

PR #694, opened 2016-01-19
virsh.cpu_compare: Prepare cpu related elements before testing

will-Do:
The cpu or cpu model element may not present in VM xml, but both of them
are necessary for guest-cpu related testing.
And avoid hardcode guest cpu model as 'core2duo' is only belong to Intel
platform.


waynesun09:
ACK and merge now, thx

PR #695, opened 2016-01-20
migrate_vm: Do fixes for some test cases

dzhengfy:
- '--downtime' option is invalid now for 'virsh migrate', so remove it
  in migrate_vm.cfg.

- Change grep parameter.
  When executing 'virsh migrate-setmaxdowntime 3 1000', below message
  will be written in libvirtd.log.
  qemuMonitorSend:1033 : QEMU_MONITOR_SEND_MSG:.....
  msg={"execute":"migrate_set_downtime","arguments":{"value":1.000000}
  So change 'max_down_time' to '1.000' from '1000' to grep
  "migrate_set_downtime.*1.000}". The down time now is in seconds, not
  millseconds.

- It is incorrect to use 'disk size' to create an image using 'qemu-img
  create'. 'virtual size' should be used which can be gotten from
  'qemu-img info'.

- Doing dompmwakeup quickly after doing dompmsuspend without any delay
  will cause the guest can not turn to running state and always stays in
  pmsuspended state. So it is necessary to sleep some seconds between
  the two commands.

- The functions in ssh_key are move from autotest to avocado-vt already.
  So it is necessary to change the import.

- The 'pool_name' should be set in function
  setup_or_cleanup_gluster(xxxx, pool_name), otherwise the /tmp
  directory will be cleaned up totally.

- When logging on remote machine with ssh, it needs enough time to wait
  for the prompt. The default timeout value is too little.

- '_' is invalid char for targetcli command which will cause target
  creation failure. So change it to 'emulated-iscsi'.

- It needs longer period to write data with cache 'writethrough' than
  cache 'None'. So it is neccessary to set bigger timeout value to check
  the migration output.

- 'swapon -s' command output should not be taken as 'if' check.
  Only command status's checking is correct.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_vm.cfg
+++ b/libvirt/tests/cfg/migration/migrate_vm.cfg
@@ -125,7 +125,7 @@
                                         - iozone_with_set_max_downtime:
                                             stress_type = "iozone_in_vms"
                                             stress_args = "-a"
-                                            max_down_time = 1000
+                                            max_down_time = 1.000
                                             set_migration_speed = 5
                                             check_job_info_time = 60
                                             run_migrate_cmd_in_front = "no"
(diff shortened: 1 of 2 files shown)


will-Do:
For this kind of fix, what we want most in the commit message is WHY you did that. I suggest you read some docs about that, such as http://chris.beams.io/posts/git-commit/

dzhengfy:
@will-Do , just update the commit message. Please review again. Thanks.

will-Do:
The commit message is much clear now, and the codes also look good. So ACK this patch.

PR #697, opened 2016-01-25
v2v: fix a wrong func name

will-Do:


waynesun09:
ACK and merge now, thx

PR #698, opened 2016-01-25
virsh.emulatorpin: Fix the range type cpu list testing

will-Do:
By default, emulator is pined to all cpus, and element
'cputune/emulatorpin' may not exist in VM's XML. And libvirt
will do nothing if pin emulator to the same cpus, that means
VM's XML still have that element. So for testing, we should
avoid that value(0-$cpu_max).


waynesun09:
ACK and merge now, thx

PR #699, opened 2016-02-15
migrate_vm: Changes for two parameters

dzhengfy:
- Some migration cases need more than 30 seconds to finish the
  migration, so a default migration_timeout with longer period
  is added.
- Remove vcpu_cpuset parameter.
  Local and remote hosts may have different cpu processor ids.
  The guest fails to start with specified cpuset statically
  which does't exist on that host. So the new implementation is
  to select a processor id automatically which exists on both
  local and remote host and then set it to the guest.


waynesun09:
ACK

PR #700, opened 2016-02-15
virsh_domstate: Bugs fix and case enhancement

will-Do:
1. There is a middle state 'crashed (panicked)' for 'coredump-destroy'
and 'coredump-restart', so wait 60 seconds for final state.
2. Create the dump path to avoid domain dump fail.
3. Use exist module to config qemu.conf.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
@@ -14,12 +14,10 @@ from virttest import libvirt_vm
 from virttest import remote
 from virttest import virsh
 from virttest import utils_libvirtd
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
```diff
@@ -14,13 +14,12 @@
 from virttest import remote
 from virttest import virsh
 from virttest import utils_libvirtd
+from virttest import utils_misc
+from virttest import utils_config
 from virttest.utils_misc import kill_process_by_pattern
```
waynesun09: could remove this and use utils_misc.kill_process_by_pattern in code

waynesun09:
ACK and merge now, thx

PR #705, opened 2016-02-23
Fix 'graphic_act' is not defined problem.

rbian:
The defination of 'graphic_act' was removed by last commit, which caused
the exception of 'NameError'.


waynesun09:
ACK and merge this now, thx.

PR #709, opened 2016-02-29
volume_application: Avoid using underscore in iqn

Hao-Liu:
Underscore are not accepted in iqn. Use hyphen instead.


waynesun09:
ACK and merge now, thx

PR #710, opened 2016-02-29
Fix virsh pool info check deviation problem

waynesun09:
The pool info with 'Capacity', 'Allocation' and 'Available' are
caculated from bytes to GiB, deviation in 1% will happen, so fix
this problem in check and deviation in 1% will not fail testing.


Hao-Liu:
ACK

will-Do:
ACK

PR #711, opened 2016-02-29
freepages: Don't append None into cellno list

Hao-Liu:
If cellno is not set, don't append None to cellno list. This will cause
error when doing final comparison. Just do nothing instead.


waynesun09:
ACK and merge now, thx

PR #715, opened 2016-03-02
setmem: fix dmidecode not supported problem

waynesun09:
dmidecode is not supported in ppc, and also it could be not present
in some vm, that'll cause test fail. Fix it by use max_mem from
vm xml to get physical memory, as the value is needed only at the
beginning of the test, the xml memory setting refected the physical
memory already.


Hao-Liu:
ACK

will-Do:
LGTM, will merge this fix.

PR #716, opened 2016-03-02
start: Long names must start correctly

nertpinx:
Until now the test assumed that VMs with long names should not be
startable.  There is no reason for that as that was just a temporary
bug.  And because that is now fixed upstream (with varios commits) this
test now fails due to that invalid presumption.

Signed-off-by: Martin Kletzander <mkletzan@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_start.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_start.cfg
@@ -17,7 +17,7 @@
                     vs_new_vm_name = "12345678"
                 - vm_named_long:
                     vs_pre_operation = "rename"
-                    vs_new_vm_name = "1234567890123456789012345678901234567890123456789012345"
+                    vs_new_vm_name = "12345678901234567890123456789012345678901234567890123456789012345678901234567890"
                 - vm_name_start_with_special:
                     vs_pre_operation = "rename"
                     vs_new_vm_name = "-vm1"
@@ -45,9 +45,6 @@
                     vm_ref = ""
                 - dom_name_not_found:
                     vs_pre_operation = "undefine"
-                - vm_named_too_long:
-                    vs_pre_operation = "rename"
-                    vs_new_vm_name = "12345678901234567890123456789012345678901234567890123456789012345678901234567890"
                 - vm_running:
                     start_vm = yes
                 - vm_paused:


Hao-Liu:
Hi @nertpinx, Thank you very much for you fix.

This fix looks good for upstream libvirt, but due to these tests are also running on various versions of downstream libvirt, we must make sure they work well on those. 

Add a libvirt version check in `libvirt/tests/src/virsh_cmd/domain/virsh_start.py` is what we are using under similar situation in many other tests. We are using a helper function `libvirt_version.version_compare` to check libvirt version. You can reference usage of this function in other libvirt test cases.

Instead of hardcoded `vs_new_vm_name` in `cfg` file, you might also need to pass a placeholder like `LONG_NAME` or `TOO_LONG_NAME` into python script and generate different vm name according to current libvirt version.

And for removing too long name test case, do libvirt accept infinite-length domain name? If not, we still need a test case to confirm failure of starting a too long domain and prevent potentially bad thing from happening.

If there's anything we can help, please let us know.

nertpinx:
Well, if that fails on some versions, then those versions are still buggy. Adding an exception kind of defeats the whole purpose of the testing, doesn't it? What's even worse is deciding this based on a libvirt version (something can be back-ported to maintenance branch, downstream packages might be different for various distros). Instead, I would imagine having a configuration that says which tests are OK to fail and thos tests will either be skipped or their error will be ignored. Such configuration can then be based on whatever any user is running the test upon. And latest upstream would match latest upstream without dragging in additional unrelated configuration.
And yes, libvirt supports vm names of arbitrary length as it supported way back as well. The limitations were not even in libvirt itself, but caused due to incompatibility of other system parts. To name one, for example, lately it was the systemd. That's another reason for not configuring exceptions based on libvirt's version because older libvirt might fail with systemd but will work without it.

will-Do:
So what we got here are:
1. The original negative case should be positive for the latest upstream libvirt, that's what the patch has done, no problem. And for downstream test, let it fail as the bug not verify yet.
2. From the test coverage purpose, we want to have a negative case to test a long VM name(long enough to trigger an start error, no matter systemd or what deny this action), so you can include it in this patch or we can propose a new one.
@nertpinx @Hao-Liu What do you think?

nertpinx:
That would mean creating a test that needs to be pretty variable. I mean for example on my system, that would be limited by the path length; with my configuration I need the name to have about 250 characters for some path to exceed the limit. However be it on different system or with slightly different configuration that limit can be very different. And the errors might or might not be correct. Unless there is some exact case that needs to be tested, creating a test that fails just for the sake of failing is pointless IMO.

will-Do:
@nertpinx Thanks for your reply.
I didn't consider well about that, as I learned from systemd, both the length(64 ch max) and characters(a-zA-Z0-9_.) are limited. And there are several bugs(maybe closed as NOTABUG) about this kind of scenarios, so I think a regression test case is good for us. But as your explanation, the length is variable, we will reconsider about this issue.

nertpinx:
Yes, systemd poses some limitations, but since that would break libvirt, we went around those limitations on purpose.

Hao-Liu:
I totally agree that buggy versions don't deserve a pass. But I still think it'll be risky that upper bound of name length not checked. There will always be a limit, even if it is not bound by libvirt itself. If it's not tested, and something happens with this limitation, we don't even be notice. I'll merge this PR firstly, and  remind myself to figure a way to check the variable limit.
Thank's very much for the clarification @nertpinx @will-Do

PR #718, opened 2016-03-03
virsh.undefine: Avoid possibility of removing guest image

Hao-Liu:
When undefine with option of removing storage, if current guest image
is managed by libvirt, the guest image will be removed as well.

This will cause following test cases using guest image fail.

This fix try to backup all current guest images before testing, and move
them back when test finished to avoid this problem.

Also a typo in configuration file is fixed along with some code cleaned
up.


will-Do:
ACK

waynesun09:
ACK and merge now, thx

PR #722, opened 2016-03-09
nodeinfo: Fix nodeinfo check problem

waynesun09:
Update check on CPU frequency, CPU(s) and CPU Sockets.

The CPU frequency in ppc is under clock, update the check for it.

The CPU(s) in nodeinfo is Online threads in the system, check it
with all online cpu in sysfs.

The CPU Sockets in nodeinfo is Total sockets in each node, check
it with sockets number from one node.


will-Do:
ACK

Hao-Liu:
ACK and merge this. Thanks @waynesun09 @will-Do

PR #723, opened 2016-03-09
setvcpus: Fix vm xml update problem

waynesun09:
The set_vm_vcpus did not take effect since the vmxml obj not
updated, fix it by reset it.

The feature part may not exist in xml, e.g. on ppc testing, so
caught the exception and make test ongoing.


Hao-Liu:
ACK

PR #724, opened 2016-03-10
virsh.domxml_to_native.expected_option dir name filter fix

svirt:
domxml_to_native fails during virsh.domxml_to_native.expected_option
It's due to libvirt commits a89f05ba8df095875f5ec8a9065a585af63a010b
and f1f68ca33433825ce0deed2d96f1990200bc6618.
A new way of dir/path is introduced , so the tests must be compatible
with it.
Now the "socket,id=" param is added to filtlist() function to be
filtered out.


waynesun09:
ACK
Test pass on downstream x86 and ppc also.

Hao-Liu:
ACK and Merge this. Thanks @svirt @waynesun09 .

PR #725, opened 2016-03-11
libvirt_vcpu_plug_unplug: Don't apply s3/s4 config to ppc

waynesun09:
The s3/s4 is not valid on ppc, so if host with power cpu, did not
update vm with the config.

### DIFF
--- a/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
+++ b/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
@@ -5,6 +5,8 @@ import logging
 from autotest.client import utils
 from autotest.client.shared import error
 
+from avocado.utils import cpu as cpu_util
+
 from virttest import virsh
 from virttest import data_dir
 from virttest import utils_misc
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -30,7 +32,7 @@ def check_vcpu_number(vm, expect_vcpu_num, expect_vcpupin, setvcpu_option=""):
     :param setvcpu_option: Option for virsh setvcpus command
     """
     logging.debug("Expect vcpu number: %s", expect_vcpu_num)
-    #if setvcpu_option != "--guest":
+    # if setvcpu_option != "--guest":
```
Hao-Liu: I think this is just a line of abandoned code, not a part of the comment. So we'd better remove this.

waynesun09: Ohh, got it, update coming

Hao-Liu:
ACK

PR #728, opened 2016-03-15
remote_access: Fix ipv6 test configuration problem

dzhengfy:
For no ipv6 configuration test and ipv6 address unreachable test,
currently the test configuration is set with actual ipv6 address
of the target host which is not to be expected. The fix is to set
an invalid value to the configuration for target host.


waynesun09:
ack and merge now, thx

PR #730, opened 2016-03-21
Fixes for remote job

dzhengfy:
1. Fix incorrectly matching error messages
The original pattern to match error messages ('forbidden: xxx') only
applies to 7.3, not 6.8. In order to apply to both 6.8 and 7.3, the
pattern to match is updated with the common part.

2. Fix tls_allowed_dn_list for RHEL6

The dn list order in tls_allowed_dn_list is opposite between RHEL6 and
RHEL7. Current order for 'tls_allowed_dn_list' only applies to RHEL7.
The fix is to make the test correctly work on RHEL6 via reversing the
order of dn when on RHEL6.

### DIFF
--- a/libvirt/tests/src/remote_access/remote_access.py
+++ b/libvirt/tests/src/remote_access/remote_access.py
@@ -17,6 +17,8 @@ from virttest.utils_net import check_listening_port_remote_by_service
 from virttest.utils_test.libvirt import remotely_control_libvirtd
 from virttest.utils_test.libvirt import connect_libvirtd
 
+from provider import libvirt_version
+
 
 def remote_access(params):
     """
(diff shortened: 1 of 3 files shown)


waynesun09:
ACK

libvirt/tests/src/remote_access/remote_access.py
```diff
@@ -126,7 +128,15 @@ def run(test, params, env):
...
     allowed_dn_list = params.get("tls_allowed_dn_list")
     if allowed_dn_list:
-        test_dict['tls_allowed_dn_list'] = eval(allowed_dn_list)
+        if not libvirt_version.version_compare(1, 0, 0):
+            # Reverse the order in the dn list to workaround the
+            # feature changes between RHEL 6 and RHEL 7
+            allowed_dn_list = eval(allowed_dn_list)
+            dn_list = allowed_dn_list[0].split(",")
+            allowed_dn_list[0] = dn_list[1] + "," + dn_list[0]
+            test_dict['tls_allowed_dn_list'] = allowed_dn_list
```
will-Do: If you want to reverse a list, list.reverse() can do the right thing.
And for this case, it seems `tls_allowed_dn_list` only has one element, so what use a list? I'm not familiar with this case, please correct me.

dzhengfy:
@will-Do , update and please review again.

will-Do:
ACK

PR #731, opened 2016-03-22
migrate_vm: Do umount to ensure nfs dir work well

dzhengfy:
The mount dir on remote host sometimes is in mount state unexpectly
    maybe due to the failure of previous test. That will cause mount
    operation fail in current test. The fix is to check and 'umount' it
    if mounted already on remote host after 'mkdir' to ensure the mount
    dir is ready to mount again.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -659,6 +659,7 @@ def run(test, params, env):
     driver = test_dict.get("test_driver", "qemu")
     uri_path = test_dict.get("uri_path", "/system")
     nfs_mount_dir = test_dict.get("nfs_mount_dir", "/var/lib/libvirt/migrate")
+    nfs_mount_src = test_dict.get("nfs_mount_src", "/usr/share/avocado/data/avocado-vt/images")
     setup_ssh = test_dict.get("setup_ssh", "yes")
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -930,6 +930,9 @@ def run(test, params, env):
             if status:
                 raise error.TestFail("Failed to run '%s' on the remote: %s"
                                      % (cmd, output))
+            cmd = "umount -l %s" % nfs_mount_dir
+            status, output = run_remote_cmd(cmd, server_ip, server_user, server_pwd)
```
waynesun09: umount the dir right after it's created on remote host?

will-Do:
Instead of doing `umount` in a rude way, I think you should point out the root problem which cause the failure.
And before(or after) create the mount dir, you can check the dir state.

dzhengfy:
@will-Do @waynesun09 , update and please review again.

waynesun09:
@dzhengfy one thing still not clear is why the nfs mount left on remote host not cleared, so if it's introduced in this test, you should consider add a clean up in finally for each test, rather then check it in the following test.

waynesun09:
since clean up did exist in finally and this is a bypass work on some uncertain failure in our CI env, I'm ok with this to keep running continue while digging more. So merge now, thx.

PR #732, opened 2016-03-22
libvirt_qemu_cmdline: set hard limit when enable mlock

waynesun09:
memtune hard limit is required when enable mlock, or else test
could fail.


will-Do:
ACK

PR #734, opened 2016-03-25
Wait a while after attach/detach operation for xml to be synchronized.

rbian:


will-Do:

PR #735, opened 2016-03-28
svirt_install: fix virt-install fail on ppc

waynesun09:
ppc vm did not support spice, and without given graphics
virt-install by default use vmport which also not supported
and fail the install.

To avoid this failure, specify graphics as vnc which is
supported by both ppc and x86.


Hao-Liu:
ACK

PR #737, opened 2016-03-31
nodeinfo: Fix check with online and total cpus

waynesun09:
As on RHEL6 the online file could missing under cpu0, and offline
it is not allowed on cpu0, the calculation is wrong then. Fix it
by check the online path under cpu0 and add 1 in both total and
online if not found.


Hao-Liu:
ACK and merge this

PR #738, opened 2016-03-31
Fix partition couldn't be found problem.

rbian:
Run 'partprobe' command after adding new partitions for the change take effect.


will-Do:
ACK

PR #739, opened 2016-04-01
Remove non-option-related cases

chloerh:
Left only v2v options-related cases, will move those cases to
athother case file.


will-Do:
Nice job.

PR #740, opened 2016-04-05
Fix wait_for_serial_login arguments problem

waynesun09:
- Batch update for wait_for_serial_login failure
- iface_hotplug: hotplug after vm fully started


will-Do:
LGTM

PR #742, opened 2016-04-07
Update use main_vm in tests

waynesun09:
Only one vm needed in these tests, so use main_vm for get
vm.


will-Do:
ACK

PR #744, opened 2016-04-08
domif_setlink_getlink: Avoid ShellStatusError for session check

waynesun09:
Some tests failed with ShellStatusError when check with session,
the command excuted succeed but following 'echo $?' command check
status failed.

To avoid this failure, catch the exception and put the func in
wait_for with multiple try.


Hao-Liu:
LGTM. Ack

will-Do:
ACK

PR #746, opened 2016-04-08
domif_setlink_getlink: Fix cmd check with serial login

waynesun09:
Not only could exit status check fail in serial login, also could
get ShellTimeoutError after can't get expected prompt but cmd did
succeed.
Fix this by check the cmd output with regex to confirm the command
succeed and do not fail the testing.


will-Do:
ACK

PR #747, opened 2016-04-11
migrate_vm: use ssh_timeout to avoid of timeout

dzhengfy:
NFSClient can read the parameter 'ssh_timeout' to setup the ssh
connection. So the ssh_timeout value is provided here to decrease the
time out possibility.


will-Do:

PR #749, opened 2016-04-11
case vm_shutoff fail fix

svirt:
there was a patch which added vm.wait_for_login().close(), but there
is a missing check of vm start status which causes a failure

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface.py
@@ -136,7 +136,7 @@ def run(test, params, env):
     if start_vm == "no":
         if vm.is_alive():
             vm.destroy()
-    else:
+    elif start_vm == "yes":
         vm.wait_for_login().close()
 
     # Test both detach and attach, So collect info


libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface.py
```diff
@@ -136,7 +136,7 @@ def run(test, params, env):
     if start_vm == "no":
         if vm.is_alive():
             vm.destroy()
-    else:
+    elif start_vm == "yes":
```
will-Do: Generally, there are only 'yes'(default) and 'no' filled for ``` start_vm```, I'm curious why you need this condition here, can you explain this in detail?
BTW, if you worry about this value, it's better to fix it at line 98, like:
```
    start_vm = "yes" == params.get("start_vm", "yes")
````

svirt:
Hi,
when this condition is not checked properly then I constantly meet an error due to vm.wait_for_login().close() saying that VMDeadError: **VM is dead    reason: Domain avocado-vt-vm2 is inactive    detail: 'shut off'**   and 
**ERROR type_specific.io-github-autotest-libvirt.virsh.attach_detach_interface.error_test.vm_shutoff -> VMDeadError: VM is dead    reason: Domain avocado-vt-vm2 is inactive    detail: 'shut off'**

waynesun09:
@svirt The error you met is fixed in privious config 66d20da, so it should work now.

svirt:
ah ok, i will check with that config then and will let you know. thanks

svirt:
@waynesun09 
I checked and that patch works fine in my case, so I will close the PR.
Thanks

PR #750, opened 2016-04-11
fix for acquiring block device names list (discovered on ppc64 case)

svirt:
checking condition for virtual block devices and when SERIAL
is not defined. Particular case is on ppc64.
In case of empty SERIAL the block device name is formed in a simple way.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_list.py
+++ b/libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_list.py
@@ -48,7 +48,8 @@ def get_storage_devices():
                 timeout=5, ignore_status=True).stdout
             # Only disk devices are list, not partition
             dev_type = re.search(r'(?<=E: DEVTYPE=)\S*', info)
-            if dev_type:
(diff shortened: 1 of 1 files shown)


waynesun09:
ACK

libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_list.py
```diff
@@ -58,6 +59,9 @@ def get_storage_devices():
                         dev_name = re.sub(
                             r'\W', '_', 'block_%s_%s' % (device, serial))
                         devices.append(dev_name)
+                    else:
+                        dev_name = 'block_' + device.replace(':', '_')
+                        devices.append(dev_name)
```
will-Do: @waynesun09 
Is this ppc specific? and it will not affect x86 tests, right?

waynesun09: @will-Do test on x86 all pass

svirt:
yes, it's ppc specific

will-Do:
Okay, then, no more comment.

PR #751, opened 2016-04-12
V2v update

will-Do:
1. v2v: Rewrite VM check module
    
    Running vm_check as a sepeate case after convert VM is inappropriate, it
    may cause extra work for setup/cleanup env. So rewrite it as a utils
    module which can be used in the convert VM case.

2. v2v: Update test case to make it more clear


rbian:
Code style is clear, you'd better post your test results.

waynesun09:
ACK and merge now, we need put this in our CI jobs

PR #752, opened 2016-04-12
blockpull: Reset expect exit status for timeout option

will-Do:
If pull job aborted as timeout, the exit status is different on RHEL6
and RHEL7


waynesun09:
ACK

PR #753, opened 2016-04-15
Replace code to capture avocado process.CmdError

Hao-Liu:
virsh_instance.connect now raise a process.CmdError when fail.


waynesun09:
ACK

PR #754, opened 2016-04-15
virsh.domblkstat: It's acceptable not specifying dev

Hao-Liu:
After bug https://bugzilla.redhat.com/show_bug.cgi?id=1142636 is fixed.
virsh domblkstat will generate summary statistics of domblkstat for QEMU
domain instead of failing.


will-Do:
Yes, this's should be positive test

PR #755, opened 2016-04-15
svirt_install: add --video option for install

waynesun09:
Add --video vga for install to avoid the primary video type
problem.


will-Do:
ACK

will-Do:
Merge this trivial fix

PR #756, opened 2016-04-15
remote_with_unix: Correct wrong regular expression

dzhengfy:
Existing regular expression can not match both 'forbidden for read only
access' and 'forbidden: read only access'. The fix is to correct it.


waynesun09:
ACK

PR #757, opened 2016-04-15
virsh.blockcopy: Bug fix and case update

will-Do:
In some cases, the emulated iscsi target not cleanup, which will cause
the following case fail, so using a list to recorde all emulated iscsi
and cleanup them in the end.
Remove autotest dependency.
Skip or fail case as early as possible.
And other code cleanup update.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -5,8 +5,8 @@ import re
 
 import aexpect
 
-from autotest.client import utils
-from autotest.client.shared import error
+from avocado.utils import process
+from avocado.core import exceptions
 
 from virttest import utils_libvirtd
 from virttest import utils_config
@@ -32,7 +32,7 @@ class JobTimeout(Exception):
         self.timeout = timeout
 
     def __str__(self):
-        return "Block job timeout in %s seconds." % self.timeout
+        return "Block job timeout in %s seconds" % self.timeout
 
 
 def check_xml(vm_name, target, dest_path, blk_options):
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -230,12 +223,31 @@ def run(test, params, env):
...
+    # Skip/Fail early
+    if with_blockdev and not libvirt_version.version_compare(1, 2, 13):
+        raise exceptions.TestSkipError("--blockdev option not supported in "
+                                       "current version")
+    if not target:
+        raise exceptions.TestSkipError("Require target disk to copy")
+    if setup_libvirt_polkit and not libvirt_version.version_compare(1, 1, 1):
+        raise exceptions.TestSkipError("API acl test not supported in current"
+                                       " libvirt version")
+    if copy_to_nfs:
```
waynesun09: destination to nfs cases are working fine and not related to the bug, why skip all dest_nfs cases?

libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -423,10 +427,11 @@ def _make_snapshot():
             vm_name, options, debug=True)
 
         if snapshot_result.exit_status != 0:
-            raise error.TestFail(snapshot_result.stderr)
+            raise (snapshot_result.stderr)
```
waynesun09: exceptions.TestFail

libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -438,18 +443,22 @@ def _make_snapshot():
...
                                                        emulated_image=blkdev_n)
+                emulated_iscsi.append(blkdev_n)
             else:
                 if copy_to_nfs:
                     tmp_dir = "%s/%s" % (tmp_dir, mnt_path_name)
                 dest_path = os.path.join(tmp_dir, tmp_file)
 
         # Domain disk replacement with desire type
         if replace_vm_disk:
+            # Calling'set_vm_disk' is bad idea as it left lots of cleanup job
```
waynesun09: s/Calling'set_vm_disk'/Calling 'set_vm_disk'/

waynesun09:
ACK

PR #762, opened 2016-04-21
virsh_blockcopy: Fix a code bug.

rbian:


will-Do:
Yes, it's a bug I just fixed in my local repo. Actually, the last patch for blockcopy is part of fix for all the issues, and I'm still working on that. Anyway, we can get this fix first, thanks.

PR #681, opened 2015-12-21
virsh_migrate: update basic migration test

dzhengfy:
The current existing migration cases can't be ran automitically, you need manually setup sharing storage and enable virt_use_nfs seboolean if you're using a NFS storage to share VM image, in addition, if the migration URI transport is SSH, you also need to configure SSH passwordless manually, this patch is used for solving these question and let tester run migartion testing automitically with simple test configuration. Meanwhile, to add new test scenarios for migration testing.

  1. Automitically setup NFS sharing ENV and enable virt_use_nfs seboolean

  2. Delete complex testing scenarios, such as ping-pong migration, it needs we setup passwordless access
    from local/remote to remote/local firstly, we may put these scenarios into ping-pong migration testing.

  3. Add some new testing scenarios to cover migration options or options combination, such as "--graphicsuri", "--live --suspend --undefinesource",.

  4. Enhance VM network connectivity check after migariton

    use local URI qemu:///system to instead of qemu+ssh://${migrate_source_host}/system, it's not necessary to use qemu+ssh://${migrate_source_host}/system URI for the local connecting, if you do this, you also need to setup passwordless access local URI firstly.

  5. Add parameters setting for migration test

  6. Change the nfs env setting order

 7.Change to use some neccesary parameters in avocado/avocado-vt/shared/cfg/base.cfg.
 8.Remove the hardcoded parameter for shared storage
9.Remove hardcoded command and use process instead of subprocess

PR #763, opened 2016-04-21
Fixup for virsh.migrate

balamuruhans:
Changed test case variant names for virsh migrate as when virsh_migrate_options="", it performs online migration.


dzhengfy:
ACK

waynesun09:
Thank and merge now, @balamuruhans you might wanna check PR #681 also as a lot of change in it.

balamuruhans:
@waynesun09 sure, will check it out

PR #764, opened 2016-04-24
daemon.kill_qemu: Enable sleep to memory before pmsuspend

Hao-Liu:
Pmsuspend will fail if VM does not have S3 enabled. So make sure that
the VM have this been set.


waynesun09:
ACK

will-Do:
ACK

PR #767, opened 2016-04-25
virsh.blkdeviotune: Change boundary according to fixed bug

Hao-Liu:
According to bug: https://bugzilla.redhat.com/show_bug.cgi?id=1329041
Current blkdeviotune should be 1e15.


waynesun09:
ACK, since it fails now and the fix will be in 1.3.4, current 1.3.3 testing will keep failing as the incremental problem so merge it seems fine also.

will-Do:
Does this change will affect RHEL6?

Hao-Liu:
@will-Do . Yes, the negative upper boundary check will fail on RHEL6 since it's still using 9223372036854775808

PR #768, opened 2016-04-26
virsh.domstate: Fix shadowed variable and remove autotest dependency

Hao-Liu:
Parameter `libvirtd` is shadowed by a following assignment of libvirtd
service instance. Thus later comparison with `"off"` will fail
definitely.

This patch also removed autotest dependency and fixed some style issues.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
@@ -6,8 +6,7 @@ import logging
 from aexpect import ShellTimeoutError
 from aexpect import ShellProcessTerminatedError
 
-from autotest.client.shared import error
-
+from avocado.core import exceptions
 from avocado.utils import process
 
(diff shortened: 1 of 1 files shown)


waynesun09:
ACK

libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
```diff
@@ -177,22 +177,25 @@ def run(test, params, env):
...
                 session.close()
-        except process.CmdError, e:
-            raise error.TestError("Guest prepare action error: %s" % e)
+        except process.CmdError, detail:
+            raise exceptions.TestError(
+                "Guest prepare action error: %s" % detail)
 
-        if libvirtd == "off":
+        if libvirtd_state == "off":
             libvirtd_service.stop()
```
will-Do: Just check the case, ```libvirtd_service``` and ```libvirtd``` should be duplicate definition

PR #769, opened 2016-04-26
conf_file.qemu_conf.auto_dump: Use shutil to remove directory

Hao-Liu:
Dump path is a folder and need to be removed by shutil.rmtree


waynesun09:
ack and merge now, thx

PR #770, opened 2016-04-26
conf_file.qemu_conf.auto_dump  & managedsave

crazyAxe:
1.add a command to stop kdump service on rhel7's guest
2.set vcpus for guest managedsave

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -404,7 +404,7 @@ def run(test, params, env):
                                                        is_login=True,
                                                        image_size="1G",
                                                        emulated_image=back_n)
-                emulated_iscsi.append(blkdev_n)
+                emulated_iscsi.append(back_n)
                 cmd = "qemu-img create -f qcow2 %s 1G" % back_path
                 process.run(cmd, shell=True)
                 new_attrs.update({'dev': back_path})


libvirt/tests/src/conf_file/qemu_conf/auto_dump.py
```diff
@@ -79,7 +79,10 @@ def run(test, params, env):
 
         session = vm.wait_for_login()
         # Stop kdump in the guest
+        # for rhel6
         session.cmd("service kdump stop", ignore_all_errors=True)
+        # for rhel7 
+        session.cmd("systemctl stop kdump", ignore_all_errors=True)
```
waynesun09: as systemd have backward support for service command, so service will work on both 6&7, you don't need adding this

libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -289,11 +289,11 @@ def build_vm_xml(vm_name, **dargs):
                 if cpu_topology:
                     cpuxml.topology = cpu_topology
                 vmxml.cpu = cpuxml
+                vmxml['vcpu'] = 16
```
waynesun09: why hard code set vcpu number to 16?

waynesun09:
You have 2 commit for 2 cases fix, you could raise 2 PRs. For fix on same problem you could squash 1 commit.

waynesun09:
and in one commit you sign-off not right, also you need fix Travis CI failure

crazyAxe:
@waynesun09  This is my first try to commit something,  apparently I messed  up. Sorry for that  and thanks for your advice. i will be more cautious.  About the fixes, i will discuss them with my mentor to catch the beautiful code style. Thanks anyway.

Hao-Liu:
Hi @crazyAxe, 

Thank you for your contribution to this project.

You don't need to be sorry at all. Actually we are very glad that new guys like you are interested in this project and taking your first step to contribute to it. Don't be frustrated if you are requested to give a reason or change your code. Seldom people don't get it right for their first attempt. BTW, don't be too cautious, we like bold ideas, although you do need to think twice before leap.

Thanks again for your supporting and looking forward to your following efforts.

PR #771, opened 2016-04-27
managedsave: add expected vcpus for cpu_mode test

crazyAxe:
the amount of vcpus should be equal or more than (sockets * cores * threads)

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
@@ -289,7 +289,7 @@ def run(test, params, env):
                 if cpu_topology:
                     cpuxml.topology = cpu_topology
                 vmxml.cpu = cpuxml
-                vmxml["vcpu"] = int(params.get("vcpus_nums"))
+                vmxml.vcpu = int(params.get("vcpus_nums"))
             if dargs.get("sec_driver"):
                 seclabel_dict = {"type": "dynamic", "model": "selinux",
                                  "relabel": "yes"}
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -289,6 +289,7 @@ def build_vm_xml(vm_name, **dargs):
                 if cpu_topology:
                     cpuxml.topology = cpu_topology
                 vmxml.cpu = cpuxml
+                vmxml["vcpu"] = int(params.get("vcpus_nums"))
```
rbian: Why not use vmxml.vcpu to set the value for it to keep consistent with other code?

crazyAxe: 'vmxml.vcpu'  is indeed better. Actually i just found the vmxml["vcpu"] in function set_vm_vcpus to set the value ,  so used it here. I'll  improve it, thank.

waynesun09:
@crazyAxe You need clean up this PR, better solution is use git branches, keep your fix in develop branch rather than master. Then push from devel branch and request merge.
Also you could consider add your sign-off email address associated to your github accound, then github could identify your commit.

waynesun09:
And you could take a look of the doc:
http://autotest.readthedocs.io/en/latest/main/developer/SubmissionChecklist.html

waynesun09:
@crazyAxe paste the wrong doc link in last comment, update it to:
https://www.redhat.com/archives/virt-test-devel/2013-April/msg00190.html

the mail is quite a long time ago, and there is a lot of git/github howto, you could take this one is for reference.

PR #772, opened 2016-05-04
graphics: fix a problem that guest can't find the defined network

crazyAxe:
the  name of listening network in vmxml should be same as  defined network

### DIFF
--- a/libvirt/tests/src/graphics/graphics_functional.py
+++ b/libvirt/tests/src/graphics/graphics_functional.py
@@ -903,9 +903,10 @@ def setup_networks(params):
     # Setup networks
     for net_type in networks:
         if net_type == 'vnet':
+            net_name = 'virt-test-%s' % net_type
             address = params.get('vnet_address', '192.168.123.1')
             networks[net_type] = LibvirtNetwork(
-                net_type, address=address, persistent=True)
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -839,8 +839,12 @@ def generate_spice_graphic_xml(params, expected_result):
...
 
     if listen_type == 'network':
-        net_type = params.get("spice_network_type", "vnet")
-        listen = {'type': 'network', 'network': 'virt-test-%s' % net_type}
+        net_name = params.get("vnet_name")
+        if net_name:
+            listen = {'type': 'network', 'network': net_name}
+        else:
+            net_type = params.get("spice_network_type", "vnet")
+            listen = {'type': 'network', 'network': 'virt-test-%s' % net_type}
```
will-Do: Good catch :+1: 
As you already define the net_name in cfg, it's unnecessary to do this judgement. And you can set a default value when get 'vnet_name'(if you want to make it more safe, just like the 'if...else' block did):
```net_name = params.get("vnet_name", 'avocado-vt-vnet')```

Hao-Liu:
ACK

Hao-Liu:
Merging this patch. Thanks @crazyAxe @will-Do.

PR #773, opened 2016-05-05
pool_acl: Remove the dir pool mode check

will-Do:
According to patches series: 'tweak dir perms handling on build', if not
set permission mode explicitly, libvirt will not change the dir
permission. More detail refer to:
https://www.redhat.com/archives/libvir-list/2015-April/msg01387.html


will-Do:
This patch just remove the invalid test case, and I'm gonna to merge it.

PR #774, opened 2016-05-05
managedsave: add expected vcpus for cpu_mode test

crazyAxe:
The max cpu (vcpu) must be no less  than socks * thread * cores so that guest can successfully start.


will-Do:
ACK

waynesun09:
ACK

PR #776, opened 2016-05-09
remote_with_unix.cfg: change a match pattern

dzhengfy:
Error messages are changed to reflect the real situation about the
problem that requested mechanism is unavailable in more descriptive way.
So the match pattern need be updated accordingly.

Refer below patch:
https://www.redhat.com/archives/libvir-list/2016-February/msg00658.html


waynesun09:
ACK and merge now, thx

PR #777, opened 2016-05-09
migrate_vm.cfg: use some parameters in base.cfg

dzhengfy:
Some paramters are already shared in avocado-vt base.cfg. So it is
better to use them instead of creating again.


waynesun09:
ACK and merge as trival fix, thx

PR #778, opened 2016-05-09
virsh_migrate_setmaxdowntime: use local uri

dzhengfy:
Use local URI qemu:///system instead of
qemu+ssh://${migrate_source_host}/system. It's not necessary to use
qemu+ssh://${migrate_source_host}/system URI for the local connecting.


will-Do:
ACK

waynesun09:
ACK

PR #779, opened 2016-05-09
virsh.attach_detach_disk: PV cleanup

will-Do:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
@@ -6,7 +6,6 @@ import aexpect
 
 from autotest.client.shared import error
 from autotest.client.shared import utils
-from autotest.client import lv_utils
 
 from virttest import virt_vm
 from virttest import virsh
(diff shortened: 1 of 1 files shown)


will-Do:
And another patch https://github.com/avocado-framework/avocado-vt/pull/495 is needed to fix the vgcreate failure in this case

libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
```diff
@@ -418,6 +418,7 @@ def acpiphp_module_modprobe(vm, os_type):
             if test_logcial_dev:
                 libvirt.delete_local_disk("lvm", vgname=vg_name, lvname=lv_name)
                 lv_utils.vg_remove(vg_name)
+                utils.system("pvremove %s" % device_source, ignore_status=True)
```
waynesun09: is pvremove needed after each vg_remove? if so, maybe we could add it into vg_remove with an pv clean option to enable or disable it

will-Do: Basically, we don't need do this, after iscsi target logout, the PV will be removed automatically. But, yes, we can consider to add this step(which may fail so ignore_status=True)

waynesun09: Ok, ack and merge this now, thx

PR #780, opened 2016-05-10
at_dt_iscsi_disk: Wait for a while to get the volume info.

rbian:
In some case, volume info cann't be got immediately after pool-create
operation, add time wait for this case.


will-Do:
ACK

waynesun09:
ACK

PR #781, opened 2016-05-10
Change the skip match string as new version error output changed

kylazhang:
From "this function is not supported by the connection driver" to
"Operation not supported"


Hao-Liu:
Ack

will-Do:
ack

PR #782, opened 2016-05-11
v2v_vmcheck_helper: fix small problems

chloerh:
1.provide default value of os version.
2.ignore checking xorg log file if it doesn't exist
3.modify return number of failed checkpoints

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -24,7 +24,7 @@ class VMChecker(object):
         self.hypervisor = params.get("hypervisor")
         self.target = params.get('target')
         self.os_type = params.get('os_type')
-        self.os_version = params.get('os_version')
+        self.os_version = params.get('os_version', 'OS_VERSION_V2V_EXAMPLE')
         self.original_vmxml = params.get('original_vmxml')
         self.virsh_session = None
         self.virsh_session_id = None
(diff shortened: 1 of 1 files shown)


provider/v2v_vmcheck_helper.py
```diff
@@ -71,7 +71,7 @@ def run(self):
         else:
             logging.warn("Unspported os type: %s", self.os_type)
         if len(self.errors):
-            return 1
+            return len(self.errors)
         else:
             return 0
```
will-Do: I think ```return len(self.errors)``` should be enough, no need the ```if...else``` block

provider/v2v_vmcheck_helper.py
```diff
@@ -212,16 +212,20 @@ def check_linux_vm(self):
         logging.info("Checking video device inside the VM")
         expect_video = self.get_expect_video_model()
         # Since RHEL7, it use 'kms' instead of 'cirrus'
-        if 'rhel7' in self.os_version and expect_video == 'cirrus':
-            expect_video = 'kms'
+        if self.os_version:
+            if 'rhel7' in self.os_version and expect_video == 'cirrus':
```
will-Do: It's better set a default vaule('') to os_version when get it, then no need update here

will-Do:
ACK

PR #783, opened 2016-05-12
migrate_vm: Fix the incorrect error message

dzhengfy:
When domjobinfo on remote host does not match those on local host, the
error message is inaccurate.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1615,14 +1615,18 @@ def run(test, params, env):
                     check_virsh_command_and_option("domjobinfo", opts)
                     jobinfo = virsh.domjobinfo(args, debug=True,
                                                ignore_status=True).stdout
-                    logging.debug("Local job info: %s", jobinfo)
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1615,14 +1615,18 @@ def run(test, params, env):
...
                     status, output = run_remote_cmd(cmd, server_ip, server_user,
                                                     server_pwd)
-                    if status or not re.search(jobinfo, output):
+                    if status:
                         raise error.TestFail("Failed to run '%s' on the remote"
                                              " : %s" % (cmd, output))
+                    elif not re.search(jobinfo, output):
+                        logging.debug("Remote job info:\n%s", output)
+                        raise error.TestFail("The jobinfo on both local "
+                                             "and remote are not matched.")
```
waynesun09: The job info from local and remote are not matched

dzhengfy: @waynesun09 , updated. Thanks.

waynesun09:
ACK

PR #784, opened 2016-05-12
blockcopy: restart libvirtd after clean the VM

will-Do:
For some cases, libvirt will lose control of the transient VM when
restart libvirtd, which means the VM will disappear but the qemu-kvm process
still exist. So adjust the order to restart libvirtd later.


waynesun09:
ACK and merge now, thx

PR #785, opened 2016-05-13
migrate_vm.cfg: Fix one parameter

dzhengfy:
Use shared value in avocado-vt/share/cfg/base.cfg.


waynesun09:
ACK and merge as trivial fix, thx

PR #792, opened 2016-05-19
virsh_update_device: Make sure vm is stopped before testing.

rbian:


waynesun09:
ACK and merge as trivial fix, thanks

PR #795, opened 2016-05-27
blockcopy: Fix fail on virtlogd command not found

waynesun09:
virtlogd is introduced in libvirt 1.3.1, if not found in earlier
version avocado.utils.path.CmdNotFoundError will be raised, need
caught it in exception and let test pass in such situation.


will-Do:
ACK

waynesun09:
Thanks, merge as trivial fix

PR #798, opened 2016-05-27
libvirt_mem: Fix for undeclared 'cur_mem'.

rbian:


waynesun09:
ACK and merge as trivial fix, thanks

libvirt/tests/src/libvirt_mem.py
```diff
@@ -281,7 +281,7 @@ def modify_domain_xml():
     max_mem = params.get("max_mem")
     max_mem_rt = params.get("max_mem_rt")
     max_mem_slots = params.get("max_mem_slots", "16")
-    #cur_mem = params.get("current_mem")
+    cur_mem = params.get("current_mem")
```
will-Do: Does it need a default value?

rbian: We don't have a default value for it, it must be got from configuration file.

PR #802, opened 2016-06-02
iface_options: update the max queue number

dzhengfy:
Multiqueue is supported up to 256 by the kernel in RHEL7.3 instead of 8.
So the number should be updated.


waynesun09:
this update will fail testing on rhel6?

dzhengfy:
@waynesun09 , In RHEL6, this case is skipped for queues options is not supported in that libvirt version.

PR #803, opened 2016-06-03
iface_hotplug: fix the stress test

dzhengfy:
Detach the interface after attaching the interface for stress test.


will-Do:
I think you should add more desc about the problem(what and why)

will-Do:
@dzhengfy 
Please explain why you doing the fix in the commit message, thx

waynesun09:
@dzhengfy could you update commit msg?

PR #804, opened 2016-06-06
virsh_vol_clone_wipe: Infinity loop fix

will-Do:


waynesun09:
ACK and merge now, thanks

PR #805, opened 2016-06-07
virsh_migrate_setmaxdowntime: Fix syntax error

dzhengfy:
utils_misc.wait_for function is invoked, so there is a need to import
utils_misc. Therefore, it seems no need to include
"from virttest.utils_misc import SELinuxBoolean".


libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py
```diff
@@ -14,7 +14,7 @@
 from virttest.utils_test import libvirt
 from virttest import utils_libvirtd
 from virttest import utils_config
-from virttest.utils_misc import SELinuxBoolean
```
will-Do: Is this invalid?

will-Do:
Okay

PR #806, opened 2016-06-08
Handle cache directory missing for qemu capability

dzhengfy:
- To determine if a certain qemu capability is supported in current
environment, we check the cache directory which is not implemented on
RHEL6. This only works on RHEL7. So an exception will be raised out on
RHEL6 which we should handle it with correct logic.

- A logic error fix
There is a need to create image or create pool before migration.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
@@ -3,6 +3,7 @@ import logging
 
 from autotest.client import lv_utils
 from autotest.client.shared import error
+from avocado.core import exceptions
 
 from virttest import ssh_key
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
```diff
@@ -3,6 +3,7 @@
 
 from autotest.client import lv_utils
 from autotest.client.shared import error
```
waynesun09: since add avocado exceptions, maybe u could replace the autotest error part

waynesun09: ACK with the fix, but depend on both autotest and avocado is weird in this case

will-Do:
ACK

PR #807, opened 2016-06-10
Fixup virsh_migrate.py

balamuruhans:
To use default jeos image if migrate_shared_storage is not configured


will-Do:
@dzhengfy please help to review this

balamuruhans:
@will-Do @dzhengfy without this check, it was using the same default jeos-23-64 image for all other guests.

waynesun09:
ACK and confirmed with @dzhengfy, merge now, thanks

balamuruhans:
@waynesun09 Thank you

PR #809, opened 2016-06-12
convert_vm_to_ovirt: assign None to parameters for hypervisor kvm

chloerh:
And remove parameters that are no longer needed.


v2v/tests/src/convert_vm_to_ovirt.py
```diff
@@ -36,7 +36,6 @@ def run(test, params, env):
     vpx_pwd_file = params.get("vpx_pwd_file")
     vpx_dc = params.get("vpx_dc")
     esx_ip = params.get("esx_ip")
-    hypervisor = params.get("hypervisor")
```
will-Do: Why remove this var

chloerh: It's duplicate with line 26

chloerh:
@will-Do add v2v_tiimeout for import to ovirt

will-Do:
ACK

PR #810, opened 2016-06-13
iface_options: update the max queue number

dzhengfy:
Multiqueue is supported up to 256 by the kernel in RHEL7.3
 instead of 8. So the number should be updated.


waynesun09:
As it's negative case, the update will not fail on earlier version, so ACK and merge this now, thanks.

PR #815, opened 2016-06-20
vol_clone_wipe: Case update

will-Do:
1. Fail case if the testing pool is exist, and it's no need to check volume
then. And check if the new create volume exist in the testing pool
before the next step testing.
2. Adjust the pre-check order to skip the case ASAP.
3. And rename an instance name and the pool/vol name to make them more readable.


libvirt/tests/src/virsh_cmd/volume/virsh_vol_clone_wipe.py
```diff
@@ -78,35 +58,52 @@ def run(test, params, env):
...
+                                vol_format=vol_format,
+                                capacity=vol_capability,
+                                allocation=None,
+                                pool_name=pool_name)
         elif vol_format == 'partition':
-            vol_name = libv_vol.list_volumes().keys()[0]
-            logging.debug("Partition %s in disk pool is volume" % vol_name)
+            if not libvirt_vol.list_volumes().keys():
+                raise error.TestError("No volume in pool: %s" % pool_name)
+                vol_name = libvirt_vol.list_volumes().keys()[0]
```
Hao-Liu: I think these two lines are over-indented.

Hao-Liu:
I think we'd better add a brief description of what we did and why in the commit message

libvirt/tests/src/virsh_cmd/volume/virsh_vol_clone_wipe.py
will-Do: Yes, will update it, thx

will-Do:
Thx @Hao-Liu

libvirt/tests/src/virsh_cmd/volume/virsh_vol_clone_wipe.py
```diff
@@ -78,35 +61,50 @@ def run(test, params, env):
...
+    # Choose an algorithm randomly
+    if wipe_algorithms:
+        alg = random.choice(wipe_algorithms.split())
+    else:
+        alg = random.choice(valid_algorithms)
+
+    libvirt_pvt = utlv.PoolVolumeTest(test, params)
+    libvirt_pool = libvirt_storage.StoragePool()
+    if libvirt_pool.pool_exists(pool_name):
+        raise error.TestError("Pool '%s' already exist", pool_name)
```
kylazhang: The above judgement should be in try as if pool already exists, we need to clean the env for following case

kylazhang:
Looks good to me for the left part.

libvirt/tests/src/virsh_cmd/volume/virsh_vol_clone_wipe.py
will-Do: As there's no prepare steps above, so it's no need to do cleanup

kylazhang: If pool exists, it is not need to destroy/undefine it?

will-Do: Yes, that should be done by the previous case, we suppose the testing env is clean as we can't do every cleanup steps before running the case.

kylazhang: ok, then no problem

Hao-Liu:
ACK and merge this, Thanks @will-Do @kylazhang

PR #816, opened 2016-06-21
domxml_to_native: Add '-object secret,id' to filter list

waynesun09:
For fixing bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1182074

libvirt commit a1344f7 introduced default secret object created
with a random AES-256 key for domain, the key is under vm dir by
default which depends on the env, so add this part to filter list
and not check it.


Hao-Liu:
ACK

will-Do:
ACK

PR #817, opened 2016-06-22
Modify return message of v2v guest check

chloerh:
1. Change the return value from number of failed checkpoints to the list of checkpoints to present clearer error message.
2. Move cleanup to final part. The cleanup part in original code is not placed in final part which will cause skipping cleaning up when test fails ahead of cleanup code. Move to final part will solve the problem.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -1,5 +1,6 @@
 import re
 import logging
+import time
 from distutils.version import LooseVersion
 
 from avocado.utils import process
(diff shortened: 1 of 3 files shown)


Hao-Liu:
It seems that your works here doesn't only include what you described in the commit message. Please make them match and better describe why you did this.

chloerh:
@Hao-Liu , updated, please review, thanks.

will-Do:
Well, CI/blacklist it's more like a internal terminology, it's better to get rid of it.

chloerh:
@will-Do updated commit message, please review, thanks.

provider/v2v_vmcheck_helper.py
```diff
@@ -70,13 +71,14 @@ def run(self):
...
             logging.warn("Unspported os type: %s", self.os_type)
-        return len(self.errors)
+        return self.errors
 
     def compare_version(self, compare_version):
         """
         Compare virt-v2v version against given version.
         """
         cmd = 'rpm -q virt-v2v'
+        process.run(cmd)
```
will-Do: What's the purpose of the line?

chloerh: there's a lock on rpmdb,  run 'rpm -q virt-v2v'  the first time will get an output like this:
"BDB2053 Freeing read locks for locker 0x20c: 17216/140399824816192
BDB2053 Freeing read locks for locker 0x20d: 17216/140399824816192
BDB2053 Freeing read locks for locker 0x20e: 17216/140399824816192
BDB2053 Freeing read locks for locker 0x20a: 17216/140399824816192
"
which won't fit the format of version string. Run this command 1 extra time will skip this problem

will-Do:
ACK

PR #819, opened 2016-06-27
Sync paramters with earlier cases

chloerh:
Unify names of parameters in different cfg files but with same
function to reduce number of config lines of replace part in ci.


will-Do:
okay

PR #820, opened 2016-06-28
migrate_vm: increase the migration timeout

dzhengfy:
For storage migration, the current timeout duration is too short which
will cause timeout exception. So now it is increased to a larger value.


will-Do:
Make sense

PR #821, opened 2016-07-04
nodeinfo: Get host memory size from numa nodes

waynesun09:
As libvirt commit 912813d in 2.0.0 change the api nodeGetInfo with
using common api nodeGetMemory to collect host memory size, current
check with nodeinfo memory size failed.

Update to collect memory size from online nodes meminfo to fix the
case.


kylazhang:
ACK.

Hao-Liu:
ACK and merge. Thanks @waynesun09 @kylazhang

PR #822, opened 2016-07-05
Storage related cases fix and update

will-Do:
For virsh_pool_acl test, iSCSI type pool always start, that because
libvirt just calling iscsiadm command to login the target which is not
discovery first. So hardcode the discovery command for it.
For virsh_vol_create_from test, the default volume in iSCSI pool is the
iSCSI lun, so the input volume size is the pool size, which require the
destination pool size should larger than the input volume size. And LVM
allocates storage in multiples of the physical extent size, which is 4MB
by default.
BTW, this patch also rename the pool/volume name in different tests to
make the debug message more clear.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/volume/virsh_vol_create_from.py
+++ b/libvirt/tests/src/virsh_cmd/volume/virsh_vol_create_from.py
@@ -52,7 +52,7 @@ def run(test, params, env):
 
         if src_pool_type != dest_pool_type:
             pvt.pre_pool(dest_pool_name, dest_pool_type, dest_pool_target,
-                         dest_emulated_image, image_size="50M",
+                         dest_emulated_image, image_size="100M",
                          pre_disk_vol=["1M"])
 
         # Print current pools for debugging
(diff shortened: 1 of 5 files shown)


libvirt/tests/src/virsh_cmd/volume/virsh_vol_create_from.py
```diff
@@ -52,27 +52,22 @@ def run(test, params, env):
...
                         capacity=vol_size, allocation=None,
                         pool_name=src_pool_name)
         else:
-            src_pv = libvirt_storage.PoolVolume(src_pool_name)
-            src_vols = src_pv.list_volumes().keys()
-            if src_vols:
-                src_vol_name = src_vols[0]
-            else:
-                raise error.TestFail("No volume in pool: %s" % src_pool_name)
+            src_vol_name = utlv.get_vol_list(src_pool_name).keys()[0]
```
kylazhang: No error raise if no volume in pool?

will-Do: Yes, we should discussed this in https://github.com/avocado-framework/avocado-vt/pull/584

kylazhang: ok, then no problem. thanks

will-Do:
Merge this patch now

PR #823, opened 2016-07-05
blockcopy: make sure blockjob exist

will-Do:
When wait for the block copy job finish, we need make sure the job
exist.
And remove a unused var.


kylazhang:
ACK.

will-Do:
Thx, merge it now

PR #824, opened 2016-07-07
Update code of function test on xen and add cases

chloerh:
Update code: Add error list to collect errors during checking vm to
make sure all checkpoints are checked. Call vmcheck object as param
to avoid create it again and again or create and close session.

Add cases:
- convert guest with display type 'vnc' and 'sdl'
- with attirb 'autoport' enable
- with scsi disk and ide disk, with 2 'program file' on win guest
- with empty or wrong environment LIBGUESTFS_BACKEND
- with a same name guest at destination
- with no passwordless SSH access configured on host
- with no regular kernel installed
- with input mode 'libvirtxml' without image file

### DIFF
--- a/v2v/tests/src/function_test_xen.py
+++ b/v2v/tests/src/function_test_xen.py
@@ -11,6 +11,7 @@ from virttest import utils_sasl
 from virttest import ssh_key
 from virttest import remote
 from virttest import data_dir
+from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
 
 from provider.v2v_vmcheck_helper import VMChecker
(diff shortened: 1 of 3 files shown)


will-Do:
Please update the commit message before we review the code

v2v/tests/src/function_test_xen.py
```diff
@@ -11,10 +11,13 @@
 from virttest import ssh_key
 from virttest import remote
 from virttest import data_dir
+from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
 
 from provider.v2v_vmcheck_helper import VMChecker
 
+DEFAULT_VM = 'avocado-vt-vm1'
```
will-Do: This should be the default vm of avocado-vt, why you need specify it here?

v2v/tests/src/function_test_xen.py
```diff
@@ -119,12 +149,32 @@ def check_result(result, status_error):
...
+            ret = vmchecker.run()
+            if len(ret) == 0:
+                logging.info("All common checkpoints passed")
+            # Check specific checkpoints
+            if checkpoint == 'rhev_file':
+                check_rhev_file_exist(vmchecker.checker)
+            elif checkpoint == 'console_xvc0':
+                check_grub_file(vmchecker.checker, 'console_xvc0')
+            elif checkpoint in ('vnc_autoport', 'vnc_encrypt'):
+                vmchecker.check_graphics(params[checkpoint])
```
will-Do: Why not use 'checkpoint' directly ?

will-Do: I see here ```check_graphics``` accept dict as parameter , so here the```params[checkpoint]``` is a dict?

chloerh: Yes, have to check more than 1 attributes' values sometimes

PR #828, opened 2016-07-15
Add cache=directsync option if io=native is used for disk driver.

rbian:
It's a feature of qemu-kvm. Qemu will report "aio=native was specified,
but it requires cache.direct=on, which was not specified" if there is
no cache=directsync option.


will-Do:

PR #829, opened 2016-07-15
Add v2v case of paramters

chloerh:
Add cases of conflict paramters, convert from xen guest and check
output format, quiet mode case convert from xen without output
format, check length of each line of v2v error message.
Update code, use avocado instead of autotest.

### DIFF
--- a/v2v/tests/cfg/v2v_options.cfg
+++ b/v2v/tests/cfg/v2v_options.cfg
@@ -13,7 +13,7 @@
     vpx51_hostname = VPX_51_HOSTNAME_V2V_EXAMPLE
     vpx51_dc = VPX_51_DC_V2V_EXAMPLE
     esx51_hostname = ESX_51_HOSTNAME_V2V_EXAMPLE
-    ovirt_engine_url = "OVIRT_ENGINE_URL_V2V_EXAMPLE"
+    ovirt_engine_url = "https://OVIRT_ENGINE_URL_V2V_EXAMPLE/api"
     ovirt_engine_user = "OVIRT_ENGINE_USER_V2V_EXAMPLE"
     ovirt_engine_password = "OVIRT_ENGINE_PASSWORD_V2V_EXAMPLE"
     remote_shell_client = "ssh"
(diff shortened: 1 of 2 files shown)


v2v/tests/cfg/v2v_options.cfg
```diff
@@ -231,14 +240,43 @@
...
+                    only input_mode.libvirt.xen
+                    only output_mode.rhev, output_mode.libvirt
+                    checkpoint = 'quiet'
         - negative_test:
             status_error = "yes"
             variants:
                 - conflict_options:
                     only input_mode.none
                     only output_mode.none
+                    checkpoint = 'conflict'
```
will-Do: How about using 'conflict_options'

v2v/tests/cfg/v2v_options.cfg
```diff
@@ -231,14 +240,43 @@
...
+                        - 2_vdsm-vm-uuid:
+                            option = '-vdsm-vm-uuid'
+                        - 2_vdsm-ovf-output:
+                            option = '-vdsm-ovf-output'
+                        - 2_vmtype:
+                            option = '-vmtype'
+                - xen_no_of:
+                    only input_mode.libvirt.xen
+                    only output_mode.libvirt
+                    checkpoint = 'xen_no_of'
```
will-Do: Please make sure the case name and checkpoint value here more readable, 'xen_no_of' it really hard to read.

v2v/tests/src/v2v_options.py
```diff
@@ -259,20 +262,25 @@ def check_result(cmd, result, status_error):
...
     backup_xml = None
     vdsm_domain_dir, vdsm_image_dir, vdsm_vm_dir = ("", "", "")
     try:
         # Build input options
         input_option = ""
         if input_mode is None:
-            pass
+            option = params.get('option')
+            if option in option_list:
+                v2v_options = "-i libvirt -%s abc -%s abc" % (option, option)
```
will-Do: I understand here you try to assemble options for negative cases, but it might confuse people who just read the code.

v2v/tests/src/v2v_options.py
```diff
@@ -68,6 +60,9 @@ def run(test, params, env):
     restore_image_owner = False
     address_cache = env.get('address_cache')
     params['vmcheck_flag'] = False
+    checkpoint = params.get('checkpoint')
+    option_list = ['of', 'ic', 'on', 'if', 'oc', 'os', '-password-file',
+                   '-vdsm-vm-uuid', '-vdsm-ovf-output', '-vmtype']
```
will-Do: the ```option_list``` here it quite confusing

v2v/tests/cfg/v2v_options.cfg
```diff
@@ -231,14 +240,43 @@
...
+                        - 2_on:
+                            option = 'on'
+                        - 2_if:
+                            option = 'if'
+                        - 2_oc:
+                            option = 'oc'
+                        - 2_os:
+                            option = 'os'
+                        - 2_password-file:
+                            option = '-password-file'
```
will-Do: It's better unify the option values

v2v/tests/src/v2v_options.py
```diff
@@ -328,16 +345,22 @@ def check_result(cmd, result, status_error):
                 os.mkdir(vdsm_image_dir)
                 os.mkdir(vdsm_vm_dir)
 
-        # Output more messages
-        v2v_options += " -v -x"
+        # Output more messages except quiet mode
+        if checkpoint == 'quiet':
+            v2v_options += ' -q'
+            logging.debug(v2v_options)
```
will-Do: Maybe you can log the v2v_option at the end part, not only for 'quit' condition.

chloerh: Turns out it's unnecessary log, already remove it.

will-Do:
ACK

PR #830, opened 2016-07-18
convert_vm_to_ovirt:set output format to 'qcow2' for qcow2 guests

chloerh:


v2v/tests/src/convert_vm_to_ovirt.py
```diff
@@ -106,6 +106,8 @@ def run(test, params, env):
         v2v_params.update({"esx_ip": esx_ip})
     if v2v_opts:
         v2v_params.update({"v2v_opts": v2v_opts})
+    if vm_name.endswith('qcow2'):
```
will-Do: Is it the safe way to determine the disk format?

will-Do:
ACK

PR #831, opened 2016-07-18
libvirt_mem: Change hotplug memory size to multiple of 128MiB

rbian:
According to the linux kernel, backend memory size must be
multiple of 128MiB.


Hao-Liu:
@rbian according to https://lists.gnu.org/archive/html/qemu-devel/2015-01/msg02164.html. This should be a bug for one or more in libvirt, qemu, kernel. So I think beside changing this size to multiple of 128, we should still cover the original "maybe-negative" scenario.

waynesun09:
@rbian https://bugzilla.redhat.com/show_bug.cgi?id=1358682
As peter updated the bug to assign, the bug could be fixed in 7.4, then we might need the negative case when with 2MiB not 128MiB libvirt will let us know it failed rather than succeed as now but not inside the guest.
So 1 negative test in this patch will work, and put it in our CI blacklist.

Hao-Liu:
ACK this. Thanks @rbian @waynesun09.

waynesun09:
ACK and merge now, thanks

PR #832, opened 2016-07-19
graphics_functional: Case update for some new changes

will-Do:
For new version libvirt, there are some changes, so update test case
to adapt them:
1. A new listen type 'none' is added. When no port given and autoport is
'no', libvirt will add listen type='none' automatically, so these two
autoport='no' and insecure mode/channel spice cases are valid now.
2. Now spice port accept negative number less than -1, If spice port
less than -1, the VM can start normally, but both listen address and port
will be omitted when pass to qemu.
3. Some error messages and vnc socket path also changed, so add new
match patterns and change two boolean values to string type to use
re.match for result compare.

Signed-off-by: Yanbing Du ydu@redhat.com

### DIFF
--- a/libvirt/tests/src/graphics/graphics_functional.py
+++ b/libvirt/tests/src/graphics/graphics_functional.py
@@ -20,6 +20,8 @@ from virttest import utils_libvirtd
 from virttest.utils_test.libvirt import LibvirtNetwork
 from virttest.libvirt_xml.devices.graphics import Graphics
 
+from provider import libvirt_version
+
 
 class PortAllocator(object):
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -952,6 +981,29 @@ def run(test, params, env):
...
+    #    listen address and port will be omitted
+    spice_listen_type = params.get('spice_listen_type', 'not_set')
+    vnc_listen_type = params.get('vnc_listen_type', 'not_set')
+    spice_port = params.get('spice_port', 'not_set')
+    spice_tlsPort = params.get('spice_tlsPort', 'not_set')
+    if not libvirt_version.version_compare(2, 0, 0):
+        for listen_type in [spice_listen_type, vnc_listen_type]:
+            if listen_type == 'none':
+                params[listen_type] = 'not_set'
+    if libvirt_version.version_compare(2, 0, 0):
```
waynesun09: since checked on line 996, use "else:" could be enough

will-Do:
Just merge this fix for now

PR #833, opened 2016-07-20
Close and recreate session to avoid connection reset on windows

chloerh:
For ncat connection might be reset before any commands could be
sent, or during the sleeping time while waiting for drivers to be
fully installed.Therefore close and recreate session and send one
test command to make sure the connection works fine.
Also, create virsh SASL session might timeout rarely, therefore
retry it several times till get the expected result.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -3,6 +3,7 @@ import logging
 import time
 from distutils.version import LooseVersion
 
+from avocado.core import exceptions
 from avocado.utils import process
 
 from virttest import utils_v2v
@@ -10,6 +11,7 @@ from virttest import utils_sasl
 from virttest import virsh
 
 V2V_7_3_VERSION = 'virt-v2v-1.32.1-1.el7'
+RETRY_TIMES = 10
 
 
 class VMChecker(object):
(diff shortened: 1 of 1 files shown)


will-Do:
Please give a detail description about the problem in your commit message

provider/v2v_vmcheck_helper.py
```diff
@@ -53,12 +55,21 @@ def close_virsh_session(self):
             self.virsh_session.close_session()
 
     def setup_session(self):
-        if self.target == "ovirt":
-            self.virsh_session = utils_sasl.VirshSessionSASL(self.params)
-            self.virsh_session_id = self.virsh_session.get_id()
-        else:
-            self.virsh_session = virsh.VirshPersistent()
-            self.virsh_session_id = self.virsh_session.session_id
+        for index in range(RETRY_TIMES):
```
will-Do: Could you figure out why the session timeout, always using retry is not a good solution

provider/v2v_vmcheck_helper.py
```diff
@@ -230,9 +241,23 @@ def check_windows_vm(self):
...
-        self.checker.create_session()
+        try:
+            self.checker.create_session()
+        except Exception as detail:
+            raise exceptions.TestError('Failed to connect to windows guest: %s' %
+                                       detail)
         logging.info("Wait 60 seconds for installing drivers")
         time.sleep(60)
+        # Close and re-create session in case connection reset by peer during
+        # sleeping time. Keep trying until the test command runs successfully.
```
will-Do: So the created session is not used and has such problem, why not create it later, and I hope we can get rid of the retry times.

provider/v2v_vmcheck_helper.py
chloerh: I could try, but this happens rarely, only once after one or two hundreds times

will-Do:
ACK

PR #836, opened 2016-07-27
Remove win8 on xen from matrix

chloerh:
For some windows 8 guests could not be successfully installed on
xen host, remove the cases from matrix as manual team required.


will-Do:
It's better to describe why remove it in the commit message

will-Do:
Okay, ACK

PR #837, opened 2016-07-27
Fix the wrong assignment

will-Do:
The 'get_vol_list' function return a dict, what we want here is the
first key value.
And fix the message when raise an exception.


waynesun09:
ACK

Hao-Liu:
ACK and merging, thanks @will-Do @waynesun09

PR #838, opened 2016-07-28
Add 6 v2v function cases

chloerh:
2 with spice type
1 with wrong fstab file
3 with selinux configuration

### DIFF
--- a/v2v/tests/cfg/specific_kvm.cfg
+++ b/v2v/tests/cfg/specific_kvm.cfg
@@ -13,7 +13,7 @@
     password = 'redhat'
 
     # RHEVM parameters
-    ovirt_engine_url = 'OVIRT_ENGINE_URL_V2V_EXAMPLE'
+    ovirt_engine_url = 'https://OVIRT_ENGINE_URL_V2V_EXAMPLE/api'
     ovirt_engine_user = 'OVIRT_ENGINE_USER_V2V_EXAMPLE'
     ovirt_engine_password = 'OVIRT_ENGINE_PASSWORD_V2V_EXAMPLE'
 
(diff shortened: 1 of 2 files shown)


v2v/tests/src/specific_kvm.py
```diff
@@ -576,75 +609,90 @@ def check_result(result, status_error):
...
+        elif checkpoint == 'host_selinux_on':
+            params['selinux_stat'] = utils_selinux.get_status()
+            utils_selinux.set_status('enforcing')
+        elif checkpoint.startswith('selinux'):
+            if set_selinux(checkpoint[8:]):
+                vm = libvirt_vm.VM(vm_name, params, test.bindir,
+                                   env.get('address_cache'))
+                # Reboot vm to make configuration effect
+                vm.reboot()
+                vm.shutdown()
```
will-Do: Why not shutdown directly, and next time why VM start, your configuration will effect, right?

v2v/tests/src/specific_kvm.py
```diff
@@ -450,6 +452,22 @@ def cleanup_fs():
...
+        session = kwargs['session']
+        current_stat = session.cmd_output('getenforce').strip()
+        logging.debug('Current selinux status: %s', current_stat)
+        if current_stat != value:
+            cmd = "sed -E -i 's/(^SELINUX=).*?/\\1%s/' /etc/selinux/config" % value
+            logging.info('Set selinux stat with command %s', cmd)
+            session.cmd(cmd)
+            return True
+        else:
+            return False
```
will-Do: It's weird to return False here

chloerh:
@will-Do updated, please review, thanks.

v2v/tests/cfg/specific_kvm.cfg
```diff
@@ -123,6 +121,8 @@
                     checkpoint = 'fstab_uuid'
                 - virtio:
                     checkpoint = 'fstab_virtio'
+                - wrong:
+                    checkpoint = 'fstab_wrong'
```
will-Do: So this should be a negative test case, right? How about name it to 'invalid'

will-Do:
ACK

PR #839, opened 2016-07-28
Update code of v2v function test on esx

chloerh:
Check common checkpoints first and then check specific checkpoints
and merge errors to a list instead of raising exception immediately.


will-Do:
Patch is okay for me.
But, please update the commit summary and message

will-Do:
Okay, hope you can write a better patch summary and description next time.

PR #845, opened 2016-08-05
Add iothread commands tests

rbian:


will-Do:
Let's see the test results

waynesun09:
Merge this new cases now and let's fix problems if CI failed on this.

PR #846, opened 2016-08-05
vfio: Add tests for attach hostdev interface with managed option

rbian:


will-Do:
ACK

PR #851, opened 2016-08-11
Fixup virsh_hostname

balamuruhans:
For hostname with name abc.github.com, shell cmd "hostname" returns only
abc but "virsh hostname" returns full name abc.github.com, which fails
the test case, "hostname -f" fixes this error.


will-Do:
ACK

balamuruhans:
Thank you @will-Do

PR #855, opened 2016-08-19
Add v2v function cases of converting from file

chloerh:
Including
- convert common ova file
- convert ova file with special configurations

### DIFF
--- a/v2v/tests/cfg/convert_from_file.cfg
+++ b/v2v/tests/cfg/convert_from_file.cfg
@@ -1,0 +1,131 @@
+- convert_from_file:
+    type = 'convert_from_file'
+    vm_type = 'libvirt'
+    start_vm = 'no'
+    take_regular_screendumps = no
+    v2v_timeout = '1200'
+
+    # Regular kvm guest parameters
+    os_type = 'linux'
+    input_mode = 'libvirt'
+    hypervisor = 'kvm'
+    username = 'root'
+    password = 'redhat'
+
+    # RHEVM parameters
+    ovirt_engine_url = 'https://OVIRT_ENGINE_URL_V2V_EXAMPLE/api'
+    ovirt_engine_user = 'OVIRT_ENGINE_USER_V2V_EXAMPLE'
+    ovirt_engine_password = 'OVIRT_ENGINE_PASSWORD_V2V_EXAMPLE'
+
+    # Using NFS storage here
+    export_name = 'NFS_EXPORT_NAME_V2V_EXAMPLE'
+    storage_name = 'NFS_STORAGE_NAME_V2V_EXAMPLE'
+    cluster_name = 'NFS_CLUSTER_NAME_V2V_EXAMPLE'
+    nfs_storage = 'NFS_EXPORT_STORAGE_V2V_EXAMPLE'
+
+    # Shell paramters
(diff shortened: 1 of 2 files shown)


v2v/tests/cfg/convert_from_file.cfg
```diff
@@ -0,0 +1,131 @@
...
+    # RHEVM parameters
+    ovirt_engine_url = 'https://OVIRT_ENGINE_URL_V2V_EXAMPLE/api'
+    ovirt_engine_user = 'OVIRT_ENGINE_USER_V2V_EXAMPLE'
+    ovirt_engine_password = 'OVIRT_ENGINE_PASSWORD_V2V_EXAMPLE'
+
+    # Using NFS storage here
+    export_name = 'NFS_EXPORT_NAME_V2V_EXAMPLE'
+    storage_name = 'NFS_STORAGE_NAME_V2V_EXAMPLE'
+    cluster_name = 'NFS_CLUSTER_NAME_V2V_EXAMPLE'
+    nfs_storage = 'NFS_EXPORT_STORAGE_V2V_EXAMPLE'
```
balamuruhans: @chloerh where are export_name, storage_name and cluster_name params are used, also base.cfg have NFS related params you can use it to avoid duplication.

v2v/tests/cfg/convert_from_file.cfg
```diff
@@ -0,0 +1,131 @@
...
+
+    # Shell paramters
+    remote_shell_client = 'ssh'
+    remote_shell_port = 22
```
balamuruhans: shell_port in base.cfg can reused

v2v/tests/cfg/convert_from_file.cfg
chloerh: These parameters are used by ovirt-python module while importing converted vm to rhevm after v2v conversion and they are different from params in base.cfg and necessary for v2v tests.

v2v/tests/cfg/convert_from_file.cfg
chloerh: Unlike other tests, basically there's just one cfg entry for v2v tests. All param updatings are through the test cfg file. And the param in base.cfg is with the namd 'shell_port', not exactly the same with this param. So maybe keeping this param could keep the param group complete and easy to maintain in the future.

v2v/tests/src/convert_from_file.py
```diff
@@ -0,0 +1,148 @@
...
+        if process.run(cmd_man_page, shell=True).exit_status != 0:
+            log_fail('Man page not contain boot failure msg')
+        for i in range(100):
+            virsh.screenshot(vm_name, screenshot)
+            similar = ppm_utils.image_histogram_compare(screenshot, match_img)
+            if similar > bar:
+                logging.info('Meet BSOD with similarity %s' % similar)
+                return
+            time.sleep(1)
+        log_fail('No BSOD as expected')
```
will-Do: Can you explain this scenario more detail?

chloerh: Convert a paticular windows guest in ova file to libvirt and check if it boots up to blue screen failure(Check if screenshots are similar with blue screenshot image). Also there should be certain content in v2v man page like "Boot failure: 0x0000007B"

PR #858, opened 2016-08-24
migrate_vm: Fix some message match mistakes

dzhengfy:
- "Cannot access storage file..." error message should not be skipped by
  some cases. Fix the wrong matching method.
- Give the migration output to the final exception for more clear
  recognition.


waynesun09:
ACK

kylazhang:
ACK

PR #859, opened 2016-08-25
virsh.emulatorpin: Restore VM xml incase of failure for vcpu placement

sathnaga:
Before this patch if VM fails to start after a vcpu placement
changes in xml, it leaves the buggy xml in the system and results in
further tests failing, this patch will restore the backupxml incase
of failout.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_emulatorpin.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_emulatorpin.py
@@ -9,6 +9,7 @@ from virttest.libvirt_xml import vm_xml
 from virttest import utils_libvirtd, virsh
 from virttest.utils_test.libvirt import cpus_parser
 from virttest.staging import utils_cgroup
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_emulatorpin.py
```diff
@@ -194,7 +194,14 @@ def run(test, params, env):
         vm.destroy()
         vmxml.placement = emulatorpin_placement
         vmxml.sync()
-        vm.start()
+        try:
+            vm.start()
+        except:
```
balamuruhans: it is better to except for specific error "LibvirtXMLError"

sathnaga:
@balamuruhans Thanks for the comment, have fixed :-)

sathnaga:
Fixed CI failure

balamuruhans:
ACK

waynesun09:
ACK

PR #863, opened 2016-09-01
Add checkpoint of snapshot id when convert to rhev

chloerh:
- Snapshot id in ovf file should not consists of '0's.
- Remove paramters of esx5.1 and esx5.5 for no longer needed.
- Update search pattern of uuids, for content of log changes from
"qemu convert" to "qemu 'convert'" with an extra pair of single
quote.

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -6,10 +6,12 @@ import re
 import pwd
 import logging
 import shutil
+import string
 
 from avocado.core import exceptions
 from avocado.utils import process
 
+from virttest import data_dir
 from virttest import ssh_key
 from virttest import utils_misc
 from virttest import utils_v2v
@@ -93,7 +95,7 @@ def run(test, params, env):
         """
         Get export domain uuid, image uuid and vol uuid from command output.
         """
-        tmp_target = re.findall(r"qemu-img\sconvert\s.+\s'(\S+)'\n", output)
+        tmp_target = re.findall(r"qemu-img\s'convert'\s.+\s'(\S+)'\n", output)
         if len(tmp_target) < 1:
             raise exceptions.TestError("Fail to find tmp target file name when"
                                        " converting vm disk image")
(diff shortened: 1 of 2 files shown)


v2v/tests/src/v2v_options.py
```diff
@@ -300,7 +314,7 @@ def check_result(cmd, result, status_error):
         elif input_mode == "disk":
             input_option += "-i %s %s" % (input_mode, disk_img)
         elif input_mode in ['libvirtxml', 'ova']:
-            raise exceptions.TestNAError("Unsupported input mode: %s" % input_mode)
+            raise exceptions.TestAbortError("Unsupported input mode: %s" % input_mode)
```
will-Do: should be 'TestSkipError'

v2v/tests/src/v2v_options.py
```diff
@@ -370,7 +384,7 @@ def check_result(cmd, result, status_error):
                 os.chown(disk_img, user_info.pw_uid, user_info.pw_gid)
                 restore_image_owner = True
             else:
-                raise exceptions.TestNAError("Only support convert local disk")
+                raise exceptions.TestAbortError("Only support convert local disk")
```
will-Do: And here

chloerh:
@will-Do updated, please review, thanks.

v2v/tests/src/v2v_options.py
```diff
@@ -236,7 +251,7 @@ def check_result(cmd, result, status_error):
                         v2v_start = True
                     if line.startswith('libvirt:'):
                         v2v_start = False
-                    if v2v_start and line > 72:
+                    if v2v_start and len(line) > 72:
```
balamuruhans: "if v2v_start and (len(line) > 72):" gives good readability

will-Do: @chloerh Please leave your comment or update it so we merge this patch later

chloerh: Since 'and' has lower priority than most operators, I felt this modification not quite necessary. And the old way looks prettier :)

will-Do:
Okay, lets merge this patch

PR #864, opened 2016-09-05
Fix start vm failure when use libvirtd session

waynesun09:
As libvirtd session start with root and virtlogd by default start
by systemd, when start vm will get selinux denial as change
virtlogd to unconfined type from default virt_t type. Resolve this
by also start virtlogd with root as unconfined.


kylazhang:
ACK.

will-Do:
ACK

PR #865, opened 2016-09-07
domfstrim: Sync disk before checking disk map

Hao-Liu:
Disk map could possible not changed due to disk cache exists in
guest. This will make the disk map got after executing commands
not up to date. Run `sync` after every commands to make sure cached
data is flushed into disk before checking disk map.

Also fixing a typo and making log more verbose.


waynesun09:
ACK

PR #866, opened 2016-09-07
attach_device: Throw out detail when VM can't start

Hao-Liu:


waynesun09:
ACK and merge as trivial update

PR #867, opened 2016-09-07
virsh_net_dhcp_leases: Fix VM login error

will-Do:
We want to remove all the exist interfaces of the VM before test to make
sure the test env clean, but free_mac_address() is not the correct
function we should call, and the interface number may larger than 1.
After the new interface attached, the 'virtnet' attribute of the vm is
not updated automatically, which will lead to vm login fail, so we need
update it explicitly.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/network/virsh_net_dhcp_leases.py
+++ b/libvirt/tests/src/virsh_cmd/network/virsh_net_dhcp_leases.py
@@ -129,7 +129,8 @@ def run(test, params, env):
     # Remove all interfaces of the VM
     if vm.is_alive():
         vm.destroy(gracefully=False)
-    vm.free_mac_address(0)
+    for idx in range(len(vmxml.get_iface_all())):
+        vm.del_nic(idx)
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/network/virsh_net_dhcp_leases.py
```diff
@@ -152,15 +153,19 @@ def check_net_lease(net_leases, expected_find=True):
             if filter_by_mac:
                 nic_mac = iface_mac
             op = "--type network --source %s --mac %s" % (net_name, iface_mac)
+            nic_params = {'mac': iface_mac, 'nettype': 'bridge',
+                          'ip_version': 'ipv4'}
```
balamuruhans: does nettype and ip_version will always be the same, or is it better to have param for it.

will-Do: For this case yes, the value should be fixed.

balamuruhans:
fine then, ACK

waynesun09:
ACK

PR #872, opened 2016-09-18
daemon.host_uuid: All same UUID is a negative case

Hao-Liu:
Host UUID with all same digits should not be started and will be a
negative test case.

This bug has been fixed by following libvirt commit
    6fe81c3: daemon: add option to read host uuid from /etc/machine-id


will-Do:
So it will fail on older versions?

Hao-Liu:
Yes, It should be a bug on older versions, and should fail.

will-Do:
Got it, thanks.

PR #880, opened 2016-09-26
setvcpus: Fix a TypeError bug and move a case to negative part

will-Do:
1. getAttribute("current") will return an unicode type value, so need format it.
2. After set CPU topology, setvcpus should fail, then the VM can start
normally.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_setvcpus.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_setvcpus.cfg
@@ -48,16 +48,6 @@
                         - option_maximum_config:
                             setvcpus_options = "--maximum --config"
                             setvcpus_count = "5"
-                        - exceeding_topology_limit:
-                            setvcpus_pre_vm_state = "shut off"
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virsh_cmd/domain/virsh_setvcpus.cfg
```diff
@@ -97,6 +87,16 @@
                             setvcpus_options = "--live --config"
                         - current_config_option:
                             setvcpus_options = "--current --config"
+                        - exceeding_topology_limit:
```
waynesun09: The case is pass on rhel6 now, will this change the result on rhel6?

will-Do: As the new design, we don't need test it on RHEL6

waynesun09: saw it run in rhel6 jobs, I guess it'll fail then, I'm gonna merge this and please update blacklist then.

PR #881, opened 2016-09-29
Set LIBGUESTFS_BACKEND=direct for v2v

chloerh:


will-Do:
The commit message is not 100% match your changes, please pay attention next time.

PR #891, opened 2016-10-17
Add function case on hypervisor esx

chloerh:
- convert EFI windows system from esx to RHEV
- convert windows vm with GPO and Anti-virus installed
- convert guest with '#' in name to kvm
- convert guest with vmware tools installed

### DIFF
--- a/v2v/tests/src/function_test_esx.py
+++ b/v2v/tests/src/function_test_esx.py
@@ -1,7 +1,9 @@
 import os
 import logging
+import re
 
 from avocado.core import exceptions
+from avocado.utils import process
 
 from virttest import virsh
 from virttest import utils_v2v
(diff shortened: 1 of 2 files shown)


v2v/tests/src/function_test_esx.py
```diff
@@ -97,20 +139,36 @@ def check_result(result, status_error):
...
+            v2v_sasl.setup(remote=True)
+
+        # Create libvirt dir pool
+        if output_mode == 'libvirt':
+            pvt.pre_pool(pool_name, pool_type, pool_target, '')
+
+        if checkpoint == 'ovmf':
+            url = params.get('ovmf_url')
+            if url and url.endswith('.rpm'):
+                process.run('rpm -iv %s' % url)
```
balamuruhans: if this test supported on ubuntu,  can have a check for distro and apt can be used here.

v2v/tests/src/function_test_esx.py
```diff
@@ -97,20 +139,36 @@ def check_result(result, status_error):
...
         v2v_result = utils_v2v.v2v_cmd(v2v_params)
         check_result(v2v_result, status_error)
 
     finally:
         if params.get('vmchecker'):
             params['vmchecker'].cleanup()
+        if output_mode == 'libvirt':
+            pvt.cleanup_pool(pool_name, pool_type, pool_target, '')
+        if checkpoint == 'ovmf':
+            process.run('rpm -q OVMF&&rpm -e OVMF', shell=True)
```
balamuruhans: same here, check to support ubuntu can be added.

v2v/tests/src/function_test_esx.py
chloerh: virt-v2v test is mainly about rhel, not even fedora. Therefore ubuntu is not supported and we haven't make support for other package management tool

PR #892, opened 2016-10-17
Fixup virsh_net_dhcp_leases testcase

bssrikanth:
This patch will resolve two issues:
1. `IndexError: list index out of range` hit by testcases which are trying to delete nic entries of vm instance.
2. Guests unable to detect the hotplggged nic, since nics were hotplugged too early.. before vm was fully started.

Signed-off-by: Srikanth Aithal <sraithal@linux.vnet.ibm.com>

PR #894, opened 2016-10-18
virsh_net_dhcp_leases: Don't remove the original interfaces

will-Do:
Remove interfaces outside the VM is not enough to update the virtnet
info. Actually, there's no need to remove all the interfaces before
testing , as we will also create a new network for the testing.
And if login VM by IP address failed, we should try serial and run
'dhclient' inside the VM to assign a new IP address for the NIC.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/network/virsh_net_dhcp_leases.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/network/virsh_net_dhcp_leases.cfg
@@ -27,6 +27,4 @@
                     prepare_net = "no"
                     net_name = "xyz"
                 - invalid_mac:
-                    # There's a bug BZ#1205045 about this case
-                    exist_bug = "1205045"
-                    nic_mac = "xyz"
+                    invalid_mac = "yes"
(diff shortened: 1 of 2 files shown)


bssrikanth:
@will-Do I thought removing original interfaces -> hotplugging new interfaces and checking of they get dhcp leases was part of the test case flow. 
So I had fixed the restore/sync part in finally section @ #892 .. Please let me know your thoughts.

will-Do:
@bssrikanth 
Not really, remove the NIC before test start has nothing with this command, the original design is trying to avoid the complexity when check the result.  But as we using a new defined network for testing, the remove step is unnecessary.

bssrikanth:
@will-Do ok. Acked. 

Can you also please fixup issue 2 which I had pointed #892? before hotplugging interface we need to make sure the guest is fully up and running. This can be done by `vm.wait_for_serial_login()`
If you agree I will close #892.

will-Do:
@bssrikanth 
Of course, I think which already in the first patch, line 177. And thanks for your confirm.

will-Do:
Dependent on https://github.com/avocado-framework/avocado-vt/pull/732

PR #895, opened 2016-10-18
dac_start_destroy: libvirt version check added

sudeeshjohn:
libvirt is using "-name guest=" in the qemu command line by default from libvirt version 1.3.5. Added a libvirt version check in the testcase.
Signed-off-by: Sudeesh John <sudheeshjohn@gmail.com>


bssrikanth:
ACK

PR #897, opened 2016-10-21
migrate_vm: multiple fixes for attached disks

dzhengfy:
1. Use '--unsafe' option to avoid unsafe migration error when cache
   != none
2. Fix the raw disk should be migrated correctly with same size on
   target host
3. Remove 'error.TestFail' usage

Signed-off-by: Dan Zheng dzheng@redhat.com

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_vm.cfg
+++ b/libvirt/tests/cfg/migration/migrate_vm.cfg
@@ -15,7 +15,6 @@
     client = "ssh"
     start_vm = "no"
     ssh_port = "${port}"
-    virt_image_dir = /var/lib/libvirt/images
     # setup NFS test environment
     nfs_client_ip = "${server_ip}"
     nfs_server_ip = "${client_ip}"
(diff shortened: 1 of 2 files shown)


dzhengfy:
@waynesun09 @will-Do , could you help review this one?

libvirt/tests/cfg/migration/migrate_vm.cfg
```diff
@@ -478,15 +477,14 @@
                             nfs_mount_dir =
                             disk_driver_cache = "writethrough"
                             virsh_options = "--live --verbose --copy-storage-all"
-                        - raw_to_qcow2_target_image_format:
```
kylazhang: Is that a test scenario? Do you confirmed with feature owner? If you changed it will mean lose this type of cases.

dzhengfy: @kylazhang , I have confirmed with the feature owner. He said no this kind of test case. And from the corresponding manual test case, I judged old automated scripts did not match the test point at all. So I decided to change it according to the manual case step. The old script did not test any special scenario seen from running history.

kylazhang: @dzhengfy ok, Thanks for your confirm. No other problem from my side.

PR #900, opened 2016-10-21
Revert "Fixup: Changed maxvcpus value based on libvirt commit"

will-Do:
This reverts commit 8dac28a5855c82d2221ebd5907f69deb55d496f4.

This patch is not complete:
    1). The 'maxCpus' in capabilities XML only for kvm type domain.
    2). And it needs a remote virsh instance for remote host testing.
    3). This also break the test for old version libvirt.


will-Do:
@sathnaga 
This patch break most case of virsh_maxvcpus, please raise a new patch if you want

sathnaga:
@will-Do oh, okies, i will address those items, it did work for me as I was using the latest libvirt.

PR #902, opened 2016-10-24
Add function cases for v2v_options

chloerh:
- Check v2v dependency
- convert vm with libvirtxml and convert format


v2v/tests/src/v2v_options.py
```diff
@@ -424,14 +433,17 @@ def check_result(cmd, result, status_error):
...
             os.environ['LIBGUESTFS_BACKEND'] = 'direct'
 
         # Running virt-v2v command
         cmd = "%s %s %s %s" % (utils_v2v.V2V_EXEC, input_option,
                                output_option, v2v_options)
         if v2v_user:
             cmd = su_cmd + "'%s'" % cmd
+
+        if checkpoint == 'dependency':
+            cmd = 'rpm -qR virt-v2v'
```
will-Do: Let the command config is better

PR #905, opened 2016-10-24
virsh_migrate: Skip the special migration error

dzhengfy:
"unsupported configuration" error got from the migration command is due
to the test environment. So we should skip this error.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
@@ -67,14 +67,26 @@ def run(test, params, env):
         if src_uri:
             vm.connect_uri = src_uri
 
+    def check_migration_result(migration_res):
+        """
+        Check the migration result.
+
+        :param migration_res: the CmdResult of migration
+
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -67,13 +67,26 @@ def cleanup_dest(vm, src_uri=""):
...
     def do_migration(delay, vm, dest_uri, options, extra):
         logging.info("Sleeping %d seconds before migration" % delay)
         time.sleep(delay)
         # Migrate the guest.
-        successful = vm.migrate(
-            dest_uri, options, extra, True, True).exit_status
-        logging.info("successful: %d", successful)
+        migration_res = vm.migrate(dest_uri, options, extra, True, True)
+        successful = migration_res.exit_status
+        logging.info("Migration exit status: %d", successful)
```
will-Do: Here 'successful' looks some kind of weird. Actually, no need define a new var here as you just use it once

dzhengfy:
@will-Do , updated.

PR #910, opened 2016-11-08
numad_vcpupin: only test pin vcpu to host online cpus

waynesun09:
Test vcpupin with only online host cpus, also as the bug is not
valid anymore, move it into error log and raise the command fail
error msg instead.

### DIFF
--- a/libvirt/tests/src/numa/numad_vcpupin.py
+++ b/libvirt/tests/src/numa/numad_vcpupin.py
@@ -65,12 +65,15 @@ def run(test, params, env):
         vm.start()
         vm.wait_for_login()
 
-        host_cpu_count = utils.count_cpus()
-        for i in range(host_cpu_count):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/numa/numad_vcpupin.py
```diff
@@ -65,12 +65,14 @@ def run(test, params, env):
         vm.start()
         vm.wait_for_login()
 
-        host_cpu_count = utils.count_cpus()
-        for i in range(host_cpu_count):
+        # Test vcpupin to the alive cpus list
+        cpus_list = utils.cpu_online_map()
+        logging.info("active cpus in host are %s", cpus_list)
+        for i in cpus_list:
```
balamuruhans: name can be given like 'cpu' instead of 'i ' for good readability.

balamuruhans:
ACK

PR #899, opened 2016-10-21
Fixup virsh_list.py

balamuruhans:
1. ssh autologin was not handled to execute commands remotely.
2. clean up and recovering the environment was not handled.

This patch fixes it.

Signed-off-by: Balamuruhan S bala24@linux.vnet.ibm.com

PR #911, opened 2016-11-08
Fixup virsh_shutdown.py

balamuruhans:
Autologin from remote machine to test machine was not handled
to run commands remotely.


balamuruhans:
Similar to PR #899

PR #913, opened 2016-11-09
Fixup: virsh_attach_device.py

harish-24:
Currently, test case maps disks attached to guest in same order
disk_VirtualDiskBasic_rcmM_0.raw -> vda, 1.raw -> vdb, 2.raw -> vdc

When guest starts, disks are attached in reverse order, i.e.
disk_VirtualDiskBasic_rcmM_2.raw may correspond to vda etc.
Test cases fails in this scenario. This patch fixes it.

Signed-off-by: Harish S <harisrir@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
@@ -509,6 +509,7 @@ class VirtualDiskBasic(AttachDeviceBase):
     devidx = 1  # devnode name index to start at (0 = vda/sda, 1 = vdb/sdb, etc)
     devtype = 'file'
     targetbus = "virtio"
+    test_data_list = []  # list of all created disks
 
     @staticmethod
     def devname_suffix(index):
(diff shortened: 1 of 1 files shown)


balamuruhans:
ACK

will-Do:
Could you give a more detail description about this?
>  When guest starts, disks are attached in reverse order, i.e.
disk_VirtualDiskBasic_rcmM_2.raw may correspond to vda etc

harish-24:
@will-Do Lets say when three disks are attached using  --config option
disk_VirtualDiskBasic_rcmM_0.raw, target vda
disk_VirtualDiskBasic_rcmM_1.raw, target vdb
disk_VirtualDiskBasic_rcmM_2.raw, target vdc
when the guest starts up, the test cases checks for vda->disk_VirtualDiskBasic_rcmM_0.raw but order of the attached disks has changed
disk_VirtualDiskBasic_rcmM_2.raw corresponds to vda
disk_VirtualDiskBasic_rcmM_1.raw -> vdb
disk_VirtualDiskBasic_rcmM_0.raw -> vdc
and the test cases fails because of this

will-Do:
@harish-24 
Thanks, actually I just want to ask the reason why the order changed

harish-24:
@will-Do Oh, thought of detailing the case :-)

PR #920, opened 2016-11-18
fixup ceph related cases

will-Do:


libvirt/tests/src/virtual_disks/virtual_disks_ceph.py
```diff
@@ -3,8 +3,11 @@
 import time
 import logging
 import aexpect
+
 from autotest.client.shared import error
 from avocado.utils import process
+from avocado.core import exceptions
```
waynesun09: exceptions from avocado and error from autotest are both used, better make full switch

PR #921, opened 2016-11-21
dompmsuspend: Add two more fail conditions

will-Do:
And remove the inappropriate step for agent_error_test


libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
```diff
@@ -144,16 +151,6 @@ def run(test, params, env):
                                          "failed with:\n%s" % (fail_pat, result))
 
             utils_misc.wait_for(lambda: vm.state() == 'pmsuspended', 30)
-            if agent_error_test:
```
waynesun09: seems this is about redo dompmsuspend the second time after already suspended, with this update this scenario is gone. If that's what in our case, i'm ok with this update.

will-Do: move it back.

PR #925, opened 2016-11-25
virsh.blockcommit: Increase timeout to avoid failure

Hao-Liu:
Setting timeout to 10s will be enough for most systems complete
blockcommit.

But do note that the test scenario that a blockcommit job
will be aborted if the job duration exceeds this timeout is
still not covered.

This test scenario is not trivial to implement since the
blockcommit duration heavily depends on the IO performance of
the hardware.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
@@ -239,7 +239,15 @@ def run(test, params, env):
         blockcommit_options = "--wait --verbose"
 
         if with_timeout:
-            blockcommit_options += " --timeout 1"
+            # Setting timeout to 10s will be enough for most systems complete
+            # blockcommit.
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
```diff
@@ -239,7 +239,15 @@ def make_disk_snapshot(postfix_n):
         blockcommit_options = "--wait --verbose"
 
         if with_timeout:
-            blockcommit_options += " --timeout 1"
```
waynesun09: The --timeout option:
```
    --timeout <number>  implies --wait, abort if copy exceeds timeout (in seconds)
```
it's not only for let the blockcommit done, but also for abort it if in the certain time commit not done, need to confirm the feature owner for this change.

waynesun09:
close as no more update

PR #927, opened 2016-11-25
Check virsh command option before run test

will-Do:
For old version libvirt(RHEL6), these cases may fail as virsh command
not support specific option, so check them before run test and raise
TestSkipError if needed.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_hotplug.py
+++ b/libvirt/tests/src/virtual_network/iface_hotplug.py
@@ -60,10 +60,36 @@ def run(test, params, env):
     username = params.get("username")
     password = params.get("password")
 
+    # stree_test require detach operation
+    stress_test_detach_device = False
+    stress_test_detach_interface = False
+    if stress_test:
+        if attach_device:
+            stress_test_detach_device = True
+        if attach_interface:
+            stress_test_detach_interface = True
+
+    # The following detach-device step also using attach option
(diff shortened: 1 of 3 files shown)


will-Do:
Dependent on:
https://github.com/avocado-framework/avocado-vt/pull/789

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -60,11 +60,37 @@ def create_iface_xml(mac):
     username = params.get("username")
     password = params.get("password")
 
+    # stree_test require detach operation
+    stress_test_detach_device = False
+    stress_test_detach_interface = False
+    if stress_test:
+        if attach_device:
+            stress_detach_device = True
```
waynesun09: never used

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -60,11 +60,37 @@ def create_iface_xml(mac):
...
+        if attach_interface:
+            stress_detach_interface = True
```
waynesun09: also this

libvirt/tests/src/virtual_network/iface_hotplug.py
will-Do: Yes, should be stress_test_detach_device

PR #934, opened 2016-11-30
change_media: Fix a module not import bug

will-Do:
And also correct a parameter name.


Hao-Liu:
ACK and merge this simple fix. Thanks @will-Do

PR #935, opened 2016-11-30
capabilities: Fix wordsize and default hypervisor check

waynesun09:
Update both check use guest capabilities dict, and check wordsize
from dict same as libvirt struct.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_capabilities.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_capabilities.py
@@ -1,10 +1,8 @@
 import logging
-import re
 
 from autotest.client import os_dep
 from autotest.client.shared import error
 
-from avocado.utils import path as utils_path
 from avocado.utils import process
 
 from virttest import libvirt_vm
@@ -27,7 +25,7 @@ def run(test, params, env):
 
         # Check that host has a non-empty UUID tag.
         xml_uuid = cap_xml.uuid
-        logging.debug("Host UUID (capabilities_xml): %s" % xml_uuid)
+        logging.debug("Host UUID (capabilities_xml): %s", xml_uuid)
         if xml_uuid == "":
             raise error.TestFail("The host uuid in capabilities_xml is none!")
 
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_capabilities.py
```diff
@@ -36,59 +34,58 @@ def compare_capabilities_xml(source):
...
-            try:
-                # For Ubuntu /usr/bin/kvm is available instead of qemu-kvm
-                img = utils_path.find_command("kvm")
-            except utils_path.CmdNotFoundError:
-                raise error.TestNAError("Cannot find {qemu-}kvm binary in host")
-        if re.search("ppc", process.run("arch", shell=True).stdout):
-            cmd = img + " --cpu ? | grep ppc"
-        else:
-            cmd = img + " --cpu ? | grep qemu"
-        cmd_result = process.run(cmd, shell=True)
```
Hao-Liu: Since the guests shown in `virsh capabilities` supported by QEMU is directly correlate with CPU supported by qemu command. Is it possible to retain this test point in this test case?

libvirt/tests/src/virsh_cmd/host/virsh_capabilities.py
```diff
@@ -36,59 +34,58 @@ def compare_capabilities_xml(source):
...
+        domain_type = "domain_" + uri_type
+        for arch_dict in guest_capa.values():
+            for arch, val_dict in arch_dict.items():
+                # Check wordsize
+                if arch not in wordsize[val_dict['wordsize']]:
+                    raise error.TestFail("'%s' wordsize '%s' in "
+                                         "capabilities_xml not expected" %
+                                         (arch, val_dict['wordsize']))
+                # Check the type of hypervisor
+                if arch == host_arch:
```
Hao-Liu: I thinks guest arch don't have relation with host_arch but only with supported CPU arch by emulator

waynesun09: Yes, arch is not related.
The test point here is check the default hypervisor value (domain type value in guest capabilities xml).

For all arches, each one have this value, and match with uri result, I only check the matching arch with host, which is similar to original design. I'm fine to extend it to check on each arch though.

waynesun09: Updated to check under all archs

PR #937, opened 2016-12-01
migrate_vm: Load stress tool on host

dzhengfy:
This fix is to load stress tool on the host before migration with
stress on host.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1778,6 +1778,8 @@ def run(test, params, env):
             elif s_list and s_list[-1] == "host":
                 logging.info("Run '%s %s' in %s", s_list[0],
                              stress_args, s_list[-1])
+                pkg_list = ['stress']
+                libvirt.yum_install(pkg_list)
                 err_msg = utils_test.load_stress(stress_type,
                                                  [vm], test_dict)
                 if len(err_msg):


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -39,6 +43,28 @@
 MIGRATE_RET = False
 
 
+def prepare_stress_tool(params):
+
+    rpm_utils = RpmBackend()
+    if rpm_utils.check_installed('stress'):
+        return
+    os_ver_cmd = 'cat /etc/redhat-release'
```
balamuruhans: Is this stress test will only works for RHEL, then we have to skip the test for other distros. if it can be used for other distros then /etc/os-release can be used

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1778,6 +1804,7 @@ def run(test, params, env):
             elif s_list and s_list[-1] == "host":
                 logging.info("Run '%s %s' in %s", s_list[0],
                              stress_args, s_list[-1])
+                prepare_stress_tool(params)
```
balamuruhans: For a single variable "stress_download_url", its better to retrieve and pass the url to stress_download_url(url), instead of entire params

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -39,6 +43,28 @@
 MIGRATE_RET = False
 
 
+def prepare_stress_tool(params):
```
balamuruhans: docstring can be included

dzhengfy:
This pr depends on https://github.com/avocado-framework/avocado-vt/pull/806/

dzhengfy:
@will-Do , as we discussed, 'stress' package would be added into our repo, so I closed https://github.com/avocado-framework/avocado-vt/pull/806  and updated this PR, please review, thanks.

PR #939, opened 2016-12-02
domstate: Fix coredump file path

waynesun09:
The vm coredump name updated with start with number, update the
regex for 'ls' command, which also works on prior libvirt.


balamuruhans:
commit msg have a typo, %s/udpated/updated, otherwise LGTM

waynesun09:
@balamuruhans thx, updated

PR #945, opened 2016-12-05
attach_device: Handle devidx based on guest disks

harish-24:
Get the index of the attaching disk based on existing disks of guest in VirtualDiskBasic class.


waynesun09:
For commit message:
- Limit the subject line to 50 characters
- Wrap the body at 72 characters

As you are fixing the code, you don't need start with "Fix:"
So I suggest update your commit title to:

attach_device: Handle the devidx based on existing disks of guest

also no need add ".py" after the test module, the test module name will be enough.
Thanks

harish-24:
@waynesun09 Updated. Thanks.

PR #953, opened 2016-12-09
libvirt_mem: Add without_numa as positive test

harish-24:
Guest is now able to boot without numa, moving it as a positive test case in libvirt_mem.cfg

Signed-off-by: Harish harisrir@linux.vnet.ibm.com


libvirt/tests/cfg/libvirt_mem.cfg
```diff
@@ -71,6 +71,8 @@
                             numa_memnode = "{'cellid':'0','mode':'interleave','nodeset':'0'}"
                             memory_addr = "{'type':'dimm','slot':'0','base':'0x100000000'}"
                             node_mask = 0
+                - without_numa_save_restore:
```
will-Do: Just wonder the suffix(''_save_restore') here

harish-24: added _save_restore as part of what the test case does

PR #954, opened 2016-12-09
Add v2v function case and update code of check log

chloerh:
Add case:
- with less than 1G space left on conversion server
- with spice/vnc and cirrus/qxl mix match

Update code:
- move error message content to cfg file controlled by 2 params to
  make check log function more clear
- modify function of create large file to adapt to the need of
  creating large file in host
- replace 'elif' with 'if'

### DIFF
--- a/v2v/tests/src/specific_kvm.py
+++ b/v2v/tests/src/specific_kvm.py
@@ -3,7 +3,8 @@ import re
 import logging
 import string
 
-from avocado.core import exceptions
+import aexpect
+
 from avocado.utils import service
 from avocado.utils import process
 
@@ -13,6 +14,7 @@ from virttest import utils_misc
 from virttest import utils_sasl
 from virttest import libvirt_vm
 from virttest import data_dir
+from virttest import utils_package
 from virttest import utils_selinux
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt as utlv
@@ -26,7 +28,7 @@ def run(test, params, env):
     """
     for v in params.itervalues():
         if "V2V_EXAMPLE" in v:
-            raise exceptions.TestSkipError("Please set real value for %s" % v)
+            test.skip("Please set real value for %s" % v)
     if utils_v2v.V2V_EXEC is None:
         raise ValueError('Missing command: virt-v2v')
     vm_name = params.get('main_vm', 'EXAMPLE')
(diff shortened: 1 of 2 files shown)


v2v/tests/src/specific_kvm.py
```diff
@@ -354,19 +361,20 @@ def specify_fstab_entry(type, **kwargs):
                 cmd_fstab = "sed -i 's|%s|%s|' /etc/fstab" % (origin, replace)
                 session.cmd(cmd_fstab)
 
-    @vm_shell
-    def create_large_file(**kwargs):
+    def create_large_file(session, left):
         """
-        Create a large file to make left space of root less than 20m
+        Create a large file to make left space of root less than $left MB
```
will-Do: Could you please pick up a more readable var name instead of 'left', and desc it more clear

v2v/tests/src/specific_kvm.py
```diff
@@ -354,19 +361,20 @@ def specify_fstab_entry(type, **kwargs):
...
-        if avail > 19:
-            params['large_file'] = '/file.large'
+        if avail > left - 1:
+            tmp_dir = data_dir.get_tmp_dir()
+            if session.cmd_status('ls %s' % tmp_dir) != 0:
+                session.cmd('mkdir %s' % tmp_dir)
+            large_file = os.path.join(tmp_dir, 'file.large')
             cmd_create = 'dd if=/dev/zero of=%s bs=1M count=%d' % \
-                         (params['large_file'], avail - 18)
+                         (large_file, avail - left + 2)
```
will-Do: what's the strategy here?

chloerh: In order to make left space less than Xm, a large file with size (X-1)m should be created. But sometimes the amount of the left space is not quite accurate by Megabites, therefore, create a file 1M larger to be safer.

waynesun09:
so many if with checkpoints @.@, LGTM though

v2v/tests/src/specific_kvm.py
```diff
@@ -583,39 +550,56 @@ def check_result(result, status_error):
...
-            check_v2v_log(output, checkpoint)
             # Merge 2 error lists
             error_list.extend(vmchecker.errors)
-            if len(error_list):
-                log_fail('%d checkpoints failed: %s' % (len(error_list), error_list))
+        log_check = utils_v2v.check_log(params, output)
+        if log_check:
+            log_fail(log_check)
+        if len(error_list):
+            raise exceptions.TestFail('%d checkpoints failed: %s' %
```
chloerh: Thanks for reminding

PR #958, opened 2016-12-12
virsh.freepages: Fix fail when node without memory

waynesun09:
For some ppc machines, not all nodes are provided with memory,
freepages command will fail when with '--all' in such situation,
which I think is acceptable.

Fix this fail situation by update status_error value.


libvirt/tests/src/virsh_cmd/host/virsh_freepages.py
```diff
@@ -124,6 +126,11 @@ def run(test, params, env):
                                      pagesize=page,
                                      options=option,
                                      debug=True)
+
+            # the node without memory will fail command and is expected
```
balamuruhans: the node without memory will fail and it is expected

waynesun09:
@balamuruhans updated, thx

balamuruhans:
ACK

Hao-Liu:
Thanks @waynesun09 @balamuruhans, Merging this.

PR #962, opened 2016-12-16
Fix uri test issues

kylazhang:
1. change the target_uri get from param because complete_uri also call virsh uri, we can not use virsh to verify virsh
2. resolve the local uri execute issues, the previous change cause local execute disappear.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/host/virsh_uri.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/host/virsh_uri.cfg
@@ -15,6 +15,7 @@
                     remote_ip = "REMOTE.EXAMPLE.COM"
                     remote_user = "root"
                     remote_pwd = "password"
+                    target_uri = "qemu+ssh://${remote_ip}/system"
         - unexpect_option:
             virsh_uri_options = "xyz"
             status_error = "yes"
--- a/libvirt/tests/src/virsh_cmd/host/virsh_uri.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_uri.py
@@ -1,8 +1,7 @@
 import logging
 
-from autotest.client.shared import error
-
 from avocado.utils import process
+from avocado.core import exceptions
 
 from virttest import libvirt_vm
 from virttest import virsh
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_uri.py
```diff
@@ -30,10 +30,10 @@ def run(test, params, env):
...
 
     # Forming the uri using the api
-    target_uri = libvirt_vm.complete_uri(params.get("remote_ip"))
+    target_uri = params.get("target_uri")
     remote_ref = params.get("uri_remote_ref", "")
 
-    if target_uri:
+    if remote_ref:
         if target_uri.count('EXAMPLE.COM'):
             raise error.TestNAError(
```
balamuruhans: exceptions from can be used instead of error

libvirt/tests/cfg/virsh_cmd/host/virsh_uri.cfg
```diff
@@ -15,6 +15,7 @@
                     remote_ip = "REMOTE.EXAMPLE.COM"
                     remote_user = "root"
                     remote_pwd = "password"
+                    target_uri = "qemu+ssh://${remote_ip}/system"
```
balamuruhans: you can have a look at https://github.com/avocado-framework/avocado-vt/pull/795 and form the uri dynamically instead of using it from config

kylazhang: Thanks, I will check it and update.

libvirt/tests/src/virsh_cmd/host/virsh_uri.py
kylazhang: ok, thanks for remind.

libvirt/tests/cfg/virsh_cmd/host/virsh_uri.cfg
kylazhang: Hi @balamuruhans you change can not met my requirements, we do not use port here and I do not want to use virsh.canonical_uri because we can not use virsh to verify virsh. The target_uri here is not only the connect_uri but also the compare uri for result check.

kylazhang:
@will-Do please help check and merge, thanks

PR #963, opened 2016-12-23
Enable --postcopy options for migration cases

dzhengfy:
Using --postcopy option when migration should not impact the migration
result. This is to enable this option for most existing migration cases.


dzhengfy:
@will-Do could you help review this item?

balamuruhans:
@dzhengfy we already have "postcopy_migration" test case enabled for virsh_migrate.py covering both --postcopy and --postcopy-after-precopy, we could have reused it by bringing it to 1 level up

PR #968, opened 2016-12-29
migrate_vm: Fix domjobinfo comparison

dzhengfy:
It is unreasonable that the domjobinfo when completion on source host
should be equal to that on target host because some fields may have
differences and other values should be equal on both sides.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -865,6 +865,97 @@ def check_iothread_after_migration(vm_name, params, iothread):
         remote_virsh.close_session()
 
 
+def check_domjobinfo_on_complete(test, source_jobinfo, target_jobinfo):
+    """
+    Compare the domjobinfo outputs on source and target hosts.
+    The domjobinfo currently includes below items:
+     Job type
+     Operation
+     Time elapsed
+     Time elapsed w/o network
+     Data processed
+     Data remaining
+     Data total
+     Memory processed
+     Memory remaining
+     Memory total
+     Memory bandwidth
+     Dirty rate
+     Iteration
+     Constant pages
+     Normal pages
+     Normal data
+     Total downtime
+     Downtime w/o network
+     Setup time
+
+    Most fields are required to be same between source and target hosts,
+    except below:
+        Operation
+        Time elapsed
+        Time elapsed w/o network
+
+    :param test: avocado.core.test.Test object
+    :param local_jobinfo: The domjobinfo output on source host
+    :param remote_jobinfo: The domjobinfo output on target host
+    """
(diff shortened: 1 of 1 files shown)


dzhengfy:
@will-Do please help review this item

dzhengfy:
@waynesun09 , could you help review this?

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -865,6 +865,78 @@ def check_iothread_after_migration(vm_name, params, iothread):
...
 
 
+def check_domjobinfo_on_complete(test, source_jobinfo, target_jobinfo):
+    """
+    Check the domjobinfo outputs on source and target hosts.
+    They should be same except following fields:
+        Time elapsed:
+        Time elapsed w/o network
+    Above values on source host should be larger than that on
+    target host with about 2 seconds
```
will-Do: Could you please declare the above desc more clear

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -865,6 +865,78 @@ def check_iothread_after_migration(vm_name, params, iothread):
...
+    :param test: avocado.core.test.Test object
+    :param local_jobinfo: The domjobinfo output on source host
+    :param remote_jobinfo: The domjobinfo output on target host
+    """
+    source_info = read_domjobinfo(test, source_jobinfo)
+    target_info = read_domjobinfo(test, target_jobinfo)
+
+    for pair in source_info.items():
+        key = pair[0]
+        value = pair[1]
```
will-Do: key, value = pair

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -865,6 +865,78 @@ def check_iothread_after_migration(vm_name, params, iothread):
...
+    """
+    Read the domjobinfo into a dict
+
+    :param test: avocado.core.test.Test object
+    :param domjobinfo: The domjobinfo command output
+
+    :return: A dict contains the domjobinfo
+    """
+    jobinfo_dict = {}
+    domjobinfo_list = domjobinfo.split("\n")
```
will-Do: splitlines() doese the same thing

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @will-Do updated.

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @will-Do updated.

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @will-Do updated.

PR #972, opened 2017-01-05
virsh_migrate_copy_storage: Check the pool firstly

dzhengfy:
The fix will check the existence of the target pool on remote host
before creating it. If it already exists, we prefer to reusing it
instead of recreating.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
@@ -39,6 +39,15 @@ def create_destroy_pool_on_remote(action, params):
     prompt = params.get("prompt", r"[\#\$]\s*$")
 
     if action == 'create':
+        # Firstly check if the pool already exists
+        all_pools = new_session.pool_list(option="--all")
+        logging.debug("Pools on remote host:\n%s", all_pools)
+        if all_pools.stdout.find(pool_name) >= 0:
+            logging.debug("The pool %s already exists and skip "
+                          "to create it.", pool_name)
+            new_session.close_session()
+            return True
+
         pool_type = params.get("precreation_pool_type", "dir")
         pool_target = params.get("precreation_pool_target")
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
```diff
@@ -39,6 +39,15 @@ def create_destroy_pool_on_remote(action, params):
     prompt = params.get("prompt", r"[\#\$]\s*$")
 
     if action == 'create':
+        # Firstly check if the pool already exists
+        all_pools = new_session.pool_list(option="--all")
+        logging.debug("Pools on remote host:\n%s" % all_pools)
```
balamuruhans: `logging.debug("Pools on remote host:\n%s", all_pools)`

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
```diff
@@ -39,6 +39,15 @@ def create_destroy_pool_on_remote(action, params):
...
+        if all_pools.stdout.find(pool_name) >= 0:
+            logging.debug("The pool %s already exists and skip to create it."
+                          % pool_name)
```
balamuruhans: same here

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
dzhengfy: @balamuruhans Thanks and I have updated them.

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
dzhengfy: @balamuruhans Thanks and I have updated them.

balamuruhans:
@dzhengfy Thanks for the patch. @nasastry Thanks for review.

PR #973, opened 2017-01-06
Skipping test cases which needs host SELinux to be Enforcing.

sudeeshjohn:
In Ubuntu SELinux is not working at the moment. So Skipping testcases which needs the SELinux to be in Enforcing mode. This patch will skip(insted of ERRORing out) the the testcase if SELinux is disabled.

### DIFF
--- a/libvirt/tests/src/svirt/dac_nfs_disk.py
+++ b/libvirt/tests/src/svirt/dac_nfs_disk.py
@@ -112,6 +112,8 @@ def run(test, params, env):
 
         # Set selinux of host.
         utils_selinux.set_status(host_sestatus)
+        if backup_sestatus == "disabled":
+            raise error.TestNAError("SKIP: SELinux is set to enforcing")
 
         # set qemu conf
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/svirt/dac_nfs_disk.py
```diff
@@ -112,6 +112,8 @@ def run(test, params, env):
 
         # Set selinux of host.
         utils_selinux.set_status(host_sestatus)
+        if backup_sestatus == "disabled":
+            raise error.TestNAError("SKIP: SELinux is set to enforcing")
```
sathnaga: use avocado.exceptions instead of error, TestNAError-->TestSkipError.
Message should convey, test skipped because of it disabled, rather than "SELinux is set to enforcing"?

sudeeshjohn: @sathnaga , sure, I'll do that.

sudeeshjohn:
New PR #974. closing this.

PR #975, opened 2017-01-16
Support to test macvtap migration

balamuruhans:
This patch adds support to test VMs with macvtap configured and
to perform migration to validate the macvtap network connectivity
before and after migration

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
@@ -1,15 +1,18 @@
 import logging
 import threading
 import time
+import os
 
 from avocado.core import exceptions
 
+from virttest import data_dir
 from virttest import libvirt_vm
 from virttest import virsh
 from virttest import remote
 from virttest import utils_test
 from virttest import nfs
 from virttest import ssh_key
+from virttest import utils_net
 from virttest.libvirt_xml import vm_xml
 
 
@@ -20,11 +23,15 @@ global ret_migration
 global ret_jobabort
 # If downtime is tolerable
 global ret_downtime_tolerable
+# List of vms static ip address list
+global vm_ip_check_list
+
 # True means command executed successfully
 ret_migration = True
 ret_jobabort = True
 ret_downtime_tolerable = True
 flag_migration = True
+vm_ip_check_list = []
 
 
 def make_migration_options(method, optionstr="", timeout=60):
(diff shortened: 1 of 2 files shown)


balamuruhans:
This PR depends on https://github.com/avocado-framework/avocado-vt/pull/825

will-Do:
@dzhengfy
Please help to review this patch

dzhengfy:
Ok, I will have a look at it. @will-Do

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
```diff
@@ -174,6 +181,222 @@ def multi_migration(vm, src_uri, dest_uri, options, migrate_type,
...
+    """
+    logging.info("Cleaning up VMs on %s" % dest_uri)
+    try:
+        if virsh.domain_exists(vm.name, uri=dest_uri):
+            vm_state = vm.state()
+            if vm_state == "paused":
+                vm.resume()
+            elif vm_state == "shut off":
+                vm.start()
+            vm.destroy(gracefully=False)
```
nasastry: Didn't get why there is a need to start the 'vm' when it is in 'shut off' state and then destroying it?

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
```diff
@@ -174,6 +181,222 @@ def multi_migration(vm, src_uri, dest_uri, options, migrate_type,
...
+
+        # mac address of macvtap interface
+        new_mac = vmxml_backup.get_devices(device_type="interface")[-1]
+        new_mac = new_mac.mac_address
+        logging.debug("mac address of macvtap interface - %s", new_mac)
+
+        # network script for base interface
+        network_script = utils_net.get_vm_network_script(vm, vm_iface, mac,
+                                                         'dhcp',
+                                                         '255.255.255.0')
```
nasastry: Is there a chance to get a different 'netmask'?

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
```diff
@@ -269,31 +528,83 @@ def run(test, params, env):
...
-        if not status_error:
-            raise exceptions.TestFail("Abort migration failed")
-    if not ret_downtime_tolerable:
-        raise exceptions.TestFail("Downtime during migration is intolerable")
+    finally:
+        # NFS cleanup
+        if nfs_shared_disk:
+            logging.info("NFS cleanup")
+            nfs_client.cleanup(ssh_auto_recover=False)
+
```
dzhengfy: I did not see any code to setup/cleanup nfs server, but only nfs client setup and cleanup. Can that migration work? Or I miss something?

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
```diff
@@ -174,6 +181,222 @@ def multi_migration(vm, src_uri, dest_uri, options, migrate_type,
...
+    """
+    global vm_ip_check_list
+    session = macvtap_vm.wait_for_login()
+    for each_macvtap_ip in vm_ip_check_list:
+        ret, out = utils_test.ping(each_macvtap_ip, count=5, timeout=5,
+                                   session=session)
+        if ret:
+            raise exceptions.TestFail("From %s failed to ping-%s: %s" %
+                                      (macvtap_vm.name, each_macvtap_ip, out))
+
```
dzhengfy: The vepa_test is called twice, before and after migration.  I do not get the idea for this function. Which VM does each_macvtap_ip belong to? And do you intent to ping from one guest to another guest? If yes, before migration, the ping should fail because they are in same host using direct network. Anyway, the ping should be ok after migration if you use one guest on source host to ping the migrated vm on target host.
So it is appreciated if you can explain in more details.

balamuruhans: > The vepa_test is called twice, before and after migration. I do not get the idea for this function. Which VM does each_macvtap_ip belong to? And do you intent to ping from one guest to another guest?

So what I am trying to do is, 
1. User can have min 2 VMs to any number of VMs
2. Configure macvtap interface to all the guest and assign IP for that interface so that all the VMs are in same subnet.
3. we will have 1 VM in source host and wont migrate it and ping from this VM to other VMs before migration (`each_macvtap_ip` belongs to these VMs which are going to migrate) to validate the macvtap is configured and working.
4. Except 1 VM in source host, perform migration of all other VMs to destination host and ping the migrated VMs from the VM from source host will be our validation statement

> If yes, before migration, the ping should fail because they are in same host using direct network. Anyway, the ping should be ok after migration if you use one guest on source host to ping the migrated vm on target host.
> So it is appreciated if you can explain in more details.

Sure, Even if the guests are in same host, the ping will succeed before migration as vepa configured switch will have hair-pin routing configured, so ping packet will go through direct network of host, goes outside to VEPA switch and comes back the the host and reach the other guest. ideally ping should work before as well as after migration which is the validation point for this test case.

dzhengfy: @balamuruhans thanks for your so many detailed comments. But the cases still need the switch configured with hair-pin mode to make the ping successful before and after the migration. I don't think switch with hair-pin mode exists in common test environments, for example, we don't have it. The cases will always fail in my side then. So that's my concern. However, the codes look good to me.

sathnaga: @balamuruhans @dzhengfy I hope a doc string of this test captures the pre-requisite would be a good idea to make aware of the usage

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
balamuruhans: okay, so I have made several network configuration changes, during clean up I will revert the settings and ensure vm gets started with proper IP address as how it starts before the test, so better I will use `vm.wait_for_login()` instead of just `vm.start()`. 
This is a good catch @nasastry Thank you.

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_multi_vms.py
balamuruhans: @dzhengfy Thank you for the review and sharing the thoughts. As @sathnaga suggests lets have doc string for the pre-requisite.

iccaszhulili:
rh-caselink retest please

chunfuwen:
@balamuruhans , do you still need this PR ?, otherwise we would like to close it  firstly since it is pending for long time

dzhengfy:

PR #982, opened 2017-02-06
libvirtd_start.py: Added autostart of inactive storage pools to the list of accepted errors.

santwana:
This patch fixes the issue of libvirtd service failing due to autostart of inactive storage pools,
which is irrelevant to the current test-cases of iptables/firewalld start and stop along with libvirtd.

Also modified the testcase to SKIP where Selinux needs to be in enforcing mode, as its not supported
on certain distributions like Ubuntu.

### DIFF
--- a/libvirt/tests/src/libvirtd_start.py
+++ b/libvirt/tests/src/libvirtd_start.py
@@ -38,6 +38,8 @@ def run(test, params, env):
         logging.info('Checking errors in libvirtd log')
         accepted_error_patterns = [
             'Cannot access storage file',
+            'Failed to autostart storage pool',
+            'cannot open directory',
         ]
 
         if (not iptables_service and not firewalld_service and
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirtd_start.py
```diff
@@ -86,9 +88,9 @@ def _check_errors():
     backup_iptables_status = service_mgr.status('iptables')
     backup_firewalld_status = service_mgr.status('firewalld')
 
-    # iptables service should always exists
+    # iptables service does not exists on many distros
     if iptables_service and backup_iptables_status is None:
-        raise exceptions.TestError('iptables service not found')
+        raise exceptions.TestSkipError('iptables service not found')
```
balamuruhans: it will be good to skip test based on distro which doesn't support iptables and it should be TestError if iptables are not available in other case

santwana: By default RHEL7 comes with firewalld instead of iptables-services, also in Ubuntu there is no iptables-service, so skipping the test if iptables-service is not present.

balamuruhans: Thanks for clarifying

libvirt/tests/src/libvirtd_start.py
```diff
@@ -86,9 +88,9 @@ def _check_errors():
     backup_iptables_status = service_mgr.status('iptables')
     backup_firewalld_status = service_mgr.status('firewalld')
 
-    # iptables service should always exists
+    # iptables service does not exists on many distros
```
balamuruhans: Then can it be "# iptables-service got deprecated in newer distros"

balamuruhans:
Looks good to me

PR #990, opened 2017-02-22
virsh_migrate: avoid guest to pin its numa with host numa having low memory

balamuruhans:
If host numa doesn't have enough memory to guest, then skip to use other
numa node to avoid test failure.


balamuruhans:
@sathnaga Thanks for review

PR #991, opened 2017-02-28
virsh_net_create: Catch LibvirtXMLError

will-Do:
It will raise LibvirtXMLError when create network by an invalid XML.


will-Do:
@sathnaga THX

PR #993, opened 2017-03-01
virt_top.py: Added option to install "virt-top" package, if missing.

santwana:
This patch enables the test-case to install the "virt-top" package, before test-case execution.
It's ideal to initially check for the package installed and later the command present, before execution.

### DIFF
--- a/libvirt/tests/src/virt_cmd/virt_top.py
+++ b/libvirt/tests/src/virt_cmd/virt_top.py
@@ -1,8 +1,10 @@
 import os
+import logging
 
 from autotest.client import os_dep
 from autotest.client import utils
 from autotest.client.shared import error
+from avocado.utils import software_manager
 
 from virttest import virsh
 from virttest import data_dir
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virt_cmd/virt_top.py
```diff
@@ -1,8 +1,10 @@
 import os
+import logging
 
 from autotest.client import os_dep
 from autotest.client import utils
 from autotest.client.shared import error
```
balamuruhans: you can use this PR to replace autotest APIs to Avocado APIs too,
`exceptions` can be used instead of `error`

libvirt/tests/src/virt_cmd/virt_top.py
```diff
@@ -13,6 +15,13 @@ def run(test, params, env):
     Test for virt-top, it is a top like tool for virtual
     machine.
     """
+    # Install virt-top package if missing.
+    software_mgr = software_manager.SoftwareManager()
+    for pkg in ['virt-top']:
```
santwana: Made changes as suggested.

sathnaga:
@santwana LGTM, merged.
As @balamuruhans suggested you can wish to create a new PR for clean up autotest imports to avocado. :-)

PR #996, opened 2017-03-03
In Ubuntu there is no "nobody" group, "nogroup" instead.

sudeeshjohn:
Ubuntu used "nogroup" group name which is equivalent to "nobody" group in Redhat distros. 

### DIFF
--- a/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
+++ b/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
@@ -2,6 +2,7 @@ import grp
 import stat
 import logging
 import os
+import platform
 
 from autotest.client.shared import error
 
@@ -134,6 +135,11 @@ def run(test, params, env):
     rw_perms = params.get('unix_sock_rw_perms', '0777')
     path = params.get('unix_sock_dir', '/var/run/libvirt')
     expected_result = params.get('expected_result', 'success')
+    # In Ubuntu there is no group called "nobody", "nogroup" instead.
+    if group == 'nobody':
+        distro = platform.dist()
+        if 'Ubuntu' in distro:
+            group = 'nogroup'
     try:
         # Change params in libvirtd.conf
         config.unix_sock_group = group


libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
```diff
@@ -2,6 +2,7 @@
 import stat
 import logging
 import os
+import platform
```
sathnaga:
```
from avocado.utils import distro
 distro_details = distro.detect()
if distro_details.name == 'Ubuntu':

```

Instead this avocado library can be used

libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
```diff
@@ -134,6 +135,11 @@ def check_all_unix_sock(group, ro_perms, rw_perms, root_path):
     rw_perms = params.get('unix_sock_rw_perms', '0777')
     path = params.get('unix_sock_dir', '/var/run/libvirt')
     expected_result = params.get('expected_result', 'success')
+    # In Ubuntu there is no group called "nobody", "nogroup" instead.
+    if group == 'nobody':
```
balamuruhans: I feel this is not right way to fix it, if user give other value in params it might not work.
instead,
```
if 'Ubuntu' in platform.dist():
    group = params.get('unix_sock_group', 'nogroup')
else:
    group = params.get('unix_sock_group', 'nobody')
```

libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
```diff
@@ -2,6 +2,7 @@
 import stat
 import logging
 import os
+import platform
 
 from autotest.client.shared import error
```
balamuruhans: please take this opportunity to use `exceptions` from `avocado` instead of `error`

libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
balamuruhans: negative scenarios can be affected with current fix

sudeeshjohn: if 'Ubuntu' in platform.dist():
    group = params.get('unix_sock_group', 'nogroup')

with this,  it is just setting the default value to 'nogroup' right ?.

balamuruhans: yes, we have to make corresponding changes in cfg file as well.
Let tests related to nobody or nogroup can take default value from script based on distro,
other tests can take from cfg file, which makes the tests generic.

sudeeshjohn:
https://github.com/autotest/tp-libvirt/pull/1007

PR #999, opened 2017-03-06
Add the case of server-clients-set for virt-admin

iccaszhulili:

### DIFF
--- a/libvirt/tests/cfg/virt_admin/management/virt_admin_server_clients_set.cfg
+++ b/libvirt/tests/cfg/virt_admin/management/virt_admin_server_clients_set.cfg
@@ -1,0 +1,25 @@
+- virt_admin.server_clients_set:
+    type = virt_admin_server_clients_set
+    start_vm = no
+    server_name = "libvirtd"
+    variants:
+        - normal_test:
+            variants:
+                - max_clients_test:
+                    options_ref = "--max-clients"
+                    variants:
+                        - max_clients_gt_nclients:
+                            nclients_maxi = 3000
+                            nclients = 0
+                            nclients_unauth_maxi = 2
+                            connect_able = yes
+                        - max_clients_eq_nclients:
+                            nclients_maxi = 2
+                            nclients = 2
+                            nclients_unauth_maxi = 2
+                            connect_able = yes
+                        - max_clients_lt_nclients:
+                            nclients = 3
+                            nclients_maxi = 2
+                            nclients_unauth_maxi = 2
+                            connect_able = no
--- a/libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
+++ b/libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
@@ -1,0 +1,90 @@
+import logging
+from virttest import virt_admin
+from virttest import virsh
+from virttest import utils_libvirtd
+from virttest import utils_config
+
+
+def run(test, params, env):
+    """
+    Test virt-admin  server-clients-set
+    2) Change max_clients to a new value;
+    3) get the current clients info;
+    4) check whether the clients info is correct;
+    5) try to connect other client onto the server;
+    6) check whether the above connection status is correct.
+    """
+
+    server_name = params.get("server_name")
+    options_ref = params.get("options_ref")
+    nclients_max = params.get("nclients_maxi", "5000")
+    nclients = params.get("nclients", "0")
+    nclients_unauth_max = params.get("nclients_unauth_maxi", "")
+    connect_able = params.get("connect_able", "yes")
+
+    config = utils_config.LibvirtdConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    virsh_instance = []
+
+    def clients_info(server):
+        """
+        check the attributes by server-clients-set.
+        1) get the output  returned by server-clients-set;
+        2) split the output to get a dictionary of those attributes;
+        :params server: print the info of the clients connecting to this server
+        :return: a dict obtained by transforming the result_info
+        """
+        result_info = vp.srv_clients_info(server, ignore_status=True,
+                                          debug=True)
+        out = result_info.stdout.strip().splitlines()
+        out_split = [item.split(':') for item in out]
+        out_dict = dict([[item[0].strip(), item[1].strip()] for item in out_split])
+        return out_dict
+
+    try:
+        if "max_clients" in options_ref:
+            if int(nclients_max) > int(nclients):
+                vp = virt_admin.VirtadminPersistent()
+                result = vp.srv_clients_set(server_name, max_clients=nclients_max,
+                                            ignore_status=True, debug=True)
+            elif int(nclients_max) == int(nclients):
+                config.max_clients = nclients_max
+                config.max_anonymous_clients = nclients_unauth_max
+                libvirtd.restart()
+                vp = virt_admin.VirtadminPersistent()
+                nclients_max_set = int(nclients_max) + 3
+                result = vp.srv_clients_set(server_name, max_clients=nclients_max_set,
+                                            ignore_status=True, debug=True)
+            else:
+                vp = virt_admin.VirtadminPersistent()
+                for _ in range(int(nclients)):
+                    virsh_instance.append(virsh.VirshPersistent(uri='qemu:///system'))
+                result = vp.srv_clients_set(server_name, max_clients=nclients_max,
+                                            max_unauth_clients=nclients_unauth_max,
+                                            ignore_status=True, debug=True)
+
+            outdict = clients_info(server_name)
+            if result.exit_status:
+                test.fail("This operation should success "
+                          "but failed! output:\n%s " % result)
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+    virsh_instance = []
+
+    def clients_info(server):
+        """
+        check the attributes by server-clients-set.
+        1) get the output  returned by server-clients-set;
+        2) split the output to get a dictionary of those attributes;
+        """
+        result_info = vp.srv_clients_info(server, ignore_status=True,
+                                          degbug=True)
```
balamuruhans: %s/degbug/debug

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+    nclients_max = params.get("nclients_maxi", "5000")
+    nclients = params.get("nclients", "0")
+    nclients_unauth_max = params.get("nclients_unauth_maxi", "")
+    connect_able = params.get("connect_able", "yes")
+
+    config = utils_config.LibvirtdConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    virsh_instance = []
+
+    def clients_info(server):
```
balamuruhans: threadpool_info(server) in #1000 and clients_info(server) looks similar, you can keep this as API in avocado-vt to use it from tests to avoid redundancy,
```
def check_virtadmin_attributes(server, func):
    try:
        result_info = func(server, ignore_status=True,
                           debug=True)
        out = result_info.stdout.strip().splitlines()
        out_split = [item.split(':') for item in out]
        out_dict = dict([[item[0].strip(), item[1].strip()] for item in out_split])
        return out_dict
    except Exceptions, info:
        exceptions.TestError(msg)
```

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+                for _ in range(int(nclients)):
+                    virsh_instance.append(virsh.VirshPersistent(uri='qemu:///system'))
+                result = vp.srv_clients_set(server_name, max_clients=nclients_max,
+                                            max_unauth_clients=nclients_unauth_max,
+                                            ignore_status=True, debug=True)
+
+            outdict = clients_info(server_name)
+            if result.exit_status:
+                raise exceptions.TestFail("This operation should success "
+                                          "but failed! output:\n%s ", result)
```
balamuruhans: For exceptions,
```
raise exceptions.TestFail("This operation should success "
                          "but failed! output:\n%s " % result)
```

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+            elif outdict["nclients_max"] != str(nclients_max_set):
+                raise exceptions.TestFail("attributes set by server-clients-set "
+                                          "is not correct!")
+
+            try:
+                virsh_instance.append(virsh.VirshPersistent(uri='qemu:///system'))
+            except Exception, info:
+                if connect_able == "yes":
+                    raise exceptions.TestFail("Connection not success, error:\n %s", info)
```
balamuruhans: same here

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
iccaszhulili: @balamuruhans Thx very much :)
Yeah, I will consider about this issue.
However, maybe I will not do this right now.
I had discussed this with my mentor in our team.
Maybe we would carry out some large scale modifications on this kinds of issues in the near future.

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
iccaszhulili: @balamuruhans This issue and the following two ones have been modified according to your comments. Thanks very much.

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
balamuruhans: @will-Do @waynesun09 please share your comments on this

will-Do: @balamuruhans 
If the require of parse the command output is not such frequently, we can keep it in the test cases, it's not forced. And along with the requirement, we can abstract it and move to upper layer.
About the upper layer, maybe we can keep it in tp-libvirt instead of avocado-vt, as such usage is only related to libvirt, and the avocado-vt is quite bloated for its backends.

balamuruhans: @will-Do okay, Thanks for the clarification.

libvirt/tests/cfg/virt_admin/management/virt_admin_server_clients_set.cfg
```diff
@@ -0,0 +1,26 @@
...
+                            nclients_maxi = 2
+                            nclients = 2
+                            nclients_unauth_maxi = 2
+                            connect_able = yes
+                        - max_clients_lt_nclients:
+                            nclients = 3
+                            nclients_maxi = 2
+                            nclients_unauth_maxi = 2
+                            connect_able = no
+
```
balamuruhans: please remove this empty line,
```
warning: 1 line adds whitespace errors.
```

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+    config = utils_config.LibvirtdConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    virsh_instance = []
+
+    def clients_info(server):
+        """
+        check the attributes by server-clients-set.
+        1) get the output  returned by server-clients-set;
+        2) split the output to get a dictionary of those attributes;
+        """
```
balamuruhans: docstring for param and return can be added

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,88 @@
...
+            if int(nclients_max) > int(nclients):
+                vp = virt_admin.VirtadminPersistent()
+                result = vp.srv_clients_set(server_name, max_clients=nclients_max,
+                                            ignore_status=True, debug=True)
+            elif int(nclients_max) == int(nclients):
+                config.max_clients = nclients_max
+                config.max_anonymous_clients = nclients_unauth_max
+                libvirtd.restart()
+                vp = virt_admin.VirtadminPersistent()
+                nclients_max_set = int(nclients_max) + 3
```
balamuruhans: why do we add 3 by hard-coding it? as by default `nclients_max > nclients` and if user gives it same is there any specific reason to change `max_clients`

iccaszhulili: @balamuruhans @will-Do This is a case to test the scenario about nclients_max=nclients. After setting  nclients_max=nclients, to check
whether the previous setting was successful, one has to attempt to connect another client to libvirtd. And thus here it is needed to increase nclients_max. I also got tangled by this for a long while, I do not
think it is good to introduce another parameter, it will destroy the consistency of the config file.  Maybe you can provide better solution. Thanks very much.

libvirt/tests/cfg/virt_admin/management/virt_admin_server_clients_set.cfg
iccaszhulili: updated, thx very much

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
iccaszhulili: updated

libvirt/tests/src/virt_admin/management/virt_admin_server_clients_set.py
```diff
@@ -0,0 +1,91 @@
+import logging
+from avocado.core import exceptions
```
balamuruhans: As per https://github.com/avocado-framework/avocado-vt/pull/966, let us use test.fail(), test.error() and test.cancel() instead of using exceptions

will-Do: @iccaszhulili Could you please resolve this comment?

iccaszhulili: @balamuruhans @will-Do updated the code according to your comment. Thanks very much

balamuruhans:
Thanks @iccaszhulili

PR #1001, opened 2017-03-06
Mov "Try" to upper place

kylazhang:
To deal with the situation that the guest do not support cirrus graphic device and the xml can be recovered


balamuruhans:
Looks good to me, can you change Autotest API `error` to `exceptions` from Avocado

PR #1004, opened 2017-03-08
modify blkdeviotune as group_name added with non-digit value

kylazhang:


lento-sun:
@will-Do pls help to merge this, it effects some of our acceptance cases

PR #1005, opened 2017-03-08
Add qemu logging config test cases

chunfuwen:
Change stdio_handler value in qemu.conf to configure qemu logging
Check associated log is written into log file correctly.

### DIFF
--- a/libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
+++ b/libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
@@ -1,0 +1,222 @@
+import logging
+import os
+import stat
+
+from pwd import getpwuid
+
+from avocado.utils import process
+
+from virttest import utils_config
+from virttest import utils_libvirtd
+from virttest import virt_vm
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest.libvirt_xml.devices.console import Console
+
+# Define qemu log path.
+QEMU_LOG_PATH = "/var/log/libvirt/qemu"
+
+
+def run(test, params, env):
+    """
+    Test stdio_handler parameter in qemu.conf to use for handling stdout/stderr
+    output from QEMU processes.
+
+    1) Change stdio_handler in qemu.conf;
+    2) Restart libvirtd daemon;
+    3) Check if libvirtd successfully started;
+    4) Check if virtlogd.socket is running;
+    5) Configure pty serial and console;
+    6) Check if VM log file exists and has correct permission and owner;
+    7) Check if VM log file is opened by virtlogd;
+    8) Check if VM start log is written into VM log file correctly;
+    9) Check if QEMU use pipe provided by virtlogd daemon for logging;
+    10) Check if VM shutdown log is written into VM log file correctly;
+    11) Check if pipe file can be closed gracefully after VM shutdown;
+    12) Check if VM restart log can be appended to the end of previous log file;
+    """
+
+    def clean_up_vm_log_file(vm_name):
+        """Clean up VM log file."""
+        # Delete VM log file if exists.
+        global QEMU_LOG_PATH
+        guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+        if os.path.exists(guest_log_file):
+            os.remove(guest_log_file)
+
+    def configure(cmd, guest_log_file=None, errorMsg=None):
+        """
+        Configure qemu log.
+        :param cmd. execute command string.
+        :param guest_log_file. the path of VM log file.
+        :param errorMsg. error message if failed
+        :return: pipe node.
+        """
+        # If guest_log_file is not None,check if VM log file exists or not.
+        if guest_log_file and not os.path.exists(guest_log_file):
+            test.error("Expected VM log file: %s not exists" % guest_log_file)
+        # If errorMsg is not None, check command running result.
+        elif errorMsg:
+            if process.run(cmd, ignore_status=True, shell=True).exit_status:
+                test.error(errorMsg)
+        # Get pipe node.
+        else:
+            result = process.run(cmd, timeout=90, ignore_status=True, shell=True)
+            ret, output = result.exit_status, result.stdout
+            if ret:
+                test.fail("Failed to get pipe node")
+            else:
+                return output
+
+    def configure_serial_console(vm_name):
+        """Configure serial console"""
+        # Check the primary serial and set it to pty.
+        VMXML.set_primary_serial(vm_name, 'pty', '0', None)
+        # Configure VM pty console.
+        vm_pty_xml = VMXML.new_from_inactive_dumpxml(vm_name)
+        vm_pty_xml.remove_all_device_by_type('console')
+
+        console = Console()
+        console.target_port = '0'
+        console.target_type = 'serial'
+        vm_pty_xml.add_device(console)
+        vm_pty_xml.sync()
+
+    def check_vm_log_file_permission_and_owner(vm_name):
+        """Check VM log file permission and owner."""
+        # Check VM log file permission.
+        global QEMU_LOG_PATH
+        guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+        logging.info("guest log file: %s", guest_log_file)
+        if not os.path.exists(guest_log_file):
+            test.error("Expected VM log file: %s not exists" % guest_log_file)
+        permission = oct(stat.S_IMODE(os.lstat(guest_log_file).st_mode))
+        if permission != '0600':
+            test.fail("VM log file: %s expect to get permission:0600, got %s ."
+                      % (guest_log_file, permission))
+        # Check VM log file owner.
+        owner = getpwuid(stat.S_IMODE(os.lstat(guest_log_file).st_uid)).pw_name
+        if owner != 'root':
+            test.fail("VM log file: %s expect to get owner:root, got %s ."
+                      % (guest_log_file, owner))
+
+    def check_info_in_vm_log_file(vm_name, cmd=None, matchedMsg=None):
+        """
+        Check if log information is written into log file correctly.
+        """
+        # Check VM log file is opened by virtlogd.
+        global QEMU_LOG_PATH
+        guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+        if not os.path.exists(guest_log_file):
+            test.fail("Expected VM log file: %s not exists" % guest_log_file)
+
+        if cmd is None:
+            cmd = ("grep -nr '%s' %s" % (matchedMsg, guest_log_file))
+        else:
+            cmd = (cmd + " %s |grep '%s'" % (guest_log_file, matchedMsg))
+        if process.run(cmd, ignore_status=True, shell=True).exit_status:
+            test.fail("Failed to get VM started log from VM log file: %s."
+                      % guest_log_file)
+
+    def check_pipe_closed(pipe_node):
+        """
+        Check pipe used by QEMU is closed gracefully after VM shutdown.
+        """
+        # Check pipe node can not be listed after VM shutdown.
+        cmd = ("lsof  -w |grep pipe|grep virtlogd|grep %s" % pipe_node)
+        if not process.run(cmd, timeout=90, ignore_status=True, shell=True).exit_status:
+            test.fail("pipe node: %s is not closed in virtlogd gracefully." % pipe_node)
+
+        cmd = ("lsof  -w |grep pipe|grep qemu-kvm|grep %s" % pipe_node)
+        if not process.run(cmd, timeout=90, ignore_status=True, shell=True).exit_status:
+            test.fail("pipe node: %s is not closed in qemu gracefully." % pipe_node)
+
+    vm_name = params.get("main_vm", "avocado-vt-vm1")
+    expected_result = params.get("expected_result", "virtlogd_disabled")
+    stdio_handler = params.get("stdio_handler", "not_set")
+    vm = env.get_vm(vm_name)
+    vm_xml = VMXML.new_from_inactive_dumpxml(vm_name)
+    vm_xml_backup = vm_xml.copy()
+    guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+
+    config = utils_config.LibvirtQemuConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    try:
+        if stdio_handler != 'not_set':
+            config['stdio_handler'] = "'%s'" % stdio_handler
+        # Restart libvirtd to make change valid.
+        if not libvirtd.restart():
+            if expected_result != 'unbootable':
+                test.fail('Libvirtd is expected to be started '
+                          'with stdio_handler=%s' % stdio_handler)
+            return
+        if expected_result == 'unbootable':
+            test.fail('Libvirtd is not expected to be started '
+                      'with stdio_handler=%s' % stdio_handler)
+        # Stop VM if VM is already started.
+        if vm.is_alive():
+            vm.destroy()
+
+        # Remove VM previous log file.
+        clean_up_vm_log_file(vm_name)
+
+        # Check if virtlogd socket is running.
+        cmd = ("systemctl status virtlogd.socket|grep 'Active: active'")
+        configure(cmd, errorMsg="virtlogd.socket is not running")
+
+        # Configure serial console.
+        configure_serial_console(vm_name)
+
+        logging.info("final vm:")
+        logging.info(VMXML.new_from_inactive_dumpxml(vm_name))
+
+        # Start VM.
+        try:
+            vm.start()
+        except virt_vm.VMStartError, detail:
+            test.fail("VM failed to start."
+                      "Error: %s" % str(detail))
+        # Check VM log file has right permission and owner.
+        check_vm_log_file_permission_and_owner(vm_name)
+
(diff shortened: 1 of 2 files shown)


chunfuwen:
case id in polarion:https://polarion.engineering.redhat.com/polarion/#/project/RedHatEnterpriseLinux7/workitem?id=RHEL7-41977

chunfuwen:
@will-Do 
@Hao-Liu 
Could you please take a look at the review?
Thanks!

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,246 @@
...
+    12) Check if VM restart log can be appended to the end of previous log file;
+    """
+
+    def clean_up_vm_log_file(vm_name):
+        """
+        Clean up VM log file.
+
+        """
+        # Delete VM log file if exists.
+        guest_log_file = "/var/log/libvirt/qemu/%s.log" % vm_name
```
balamuruhans: as same path is used in many places, we can have const and used elsewhere,
```
QEMU_LOG_FILE = "/var/log/libvirt/qemu"
os.path.join() can be used to concat path
```

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,246 @@
...
+        vm_pty_xml.add_device(console)
+        vm_pty_xml.sync()
+
+    def check_vm_log_file_permission_and_owner(vm_name):
+        """
+        Check VM log file permission and owner.
+
+        """
+        # Check VM log file permission.
+        guest_log_file = "/var/log/libvirt/qemu/%s.log" % vm_name
```
balamuruhans: same here

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,246 @@
...
+        if owner != 'root':
+            exceptions.TestFail("VM log file: %s expect to get owner:root, got %s ." % (guest_log_file, owner))
+
+    def check_vm_log_file_opened_by_virtlogd(vm_name):
+        """
+        Check VM log file is opened by virtlogd.
+
+        """
+        # Check VM log file is opened by virtlogd.
+        guest_log_file = "/var/log/libvirt/qemu/%s.log" % vm_name
```
balamuruhans: same here

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,246 @@
...
+        if utils.run(cmd, ignore_status=True).exit_status:
+            raise exceptions.TestFail("VM log file: %s expect to be opened by:virtlogd." % guest_log_file)
+
+    def check_info_in_vm_log_file(vm_name, cmd=None, matchedMsg=None):
+        """
+        Check if log information is written into log file correctly.
+
+        """
+        # Check VM log file is opened by virtlogd.
+        guest_log_file = "/var/log/libvirt/qemu/%s.log" % vm_name
```
balamuruhans: here

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: updated

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: updated

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: updated

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: updated

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,246 @@
+import logging
+import os
+import stat
+from pwd import getpwuid
+from autotest.client import utils
```
chunfuwen: updated

chunfuwen: Sound like that process.run() can not handle command with whitespace in.
*************************************
Running 'systemctl status virtlogd.socket|grep 'Active: active''
[stderr] Unit virtlogd.socket\x7cgrep.service could not be found.
[stderr] Unit Active:\x20active.service could not be found.
Setting ignore_status to True.

*******************
so still use utils.run() in autotest.client

balamuruhans:
```
>>> output = process.system_output("systemctl status virtlogd.socket")
>>> if "Active: active" not in output:
...     print "service not available"
... else:
...     print "service available"
... 
service available
```

chunfuwen: The use case command is :systemctl status virtlogd.socket|grep 'Active: active' ,see below output:
********************************
>>> cmd = ("systemctl status virtlogd.socket|grep 'Active: active'")
>>> result = process.run(cmd, ignore_status=True)
>>> print result
Command: systemctl status virtlogd.socket|grep 'Active: active'
Exit status: 4
Duration: 0.00464391708374
Stdout:

Stderr:
Unit virtlogd.socket\x7cgrep.service could not be found.
Unit Active:\x20active.service could not be found.


>>>

balamuruhans: @chunfuwen, I think you have missed `shell=True`,
```
>>> result = process.run("systemctl status virtlogd.socket|grep 'Active: active'", shell=True)
>>> print result
Command: systemctl status virtlogd.socket|grep 'Active: active'
Exit status: 0
Duration: 0.00701689720154
Stdout:
   Active: active (running) since Mon 2017-03-13 00:21:54 EDT; 4h 12min ago

Stderr:

PID:
26310

```

chunfuwen: You are correct,after adding shell = True  it works.
So there is a little different behaviour  with utils.run()

chunfuwen:
Address review comments.

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,263 @@
...
+    9) Check if QEMU use pipe provided by virtlogd daemon for logging;
+    10) Check if VM shutdown log is written into VM log file correctly;
+    11) Check if pipe file can be closed gracefully after VM shutdown;
+    12) Check if VM restart log can be appended to the end of previous log file;
+    """
+
+    def clean_up_vm_log_file(vm_name):
+        """
+        Clean up VM log file.
+
```
will-Do: Just find a blank line here. If the desc is really short, you can use one-line docstrings. 
https://www.python.org/dev/peps/pep-0257/
And a suggestion is that you can add arguments in the docstrings, for all new functions you created

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        # Delete VM log file if exists.
+        global QEMU_LOG_PATH
```
will-Do: does this ```global``` necessary?

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: updated:
1)Fix blank line issue, and transform short doc to one-line docstrings format.
2)Agree that complex functions need add parameters in docstring,but here the newly added function parameters are very simple and and explicit , so ignore that.

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: it is suggested by  balamuruhans  in previous comments.
Anyway, there is no harm here.

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,240 @@
...
+        clean_up_vm_log_file(vm_name)
+
+        # Check if virtlogd socket is running.
+        check_virtlogd_socket_status()
+
+        # Configure primary serial.
+        configure_primary_serial(vm_name)
+
+        # Configure pty serial.
+        configure_pty_console(vm_name)
```
balamuruhans: they all looks similar too, can we have single function to solve our purpose like,
```
def configure(cmd, guest_log_file=None):
    if guest_log_file and not os.path.exists(guest_log_file):
        exceptions.TestError("Expected VM log file: %s not exists" % guest_log_file)
        
        result = process.run(cmd, timeout=90, ignore_status=True, shell=True)
        ret, output = result.exit_status, result.stdout
        if ret:
            raise exceptions.TestFail("Failed to get pipe node")
        else:
            return output
```

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,240 @@
...
+        # Check VM shutdown log is written into log file correctly.
+        check_info_in_vm_log_file(vm_name, matchedMsg="shutting down")
+
+        # Check pipe is closed gracefully after VM shutdown.
+        check_pipe_closed(pipe_node)
+
+        # Start VM again.
+        try:
+            vm.start()
+        except virt_vm.VMStartError, e:
```
balamuruhans: please have proper variable name instead of `e`

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,240 @@
...
+    stdio_handler = params.get("stdio_handler", "not_set")
+    vm = env.get_vm(vm_name)
+    vm_xml = VMXML.new_from_inactive_dumpxml(vm_name)
+    vm_xml_backup = vm_xml.copy()
+
+    config = utils_config.LibvirtQemuConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    try:
+        if stdio_handler == 'not_set':
+            pass
```
chunfuwen: updated

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: 1)I prefer to wrap >2 lines business implementation into one method, in this way other readers can have  very simple and clear picture about what it does in try ... finally{} since they wont' see many business codes there,but simple method name.
2)Moreover, it help your do debugging after separating functions into methods

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,240 @@
...
+        # Configure pty serial.
+        configure_pty_console(vm_name)
+
+        logging.info("final vm:")
+        logging.info(VMXML.new_from_inactive_dumpxml(vm_name))
+
+        # Start VM.
+        try:
+            vm.start()
+        except virt_vm.VMStartError, e:
```
chunfuwen: update e to detail

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: update e to detail

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
balamuruhans: right, I agree but we should also consider redundancy of code. I have given a example, you can have appropriate function name, docstring and comments for better readability of the implementation.
its your call @chunfuwen :)

chunfuwen:
Merge some similar function: process.run () to one method to avoid duplicate code.

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: merge some functions into one method to avoid duplicated code

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,225 @@
...
+    vm_xml_backup = vm_xml.copy()
+    guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+
+    config = utils_config.LibvirtQemuConfig()
+    libvirtd = utils_libvirtd.Libvirtd()
+    try:
+        if stdio_handler != 'not_set':
+            config['stdio_handler'] = "'%s'" % stdio_handler
+        # Restart libvirtd to make change valid.
+        if not libvirtd.restart():
```
balamuruhans: if libvirtd fails to restart, why do we check whether expected_result is stdio_handler, as the changes gets reflected only after successful restart. should this be `if libvirtd.restart():`

chunfuwen: it aims to meet negative case with setting stdio_handler as invalid.
In this situation, we expect it failed to restart

chunfuwen:
@balamuruhans ,comments updated.

chunfuwen:
@will-Do 
@Hao-Liu 
Please take a look.

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,225 @@
...
+        :param errorMsg. error message if failed
+        :return: pipe node.
+        """
+        # If guest_log_file is not None,check if VM log file exists or not.
+        if guest_log_file and not os.path.exists(guest_log_file):
+            exceptions.TestError("Expected VM log file: %s not exists" % guest_log_file)
+        # If errorMsg is not None, check command running result.
+        elif errorMsg:
+            if process.run(cmd, ignore_status=True, shell=True).exit_status:
+                raise exceptions.TestError(errorMsg)
```
balamuruhans: let us use test.error(), test.fail() and test.cancel() instead of exceptions.TestError(), exceptions.TestFail() and exceptions.TestSkipError() respective as suggested in https://github.com/avocado-framework/avocado-vt/pull/966

chunfuwen: updated,please check

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,222 @@
...
+        :param errorMsg. error message if failed
+        :return: pipe node.
+        """
+        # If guest_log_file is not None,check if VM log file exists or not.
+        if guest_log_file and not os.path.exists(guest_log_file):
+            test.error("Expected VM log file: %s not exists" % guest_log_file)
+        # If errorMsg is not None, check command running result.
+        elif errorMsg:
+            if process.run(cmd, ignore_status=True, shell=True).exit_status:
+                raise test.error(errorMsg)
```
balamuruhans: `raise` is not required itseems

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,222 @@
...
+        # Get pipe node.
+        else:
+            result = process.run(cmd, timeout=90, ignore_status=True, shell=True)
+            ret, output = result.exit_status, result.stdout
+            if ret:
+                raise test.fail("Failed to get pipe node")
```
balamuruhans: same here

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
```diff
@@ -0,0 +1,222 @@
...
+        guest_log_file = os.path.join(QEMU_LOG_PATH, "%s.log" % vm_name)
+        if not os.path.exists(guest_log_file):
+            test.fail("Expected VM log file: %s not exists" % guest_log_file)
+
+        if cmd is None:
+            cmd = ("grep -nr '%s' %s" % (matchedMsg, guest_log_file))
+        else:
+            cmd = (cmd + " %s |grep '%s'" % (guest_log_file, matchedMsg))
+        if process.run(cmd, ignore_status=True, shell=True).exit_status:
+            raise test.fail("Failed to get VM started log from VM log file: %s."
```
balamuruhans: same here and elsewhere

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: remove it

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: remove it

libvirt/tests/src/conf_file/qemu_conf/set_virtlogd.py
chunfuwen: remove them

PR #1007, opened 2017-03-09
In Ubuntu there is no group called "nobody", "nogroup" instead.

sudeeshjohn:
In Ubuntu there is no group called "nobody", "nogroup" instead.
Signed-off-by: Sudeesh John <sudeesh@linux.vnet.ibm.com>


balamuruhans:
@sudeeshjohn Thanks for the Patch

PR #1009, opened 2017-03-10
virtual_disks_multidisks: Add a new expected condition

will-Do:
IOThreads are properly validated while a domain is defined, so for
these 3 negative cases, define VM fail is expected.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -5,8 +5,8 @@ import logging
 
 import aexpect
 
-from autotest.client import utils
-from autotest.client.shared import error
+from avocado.utils import process
+from avocado.core import exceptions
 
 from virttest import virt_vm
 from virttest import virsh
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -792,9 +792,9 @@ def check_dom_iothread():
     except virt_vm.VMStartError as details:
         if not status_error:
             raise error.TestFail('VM failed to start:\n%s' % details)
-    except xcepts.LibvirtXMLError:
+    except xcepts.LibvirtXMLError as details:
         if not define_error:
-            raise error.TestFail("Failed to define VM")
+            raise error.TestFail(details)
```
balamuruhans: we shall use `exceptions` from `avocado` instead

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1023,11 +1026,11 @@ def check_dom_iothread():
             if check_patitions_hotunplug:
                 if not check_vm_partitions(devices,
                                            device_targets, False):
-                    raise error.TestFail("See device in VM after hotunplug")
+                    raise exceptions.TestFail("See device in VM after hotunplug")
 
     finally:
         # Delete snapshots.
-        libvirt.clean_up_snapshots(vm_name, domxml=vmxml_backup)
+        libvirt.clean_up_snapshots(vm_name, domxm=vmxml_backup)
```
nasastry: The above should be 'domxml' not 'domxm'

virttest/utils_test/libvirt.py:258:def clean_up_snapshots(vm_name, snapshot_list=[], domxml=None):

will-Do: Thanks

balamuruhans:
@nasastry, Thanks for review @will-Do, Thanks for the patch

PR #1010, opened 2017-03-13
virsh.resume: Now resume running VM is invalid

will-Do:
libvirt commit 3e044e6e fix the running VM can resume issue, so
we can't do it anymore.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_resume.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_resume.cfg
@@ -15,8 +15,6 @@
                     resume_vm_ref = domuuid
             variants:
                 - vm_paused:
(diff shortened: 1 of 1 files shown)


balamuruhans:
@will-Do, should we have a check for libvirt version till commit 3e044e6e to skip this test ? else test will fail with old libvirt. what do you say

will-Do:
@balamuruhans 
Thanks for you concern. The problem is that we always have these version related issues, if we keep adding the version compare in the case, the code may become  more and more hard for maintain, so for this kind of problem, we hope to make it sample(and the test runner can choose which cases to run)

balamuruhans:
@will-Do Agreed

balamuruhans:
@will-Do Thanks for the patch

PR #1012, opened 2017-03-14
Add the case of client-disconnect for virt-admin

iccaszhulili:

### DIFF
--- a/libvirt/tests/src/virt_admin/management/virt_admin_client_disconnect.py
+++ b/libvirt/tests/src/virt_admin/management/virt_admin_client_disconnect.py
@@ -1,0 +1,41 @@
+from avocado.core import exceptions
+from virttest import virt_admin
+from virttest import virsh
+from virttest import utils_libvirtd
+
+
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virt_admin/management/virt_admin_client_disconnect.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    num_clients = params.get("num_clients")
+    server_name = params.get("server_name")
+    vp = virt_admin.VirtadminPersistent()
+    libvirtd = utils_libvirtd.Libvirtd()
+
+    try:
+        virsh_instant = []
+        for _ in range(int(num_clients)):
+            virsh_instant.append(virsh.VirshPersistent(uri="qemu:///system"))
```
balamuruhans: to create virsh instance, we are not using `virsh_instant` list anywhere else, instead
```
for _ in range(int(num_clients)):
    virsh.VirshPersistent(uri="qemu:///system")
```

PR #1015, opened 2017-03-15
virt_top.py: Enabled option to install the "virt-top" package , if missing. 

santwana:
This patch enables the test-case to install the "virt-top" package from the repo, before test-case execution.

It's ideal to initially check for the package installed and later the command present, before execution.

Also replaced the Autotest API's with Avocado API's.

### DIFF
--- a/libvirt/tests/src/virt_cmd/virt_top.py
+++ b/libvirt/tests/src/virt_cmd/virt_top.py
@@ -1,8 +1,10 @@
 import os
+import logging
 
-from autotest.client import os_dep
-from autotest.client import utils
-from autotest.client.shared import error
+from avocado.utils import path
+from avocado.utils import process
+from avocado.core import exceptions
+from avocado.utils import software_manager
 
 from virttest import virsh
 from virttest import data_dir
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virt_cmd/virt_top.py
```diff
@@ -27,7 +35,7 @@ def run(test, params, env):
 
     id_result = virsh.domid(vm_name)
     if id_result.exit_status:
-        raise error.TestNAError("Get domid failed.")
+        raise exceptions.TestNAError("Get domid failed.")
```
balamuruhans: it is `exceptions.TestSkipError`

libvirt/tests/src/virt_cmd/virt_top.py
```diff
@@ -54,9 +62,9 @@ def run(test, params, env):
...
-            raise error.TestFail("Command virt-top exit successfully, but "
-                                 "domid is expected")
+            raise exceptions.TestFail("Command virt-top exit successfully, but"
+                                      "domid is expected")
     else:
         if cmd_result.exit_status != 2:
-            raise error.TestFail("Command virt-top exit successfully with"
-                                 "invalid option:%s\n", cmd_result.stdout)
+            raise exceptions.TestFail("Command virt-top exit successfully with"
+                                      "invalid option:%s\n", cmd_result.stdout)
```
balamuruhans:
```
raise exceptions.TestFail("Command virt-top exit successfully with"
                          "invalid option:%s\n" % cmd_result.stdout)
```

santwana: Made changes as suggested.

balamuruhans:
@santwana, Thanks for the patch.

PR #1016, opened 2017-03-16
Ioeventfd is only supported by virtio-scsi controller

chunfuwen:
Previous ioeventfd cases combined with without_controller_model failed
So need filter them out


chunfuwen:
@will-Do 
@Hao-Liu 

Please have a look.

will-Do:
Sure

PR #1017, opened 2017-03-20
Extent waiting time for installing drivers for windows guests

chloerh:
Update waiting time from 60 seconds to 120 seconds, for QXL driver
could take more time to finsh installing.Otherwise case would fail
on not finding QXL driver.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -13,6 +13,7 @@ from virttest.libvirt_xml import vm_xml
 
 V2V_7_3_VERSION = 'virt-v2v-1.32.1-1.el7'
 RETRY_TIMES = 10
+WAIT_FOR_DRIVERS = 120
 
 
 class VMChecker(object):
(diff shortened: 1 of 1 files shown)


provider/v2v_vmcheck_helper.py
```diff
@@ -13,6 +13,7 @@
 
 V2V_7_3_VERSION = 'virt-v2v-1.32.1-1.el7'
 RETRY_TIMES = 10
+WAIT_FOR_DRIVERS = 120
```
nasastry: What's the base of this number? what if in future it's going to take more time?

chloerh: The time to wait until drivers are all fully installed is based on experience of multiple test runs of similar cases. The number 120 is an acceptable value for now. If it's going to take more in the future, we'll balance the cost and other factors to decide the value again.

provider/v2v_vmcheck_helper.py
```diff
@@ -248,8 +249,8 @@ def check_windows_vm(self):
         except Exception as detail:
             raise exceptions.TestError('Failed to connect to windows guest: %s' %
                                        detail)
-        logging.info("Wait 60 seconds for installing drivers")
-        time.sleep(60)
+        logging.info("Wait %d seconds for installing drivers", WAIT_FOR_DRIVERS)
+        time.sleep(WAIT_FOR_DRIVERS)
```
balamuruhans: just a suggestion, instead of just to sleep inside the test, cant we check drivers got installed or if we cant check then can we poll whether driver installed in regular interval basis. so it will check for drivers got installed if not sleep 5 sec and try again, we can have timeout for this poll and error out. But just sleeping inside the test by changing the sleep time is not right way.

PR #1018, opened 2017-03-20
Filter out raw format internal snapshot cases

chunfuwen:
Internal snapshot doesn't support raw disk format


chunfuwen:
@will-Do 
@Hao-Liu 

Please take a look.

waynesun09:
so current test_snapshot with internal snapshot error regex not work in the test module line:434 ?

chunfuwen:
@waynesun09 
Remove test module line:434

chunfuwen:
@waynesun09

Could you do me a favour to take a look ?

PR #1021, opened 2017-03-20
Fixed local variable 'vms_list' referenced before assignment issue

chunfuwen:
vms_list is used to released resouces in finally block,and
sometimes create_scsi_disk failure leads to enter finally block
before vms_list is initialized


chunfuwen:
@will-Do 
@Hao-Liu 

Please take a look.

waynesun09:
ACK

PR #1022, opened 2017-03-20
dumpxml: Case update according to libvirt change

will-Do:
libvirt commit 3b6be3c0 rework cpuUpdate, so the expect result also need
update.
This patch also rework the test case:
1). Skip case early when check libvirt version for ACL testing;
2). Simplify running command and checking test result part;
3). Remove autotest dependency;
4). Add more debug messages.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_dumpxml.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_dumpxml.py
@@ -1,11 +1,10 @@
 import re
 import logging
 
-from autotest.client.shared import error
-
 from virttest import virsh
 from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml import capability_xml
+from virttest.utils_test import libvirt as utlv
 
 from provider import libvirt_version
 
@@ -51,6 +50,7 @@ def run(test, params, env):
         vmcpu_xml.add_feature('ia64', 'optional')
         vmcpu_xml.add_feature('vme', 'require')
         vmxml['cpu'] = vmcpu_xml
+        logging.debug('Custom VM CPU: %s', vmcpu_xml.xmltreefile)
         vmxml.sync()
 
     def check_cpu(xml, cpu_match):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_dumpxml.py
```diff
@@ -143,8 +156,21 @@ def check_cpu(xml, cpu_match):
...
     vm_state = params.get("dumpxml_vm_state", "running")
     security_pwd = params.get("dumpxml_security_pwd", "123456")
-    status_error = params.get("status_error", "no")
+    status_error = "yes" == params.get("status_error", "no")
     cpu_match = params.get("cpu_match", "minimum")
+
+    # acl polkit params
+    setup_libvirt_polkit = "yes" == params.get('setup_libvirt_polkit')
+    if not libvirt_version.version_compare(1, 1, 1):
+        if setup_libvirt_polkit:
```
balamuruhans: if `setup_libvirt_polkit` is checked first and then for `libvirt_version.version_compare` then we can avoid calling `libvirt_version` for other non acl tests.

waynesun09: 1. version_compare only return True or False
2. another compare happen in check_cpu function at line 118

save the result not apply

libvirt/tests/src/virsh_cmd/domain/virsh_dumpxml.py
```diff
@@ -143,8 +156,21 @@ def check_cpu(xml, cpu_match):
...
+            raise exceptions.TestSkipError("API acl test not supported in"
```
balamuruhans: let us start using test.cancel()

libvirt/tests/src/virsh_cmd/domain/virsh_dumpxml.py
```diff
@@ -153,22 +179,10 @@ def check_cpu(xml, cpu_match):
         try:
             vm_xml.VMXML.add_security_info(new_xml, security_pwd)
         except Exception, info:
-            raise error.TestNAError(info)
+            raise exceptions.TestSkipError(info)
```
balamuruhans: same here and elsewhere

will-Do: @balamuruhans 
So you are using avocado latest version(not the 36lts)?
As we are using 36lts version for daily testing, it support skip status and don't backport the cancel status, so we have to use ```skip``` for now. I will raise a new patch to backport it.

balamuruhans: @will-Do yes I am using upstream avocado.
sure. I agree

balamuruhans:
Looks good, Thanks @will-Do

PR #1023, opened 2017-03-21
Fix domain define error issue

chunfuwen:
When using "type=usb,bus=2,port=2" in address attribute
the expected result is domain define error

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -2,7 +2,6 @@ import os
 import re
 import json
 import logging
-
 import aexpect
 
 from avocado.utils import process
@@ -21,6 +20,7 @@ from virttest.libvirt_xml.devices.disk import Disk
 from virttest.libvirt_xml.devices.input import Input
 from virttest.libvirt_xml.devices.hub import Hub
 from virttest.libvirt_xml.devices.controller import Controller
+from virttest.libvirt_xml.devices.address import Address
 
 from provider import libvirt_version
 
@@ -81,7 +81,7 @@ def run(test, params, env):
             if disk_type in ["file", "floppy"]:
                 cmd = ("mkfs.ext3 -F %s && setsebool virt_use_nfs true"
                        % device_source)
-                if process.run(cmd, ignore_status=True, shell=True).exit_status:
+                if process.system(cmd, ignore_status=True, shell=True):
                     raise exceptions.TestSkipError("Format disk failed")
 
         nfs_params = {"nfs_mount_dir": mount_dir, "nfs_mount_options": "ro",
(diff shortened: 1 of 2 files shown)


chunfuwen:
update this patch:
1)Add define_error = "yes" for error_test2 since it has similar issue
2)Remove hardcode input_usb_address and hub_usb_address in cfg file to avoid duplicate usb bus issue

chunfuwen:
@will-Do 
@Hao-Liu 

Please take a look.

waynesun09:
You've removed parameters:
input_usb_address = "bus=0,port=2"
hub_usb_address = "bus=0,port=1"

but remained the check code with both options in test module

chunfuwen:
@waynesun09 

So the choices has two:
1)One is to remove check code,but sound like it will miss one check point
2)The other is to get input and hub attributes dynamically,but currently vm_xml doesn't provide functions to get detailed attributes for input and  hub.

So what's your suggestion?

chunfuwen:
@waynesun09 
Some updates on new patch:
1)add one method to get input/hub address attributes from dumpxml dynamically 
2)add one remove hub operation before creating
3)remove unused add input/hub addr_dict updating code

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -2,11 +2,10 @@
 import re
 import json
 import logging
-
 import aexpect
 
-from avocado.utils import process
-from avocado.core import exceptions
+from autotest.client import utils
```
waynesun09: why reverse back to autotest?

chunfuwen: this is override script error by accident, soon will update new patch

chunfuwen: @waynesun09 ,upload new patch to fix reverse back issue

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -793,9 +786,9 @@ def check_dom_iothread():
     except virt_vm.VMStartError as details:
         if not status_error:
             raise exceptions.TestFail('VM failed to start:\n%s' % details)
-    except xcepts.LibvirtXMLError as details:
+    except xcepts.LibvirtXMLError:
         if not define_error:
-            raise exceptions.TestFail(details)
+            raise exceptions.TestFail("Failed to define VM")
```
will-Do: More error message is meaningful for debugging

chunfuwen: @will-Do 
Updated

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -386,6 +386,18 @@ def check_dom_iothread():
         if len(iothreads_ret['return']) != int(dom_iothreads):
             raise exceptions.TestFail("Failed to check domain iothreads")
 
+    def get_device_addr(device_type, type):
```
balamuruhans: `type` is a keyword, can we have different variable name to make it distinct.
looks good otherwise

chunfuwen: change variable name type to elem_type

chunfuwen: @balamuruhans 
Could you please have a check?

balamuruhans:
Thanks @waynesun09 @will-Do for review, @chunfuwen Thanks for the patch

PR #1024, opened 2017-03-22
Update cases with new package_install function

waynesun09:
The yum_install function could only work with yum package manager,
new package_install support apt, zapper, dnf with avocado package
manager module.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_network.py
+++ b/libvirt/tests/src/virtual_network/iface_network.py
@@ -10,6 +10,7 @@ from virttest import virt_vm
 from virttest import virsh
 from virttest import utils_net
 from virttest import utils_misc
+from virttest import utils_package
 from virttest import utils_libguestfs
 from virttest.utils_test import libvirt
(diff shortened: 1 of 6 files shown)


waynesun09:
This depends on:
https://github.com/avocado-framework/avocado-vt/pull/942

libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -420,7 +420,7 @@ def run_guest_libvirt(session):
         Check guest libvirt network
         """
         # Try to install required packages
-        if not utils_misc.yum_install(['libvirt'], session):
+        if not utils_misc.package_install(['libvirt'], session):
```
balamuruhans: In `L444`, we have `if session.cmd_status("yum -y remove libvirt*"):` 
shall we also have to support removing packages

waynesun09: @balamuruhans updated depended avocado-vt PR#942, added new package_remove function and applied to line 444.

PR #1025, opened 2017-03-22
virsh_migrate: retrieve migrate_shared_storage with available params

balamuruhans:
we have params for image_name, image_format and nfs_mount_dir using
which shared image path can be retrieved without depending on the
user input for migrate_shared_storage which avoids redundancy of
getting input from user.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
@@ -466,17 +466,28 @@ def run(test, params, env):
                                                graphics_port)
         extra = "--graphicsuri %s" % graphics_uri
 
+    # Check the required parameters
+    nfs_mount_path = params.get("nfs_mount_dir")
+    shared_storage = params.get("migrate_shared_storage", "")
+    if shared_storage == "" or "EXAMPLE" in shared_storage:
+        # retrieve shared storage image path from existing params
+        try:
+            image = params.get("image_name").split("/")[-1]
+            image_format = params.get("image_format")
+            image = "%s.%s" % (image, image_format)
+            params["migrate_shared_storage"] = shared_storage = os.path.join(nfs_mount_path,
+                                                                             image)
+            logging.debug("shared storage image location: %s", shared_storage)
+        except IndexError:
+            # use default image jeos-23-64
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -395,6 +395,19 @@ def check_migration_timeout_suspend(params):
...
         extra = "--graphicsuri %s" % graphics_uri
 
+    # retrieve shared storage image path from existing params
+    try:
+        nfs_mount_path = params.get("nfs_mount_dir")
+        image = params.get("image_name").split("/")[-1]
+        image_format = params.get("image_format")
+        image = "%s.%s" % (image, image_format)
+        params["migrate_shared_storage"] = os.path.join(nfs_mount_path, image)
+        logging.debug("shared storage image location: %s",
```
nasastry: The above parameter options namely `nfs_mount_dir, image_name, image_format, migrate_shared_storage` are not seen in cfg file i.e. `libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg`. From which cfg file they are taken?

balamuruhans: nfs_mount_dir and migrate_shared_storage are taken from base.cfg
image_name - guest-os.cfg and image_format - guest-hw.cfg

libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -395,6 +395,19 @@ def check_migration_timeout_suspend(params):
...
+                      params["migrate_shared_storage"])
+    except IndexError:
+        # it will be taken care by default image jeos-23-64
```
sathnaga: can you keep the older implementation for backward compatibility, instead user wants to use the given path,

if we want to use the manipulated path like this lets us set the param `migrate_shared_storage = "" `and check if none then lets do this try otherwise silently use the user defined value?

balamuruhans: sure, I will make the changes as suggested.

balamuruhans:
@sathnaga, have done the changes as suggested.

balamuruhans:
Thank you @sathnaga

PR #1029, opened 2017-03-27
migrate_vm: Update the grep string

dzhengfy:
There are some differences between RHEL7.3 and RHEL7.4 in qemu command
line when setting write-bytes-sec, read-bytes-sec, total-bytes-sec and
total-iops-sec.
Take total-iops-sec for example:
RHEL7.3: qemu command line: iops=1000
RHEL7.4: qemu command line: iops-total=1000

So the fixes are to adjust the grep string to meet both scenarios.


balamuruhans:
LGTM

balamuruhans:
@waynesun09 @dzhengfy Thanks

PR #1032, opened 2017-03-30
Add iothread case into migration_vm

chunfuwen:
Enable iothread function on disk,
Then verify it still works after vm migration


chunfuwen:
@will-Do 
@waynesun09 
If free, please help take a look.

PR #1034, opened 2017-04-01
Add the case for tracking all events for multiple domains.

iccaszhulili:
Only life-cycle events are tracked.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_event.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_event.py
@@ -21,8 +21,7 @@ def run(test, params, env):
     3. Catch the return of virsh event and qemu-monitor-event, and check it.
     """
 
-    vm_name = params.get("main_vm")
-    vm = env.get_vm(vm_name)
+    vms = env.get_all_vms()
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_event.py
```diff
@@ -42,35 +47,43 @@ def run(test, params, env):
...
+        if dom.is_alive():
+            dom.destroy()
+
+    vmxml_backup = []
+    for dom_name in vms_name:
+        vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(dom_name)
+        vmxml_backup.append(vmxml.copy())
 
-    def trigger_events(events_list=[]):
+    def trigger_events(dom_name, dom, events_list=[]):
```
balamuruhans: you can use domain object `dom.name` to get the name of the domain instead of taking `dom_name` as separate param

iccaszhulili: Thanks very much for your comment. We have been very busy with RHEL7.4 manual testing. Sorry for the late reply. 
Code has been changed according your review, the same with the following comments.

PR #1036, opened 2017-04-07
pool: Run command in shell

will-Do:


balamuruhans:
@will-Do Thanks for the patch

PR #1040, opened 2017-04-13
Recover postcopy option for all migration

dzhengfy:
The purpose of this patch is to make sure '--postcopy' option will not impact other existing test cases. The exising test cases should have same behavior as before even though with '--postcopy'.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
@@ -34,6 +34,17 @@
     ping_count = 10
     ping_timeout = 20
     variants:
+        - with_postcopy:
+            no there_p2p_tunnelled, there_suspend, there_suspend_undefinesource
+            no there_vm_suspend_live, there_vm_suspend_online, there_timeout_suspend
+            no with_inactive_guest, there_vm_shutdown_live, there_vm_shutdown_online
+            no there_offline.with_active_guest, with_HP_only, with_HP
+            no with_numa_and_HP, with_HP_pin, there_online, there_online_with_numa
+            postcopy_options = "--postcopy"
+        - without_postcopy:
+            no postcopy_after_precopy, migrate-postcopy
+            postcopy_options = ""
+    variants:
         - with_cpu_hotplug:
             virsh_migrate_cpu_hotplug = "yes"
             variants:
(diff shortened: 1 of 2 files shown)


dzhengfy:
@will-Do @balamuruhans could you help review?

libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -493,6 +493,11 @@ def check_migration_timeout_suspend(params):
             virsh.has_command_help_match("migrate", "--graphicsuri")):
         test.cancel("Do not support 'graphicsuri' option on this version.")
 
+    # For --postcopy enable
+    postcopy_options = params.get("postcopy_options")
```
balamuruhans: "--postcopy" flag will enable postcopy but does not trigger postcopy migration unless "migrate-postcopy" command is sent from another virsh instance for that particular guest.

It is handled in `- migrate-postcopy` variant, so let us bring `- migrate-postcopy` and `postcopy_after_precopy` up in the place of `- with_postcopy`. please share your thoughts

dzhengfy: The purpose of this patch is to make sure '--postcopy' option will not impact other existing test cases. The exising test cases should have same behavior as before even though with '--postcopy'. Sorry, I should have make it more clear. I have updated it now.
My thought is  'postcopy_after_precopy'' and 'migrate-postcopy' in this fix will be only matched with ' with_postcopy'.
But I am not very sure about your sugguestion.  Could you help point out where you prefer to placing the two cases?  Under ' with_postcopy' or peer of  'with_postcopy' or else?

libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
```diff
@@ -211,16 +222,13 @@
...
-            variants:
-                - postcopy_after_precopy:
-                    virsh_migrate_options = "--live --postcopy --postcopy-after-precopy"
-                - postcopy:
-                    virsh_migrate_options = "--live --postcopy"
-                    virsh_postcopy_cmd = "migrate-postcopy"
-                    # migration thread timeout
-                    postcopy_migration_timeout = "200"
+        - postcopy_after_precopy:
+            virsh_migrate_options = "--live --postcopy-after-precopy"
```
balamuruhans: `virsh_migrate_options = "--live --postcopy --postcopy-after-precopy"`

--postcopy is to enable the postcopy migration and --postcopy-after-precopy will trigger postcopy migration automatically after 1st successful precopy migration.

libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
balamuruhans: Thanks for the explanation, if the objective is to ensure that existing test cases are not affected with enabling "--postcopy" flag without triggering postcopy migration then it is fine.

libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
dzhengfy: Yes. Do you have any other comments?

balamuruhans: --postcopy flag is missed in the `virsh_migrate_options`

dzhengfy: The  variants:  - with_postcopy: will apply '--postcopy' option into this virsh_migrate_options for case 'postcopy_after_precopy' and case 'migrate-postcopy'. Please share your thoughts about this configuration.

balamuruhans: Agreed.

PR #1044, opened 2017-04-18
libvirt_mem: Fix memory assertion for detach

waynesun09:
Fix the memory detach assertion after restore, the check need use
new_max_mem rather than original max_mem to minus detached memory.

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -162,10 +162,10 @@ def run(test, params, env):
                 assert int(tg_node) == int(mem_dev[0].target.node)
             elif dt_mem:
                 if at_times:
-                    assert int(max_mem) - (int(tg_size) *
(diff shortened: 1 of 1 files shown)


balamuruhans:
@nasastry please help in reviewing this PR

nasastry:
My testing found that when trying to do 'all slots cold plug', then this equation is getting varied. Found that when `slot are 32` then attach assert is passing
```
assert int(max_mem) + (int(tg_size) *
                                           at_times) == xml_max_mem
```
but detach assert is failing
```
assert int(new_max_mem) - (int(tg_size) *
                                           at_times) == xml_max_mem
```
because detach happened only once instead of `all the slots i.e. 32`.

nasastry:
Found the reason for 'all slots cold plug' issue. After testing will send a pull request.

balamuruhans:
Thanks @nasastry for the review, 
Thanks @waynesun09.

PR #1045, opened 2017-04-18
virsh_domstats: default vms and main_vm can be used for vm_list

balamuruhans:
list of vms can be defaulted from existing param vms for all_domains and
param main_vm can be used for specific domain, use test.fail(),
test.error() and test.cancel() instead of using error from autotest.


balamuruhans:
Thanks @will-Do

PR #1048, opened 2017-04-24
Added topology support, removed few tests for ppc64le

sathnaga:
1. Added topology support
2. Skipped power management tests for power
3. Changed cpus count method to make it work in all platforms 

Earlier, it was failing if the guest has topology
definition.

### DIFF
--- a/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
+++ b/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
@@ -108,7 +108,7 @@ def check_vcpu_number(vm, expect_vcpu_num, expect_vcpupin, setvcpu_option=""):
     libvirt.check_exit_status(result)
     vcpupin_output = result.stdout.strip().splitlines()[2:]
     if expect_vcpupin:
-        host_cpu_count = utils.count_cpus()
+        host_cpu_count = os.sysconf('SC_NPROCESSORS_CONF')
         xml_affinity_list = []
         xml_affinity = {}
         try:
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -421,8 +423,13 @@ def run(test, params, env):
             vmxml.remove_agent_channels()
         vmxml.sync()
 
-        vmxml.set_vm_vcpus(vm_name, int(vcpu_max_num), int(vcpu_current_num))
-
+        topology = vmxml.get_cpu_topology()
+        if topology and (sockets != 0 and cores != 0 and threads != 0):
```
will-Do: this should equal to ```if all([topology, sockets, cores, threads]):```

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -421,8 +423,13 @@ def run(test, params, env):
...
+            vmxml.set_vm_vcpus(vm_name, int(vcpu_max_num),
+                               int(vcpu_current_num))
+        else:
+            vmxml.set_vm_vcpus(vm_name, int(vcpu_max_num),
+                               int(vcpu_current_num), sockets, cores, threads)
```
will-Do: is this ```if ... else``` inverted?

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
sathnaga: @will-Do sure

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
sathnaga: @will-Do nope, I don't want make the topology as the must attribute, if already available, make sure we handle it.

sathnaga:
@will-Do addressed the comments.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
sathnaga: @will-Do Tnx, you are right, overlooked, will change it.

sathnaga:
@will-Do Thanks :+1: addressed the comments.

balamuruhans:
Looks good to me

PR #1053, opened 2017-05-03
Check multiple times for windows guests drivers after conversion

chloerh:
In case got in-accurate test result if drivers not fully installed
yet.
Sleep 60 seconds during each check.
Log the failure if still couldn't find driver after checks run out.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -272,7 +272,6 @@ class VMChecker(object):
 
         # 2. Check Red Hat VirtIO drivers and display adapter
         logging.info("Checking VirtIO drivers and display adapter")
-        win_dirves = self.checker.get_driver_info()
         expect_drivers = ["Red Hat VirtIO SCSI",
                           "Red Hat VirtIO Ethernet Adapte"]
         # Windows display adapter is different for each release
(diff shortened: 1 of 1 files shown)


provider/v2v_vmcheck_helper.py
```diff
@@ -283,13 +282,28 @@ def check_windows_vm(self):
         if self.os_version in ['win8', 'win8.1', 'win10', 'win2012', 'win2012r2']:
             expect_adapter = 'Basic Display Driver'
         expect_drivers.append(expect_adapter)
-        for driver in expect_drivers:
-            if driver in win_dirves:
-                logging.info("Find driver: %s", driver)
-                logging.info("PASS")
+        check_drivers = expect_drivers[:]
+        for check_times in range(5):
```
will-Do: could be use 'wait_for'

balamuruhans:
please address the comments

provider/v2v_vmcheck_helper.py
chloerh: There is this list of drivers to be checked. A driver will be removed from the list if it is found in guest during each loop of driver checking operation. The 'wait_for' function will need a function and check the return value, which in this case should be 'expect_drivers' is empty. I don't think it's more clear of the code if use the 'wait_for' function. I believe this way is more simple and clear.

chloerh:
Recent test result

kvm.windows.win10.i386.qcow2_f.NFS | 16 分 | Passed
kvm.windows.win2008.i386.raw_f.NFS | 16 分 | Passed
kvm.windows.win2016.x86_64.qcow2_f.NFS | 21 分 | Fixed
kvm.windows.win8.x86_64.raw_f.NFS | 18 分 | Passed
kvm.windows.win8_1.i386.qcow2_f.NFS | 18 分 | Passed

provider/v2v_vmcheck_helper.py
```diff
@@ -283,13 +282,28 @@ def check_windows_vm(self):
...
+            logging.info('Check drivers for the %dth time', check_times + 1)
+            win_dirves = self.checker.get_driver_info()
+            for driver in expect_drivers:
+                if driver in win_dirves:
+                    logging.info("Find driver: %s", driver)
```
balamuruhans: `logging.info("Driver %s found", driver)`

provider/v2v_vmcheck_helper.py
```diff
@@ -283,13 +282,28 @@ def check_windows_vm(self):
...
+                    logging.info("PASS")
```
balamuruhans: I dont think this is required, it might be used during dev/debug

provider/v2v_vmcheck_helper.py
```diff
@@ -283,13 +282,28 @@ def check_windows_vm(self):
...
+                    check_drivers.remove(driver)
+                else:
+                    err_msg = "Not find driver: %s" % driver
```
balamuruhans: `err_msg = "Driver %s not found" % driver`

chloerh:
@balamuruhans Thanks for the review, all updated

PR #1058, opened 2017-05-05
Add v2v function cases

chloerh:
- Check dependency of virt-v2v
- Remove options:--vm-type, --no-trim, --debug-dc
- option --help

Update cfg files for public cfg

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -30,7 +30,7 @@ def run(test, params, env):
         raise ValueError('Missing command: virt-v2v')
     for v in params.itervalues():
         if "V2V_EXAMPLE" in v:
-            raise exceptions.TestSkipError("Please set real value for %s" % v)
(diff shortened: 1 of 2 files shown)


v2v/tests/src/v2v_options.py
```diff
@@ -305,6 +305,33 @@ def check_source(output):
...
             raise exceptions.TestFail('Source info not correct for: %s' % fail)
 
+    def check_man_page(check_man):
+        """
+        Check if content of man page or help info meets expectation
+        """
+        logging.info('Checking man page of virt-v2v for "%s"', check_man)
+        man_page = process.run('man virt-v2v').stdout.strip()
+        if check_man in man_page:
+            raise exceptions.TestFail('"%s" not removed from man page', check_man)
```
balamuruhans: let us use `test.fail()`, `test.error()` in places of `exceptions.TestFail()` and `exceptions.TestError()` respectively here and elsewhere.

PR #1070, opened 2017-05-15
Add disk with duplicate controller index

chunfuwen:
If cold plug,attach expect succeed,but throw error:Found duplicate drive address for disk
when redefine VM.
On the other hand, if hot plug,attach expect fail with error:internal error:
unable to execute QEMU command '__com.redhat_drive_add': Duplicate ID '

Details refer bug:https://bugzilla.redhat.com/show_bug.cgi?id=1036715

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
@@ -526,6 +526,20 @@
                                             status_error = "yes"
                                             define_error = "yes"
                                             virt_disk_addr_options = "type=pci,domain=0x0000,bus=0x00,slot=0x0a,function=0x0"
+                - disk_duplicate_scsi_controller_index_error_test:
+                    virt_disk_path = "/var/lib/libvirt/images/seconddisk.qcow2"
+                    virt_disk_with_duplicate_scsi_controller_index = "yes"
+                    virt_disk_device = "disk"
+                    virt_disk_device_source = "disk1"
+                    virt_disk_device_target = "sda"
+                    virt_disk_device_format = "qcow2"
+                    virt_disk_device_bus = "scsi"
+                    virt_disk_device_type = "file"
+                    driver_option = "type=qcow2,cache=none"
+                    virtio_scsi_controller_model = "virtio-scsi"
+                    virtio_scsi_controller = "yes"
+                    define_error = "yes"
+                    virt_disk_addr_options = "type=drive,controller=0,bus=0,target=0,unit=0"
     variants:
         - hotplug:
             virt_disk_device_hotplug = "yes"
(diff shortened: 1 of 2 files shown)


chunfuwen:
@will-Do 
@waynesun09 

Please take a look.

libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
```diff
@@ -526,6 +526,20 @@
                                             status_error = "yes"
                                             define_error = "yes"
                                             virt_disk_addr_options = "type=pci,domain=0x0000,bus=0x00,slot=0x0a,function=0x0"
+                - disk_duplicate_scsi_controller_index_error_test:
+                    virt_disk_path = "/var/lib/libvirt/images/seconddisk.qcow2"
```
will-Do: it's better to avoid the abs dir

chunfuwen: It sound like that only /var/lib/libvirt/images/ folder can guarantee access rights by libvirt

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -398,6 +398,23 @@ def get_device_addr(device_type, elem_type):
                     break
         return addr
 
+    def create_custom_disk(disk_path):
```
balamuruhans: `create_disk_xml()` from `utils_test/libvirt.py` can help to create custom disk xml

chunfuwen: Do you mean create_disk_xml(params) in the libvirt.py? it sound like it can not support  addr_attribute fit in,which is required in this patch. Moreover, it return Disk.xml instead of Disk object ,which disallow last chance to update attribute

balamuruhans: we can use the existing API `create_disk_xml()` in our tests if we cant able to use it, we can have this `create_custom_disk()` in common as #1073 have `create_addtional_disk()` which is similar to this.

chunfuwen: updated
use create_custom_disk() in common as #1073. And after this commit is merged,the duplicated code in #1073 will be removed

chunfuwen:
@balamuruhans 
New patch had been uploaded.

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -399,6 +399,33 @@ def get_device_addr(device_type, elem_type):
...
+                              disk_device, disk_target, disk_bus):
+        """
+        Create another disk for a given path,customize some attributes.
+        :param disk_type. the type of disk.
+        :param disk_path. the path of disk.
+        :param disk_format. the format to disk image.
+        :param disk_device_type. the disk device type.
+        :param disk_device. the device of disk.
+        :param disk_target. the target of disk.
+        :param disk_bus. the target bus of disk.
```
balamuruhans: Looks good to me,
just a typo in docstring, `%s/:param disk_type./:param disk_type:` for all the params

PR #1077, opened 2017-05-19
Bugs fix and case update

will-Do:


libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -612,18 +599,16 @@ def run_guest_libvirt(session):
                         br_delay = float(bridge['delay'])
                         cmd = ("brctl showstp %s | grep 'bridge forward delay'"
                                % bridge['name'])
-                        out = utils.run(cmd, ignore_status=False).stdout.strip()
+                        out = process.system_output(cmd, ignore_status=False)
```
balamuruhans: `out = process.system_output(cmd, shell=True, ignore_status=False)`

balamuruhans: it failed to grep without `shell=True`

will-Do: thanks

PR #1080, opened 2017-05-27
Fixed mklabel issue in luks encryption cases

chunfuwen:
It need mklabel in Vm internal instead of host.


chunfuwen:
@balamuruhans 
@will-Do 
@waynesun09 

Please help take a look.

PR #1081, opened 2017-06-01
Fix the issue that Vm failed to be started

chunfuwen:
since rbd disk with snapshot need set readonly attribute explicitly

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_ceph.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_ceph.cfg
@@ -67,6 +67,7 @@
                 - attach_with_snapshot:
                     create_snapshot = "yes"
                     disk_snap_name = "s1"
+                    read_only = "yes"
         - disk_pool:
             pre_vm_state = "shut off"
             test_vm_parts = "no"


chunfuwen:
#https://github.com/avocado-framework/avocado-vt/pull/1023

chunfuwen:
@will-Do 
@balamuruhans 

Please help take a look

libvirt/tests/src/virtual_disks/virtual_disks_ceph.py
```diff
@@ -533,6 +534,7 @@ def check_in_vm(vm_obj, target, old_parts, read_only=False):
                 snap_cmd = ("rbd -m %s %s snap create %s@%s" %
                             (mon_host, key_opt, disk_src_name, snap_name))
                 process.run(snap_cmd, ignore_status=False, shell=True)
+                params.update({"readonly": read_only})
```
balamuruhans: This will get used and handled by https://github.com/avocado-framework/avocado-vt/pull/1023 in 
`L566 libvirt.set_vm_disk(vm, params)`

so param `read_only` in cfg is enough right?

chunfuwen: Yes, need set read_only attribute explicitly in cfg file.

balamuruhans: This piece of code is not necessary as it is taken care from API.

chunfuwen: In my understanding, In set_vm_disk(),the readnly value is gotten from passing in params,and configurable,so need caller set the value explicitly .

***************************************************************************
        if disk_snap_name:
            disk_params_src.update({'source_snap_name': disk_snap_name})
            disk_params.update({'readonly': params.get("read_only", "no")})

balamuruhans: right, when we have given explicitly in `read_only = "yes"` in cfg, params already have it and API can pick it up,
again updating in tests with same value is not necessary, it is redundant

chunfuwen: Removed duplicated code

chunfuwen:
@balamuruhans 
patch had been updated.

balamuruhans:
@chunfuwen Thanks

PR #1085, opened 2017-06-08
Add usb disk auto fit address attribute case

chunfuwen:
This is used to validate bugzilla:1046267 issue where
address attribute need to be fit in automatically after
usb disk plug in.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -1057,9 +1057,13 @@ def run(test, params, env):
                 dev_id = vm_xml.VMXML.get_disk_attr(vm_name, device_targets[0],
                                                     "alias", "name")
                 if devices[0] == "disk":
-                    cmd += (" | grep usb-storage,bus=usb%s.0,port=%s,"
+                    # For dev_bus ==0,it hardcode bus=usb.0,other than usb%s.0.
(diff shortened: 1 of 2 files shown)


chunfuwen:
@will-Do 
@balamuruhans 

Please help take a look.

chunfuwen:
@balamuruhans 

Could you please take a look?

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1057,9 +1057,15 @@ def test_device_update_boot_order(disk_type, disk_path, error_msg):
                 dev_id = vm_xml.VMXML.get_disk_attr(vm_name, device_targets[0],
                                                     "alias", "name")
                 if devices[0] == "disk":
-                    cmd += (" | grep usb-storage,bus=usb%s.0,port=%s,"
-                            "drive=drive-%s,id=%s"
-                            % (dev_bus, dev_port, dev_id, dev_id))
+                    # For dev_bus ==0,it hardcode bus=usb.0,other than usb%s.0.
```
will-Do: if the difference is only about 'usb', it may looks more clear if you just deal with it.

chunfuwen: Updated,Thanks!

chunfuwen:
@balamuruhans

Could you please take a look?

PR #1091, opened 2017-06-18
Vcpucount option fix

will-Do:


libvirt/tests/src/virsh_cmd/domain/virsh_vcpucount.py
```diff
@@ -184,11 +185,11 @@ def run(test, params, env):
 
     # 1.2 Check for all options
     option_list = options.split(" ")
-    for item in option_list:
-        if virsh.has_command_help_match("vcpucount", item) is None:
-            test.cancel("The current libvirt "
-                        "version doesn't support "
-                        "'%s' option" % item)
+    if not status_error:
```
balamuruhans: For negative tests, the options should supported to give invalid values. so test should be skipped for negative tests too if options are not supported by libvirt.

will-Do: No, as we know it is a negative test, skip it make no sense. Such as for testing invalid option: '--xyz', you may get:
```
TestSkipError: The current libvirt version doesn't support '--xyz' option
```

will-Do:
@sathnaga 
Thx, merge this pr now.

PR #1093, opened 2017-06-21
Fix local variable 'nfs_service' referenced before assignment

chunfuwen:
In finally block, nfs_service may be one None object,and add
None check before start

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -150,8 +150,9 @@ def run(test, params, env):
         create_large_image()
         if error_type == "unspecified error":
             # umount nfs to trigger error after create large image
-            nfs_service.stop()
-            logging.debug("nfs status is %s", nfs_service.status())
+            if nfs_service is not None:
+                nfs_service.stop()
+                logging.debug("nfs status is %s", nfs_service.status())
 
         # wait and check the guest status with timeout
         def _check_state():
@@ -177,7 +178,8 @@ def run(test, params, env):
     finally:
         logging.info("Do clean steps")
         if error_type == "unspecified error":
-            nfs_service.start()
+            if nfs_service is not None:
+                nfs_service.start()
             vm.destroy()
             if os.path.isfile("%s.bak" % export_file):
                 shutil.move("%s.bak" % export_file, export_file)


chunfuwen:
@will-Do 
@balamuruhans 

Please take a look.

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -177,7 +177,8 @@ def _check_state():
     finally:
         logging.info("Do clean steps")
         if error_type == "unspecified error":
-            nfs_service.start()
+            if nfs_service is not None:
```
balamuruhans: check is to start the nfs service if it is in stop state then,
`if not nfs_service.status():`

chunfuwen: UnboundLocalError: local variable 'nfs_service' referenced before assignment 
The error is UnboundLocalError,meaning that nfs_service is none, so use nfs_service is not None to judge condiiton

balamuruhans: nfs_service is defined elsewhere inside the condition `error_type == "unspecified error"`, I was just thinking how this error could hit.

chunfuwen: It could happen that Factory.create_service() fail on line  nfs_service = Factory.create_service("nfs")

balamuruhans: Then we could have hit UnbouldLocalError in 
`L153: nfs_service.stop()` before even coming to finally section.

chunfuwen: add condition check before call nfs_service.stop()

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -150,8 +150,9 @@ def create_large_image():
         create_large_image()
         if error_type == "unspecified error":
             # umount nfs to trigger error after create large image
-            nfs_service.stop()
-            logging.debug("nfs status is %s", nfs_service.status())
+            if nfs_service is not None:
```
balamuruhans: I don't think this is a good way to approach.
1. if `L75 nfs_service = Factory.create_service("nfs")` fails means nfs service might not available, 
but if nfs service is not available then test errors out in `L65 libvirt.setup_or_cleanup_nfs()` as mounting will fail.
2. I doubt, this can happen in Ubuntu machine as it should be nfs_service = Factory.create_service("nfs-server")
so lets have check for distro and proceed accordingly

chunfuwen: 2017-06-23 17:25:10,044 utils_misc       L1105 DEBUG| 127.0.0.1:/var/tmp/avocado_AmMZdK/images is not mounted.
2017-06-23 17:25:10,087 utils_misc       L1098 DEBUG| Searching '127.0.0.1:/var/tmp/avocado_AmMZdK/images /var/tmp/avocado_AmMZdK/mnt .*nfs\d? ' in mtab...
2017-06-23 17:25:10,088 utils_misc       L1105 DEBUG| 127.0.0.1:/var/tmp/avocado_AmMZdK/images is not mounted.
2017-06-23 17:25:10,089 utils_misc       L1029 DEBUG| 127.0.0.1:/var/tmp/avocado_AmMZdK/images is not mounted under /var/tmp/avocado_AmMZdK/mnt
2017-06-23 17:25:10,089 process          L0333 INFO | Running 'mount -o nolock,soft,timeo=1,retrans=1,retry=0 127.0.0.1:/var/tmp/avocado_AmMZdK/images /var/tmp/avocado_AmMZdK/mnt'
2017-06-23 17:25:10,107 process          L0420 DEBUG| [stderr] mount.nfs: access denied by server while mounting 127.0.0.1:/var/tmp/avocado_AmMZdK/images
********************************
On Redhat Linux, error happen before L75,so go directly to finally block ,which will can nfs_service.start()

chunfuwen:
@balamuruhans 

Any update?

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
balamuruhans: Okay, Agreed.

balamuruhans:
Thanks @chunfuwen

PR #1094, opened 2017-06-21
Fix duplicate add image issue

chunfuwen:


chunfuwen:
@will-Do 
@balamuruhans 
please take a look.

PR #1095, opened 2017-06-21
Fix occasional bus port conflict issue

chunfuwen:

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
@@ -562,7 +562,7 @@
                                         - default:
                                             add_usb_device = "no"
                                         - normal_test:
-                                            virt_disk_addr_options = "type=usb,bus=1,port=2"
+                                            virt_disk_addr_options = "type=usb,bus=1"
                                         - error_test1:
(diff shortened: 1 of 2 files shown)


chunfuwen:
@balamuruhans 
@will-Do 
Could you please take a look?

chunfuwen:
@balamuruhans 

Could you please take a look?

libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
```diff
@@ -562,7 +562,7 @@
                                         - default:
                                             add_usb_device = "no"
                                         - normal_test:
-                                            virt_disk_addr_options = "type=usb,bus=1,port=2"
+                                            virt_disk_addr_options = "type=usb"
```
will-Do: So this will not affect the usage of 'virt_disk_addr_options', right?

chunfuwen: Yes, it should not. Libvirt will auto-fit bus and port attributes if absent. Previous hardcoded those attributes can see some resource conflict in testing result.

chunfuwen:
@will-Do 
Please help take a look.

PR #1099, opened 2017-06-29
Update cfg about case virsh_create_lxc

hsmj1412:
Fixed a conflict between main_vm(avocado_vt_vm1) and vms(lxc_test_vm1)

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_create_lxc.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_create_lxc.cfg
@@ -3,7 +3,7 @@
     take_regular_screendumps = "no"
     start_vm = "no"
     connect_uri = "lxc:///"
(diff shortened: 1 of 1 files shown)


libvirt/tests/cfg/virsh_cmd/domain/virsh_create_lxc.cfg
```diff
@@ -3,7 +3,7 @@
     take_regular_screendumps = "no"
     start_vm = "no"
     connect_uri = "lxc:///"
-    vms = "lxc_test_vm1"
+    main_vm = "lxc_test_vm1"
```
balamuruhans: https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virsh_cmd/domain/virsh_create_lxc.py#L22
Test uses "vms" to get the vm name, 
`In L22: vm_name = params.get("vms")`

PR #1100, opened 2017-06-29
Modify code to keep vm obj when skip guest check

chloerh:
For future clean up process could proceed.


balamuruhans:
Thanks @chloerh @will-Do

PR #1108, opened 2017-07-04
Fix libvirt_rng failure issues

kylazhang:
The change was due to the namespace support, thus /dev/random is re-created in a different path so that it can be passed into the correct namespace monut, the device node is re-created in a diffent place, and then re-mount that directory in a separate namespace to /dev/ for that process, so the file qemu seems differs from the actual /dev/random by inode, thus lsof does not show it but lsof|grep /dev/random will show

### DIFF
--- a/libvirt/tests/src/libvirt_rng.py
+++ b/libvirt/tests/src/libvirt_rng.py
@@ -3,7 +3,7 @@ import os
 import ast
 import shutil
 import logging
-from autotest.client.shared import error
+from avocado.utils import process
 from autotest.client import utils
 from virttest import virt_vm, virsh
 from virttest import utils_package
@@ -69,7 +69,7 @@ def run(test, params, env):
                                       str(rng_xml), re.M
                                       )[0].replace("/", "\/"))])
             if not status:
-                raise error.TestFail("Failed to edit vm xml")
+                test.fail("Failed to edit vm xml")
 
     def check_qemu_cmd(dparams):
         """
@@ -101,9 +101,9 @@ def run(test, params, env):
             rate = ast.literal_eval(rng_rate)
             cmd += (" | grep 'max-bytes=%s,period=%s'"
                     % (rate['bytes'], rate['period']))
-        if utils.run(cmd, ignore_status=True).exit_status:
-            raise error.TestFail("Cann't see rng option"
-                                 " in command line")
+        if process.run(cmd, ignore_status=True, shell=True).exit_status:
+            test.fail("Cann't see rng option"
+                      " in command line")
 
     def check_host():
         """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_rng.py
```diff
@@ -111,7 +111,7 @@ def check_host():
         """
         backend_dev = params.get("backend_dev")
         if backend_dev:
-            cmd = "lsof %s" % backend_dev
+            cmd = "lsof |grep %s" % backend_dev
             ret = utils.run(cmd, ignore_status=True)
```
balamuruhans: @kylazhang does utils.run require `shell=True` if you can take this opportunity to remove autotest dependency it will be helpful.

kylazhang:
Fixed with @balamuruhans 's suggestion, but still can not avoid utils as it called AsyncJob which are not exist in process

libvirt/tests/src/libvirt_rng.py
```diff
@@ -3,7 +3,8 @@
 import ast
 import shutil
 import logging
-from autotest.client.shared import error
+from avocado.core import exceptions
```
balamuruhans: importing `exceptions` not required

libvirt/tests/src/libvirt_rng.py
```diff
@@ -69,7 +70,7 @@ def modify_rng_xml(dparams, sync=True):
                                       str(rng_xml), re.M
                                       )[0].replace("/", "\/"))])
             if not status:
-                raise error.TestFail("Failed to edit vm xml")
+                raise test.fail("Failed to edit vm xml")
```
balamuruhans: `raise` is not required.

libvirt/tests/src/libvirt_rng.py
```diff
@@ -101,21 +102,21 @@ def check_qemu_cmd(dparams):
...
-            cmd = "lsof %s" % backend_dev
-            ret = utils.run(cmd, ignore_status=True)
+            cmd = "lsof |grep %s" % backend_dev
+            ret = process.run(cmd, ignore_status=True, shell=True)
             if ret.exit_status or not ret.stdout.count("qemu"):
-                raise error.TestFail("Failed to check random device"
-                                     " on host, command output: %s",
+                raise test.fail("Failed to check random device"
+                                " on host, command output: %s" %
                                      ret.stdout)
```
balamuruhans: L120 have indentation issue to fix inspekt error

libvirt/tests/src/libvirt_rng.py
```diff
@@ -256,11 +257,11 @@ def check_guest(session):
...
             if not os.path.exists(rngd_srv_conf):
                 shutil.copy(rngd_srv, rngd_srv_conf)
-            utils.run("sed -i -e 's#^ExecStart=.*#ExecStart=/sbin/rngd"
-                      " -f -r /dev/urandom -o /dev/random#' %s"
-                      % rngd_srv_conf)
-            utils.run('systemctl daemon-reload')
-        utils.run("service rngd start")
+            process.run("sed -i -e 's#^ExecStart=.*#ExecStart=/sbin/rngd"
+                        " -f -r /dev/urandom -o /dev/random#' %s"
+                      % rngd_srv_conf, shell=True)
```
balamuruhans: indentation to be fixed to pass inspekt

kylazhang: @balamuruhans Thanks and fixed

kylazhang:
sorry, still need some fix but seems push hang without response, will push soon

balamuruhans:
Thank you @kylazhang @will-Do, let us merge this patch

PR #1109, opened 2017-07-04
iface_options: Fix inaccurate prompt

dzhengfy:
The original prompt is inaccurate which causes the log on guest failure.
This is to make the prompt more accurate.


will-Do:
Same issue with https://github.com/avocado-framework/avocado-vt/pull/986

will-Do:
Same issue with https://github.com/avocado-framework/avocado-vt/pull/986

PR #1112, opened 2017-07-07
virsh_volume_application: Fix the image format

dzhengfy:
The default format of the created volume is 'raw', but attach_disk will
treat it as qcow2 image as default. That will cause the guest can not
recognize this disk. So the fix is to specify the disk format when
attaching it.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/volume/virsh_volume_application.py
+++ b/libvirt/tests/src/virsh_cmd/volume/virsh_volume_application.py
@@ -1,4 +1,5 @@
 import os
+import time
 import logging
 
 from autotest.client.shared import error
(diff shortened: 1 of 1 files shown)


dzhengfy:
@will-Do , I did minor update.

libvirt/tests/src/virsh_cmd/volume/virsh_volume_application.py
```diff
@@ -87,6 +88,7 @@ def run(test, params, env):
                                          % vm_attach_device)
                 return
             # Test if attached disk can be used normally
+            time.sleep(10) # Need seconds for the new disk to be recognized
```
balamuruhans: inspekt style expects, E261 at least two spaces before inline comment for travis to pass

dzhengfy:
@balamuruhans thanks and updated.

PR #1116, opened 2017-07-07
virsh_vol_create_from: Fix image size for qemu-img

dzhengfy:
qemu-img refuses creating a volume's device less than 8M on PPC, 4M on
X86. So the fix is to adjust the image size to fit for both PPC and X86.


dzhengfy:
@balamuruhans thanks and updated.

PR #1120, opened 2017-07-12
Check the domain status after svirt-install completed

yafu-1:
Check the domain status after svirt-install completed to avoid the possibilies 
  that 'virsh domstate' will get the paused status while virt-install executing 
  even if virt-install failed.

### DIFF
--- a/libvirt/tests/src/svirt/svirt_install.py
+++ b/libvirt/tests/src/svirt/svirt_install.py
@@ -1,6 +1,6 @@
 import os
 
-from autotest.client import utils
+from avocado.utils import process
 from autotest.client.shared import error
 
 from virttest import data_dir
@@ -68,8 +68,8 @@ def run(test, params, env):
         if sec_relabel is not None:
             cmd += ",relabel=%s" % sec_relabel
 
-        cmd += " --noautoconsole --graphics vnc --video vga &"
-        utils.run(cmd, ignore_status=True)
+        cmd += " --noautoconsole --graphics vnc --video vga"
+        process.run(cmd, timeout=600, ignore_status=True)
 
         def _vm_alive():
             return virsh.is_alive(vm_name)


yafu-1:
Test result with libvirt-3.2.0-14.el7.x86_64.
```
# avocado run svirt_install.negative_test.guest_img_label_virt_content.sec_relabel_no.sec_type_static.host_selinux_enforcing --vt-type libvirt
JOB ID     : 88f016618f1f7ed494912f80956b60cb7668577e
JOB LOG    : /root/avocado/job-results/job-2017-07-12T13.47-88f0166/job.log
TESTS      : 1
 (1/1) type_specific.io-github-autotest-libvirt.svirt_install.negative_test.guest_img_label_virt_content.sec_relabel_no.sec_type_static.host_selinux_enforcing: PASS (7.61 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0
JOB HTML   : /root/avocado/job-results/job-2017-07-12T13.47-88f0166/html/results.html
TESTS TIME : 7.61 s

```

chloerh:
Here is another idea:
we could set timeout for install command, instead of running the command in background. Then we could check guest status after the command finished. Then there should not be this kind of problem.

yafu-1:
Good idea. I have updated the code according to your suggestion.
Please help to review it again. Thanks.

balamuruhans:
Thanks @yafu-1,  @chloerh for review

libvirt/tests/src/svirt/svirt_install.py
```diff
@@ -68,8 +68,8 @@ def run(test, params, env):
         if sec_relabel is not None:
             cmd += ",relabel=%s" % sec_relabel
 
-        cmd += " --noautoconsole --graphics vnc --video vga &"
-        utils.run(cmd, ignore_status=True)
+        cmd += " --noautoconsole --graphics vnc --video vga"
+        process.run(cmd, timeout=600, ignore_status=True)
```
will-Do: so add 'timeout' can resolve the problem? as you mentioned in the commit summary about check the domain status

will-Do:
And you can also update the Pull Request subject and summary when the commit subject changed.

libvirt/tests/src/svirt/svirt_install.py
balamuruhans: I agree we might face timing issue but do you really need 600 seconds ?
As doc string says,
```
As this test only care whether the qemu-kvm process
    can access the image. It is not necessary to install
    a full os in a vm. Just verify the vm is alive after
    virt-install command is enough. Then we can save a lot
    of time and make this test independent from unattended_install
```

yafu-1: @will-Do @balamuruhans  
Actually, virt-install do the things just like 'virsh create' and exit quickly. The timeout is used to avoid the possibility that virt-install hang, then it will exit forcely after timeout.

kylazhang:
ACK and please help to merge.

balamuruhans:
Thanks all.

PR #1133, opened 2017-07-19
Enabling SELinux mode check

sudeeshjohn:
In Ubuntu SELinux is not working at the moment.
So Skipping testcases which needs the SELinux to be in
Enforcing mode. This patch will skip(insted of ERRORing out)
the the testcase if SELinux is disabled.

### DIFF
--- a/libvirt/tests/src/svirt/dac_nfs_disk.py
+++ b/libvirt/tests/src/svirt/dac_nfs_disk.py
@@ -2,8 +2,6 @@ import os
 import re
 import logging
 
-from autotest.client.shared import error
-
 from avocado.utils import process
 
 from virttest import qemu_storage
@@ -70,8 +68,8 @@ def run(test, params, env):
         try:
             img_val.append(int(params.get(i)))
         except ValueError:
-            raise error.TestNAError("%s value '%s' is not a number." %
-                                    (i, params.get(i)))
+            test.cancel("%s value '%s' is not a number." %
+                        (i, params.get(i)))
     # False positive - img_val was filled in the for loop above.
     # pylint: disable=E0632
     img_user, img_group, img_mode = img_val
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/svirt/dac_nfs_disk.py
```diff
@@ -2,10 +2,6 @@
 import re
 import logging
 
-from autotest.client.shared import error
```
balamuruhans: `error` is removed but still the file uses it, please replace all.

libvirt/tests/src/svirt/dac_nfs_disk.py
```diff
@@ -2,10 +2,6 @@
...
-
-from avocado.utils import process
```
balamuruhans: `process` is used in the file and removing it will break the test.

libvirt/tests/src/svirt/dac_start_destroy.py
```diff
@@ -4,9 +4,6 @@
 import grp
 import logging
 
-from autotest.client import utils
-from autotest.client.shared import error
```
balamuruhans: file is using `utils` please use `process` instead and replace `error` accordingly.

libvirt/tests/src/svirt/dac_vm_per_image_start.py
```diff
@@ -4,8 +4,6 @@
 import grp
 import logging
 
-from autotest.client.shared import error
```
balamuruhans: same here

libvirt/tests/src/svirt/dac_nfs_disk.py
```diff
@@ -3,7 +3,6 @@
 import logging
 
 from autotest.client.shared import error
```
sudeeshjohn: @balamuruhans do we need to avoid "autotest.client import utils" as well ?

balamuruhans: yes it is better and suggested to use Avocado APIs.

sudeeshjohn: @balamuruhans okie..

PR #1139, opened 2017-07-21
virtual_disks: Add wait time for PPC

dzhengfy:
Add the sleep time for the VM to recognize the attached disk on PPC.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -1,8 +1,10 @@
 import os
 import re
 import json
+import time
 import logging
 import aexpect
+import platform
 
 from avocado.utils import process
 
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -174,6 +175,10 @@ def check_vm_partitions(devices, targets_name, exists=True):
         logging.info("Checking VM partittion...")
         try:
             session = vm.wait_for_login()
+            # Here the script needs wait for a while for the guest to
+            # recognize the hotplugged disk on PPC
+            if platform.platform().count('ppc64'):
```
balamuruhans: we can do this once with variable and use it in other places.

dzhengfy: @balamuruhans updated using singe function. Please check. Thanks

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -811,9 +819,21 @@ def create_addtional_disk(disk_type, disk_path, disk_format, disk_device_type,
...
 
+        def add_sleep(sleep_seconds=10):
+            """
+            Check specific arch and do sleep for specified seconds.
+            This is to workaround for some failures due to different
+            VM response.
+
+            :param sleep_seconds: seconds for sleeping
+            """
+            if platform.platform().count('ppc64'):
```
balamuruhans: but if you see even here it has to call platform() every time add_sleep invoked. we can check the arch before even test starts (as value doesn't going to change during test) and use the variable to check for more efficiency. But your code looks good to me.

PR #1142, opened 2017-07-26
Don't skip the case with seclabel having no access to the image

yafu-1:
Since the guest failed to start as expected when the seclabel having
no access to the image, so mark the cases pass.


balamuruhans:
Looks good to me

kylazhang:
ACK.

PR #1144, opened 2017-07-28
Modify the keycode 'touch abc\n' to 'touch ab\n'

yafu-1:
Since the 'virsh send-key' send all the keys simulataneously, so the
string contains the same keys will have the possibility that the key
only received one time by the guest.


kylazhang:
ACK.

PR #1149, opened 2017-08-03
Fix startup_policy "None Type" errors

hahakiki2010:
From libvirt 2.0.0 on, the disk source element is unable to check
under error phenomenons. Fix by check disk source according
libvirt version.

### DIFF
--- a/libvirt/tests/src/virtual_disks/startup_policy.py
+++ b/libvirt/tests/src/virtual_disks/startup_policy.py
@@ -1,67 +1,11 @@
 import os
 import logging
 import shutil
-
 from autotest.client import utils
-from autotest.client.shared import error
-
 from virttest import virsh
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
-
-
-def create_disk_xml(xml_file, device_type, source_file, target_dev, policy):
-    """
-    Create a disk xml file for attaching to a domain.
-
-    :prams xml_file: path/file to save the disk XML
-    :source_file: disk's source file
-    :device_type: CD-ROM or floppy
-    """
-    if device_type == "cdrom":
-        target_bus = "ide"
-        image_size = "100M"
-    elif device_type == "floppy":
-        target_bus = "fdc"
-        image_size = "1.44M"
-    else:
-        error.TestNAError("Unsupport device type in this test: " + device_type)
-    utils.run("qemu-img create %s %s" % (source_file, image_size))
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_disks/startup_policy.py
```diff
@@ -3,67 +3,12 @@
 import shutil
 
 from autotest.client import utils
```
chunfuwen: replace from autotest.client import utils module with from avocado.utils import process since the previous one is deprecated.

libvirt/tests/src/virtual_disks/startup_policy.py
```diff
@@ -84,35 +29,69 @@ def run(test, params, env):
...
-    virsh_dargs = {'debug': True, 'ignore_status': True}
-    if vm.is_alive():
-        vm.destroy()
-    try:
-        virsh.attach_device(domainarg=vm_name, filearg=disk_xml_file,
-                            flagstr="--config", **virsh_dargs)
-    except:
-        os.remove(media_file)
-        raise error.TestError("Attach %s fail", device_type)
+        :prams xml_file: path/file to save the disk XML
```
chunfuwen: Suggest keep previous method create_disk_xml since it can  be used extend to support other protocol rather than file

hahakiki2010:
This pull request is only to address startup policy "None Type" issue . As for the code re-factoring mentioned above, will open another PR to fix them

libvirt/tests/src/virtual_disks/startup_policy.py
```diff
@@ -1,67 +1,12 @@
+
```
waynesun09: what's the empty line for?

hahakiki2010: removed

libvirt/tests/src/virtual_disks/startup_policy.py
```diff
@@ -1,67 +1,11 @@
 import os
 import logging
 import shutil
-
 from autotest.client import utils
```
waynesun09: pls remove utils also

hahakiki2010: @waynesun09 , for the deprecated code , I think chunfu has already have a PR to refactor the startup Policy code. So just focus on my changed part.

hahakiki2010:
Close this PR,  since there was one big change ever merged, and this issue shall have been fixed together

PR #1154, opened 2017-08-07
iface_network.py: Fix "cp" CmdError

hahakiki2010:
This issue is caused by process.system which has two cmds in one line.
After add shell=True, this issue resolved.


libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -49,8 +49,8 @@ def prepare_pxe_boot():
         # Download pxe boot images
         process.system("wget %s -O %s/initrd.img" % (boot_initrd, tftp_root))
         process.system("wget %s -O %s/vmlinuz" % (boot_vmlinuz, tftp_root))
-        process.system("cp -f /usr/share/syslinux/pxelinux.0 {0};"
-                       " mkdir -m 777 -p {0}/pxelinux.cfg".format(tftp_root))
+        process.system("cp -f /usr/share/syslinux/pxelinux.0 {0}".format(tftp_root))
```
chunfuwen: Please add parameter:  shell=True when calling process.system to fix this issue

hahakiki2010: resolved

PR #1156, opened 2017-08-08
Rename vm in v2v conversion

chloerh:
To avoid conflict when converting to rhvm. Add sufix of 3 letters string
to original vm name.

### DIFF
--- a/v2v/tests/src/convert_vm_to_ovirt.py
+++ b/v2v/tests/src/convert_vm_to_ovirt.py
@@ -2,7 +2,6 @@ import os
 import logging
 import re
 
-from avocado.core import exceptions
 from avocado.utils import process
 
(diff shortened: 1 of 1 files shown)


kylazhang:
ACK.

v2v/tests/src/convert_vm_to_ovirt.py
```diff
@@ -121,6 +122,8 @@ def run(test, params, env):
         if v2v_ret.exit_status != 0:
             raise exceptions.TestFail("Convert VM failed")
```
kylazhang: Can you also update with changing test fail?

chloerh: Sure :)

PR #1157, opened 2017-08-08
Rename vm name when converting disk

chloerh:
In case vm with same name exists.
Also extend timeout to 2400s for esx cases since it will usually
take more time than 1200s.

### DIFF
--- a/v2v/tests/cfg/v2v_options.cfg
+++ b/v2v/tests/cfg/v2v_options.cfg
@@ -77,8 +77,8 @@
                             input_mode = disk
                         - guest:
                             input_mode = libvirt
-                            new_vm_name = ${main_vm}_new
-                            v2v_options = "-on ${new_vm_name}"
+                    new_vm_name = ${main_vm}_new
+                    v2v_options = "-on ${new_vm_name}"
                 - libvirt:
                     input_mode = "libvirt"
                     variants:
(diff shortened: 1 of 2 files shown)


v2v/tests/cfg/v2v_options.cfg
```diff
@@ -77,8 +77,8 @@
                             input_mode = disk
                         - guest:
                             input_mode = libvirt
-                            new_vm_name = ${main_vm}_new
-                            v2v_options = "-on ${new_vm_name}"
+                    new_vm_name = ${main_vm}_new
+                    v2v_options = "-on ${new_vm_name}"
```
kylazhang: I think add in above just under the variants will be more clear?

v2v/tests/cfg/v2v_options.cfg
```diff
@@ -199,11 +199,10 @@
                     variants:
                         - privileged:
                             oc_uri = "qemu:///system"
-                            v2v_options = "-oc ${oc_uri}"
                         - unprivileged:
                             oc_uri = "qemu:///session"
-                            v2v_options = "-oc ${oc_uri}"
                             unprivileged_user = "USER._V2V_EXAMPLE"
+                    v2v_options = "-oc ${oc_uri} -on ${new_vm_name}"
```
kylazhang: save as above

v2v/tests/cfg/v2v_options.cfg
chloerh: I hope to save a few lines by this way, maybe less duplicated lines?

PR #1160, opened 2017-08-09
v2v_options:fix some case failures

chloerh:
- cannot install vm in 'ic' test
- wrong params for esx test

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -593,7 +593,7 @@ def run(test, params, env):
 
         # Create password file for access to ESX hypervisor
         if hypervisor == 'esx':
-            vpx_passwd = params.get("vpx_passwd")
+            vpx_passwd = params.get("vpx_password")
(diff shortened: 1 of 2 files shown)


v2v/tests/src/v2v_options.py
```diff
@@ -617,7 +617,10 @@ def check_alloc():
             sh_install_vm = params.get('sh_install_vm')
             if not sh_install_vm:
                 test.error('Source vm installing script missing')
-            process.run('su - %s -c %s' % (v2v_user, sh_install_vm))
+            with open(sh_install_vm) as fh:
+                cmd_install_vm = fh.read().strip()
+            process.run('su - %s -c "%s"' % (v2v_user, cmd_install_vm),
+                        timeout=10, shell=True)
```
kylazhang: Why using file to get install cmd? The string can not met your requirement?

chloerh: The string is kind of long, with 300+ characters. It could be placed in cfg file, just not pretty, maybe.

PR #1162, opened 2017-08-11
Controller_functional: adapt more machine type by auto

hsmj1412:
The case should adpat more machine type instead of q35 or i440fx

### DIFF
--- a/libvirt/tests/src/controller/controller_functional.py
+++ b/libvirt/tests/src/controller/controller_functional.py
@@ -51,6 +51,8 @@ def run(test, params, env):
         """
         Prepare os part of VM XML according to params.
         """
+        if os_machine == 'auto':
+            return
         osxml = vm_xml.os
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/controller/controller_functional.py
```diff
@@ -246,7 +248,7 @@ def check_msi():
                 raise error.TestFail('Expect MSI enable with non-zero vectors,'
                                      ' but got %s' % output)
 
-    os_machine = params.get('os_machine', 'i440fx')
+    os_machine = params.get('os_machine', 'auto')
```
waynesun09: this is tricky fix with on x86 default as i440fx and pseries on ppc, and q35 need specified in cfg, still you need replace i440fx to pseries on ppc to let case run.

This fix just make explicit cfg to implicit, which I don't like, but base on current libvirt support this could work

PR #1164, opened 2017-08-14
virsh_vcpupin: Fix the max cpus

hsmj1412:
There is some difference between cpu_max and online_cpu_max when part of cpus are offline


waynesun09:
@hsmj1412 pls remove the not needed merge commit in PR

PR #1165, opened 2017-08-15
Fix mem allocate failure

hahakiki2010:
when libvirt mem testing with hugepages, it shall not use the system available memory to set hugepages. Otherwise vm may fail to start because the target mem can't allocate  

### DIFF
--- a/libvirt/tests/cfg/libvirt_mem.cfg
+++ b/libvirt/tests/cfg/libvirt_mem.cfg
@@ -57,6 +57,7 @@
                     page_unit = "KiB"
                     node_mask = 0
                     huge_pages = "{'size':'2048','unit':'KiB','nodeset':'0'}"
+                    hugepage_force_allocate = "yes"
                     variants:
                         - hot_plug:
                             max_mem = "2097152"


libvirt/tests/cfg/libvirt_mem.cfg
```diff
@@ -57,6 +57,7 @@
                     page_unit = "KiB"
                     node_mask = 0
                     huge_pages = "{'size':'2048','unit':'KiB','nodeset':'0'}"
+                    hugepage_force_allocate = "yes"
```
balamuruhans: so when sufficient hugepage mem is not available in host and guest is backed by hugepages, this param will not raise error in setup but any way VM fails to start ?

hahakiki2010: Given hugepage_force_allocate is "no", it will warn in setup but not raise error. Then VM fails to start anyway
But by setting the param to "yes", error will be raised in setup instead

balamuruhans: Agreed.

waynesun09:
ACK and merge now, thanks

PR #1167, opened 2017-08-15
Add time delay for virsh.reset to take effect

hahakiki2010:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_reset.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_reset.py
@@ -1,5 +1,6 @@
 import logging
 import commands
+import time
 
 from autotest.client.shared import error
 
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_reset.py
```diff
@@ -98,6 +99,7 @@ def change_cache(vmxml, mode):
             raise error.TestFail("Expect fail, but succeed indeed.")
 
         session.close()
+        time.sleep(5)
```
chunfuwen: not sure why you need additional wait time here since wait_for_login already consume wait time.

hahakiki2010: it's correlated to a debug failure, the tmp file still exist after virsh.reset. And adding sleep time here is expected to make sure the file check is performing on server after reset, but not still on the existing one.

waynesun09:
ACK and merge now, thx

PR #1172, opened 2017-08-21
virsh_domblkerror:add clean up step for mount point

chloerh:
Unmount nfs mount point if there is a mount point and it is mounted.


waynesun09:
NACK, the fix should be in line 186 with why nfs clean up did not work

PR #1175, opened 2017-08-22
fix a conditional judgement error

carlsplace:
this judgement error always set disk.readonly to True, which blocks the snapshot creation.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -708,7 +708,7 @@ def run(test, params, env):
 
             if bootdisk_snapshot != "":
                 disk.snapshot = bootdisk_snapshot
-                if device_readonly:
+                if device_readonly[0] == 'yes':
                     disk.readonly = True
 
             disk.target = {"dev": bootdisk_target, "bus": bootdisk_bus}


libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -708,7 +708,7 @@ def create_addtional_disk(disk_type, disk_path, disk_format, disk_device_type,
 
             if bootdisk_snapshot != "":
                 disk.snapshot = bootdisk_snapshot
-                if device_readonly:
+                if device_readonly[0] == 'yes':
```
balamuruhans: `virt_disk_boot_order = "2 1"`, so this check comes in boot disk where boot_order is 1. I doubt device_readonly[0] will have the readonly flag for boot order 2 which is not boot disk ?

chunfuwen: The is not related to virt_disk_boot_order since device_readonly value is updated based on bootdisk_snapshot != "". And only disk_virtio_snapshot under single_disk_test  will go into this logic.
But anyway, please run all related cases in order to avoid regression by this change.

carlsplace: I checked 6 related cases(`virtual_disks.multidisks.coldplug.single_disk_test..disk_virtio_snapshot..`) and the boot order case mentioned above(`virtual_disks.multidisks.coldplug.multi_disks_test.disks_bootorder`). 
All the test cases passed successfully on my local machine.
The code change only affects the single disk test part.

PR #1179, opened 2017-08-23
virsh_domblkerror:add params for cleanup step

chloerh:
nfs mount point could not be unmounted for not all necessary params
are provided including export_dir and mount_dir.


waynesun09:
ACK

PR #1182, opened 2017-08-24
make sure the additional vm is running

carlsplace:
make sure the cloned VM is running before attaching a disk, this change only affects `disk_shareable` test.


waynesun09:
ACK

PR #1186, opened 2017-08-25
fix iface_hotplug failure

kylazhang:
resolve problem:  polltime = time.time() + poll_timeout
TypeError: unsupported operand type(s) for +: 'float' and 'str'


kylazhang:
merged as it is very small change and it affect out current testing

PR #1187, opened 2017-08-25
migrate_vm: Enable required ports for migration

dzhengfy:
The firewalld service is started on source host, so migration from
target host to source host will fail due to some required ports
disabled. This is to enable those ports before this kind of migration
is executes.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1104,6 +1104,9 @@ def run(test, params, env):
     # For --migrate-disks test
     migrate_disks = "yes" == test_dict.get("migrate_disks", "no")
 
+    # For bi-directional and tls reverse test
+    src_uri = test_dict.get("migration_source_uri", "qemu:///system")
+
     # Pre-creation image parameters
     target_pool_name = test_dict.get("target_pool_name", "temp_pool_1")
     target_pool_type = test_dict.get("target_pool_type", "dir")
@@ -1231,6 +1234,7 @@ def run(test, params, env):
     remote_virsh_dargs = {'remote_ip': server_ip, 'remote_user': server_user,
                           'remote_pwd': server_pwd, 'unprivileged_user': None,
                           'ssh_remote_auth': True}
+    migrate_setup = libvirt.MigrationTest()
 
     try:
         if iscsi_setup:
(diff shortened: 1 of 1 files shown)


dzhengfy:
Depends on avocado-vt 1132

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1104,6 +1104,9 @@ def run(test, params, env):
     # For --migrate-disks test
     migrate_disks = "yes" == test_dict.get("migrate_disks", "no")
 
+    # For bi-directional and tls reverse test
+    src_uri = test_dict.get("migration_source_uri", "qemu:///system")
```
chunfuwen: I don't see migration_source_uri was defined in anywhere

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2466,23 +2475,31 @@ def run(test, params, env):
             check_vm_disk_after_migration(vm, test_dict)
 
         if migr_vm_back:
-            src_uri = test_dict.get("migration_source_uri")
+            migrate_setup = libvirt.MigrationTest()
```
chunfuwen: migrate_setup = libvirt.MigrationTest()
migrate_setup.migrate_pre_setup(src_uri, params, ports=uri_port)

those 2 statements were used at least twice,please use function to wrap up them

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2466,23 +2475,31 @@ def run(test, params, env):
...
+            # Pre migration setup for local machine
+            migrate_setup.migrate_pre_setup(src_uri, params)
             cmd = "virsh migrate %s %s %s" % (vm_name,
                                               virsh_options, src_uri)
             logging.debug("Start migrating: %s", cmd)
             status, output = run_remote_cmd(cmd, server_ip, server_user,
                                             server_pwd)
+            logging.info(output)
+            # Clean up of pre migration setup for local machine
+            migrate_setup.migrate_pre_setup(src_uri, params,
```
chunfuwen: Contrast with your code on line 2500~2501,it need some conditional cleanup, so a little wonder why here it doesn't need any condition?

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @chunfuwen , migration_source_uri  is in cfg file.

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @chunfuwen updated

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @chunfuwen updated

PR #1190, opened 2017-08-28
virsh_dump: remove os.system and utils.run

chloerh:
Use process.run from avocado instead.Therefore test log won't be
printed in console.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
@@ -4,8 +4,7 @@ import commands
 import time
 import signal
 
-from autotest.client.shared import error
-from autotest.client.shared import utils
+from avocado.utils import process
 
 from virttest import virsh
 from virttest import utils_libvirtd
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
```diff
@@ -5,7 +5,8 @@
 import signal
 
 from autotest.client.shared import error
```
waynesun09: please also help remove this

chloerh: Since the 'raise' part is not in 'run' function, to replace 'raise xxx' with 'test.xxx' is not an option, maybe need some big change.

libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
```diff
@@ -61,7 +62,7 @@ def check_bypass(dump_file):
 
     cmd1 = "lsof -w %s" % dump_file
     while True:
-        if not os.path.exists(dump_file) or os.system(cmd1):
+        if not os.path.exists(dump_file) or process.run(cmd1).exit_status:
```
balamuruhans: `process.system()` can help to return you the status directly.

libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
```diff
@@ -229,7 +230,7 @@ def check_dump_format(dump_image_format, dump_file):
             vm_pid = vm.get_pid()
             cmd = "cat /proc/%d/cmdline|xargs -0 echo" % vm_pid
             cmd += "|grep dump-guest-core=%s" % dump_guest_core
-            result = utils.run(cmd, ignore_status=True)
+            result = process.run(cmd, ignore_status=True)
```
balamuruhans: This would break, please use `shell=True`

libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
balamuruhans: I would agree with @waynesun09 suggestion, let us make it better while we get the opportunity.

PR #1192, opened 2017-08-28
virsh_snapshot_create_as: Check for the package "targetcli"

MalleshKoti:
…s installed and install if it is not there.

Got Error to install targetcli package to run the test so added code to install targetcli.

Signed-off-by: Mallesh Koti <mallesh@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_create_as.py
+++ b/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_create_as.py
@@ -8,6 +8,7 @@ import logging
 from autotest.client.shared import error
 from autotest.client import utils
 
+from virttest import utils_package
 from virttest import virsh
 from virttest import utils_misc
 from virttest import xml_utils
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_create_as.py
```diff
@@ -410,6 +411,9 @@ def run(test, params, env):
     vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     logging.debug("original xml is %s", vmxml_backup)
 
+    if not utils_package.package_install('targetcli'):
```
balamuruhans: This looks to be command, we should consider package name for installation.
`tgt` is the package name for Ubuntu based distro and `scsi-target-utils` for rhel based distro.

so IMO you can check whether this command is available or not and then proceed the package installation accordingly.

waynesun09: NACK

please check virttest/iscsi.py module, both targetcli and tgtadm will be checked and raise error to user for install the package.

You could add the package in your env preparation as required package base on OS, as iscsi is widely used in this test suite.

balamuruhans: @MalleshKoti I have handled it from framework, https://github.com/avocado-framework/avocado-vt/pull/1161, please help to check whether it resolves your issue.

iccaszhulili:
rh-caselink retest please

dzhengfy:

PR #1201, opened 2017-08-31
virsh_dompmsuspend: PowerPC doesn't support ACPI sleep states

balamuruhans:
In x86 pm tags of guest xml is to enable support for (suspend-to-mem) and
S4 (suspend-to-disk) ACPI sleep states, but it is not supported and pm tags
in guest xml are not required for PowerPC.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import platform
 
 from virttest import virt_vm
 from virttest import virsh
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_dompmsuspend.py
```diff
@@ -63,11 +65,12 @@ def run(test, params, env):
...
 
-    # Setup possible failure patterns
-    if pm_enabled == 'not_set':
-        fail_pat.append('not supported')
-    if pm_enabled == 'no':
-        fail_pat.append('disabled')
+    # Setup possible failure patterns excluding ppc
+    if "ppc64" not in arch:
+        if pm_enabled == 'not_set':
+            fail_pat.append('not supported')
```
sathnaga: @balamuruhans do we require this as it already got skipped in cfg?

balamuruhans: @sathnaga Only 1 testcase gets skipped for ppc from cfg, here we don't configure pm tags in guest xml for ppc, as guest fails to start. we have to test suspend-to-mem scenario without pm tags for ppc without affecting x86. This check is for that.

sathnaga: @balamuruhans okies

PR #1203, opened 2017-09-04
virsh_domblkerror: enable selinux for distro other than ubuntu

balamuruhans:
This patchset fixes the following,
	- test failed while performing selinux configuration on ubuntu machine
	- nfs service name for ubuntu is nfs-kernel-server
	- UnboundLocalError: local variable 'nfs_service' referenced before assignment
	  for variable in clean up part which is not initialized.


balamuruhans:
Let's close this PR as it is fixed in #1219.

PR #1208, opened 2017-09-06
virsh_sendkey: keystrokes get missed if all of them are sent simultaneously

balamuruhans:
Test fails when keystrokes are sent in a shot it gets missed or received in
random order, provision is given for multiple invocations to avoid the failures

param sendkey_simultaneous can be set to "no" to invoke each keystroke separately
and can be set to "yes" if all the keystrokes to be invoked at a time.
    
with_holdtime: holdtime with keystrokes will emulate keys in different count
for the milliseconds eachtime and will create filename with different name,
we can't validate the filename when keys are pressed with holdtime milliseconds
as it is not persistent

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
@@ -22,7 +22,9 @@ def run(test, params, env):
 
     vm_name = params.get("main_vm", "avocado-vt-vm1")
     status_error = ("yes" == params.get("status_error", "no"))
-    options = params.get("sendkey_options", "")
+    keystrokes = params.get("sendkey", "")
+    codeset = params.get("codeset", "")
+    holdtime = params.get("holdtime", "")
     sysrq_test = ("yes" == params.get("sendkey_sysrq", "no"))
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
```diff
@@ -92,9 +94,24 @@ def send_line(send_str):
...
 
-        output = virsh.sendkey(vm_name, options, readonly=readonly,
-                               unprivileged_user=unprivileged_user,
-                               uri=uri)
+        if sysrq_test:
+            output = virsh.sendkey(vm_name, keystrokes, codeset=codeset,
+                                   holdtime=holdtime, readonly=readonly,
+                                   unprivileged_user=unprivileged_user,
+                                   uri=uri)
+        else:
```
sathnaga: What if we have to add a scenario with all keystrokes send together, can we have provision for that using 
a param, if set do as similar to `sysrq_test`?

balamuruhans: yes, it is a good idea.

sathnaga:
@balamuruhans looks like it has conflict can you rebase and fix above comments, Thanks

PR #1212, opened 2017-09-07
virsh_setvcpus: fix local variable 'result' referenced before assignment

cliping:
Signed-off-by: Liping Cheng <lcheng@redhat.com>


balamuruhans:
@cliping Thanks

PR #1214, opened 2017-09-14
Add v2v function cases

chloerh:
- convert disk image using -if option with the wrong format

### DIFF
--- a/v2v/tests/cfg/v2v_options.cfg
+++ b/v2v/tests/cfg/v2v_options.cfg
@@ -396,3 +396,16 @@
                             not_in_man = --debug-gc
                         - sles_opensuse:
                             not_in_man = 'Virt-v2v is able to import Xen guests from RHEL 5 Xen or SLES and openSUSE'
+                - if_wrong_format:
+                    only output_mode.none
+                    variants:
+                        - if_qcow2:
+                            only input_mode.disk.image.raw_format.sparse
+                            input_format = qcow2
+                            msg_content = 'Image is not in qcow2 format'
+                            expect_msg = yes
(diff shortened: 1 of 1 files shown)


v2v/tests/cfg/v2v_options.cfg
```diff
@@ -396,3 +396,16 @@
                             not_in_man = --debug-gc
                         - sles_opensuse:
                             not_in_man = 'Virt-v2v is able to import Xen guests from RHEL 5 Xen or SLES and openSUSE'
+                - if_wrong_format:
+                    only output_mode.none
```
chunfuwen: is this filter conflict with line 403 ?

v2v/tests/cfg/v2v_options.cfg
```diff
@@ -396,3 +396,16 @@
                             not_in_man = --debug-gc
                         - sles_opensuse:
                             not_in_man = 'Virt-v2v is able to import Xen guests from RHEL 5 Xen or SLES and openSUSE'
+                - if_wrong_format:
```
chunfuwen: if_wrong_format to wrong_format_if ?

v2v/tests/cfg/v2v_options.cfg
chloerh: This line is about output_mode, L403 is about input_mode, therefore no conflict

v2v/tests/cfg/v2v_options.cfg
chloerh: 'if' is an option for v2v, I like my way better :)

PR #1217, opened 2017-09-15
Add v2v function case of converting from esx

chloerh:
- Convert esx guest that has vmtools installed

### DIFF
--- a/v2v/tests/src/function_test_esx.py
+++ b/v2v/tests/src/function_test_esx.py
@@ -56,17 +56,32 @@ def run(test, params, env):
             if "device='cdrom'" not in xml:
                 log_fail('CDROM no longer exists')
 
-    def check_vmtools(vmcheck):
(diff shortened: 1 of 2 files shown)


v2v/tests/src/function_test_esx.py
```diff
@@ -56,17 +56,28 @@ def check_device_exist(check, virsh_session_id):
             if "device='cdrom'" not in xml:
                 log_fail('CDROM no longer exists')
 
-    def check_vmtools(vmcheck):
+    def check_vmtools(vmcheck, check):
         """
-        Check whether vmware tools packages have been removed
+        Check whether vmware tools packages have been removed,
+        or vmware-tools service has stopped
```
kylazhang: The explain for the func params?

weikunzz:
looks good to me.

waynesun09:
ACK

PR #1219, opened 2017-09-15
virsh_domblkerror: cleanup is not done properly

balamuruhans:
when iscsi setup is not proper, test cleanup errors with,
Attribute error: 'NoneType' object has no attribute 'export_flag'.
This patch handles this condition

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -4,6 +4,7 @@ import shutil
 import os
 
 from avocado.utils import process
+from avocado.utils import distro
 
 from virttest import virsh
 from virttest import data_dir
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -67,12 +70,16 @@ def run(test, params, env):
...
                                                export_dir=img_dir)
-            selinux_bak = res["selinux_status_bak"]
+            if not ubuntu:
+                selinux_bak = res["selinux_status_bak"]
             process.run("mount -o nolock,soft,timeo=1,retrans=1,retry=0 "
                         "127.0.0.1:%s %s" % (img_dir, nfs_dir), shell=True,
                         verbose=True)
             img_path = os.path.join(nfs_dir, img_name)
-            nfs_service = Factory.create_service("nfs")
+            if not ubuntu:
```
chunfuwen: suggest use variable name such as nfs_service_package, if not ubuntu: nfs_package_name  = 'nfs',therefore here we can remove duplicate if ...else statement

PR #1221, opened 2017-09-18
virsh.domblkerror: fix guest not paused after nfs stop

hahakiki2010:
When stop nfs on host, it is expected guest could be paused
automatically. But this sometimes doesn't work, since seemingly
vm can't detect the nfs status timely. Fix by add one more write
to the vm mnt disk if vm is not paused so that vm can know the 
nfs has gone.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -162,7 +162,10 @@ def run(test, params, env):
             return (vm.state() == "paused")
 
         if not utils_misc.wait_for(_check_state, timeout):
-            test.fail("Guest does not paused, it is %s now" % vm.state())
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -162,7 +162,10 @@ def _check_state():
             return (vm.state() == "paused")
 
         if not utils_misc.wait_for(_check_state, timeout):
-            test.fail("Guest does not paused, it is %s now" % vm.state())
+            # if not paused, perform one more IO operation to the mnt disk
```
chunfuwen: if to If

hahakiki2010: @chunfuwen fixed

PR #1237, opened 2017-10-12
Fix the mistakes in USB controller tests

iccaszhulili:
Fix the following mistakes:
(1) the usb_controller params in cfg file is never been used in
the script.
(2) the usb controller can be used in both i440fx and q35 machine.
(3) the companian controller models, such as ich9-uhci[123], are
usually used only with ich9-ehci1 all together, rather than used with
ich9-ehci1 alone.
(4) qemu cmdline is never been checked in usb controller tests.


iccaszhulili:
The original manual cases related to the USB2.0 cases in this cfg file has been deleted. Hence, the cfg file changes can be made without changing anything about caselink.

PR #1240, opened 2017-10-13
remote_access: correct configuration to get vm

dzhengfy:
In configuration, vms = "" will cause env.get_vm to return zero vms.


waynesun09:
ACK

PR #1243, opened 2017-10-16
migrate_vm: skip to compare 'expected downtime'

dzhengfy:
"Expected Downtime" is newly added in domjobinfo on source host for
migration and "Total Downtime" for target host, too. Currently we do not
need to compare this two fields. So they are skipped for comparison.


kylazhang:
@chunfuwen

PR #1244, opened 2017-10-17
Fix virtual_disk_multidisks case

hsmj1412:
Boot disk couldn't be set readonly


hsmj1412:
@chunfuwen please help review

PR #1247, opened 2017-10-18
Add two guest from two new ova compress method

weikunzz:
Add pigz and xz compress methods ova file

Signed-off-by: Kun Wei <kuwei@redhat.com>


balamuruhans:
@weikunzz please squash it to single commit.

PR #1248, opened 2017-10-18
virsh_vol_resize: Correct the configuration for no_space_allocation

hahakiki2010:
To trigger no_space_allocation, it shall set the new vol capacity
by using the pool capacity, rather than some hard coding in cfg file.
The old way can fail the case too, but for "invalid argument" not for 
"no space left". 


balamuruhans:
Thanks @hahakiki2010, let us merge this.

PR #1260, opened 2017-10-31
Update code for getting size of available space

chloerh:
This could fix the problem of can't get correct output with 'df'
command.


balamuruhans:
Thanks all, let us merge this.

PR #1267, opened 2017-11-08

prudmirya:
This test case will check two scenarios on is positive scenario and other is negative scnario. pcc guest support max 32 vphbs(0 to 31) it will check with 32 controllers weather vm is able to start or not and second one is with controller index number 32  it check weather vm undefine or not


prudmirya:
[job.txt](https://github.com/autotest/tp-libvirt/files/1455248/job.txt)
job log added

libvirt/tests/cfg/libvirt_max_vphb.cfg
```diff
@@ -0,0 +1,9 @@
+- libvirt_max_vphb:
+    type = libvirt_max_vphb
+    variants:
+        - model:
+            only ppc64le
+            # ppc supports pc-root architecture model
```
sathnaga: s/pc-root/pci-root, i guess this comment can be removed

prudmirya:
closing current PR and open new PR #1277

PR #1273, opened 2017-11-14
Remove unnecessary log and rearrange code steps

lento-sun:
1. Disk xml info is already logged in libvirt.create_disk_xml(),
   no need to be logged again.
2. Starting/Destroying a vm could be put at beginning to avoid
   waste of time to prepare env such as iscsi backend if vm got
   failure.

Signed-of-by: Yi Sun <yisun@redhat.com>


lento-sun:
@chunfuwen this fix could avoid following error
`IOError: [Errno 2] No such file or directory: '/var/tmp/xml_utils_temp_tCHg7Y.xml'&#10;`
That error happened after some vm.wati_for_login(), but not always reproduced, with this fix that issue could be avoided.

PR #1274, opened 2017-11-14
Empty domain operation should not trigger logging.error

lento-sun:
For empty domain operation, just skip related test step is ok,
no need to log it as an error, it's confusing. Now change it to
logging.debug.


balamuruhans:
@lento-sun Thanks, let us merge it.

PR #1275, opened 2017-11-15
migrate_vm: Fix the machine type for remote guest

dzhengfy:
This is to change the machine type to alias value in order to leverage
libvirt to automatically create the machine type for the guest. This is
required when qemu-kvm and qemu-kvm-rhev are installed one by one which
have different machine type.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -4,6 +4,7 @@ import re
 import signal
 import time
 import aexpect
+import platform
 
 from subprocess import PIPE
 from subprocess import Popen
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2237,11 +2238,19 @@ def run(test, params, env):
...
+                logging.debug("Modify remote guest xml's disk path")
                 pattern2repl = {r".*.qcow2.*":
                                 "<source file='%s/%s'/>"
                                 % (nfs_mount_dir, image_name)}
-
-                logging.debug("Modify the config file using pattern")
+                guest_config.sub(pattern2repl)
+                logging.debug("Modify remote guest xml's machine type")
+                machine_type = "pc"
+                if platform.platform().count("ppc64le"):
```
balamuruhans: instead calling twice here and in L2253 to get Arch, it is better to have a variable and reuse it.

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2237,11 +2238,19 @@ def run(test, params, env):
                                                  password=server_pwd,
                                                  port='22',
                                                  remote_path=xml_path)
+                logging.debug("Modify remote guest xml's disk path")
                 pattern2repl = {r".*.qcow2.*":
                                 "<source file='%s/%s'/>"
                                 % (nfs_mount_dir, image_name)}
-
-                logging.debug("Modify the config file using pattern")
+                guest_config.sub(pattern2repl)
```
chunfuwen: Are you sure  that this command will not be executed when platform contains "ppc64le" ? otherwise guest_config.sub(pattern2repl) will be called twice respectively in line 2245 and line 2254.

dzhengfy: @chunfuwen , line 2245 and line 2254 are both needed to be executed as they are different modifications, one for source image path and the other is for arch and machine type.

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: @balamuruhans Thanks. I did update.

PR #1277, opened 2017-11-15

prudmirya:

### DIFF
--- a/libvirt/tests/cfg/libvirt_max_vphb.cfg
+++ b/libvirt/tests/cfg/libvirt_max_vphb.cfg
@@ -1,0 +1,9 @@
+- libvirt_max_vphb:
+    type = libvirt_max_vphb
+    variants:
+        - model:
+            only ppc64le
+            # ppc supports pci-root architecture model
+            model = "pci-root"
+            # on guest we can create maximum 32 vphbs
+            max_pci = "32"
--- a/libvirt/tests/src/libvirt_max_vphb.py
+++ b/libvirt/tests/src/libvirt_max_vphb.py
@@ -1,0 +1,53 @@
+import logging
+from virttest import virsh
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest.libvirt_xml.devices.controller import Controller
+
+
+def run(test, params, env):
+    """
+    1) get params from cfg file
+    2) create a controller with index equal to 1
+    3) add controller to xml
+    4) start the Vm
+       Vm should start with index 1 to 31, for index 32 it should not.
+    5) shutdown the Vm
+    6) repeat the steps  2 to 5, but each time
+       increase the index value in step2
+    """
+    # get the params from params
+    vm_name = params.get("main_vm")
+    max_pci = int(params.get("max_pci", "32"))
+    model = params.get("model", "ENTER_MODEL")
+    if model.count("ENTER_MODEL"):
+        test.cancel("Please enter model.")
+    vm = env.get_vm(vm_name)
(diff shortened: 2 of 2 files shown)


prudmirya:
[job(2).txt](https://github.com/autotest/tp-libvirt/files/1474089/job.2.txt)
job log added

libvirt/tests/cfg/libvirt_max_vphb.cfg
```diff
@@ -0,0 +1,9 @@
+- libvirt_max_vphb:
+    type = libvirt_max_vphb
+    variants:
+        - model:
+            only ppc64le
+            # ppc supports pci-root architecture model
```
sathnaga: comment not needed

libvirt/tests/src/libvirt_max_vphb.py
```diff
@@ -0,0 +1,53 @@
...
+            controller.index = index
+            controller.model = model
+            devices.append(controller)
+            try:
+                vmxml.set_devices(devices)
+                vmxml.sync()
+            except Exception as e:
+                if index < max_pci:
+                    test.fail("Vm fail to define with controller index %s"
+                              % index)
```
sathnaga: log exception error also

libvirt/tests/src/libvirt_max_vphb.py
```diff
@@ -0,0 +1,53 @@
...
+                else:
```
sathnaga: instead of else it can be a definite check index == 32, so that any changes in future to that value can be addressed. 
given that max_pci should be a constant=32, why to keep it as user defined param?

libvirt/tests/src/libvirt_max_vphb.py
```diff
@@ -0,0 +1,53 @@
...
+                    logging.debug("PPC supports maximum 32 vphbs(0 to 31)")
```
sathnaga: here it has failed to define vm, ideally we should exit the testcase here not continue.

prudmirya: any how it will fail in loop condition

libvirt/tests/src/libvirt_max_vphb.py
```diff
@@ -0,0 +1,53 @@
...
+        for index in range(1, max_pci+1):
+            controller = Controller("controller")
+            controller.type = "pci"
+            controller.index = index
+            controller.model = model
+            devices.append(controller)
+            try:
+                vmxml.set_devices(devices)
+                vmxml.sync()
+            except Exception as e:
```
balamuruhans: it is better to handle expected `LibvirtXMLError` exception instead.

libvirt/tests/src/libvirt_max_vphb.py
sathnaga: Sure, agree, overlooked that `continue` inside  `if..else` actually it is outside

libvirt/tests/src/libvirt_max_vphb.py
prudmirya: it is not just about libvit xml error, i seeing lot of other exception related to xml operation, so used Exception class

libvirt/tests/src/libvirt_max_vphb.py
sathnaga: Even if that is the case you should move ‘test.fail()’ out of loop for continue to really work

prudmirya:
opened new PR #1282 , closing current PR

PR #1280, opened 2017-11-17
Update nodeset for numad test

chloerh:
Instead of setting a fixed value of '0-1', we choose 2 from the
online nodes: the max and the minium. For numa node is not
continuous on power hosts.

### DIFF
--- a/libvirt/tests/src/numa/numad_vcpupin.py
+++ b/libvirt/tests/src/numa/numad_vcpupin.py
@@ -6,7 +6,6 @@ from autotest.client.shared import error
 from virttest import libvirt_xml
 from virttest import virsh
 from virttest import utils_misc
-from virttest import utils_test
 from virttest import utils_libvirtd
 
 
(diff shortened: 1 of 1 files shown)


chloerh:
**Test result on POWER 9**

 (1/1) type_specific.io-github-autotest-libvirt.numad_vcpupin: PASS (75.20 s)
JOB TIME   : 83.57 s

libvirt/tests/src/numa/numad_vcpupin.py
```diff
@@ -37,14 +36,11 @@ def run(test, params, env):
...
         logging.debug("host node list is %s", node_list)
-        if numa_memory.get('nodeset'):
-            used_node = utils_test.libvirt.cpus_parser(numa_memory['nodeset'])
-            logging.debug("set node list is %s", used_node)
-            if not status_error:
-                for i in used_node:
-                    if i > max(node_list):
-                        raise error.TestNAError("nodeset %s out of range" %
-                                                numa_memory['nodeset'])
+        node_a, node_b = min(node_list), max(node_list)
```
chunfuwen: if node_list is empty list, min(xx)will throw valueError, so maybe need do assertion for node_list

chloerh: Thanks for the review, updated.

PR #1283, opened 2017-11-20

prudmirya:

Signed-off-by: prudhvi mprudhvi@linux.vnet.ibm.com


prudmirya:
[log.txt](https://github.com/autotest/tp-libvirt/files/1487830/log.txt)
job log added

PR #1279, opened 2017-11-16
Add multiple controller case

hsmj1412:
Only support PPC

Signed-off-by: Junxiang Li <junli@redhat.com>

PR #1284, opened 2017-11-20

prudmirya:

### DIFF
--- a/libvirt/tests/cfg/libvirt_max_vphb.cfg
+++ b/libvirt/tests/cfg/libvirt_max_vphb.cfg
@@ -1,0 +1,6 @@
+- libvirt_max_vphb:
+    type = libvirt_max_vphb
+    variants:
+        - model:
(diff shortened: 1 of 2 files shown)


prudmirya:
[log.txt](https://github.com/autotest/tp-libvirt/files/1487903/log.txt)
job log added

chunfuwen:
#1279 
Sound like 1279 cover similar cases

prudmirya:
@chunfuwen #1296 is about adding max devices under controllers it seems, my test case is just check max vphb supported by ppc guest

prudmirya:
@chunfuwen my test case is specific to power architecture feature, power supports max 32 vphbs in guest

prudmirya:
this test case will start with one controller  and append the new controller after every guest reboot

balamuruhans:
@prudmirya I agree with @chunfuwen, #1279 covers the scenario

PR #1289, opened 2017-11-22
Fix domiftune case

hsmj1412:
Fix variable referenced before assignment


hsmj1412:
@chunfuwen @kylazhang

PR #1292, opened 2017-11-24
Fix for porting eroor

prudmirya:
existing test case ported to avocado from auto test. but miss the port import statement. in auto test lv_list() expected an argument and lv_list_all() not. while porting lv_list_all()
code copied to lv_list() and called lv_list. so it is expecting argument and errored out

Signed-off-by: Prudhvi Miryala <mprudhvi@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/storage_discard.py
+++ b/libvirt/tests/src/storage_discard.py
@@ -6,7 +6,7 @@ import re
 import logging
 import time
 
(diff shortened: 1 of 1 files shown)


prudmirya:
[job_storage.txt](https://github.com/autotest/tp-libvirt/files/1501617/job_storage.txt)
job log added

PR #1293, opened 2017-11-27
Add test for sriov

yafu-1:
Add test for sriov:
1.Create vfs;
2.Check the node device info of the pf and vf;
3.Plug/unPlug vfs;
4.Do operatoin with guest having vfs;

Signed-off-by: yafu <yafu@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/sriov/sriov.cfg
+++ b/libvirt/tests/cfg/sriov/sriov.cfg
@@ -1,0 +1,116 @@
+- sriov:
+    type = sriov
+    take_regular_screendumps = "no"
+    start_vm = "yes"
+    variants:
+        - driver:
+            variants:
+                - ixgbe:
+                    driver = ixgbe
+                - igb:
+                    driver = igb
+                - be2net:
+                    driver = be2net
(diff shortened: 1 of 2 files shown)


balamuruhans:
@santwana would you like to review this PR ?

libvirt/tests/cfg/sriov/sriov.cfg
```diff
@@ -0,0 +1,127 @@
...
+                    only vf.model_virtio.managed_yes
+                    variants:
+                        - multiple_times:
+                            times = 500
+                        - max_vfs:
+                            max_vfs_attached = "yes"
+                - define_network:
+                    only vf_pool, macvtap_network
+                    define_network = "yes"
+        - basic_test:
```
chunfuwen: 1)basic test and scaleblity test should be same hierarchy ,otherwise it doesn't make sense
2)coldplug and hotplug also need align with variants:normal and error_test, and in the outer layer
3)variants:vf and vf_pool also can be outer layer, equals to variants:coldplug and hotplug

PR #1294, opened 2017-11-27
Update numa cases for non-memory node

hsmj1412:
There are some numa node without memory, so need to use the new function


hsmj1412:
Depend on https://github.com/avocado-framework/avocado-vt/pull/1194/files

PR #1304, opened 2017-12-04
Modify two v2v cases from specific kvm

weikunzz:
1.Prepare a bogus kernel image for testing
2.Update multi kernel case

Signed-off-by: weikunzz <1525690718@qq.com>

### DIFF
--- a/v2v/tests/cfg/specific_kvm.cfg
+++ b/v2v/tests/cfg/specific_kvm.cfg
@@ -55,13 +55,8 @@
                     main_vm = 'VM_WIN_V2V_EXAMPLE'
                     vms = ${main_vm}
         - multi_kernel:
-            kernel_fm_url = 'KERNEL_FIRMWARE_URL_V2V_EXAMPLE'
-            kernel_url = 'KERNEL_URL_V2V_EXAMPLE'
-            variants:
-                - default_kernel:
-                    checkpoint = 'multi_kernel'
-                - debug_kernel:
-                    checkpoint = 'debug_kernel'
+            checkpoint = 'multi_kernel'
+            main_vm = 'VM_MULTI_KERNEL_V2V_EXAMPLE'
         - virtio:
             variants:
                 - windows:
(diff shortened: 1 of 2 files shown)


weikunzz:
(1/2) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.multi_kernel.output_mode.rhev: PASS (885.99 s)
 (2/2) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.bogus_kernel.output_mode.rhev: PASS (826.88 s)

v2v/tests/src/specific_kvm.py
```diff
@@ -174,19 +134,16 @@ def check_vmlinuz_initramfs(v2v_result):
...
-        logging.debug('Check debug kernel: %s' % kernel_debug)
         current_kernel = vmcheck.session.cmd('uname -r').strip()
         logging.debug('Current kernel: %s' % current_kernel)
-        logging.debug('Default kernel: %s' % default_kernel)
-        if kernel_debug:
-            if '.debug' in current_kernel:
-                log_fail('VM should choose non-debug kernel')
-        elif current_kernel not in default_kernel:
-            log_fail('VM should choose default kernel')
+        if current_kernel == '3.10.0-799.el7.x86_64':
```
chunfuwen: hardcode kernel version is not a good way since you need update script frequently if new kernel version is coming

v2v/tests/cfg/specific_kvm.cfg
```diff
@@ -55,13 +55,8 @@
                     main_vm = 'VM_WIN_V2V_EXAMPLE'
                     vms = ${main_vm}
         - multi_kernel:
-            kernel_fm_url = 'KERNEL_FIRMWARE_URL_V2V_EXAMPLE'
-            kernel_url = 'KERNEL_URL_V2V_EXAMPLE'
-            variants:
-                - default_kernel:
-                    checkpoint = 'multi_kernel'
-                - debug_kernel:
-                    checkpoint = 'debug_kernel'
```
chunfuwen: is there any reason to remove variant:debug_kernel ?

v2v/tests/src/specific_kvm.py
weikunzz: First of all, we test v2v how to choose kernel from multi kernels. I have prepared a guest image which has multi kernels and the latest kernel version is "3.10.0-799.el7.x86_64".

v2v/tests/cfg/specific_kvm.cfg
weikunzz: The case has updated, and the expect result is v2v will choose the latest kernel rather than default_kernel or debug_kernel.
I have prepare a guest image which has multi kernel and default kernel is debug-kernel.
So,the two variant could be merged into one

PR #1315, opened 2017-12-07
virsh_capabilities: Fix grep string

dzhengfy:
This is to fix the missing of ''


balamuruhans:
Let's merge it.

PR #1320, opened 2017-12-08
qemu-agent-command: Fix two issues

hahakiki2010:
1) Fix "VM is dead" issue for error_test.no_ga, the root
cause is VM is not started. Fixup by start vm withou
qemu-guest-agent
2) Fix Login timeout Error when with error_test.vm_halt
and error_test.vm_paused,  it is normal that vm can't be
login when vm is paused or halt. Fix by add judgement
before wait_for_login.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_qemu_agent_command.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_qemu_agent_command.py
@@ -22,11 +22,13 @@ def reset_domain(vm, vm_state, needs_agent=False, guest_cpu_busy=False,
     """
     if vm.is_alive():
         vm.destroy()
+    if not needs_agent:
+        vm.prepare_guest_agent(start=False)
     if needs_agent:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_qemu_agent_command.py
```diff
@@ -22,11 +22,13 @@ def reset_domain(vm, vm_state, needs_agent=False, guest_cpu_busy=False,
     """
     if vm.is_alive():
         vm.destroy()
+    if not needs_agent:
```
chunfuwen: a little confused here :why we need prepare guest agent if not needs_agent is True

hahakiki2010: prepare_guest_agent(start=False) will help to install qemu guest agent and then stop it on guest. It is to trigger "guest agent is not responding" error when "guest-ping" is executed

PR #1335, opened 2017-12-18
fix readonly disk snapshot issue

chunfuwen:
its expect result is to external snapshot failures


chunfuwen:
@balamuruhans 
@kylazhang 
please take a look

sathnaga:
@MalleshKoti ^^

PR #1340, opened 2017-12-20
Change the serial file path to fix permission denied issue

prudmirya:
Existing path '/var/log/libvirt' is doesnt have the required permissions.
the context is not of the qemu_img_t.
so i modified it to /var/lib/libvirt/images and i am not
distrubing the original


prudmirya:
[path_patch.txt](https://github.com/autotest/tp-libvirt/files/1575005/path_patch.txt)

PR #1347, opened 2017-12-28
Update remote_access case

hsmj1412:
From libvirt 3.2.0, default mechanism changed from
digest-md5 to gssapi


chunfuwen:
@hsmj1412 ,please add comment mentioning that it need set up kerbose server and client first before running cases

hsmj1412:
@chunfuwen Done

PR #1354, opened 2018-01-04
Fix reboot case imperfect

hsmj1412:
When framwork clear the env, the reboot operation
might not complete.


libvirt/tests/src/virsh_cmd/domain/virsh_reboot.py
```diff
@@ -95,6 +95,8 @@ def run(test, params, env):
                 if not virsh.has_command_help_match('reboot', '\s+--mode\s+'):
                     # old libvirt doesn't support reboot
                     status = -2
+            else:
+                vm.wait_for_login().close()
```
chunfuwen: please add comments why you need this .

hsmj1412: Done

PR #1356, opened 2018-01-08
Update for new serial dev format

hsmj1412:

### DIFF
--- a/libvirt/tests/src/serial/serial_functional.py
+++ b/libvirt/tests/src/serial/serial_functional.py
@@ -1,5 +1,4 @@
 import os
-import platform
 import logging
 import time
 import errno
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/serial/serial_functional.py
```diff
@@ -537,7 +538,12 @@ def check_qemu_cmdline():
         exp_ser_opt = ','.join(exp_ser_opts)
 
         if "ppc" in platform.machine():
-            exp_ser_devs = ['spapr-vty', 'chardev=charserial0', 'reg=0x30000000']
+            if libvirt_version.version_compare(3, 9, 0):
+                exp_ser_devs = ['spapr-vty', 'chardev=charserial0',
```
chunfuwen: suggest call exp_ser_devs.insert(2,'id=serial0') to update existed list, rather than duplicate the whole list when do assignment.

hsmj1412: Done

PR #1364, opened 2018-01-09
blockcopy:Ignore non-disk disks

chloerh:
In case making invalid snapshot that has not 'source' attrib

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -367,6 +367,10 @@ def run(test, params, env):
         # Add all disks into xml file.
         vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
         disks = vmxml.devices.by_device_tag('disk')
+        # Remove non-storage disk such as 'cdrom'
+        for disk in disks:
+            if disk.device != 'disk':
+                disks.remove(disk)
         new_disks = []
         src_disk_xml = disks[0]
         disk_xml = snap_xml.SnapDiskXML()


libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -367,6 +367,10 @@ def _make_snapshot():
         # Add all disks into xml file.
         vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
         disks = vmxml.devices.by_device_tag('disk')
+        # Remove non-storage disk such as 'cdrom'
+        for dis in disks:
```
balamuruhans: let us have the name complete
`for disk in disks:`

libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -367,6 +367,10 @@ def _make_snapshot():
         # Add all disks into xml file.
         vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
         disks = vmxml.devices.by_device_tag('disk')
+        # Remove non-storage disk such as 'cdrom'
```
chunfuwen: Beside cdrom, how about floppy? Do we need remove it too?

chloerh: Will remove all disks that 'device' is not 'disk', no matter it's 'cdrom' or 'floppy'

chunfuwen:
@chloerh ,please reply balamuruhans's comment about dis in disks:

libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
chloerh: Updated, thanks for the review

balamuruhans:
Thanks @chloerh @chunfuwen

PR #1370, opened 2018-01-11
virsh_domcapabilities: Change the code order for remote test

cliping:
Local test don't need set remote_ip. So change the order for
remote test.

PR #1369, opened 2018-01-11
virsh_cpu_models: Change the order for remote test

cliping:
Local host test don't need set remote_ip and remote_pwd. So change the code order for remote test.

PR #1371, opened 2018-01-11
Change test.error to test.cancel when remote ip not provided in 2 cases

lento-sun:
Existing virsh_cmd cases always skipped if env not ready such as
remote host ip not provided. But these 2 cases will be failed if
ip not provided. Change them to make things consistent and avoid
regression case failure.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_cpu_models.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_cpu_models.py
@@ -23,7 +23,7 @@ def run(test, params, env):
                                                               "default"))
 
     if 'EXAMPLE.COM' in remote_ip:
-        test.error("Please replace '%s' with valid remote_ip" % remote_ip)
+        test.cancel("Please replace '%s' with valid remote_ip" % remote_ip)
 
     if remote_ref == "remote":
         ssh_key.setup_ssh_key(remote_ip, "root", remote_pwd)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_domcapabilities.py
```diff
@@ -17,7 +17,7 @@ def run(test, params, env):
                                                               "default"))
 
     if 'EXAMPLE.COM' in remote_ip:
-        test.error("Please replace '%s' with valid remote_ip" % remote_ip)
+        test.cancel("Please replace '%s' with valid remote_ip" % remote_ip)
```
balamuruhans: This is taken care in https://github.com/autotest/tp-libvirt/pull/1370, also with few more betterment. so this change can be removed

libvirt/tests/src/virsh_cmd/host/virsh_cpu_models.py
```diff
@@ -23,7 +23,7 @@ def run(test, params, env):
                                                               "default"))
 
     if 'EXAMPLE.COM' in remote_ip:
-        test.error("Please replace '%s' with valid remote_ip" % remote_ip)
+        test.cancel("Please replace '%s' with valid remote_ip" % remote_ip)
```
balamuruhans: this is in https://github.com/autotest/tp-libvirt/pull/1369

lento-sun: oops, it appears in our test env so I thought it was merged. Thanks for the notice, close this one.

PR #1374, opened 2018-01-12
Fix - E722 do not use bare except in libvirt_mem.py

nasastry:
Fix E722 do not use bare except
in libvirt_mem.py

Signed-off-by: Nageswara R Sastry <rnsastry@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -257,7 +257,7 @@ def run(test, params, env):
             try:
                 del vmxml.numa_memory
                 del vmxml.numa_memnode
-            except:
+            except Exception:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_mem.py
```diff
@@ -257,7 +257,7 @@ def modify_domain_xml():
             try:
                 del vmxml.numa_memory
                 del vmxml.numa_memnode
-            except:
+            except Exception as e:
```
balamuruhans: `except Exception:` should be suffice as we dont use e

nasastry:
@balamuruhans A new version of the patch sent. Please have a look. Thanks for review.

PR #1382, opened 2018-01-18
Add --nvram option for undefine

waynesun09:
For UEFI boot VM, the --nvram option required to undefine vm.

For bios VM, use the option bring no harm.

Update case undefine and blockcopy first as they are in
acceptance test scope.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -470,7 +470,7 @@ def run(test, params, env):
 
         # Prepare transient/persistent vm
         if persistent_vm == "no" and vm.is_persistent():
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -470,7 +470,7 @@ def _make_snapshot():
 
         # Prepare transient/persistent vm
         if persistent_vm == "no" and vm.is_persistent():
-            vm.undefine()
+            vm.undefine("--nvram")
```
balamuruhans: Looks good to me, but the better way,
1. to have a check for UEFI/bios boot and perform this
2. comment explaining about introducing --nvram option to undefine helps for better readability.

PR #1387, opened 2018-01-22
Fix the serial log path

hsmj1412:
From 7.4 release, the virtlogd is used and only could write /var/log


libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
```diff
@@ -387,7 +387,7 @@ def make_filepath(self, index):
         """Return full path to unique filename per device index"""
         # auto-cleaned at end of test
         if self.type_name == 'file':
-            serial_dir = self.test_params.serial_dir
+            serial_dir = '/var/log/libvirt'
```
balamuruhans: I believe this would change with libvirt version rather than RHEL release, does this fix is backward compatible?

hsmj1412: @balamuruhans To prevent any question, I have added the version_compare

PR #1389, opened 2018-01-24
v2v matrix matching problems in v2v cfg

weikunzz:
v2v i386 x86_64 variants are  easy to repeat.
In latest auto job testing will cause the some issue.

Signed-off-by: weikunzz <kuwei@redhat.com>


chloerh:
@chunfuwen Would you please take a look at this? It will fix the problem that some v2v cases being matched multiple times when running a job.

balamuruhans:
Thanks @weikunzz @chloerh, let's merge this.

PR #1397, opened 2018-02-01
Remove non-disk disks before test

chloerh:
The whole procedure is to count disks in vm first, generate target
for the to be attached disk, attach disk, then check the status of
the attached disk.

If there's a cdrom disk in original vm, let's say its target is sda,
The attached disk will be 'sdb'. The checking step will try to find
certain string in vm device '/dev/sdb' which does not exist, since
the disk supposed to be 'sda' is acctually 'cdrom' in /dev.

Therefore remove any non-disk disks before test starts.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
@@ -808,6 +808,18 @@ def run(test, params, env):
     5) Restore domain
     6) Handle results
     """
+    # Remove non-disk disks before test to avoid influence.
+    # None-disk disks such as cdrom with bus 'SATA' will be recognized
+    # as 'cdrom', not 'sda' as expected, therefore the attached SATA
+    # disk will not be recognized as 'sdb' as expected.
+    vm_name = params.get('main_vm')
+    backup_vm_xml = vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
+    disks = vmxml.devices.by_device_tag('disk')
+    for disk in disks:
+        if disk.device != 'disk':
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
```diff
@@ -808,6 +808,15 @@ def run(test, params, env):
     5) Restore domain
     6) Handle results
     """
+    # Remove non-disk disks before test to avoid influence
```
chunfuwen: please state clearly what kind of influence you want to avoid by those steps

libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
```diff
@@ -808,6 +808,15 @@ def run(test, params, env):
...
+    vm_name = params.get('main_vm')
+    bkxml = vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
```
chunfuwen: bkxml to backup_vm_xml

balamuruhans:
Looks good to me, let us merge once @chunfuwen's comments are addressed.

libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
chloerh: It's already stated in commit message

chunfuwen: Please move some key commit message into comments

chloerh: Updated

PR #1402, opened 2018-02-08
virsh_domstate: configure less guest memory for coredump testcase

balamuruhans:
guest with huge memory takes more time for dumping core which timeout
the testcase and cause error, instead configuring less memory would
ensure test to work properly

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
@@ -108,9 +108,23 @@ def run(test, params, env):
 
     dump_path = os.path.join(test.tmpdir, "dump/")
     logging.debug("dump_path: %s", dump_path)
-    os.mkdir(dump_path)
+    try:
+        os.mkdir(dump_path)
+    except OSError:
+        # If the path already exists then pass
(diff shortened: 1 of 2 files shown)


balamuruhans:
This PR depends on https://github.com/avocado-framework/avocado-vt/pull/1340

libvirt/tests/src/virsh_cmd/monitor/virsh_domstate.py
```diff
@@ -102,6 +102,16 @@ def run(test, params, env):
     vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     backup_xml = vmxml.copy()
 
+    # Let's have guest memory less so that dumping core takes
```
chunfuwen: suggest move those implementation to line 124 to make it more understandable

balamuruhans:
@chunfuwen made the changes as suggested.

balamuruhans:
Fixed another issue observed with new commit, the guest's core file generated on crash dumped with file name having guest name, but if guest name have more than 20 characters, it clips it causing the tests to error out.

balamuruhans:
Thanks @chunfuwen!

PR #1211, opened 2017-09-07
expand virsh_vol_create to cover vol-create-as

carlsplace:
expand the test case to cover vol-create-as. Since vol-create and vol-create-as have similar functions, I only have to add some variants to cfg file, and change the py code according to the cfg file.

the original test cases and newly added test cases PASS on my local machine, the origin test results are the same with Jenkins result, so this change doesn't break old tests

Signed-off-by: yachang <yachang@redhat.com>

PR #1403, opened 2018-02-09
vol_create: Skip bug 106028

hahakiki2010:
This PR is to fix PR  #1211 for tp-libvirt,
where the exception shall be skipped instead
of test.fail


hsmj1412:
LGTM

PR #1404, opened 2018-02-11
libvirt_vcpu_plug_unplug: Fix conditions to update topology attributes

czhang0:
If topology is configured in guest xml, or new topology is needed to
    added, correct topology attributes to right value.
    
    For example:
    In some cases, we change vcpu max number, but not change topology
    manually, topology should be automatically updated if topology
    exist in original domain xml.
    
### DIFF
--- a/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
+++ b/libvirt/tests/src/libvirt_vcpu_plug_unplug.py
@@ -285,15 +285,21 @@ def run(test, params, env):
         vmxml.sync()
 
         topology = vmxml.get_cpu_topology()
-        if all([topology, sockets, cores, threads]):
-            # Check if the topology matches correct otherwise
-            if (sockets * cores * threads != vcpu_max_num):
-                if topology_correction:
+        # If topology is configured in domain xml or new topology is
+        # need to be newly set, check if the topology matches
+        # correct, if not, change topology to match expected values
+        # For example:
+        # In some case, we change vcpu max number, but not change topology
+        # manually, topology should be automatically updated if topology
+        # exist in original domain xml.
+        if topology or all([sockets, cores, threads]):
+            if topology_correction:
+                if (sockets * cores * threads != vcpu_max_num):
                     cores = vcpu_max_num
                     sockets = 1
                     threads = 1
-            vmxml.set_vm_vcpus(vm_name, vcpu_max_num, vcpu_current_num,
-                               sockets, cores, threads)
+                vmxml.set_vm_vcpus(vm_name, vcpu_max_num, vcpu_current_num,
+                                   sockets, cores, threads, add_topology=True)
         else:
             vmxml.set_vm_vcpus(vm_name, vcpu_max_num, vcpu_current_num)
         # Do not apply S3/S4 on power


libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,22 @@ def check_setvcpus_result(cmd_result, expect_error):
         vmxml.sync()
 
         topology = vmxml.get_cpu_topology()
-        if all([topology, sockets, cores, threads]):
-            # Check if the topology matches correct otherwise
-            if (sockets * cores * threads != vcpu_max_num):
-                if topology_correction:
+        if topology or all([sockets, cores, threads]):
+            # If topology is configured in domain xml or new topology is
```
chunfuwen: 1)Move comments ahead your module:before line288
2)If there are multiple lines comments, suggest use """ """

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,22 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+            # needed to newly setted, check if the topology matches
```
chunfuwen: need to be newly set

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,22 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+            # correct, otherwise change topology to match expeted values
```
chunfuwen: expeted to expected

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: 1) accept.
2) """ in the middle of function seems strange which is not docstring but used """ enclosed.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Accept. Thanks!

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Accept. Thanks!

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,21 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+        # exist in original domain xml.
+        if topology or all([sockets, cores, threads]):
+            if topology_correction:
+                if (sockets * cores * threads != vcpu_max_num):
                     cores = vcpu_max_num
                     sockets = 1
                     threads = 1
-            vmxml.set_vm_vcpus(vm_name, vcpu_max_num, vcpu_current_num,
-                               sockets, cores, threads)
+                vmxml.set_vm_vcpus(vm_name, vcpu_max_num, vcpu_current_num,
```
sathnaga: `add_topology=True` needs to be added here...

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,21 @@ def check_setvcpus_result(cmd_result, expect_error):
...
-            # Check if the topology matches correct otherwise
-            if (sockets * cores * threads != vcpu_max_num):
-                if topology_correction:
+        # If topology is configured in domain xml or new topology is
+        # need to be newly set, check if the topology matches
+        # correct, if not, change topology to match expected values
+        # For example:
+        # In some case, we change vcpu max number, but not change topology
+        # manually, topology should be automatically updated if topology
+        # exist in original domain xml.
```
balamuruhans: we can remove the example part as it is well explained in your commit msg.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Adopt, thanks!

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Keeping those lines of comments in code will help to get meanings easily. 
In general, we read code directly but  not go through each commits message to get code meanings.
So, I think it is better to keep them here. Thanks.

dzhengfy:
LGTM.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -285,15 +285,21 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+        if topology or all([sockets, cores, threads]):
+            if topology_correction:
+                if (sockets * cores * threads != vcpu_max_num):
```
hahakiki2010: Some puzzling questions about this condition judgement. 
1) we shall get value from existing topology when parameters are not configured. According current code, it seemed taking values from cfg file always. 
```
 if (sockets * cores * threads != vcpu_max_num):
```
2) ``` all([sockets, cores, threads]) ```  always return True, because their default value is 0, not None. 
3) This set of logic is similar with set_vm_vcpus in avocado-vt. Could we consider to combine them by adding a parameter topology_correction=False?

czhang0: When sockets, cores and threads are not configured:
1. sockets, cores and threads  will get default value 0. all([sockets, cores, threads]) is False.
2. if topology exist in original guest xml,  topology should be updated according to vcpu_max_num. 
The code seems work fine.

czhang0: topology_correction is only used by this tests, so handling it in avocado-vt seems not suitable.

sathnaga: @hahakiki2010 @czhang0 Please checkout https://github.com/avocado-framework/avocado-vt/pull/1488 and https://github.com/autotest/tp-libvirt/pull/1496, I am trying to address it through api itself.

czhang0: I had a look at avocado-framework/avocado-vt#1488 and #1496, seems good to correct topology issue and the code is more simple.
I'd like to close this PR.

PR #1405, opened 2018-02-11
libvirt_vcpu_plug_unplug: entend timeout and try several times for wi…

czhang0:
…th_maxvcpus

For with_maxvcpu case, hotplug/hotunplug 240 vcpus takes much more
time, so extend timeout time to ensure operation can be finished
before timeout. Timeout not means waiting time, instead of maximum
waiting time. If operation is finished in advance, terminate and
exit directly not waiting to timeout termination.

When unplug vcpus from 240 to 1, unplug sometimes exit with error.
Try several times, if it fail all, test is regarded as failure.


libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -438,11 +438,15 @@ def check_setvcpus_result(cmd_result, expect_error):
...
                     # expect_vcpupin to empty
                     expect_vcpupin = {}
-
-                result = virsh.setvcpus(vm_name, vcpu_unplug_num,
-                                        setvcpu_option,
-                                        readonly=setvcpu_readonly,
-                                        ignore_status=True, debug=True)
+                # Unplug vcpus from max vcpu number to 1 does not always
+                # succeed, so try 5 times, if failed all, case run fail.
+                for _ in range(5):
```
balamuruhans: please make this range as configurable

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -347,7 +347,7 @@ def check_setvcpus_result(cmd_result, expect_error):
                     expect_vcpu_num['guest_live'] = vcpu_plug_num
                     if not status_error:
                         if not utils_misc.wait_for(lambda: utils_misc.check_if_vm_vcpu_match(vcpu_plug_num, vm),
-                                                   120, text="wait for vcpu online") or not online_new_vcpu(vm, vcpu_plug_num):
+                                                   480, text="wait for vcpu online") or not online_new_vcpu(vm, vcpu_plug_num):
```
balamuruhans: make even this as configurable param

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Thanks for your comments, your suggestion is more reasonable. Adopt it.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: Thanks for your comments, your suggestion is more reasonable. Adopt it.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -438,11 +440,15 @@ def check_setvcpus_result(cmd_result, expect_error):
                     # As the vcpu will unplug later, so set
                     # expect_vcpupin to empty
                     expect_vcpupin = {}
-
-                result = virsh.setvcpus(vm_name, vcpu_unplug_num,
-                                        setvcpu_option,
-                                        readonly=setvcpu_readonly,
-                                        ignore_status=True, debug=True)
+                # Unplug vcpus from max vcpu number to 1 does not always
+                # succeed, so try 5 times, if failed all, case run fail.
```
balamuruhans: does this behaviour is expected ? Ideally it should be a bug if unplug couldn't succeed in single attempt right ?

dzhengfy: Similar suggestion: could you help explain in what scenarios the failure of vcpu unplug could be accepted?

hahakiki2010: Similar comments: Maybe need address the accepted failures in stderr.

czhang0: For case of unpluging from 240 to 1, virsh.setvcpus() doesn't always return 0,  sometimes it return timeout. When timeout occurs, if kernel continue to unplug guest vcpu until to 1 in limited seconds, it is expected action. If not, case is regarded to be run failed.
I will modify code to follow this logic.

balamuruhans: Sure, I too face with this situation in migration when I perform hotunplug, I check for lscpu by polling

https://github.com/autotest/tp-libvirt/blob/c6cb7b5c1898d50bdde660c305cec2b37e198677/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py#L401-L414

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -438,11 +440,40 @@ def check_setvcpus_result(cmd_result, expect_error):
...
                                         setvcpu_option,
                                         readonly=setvcpu_readonly,
                                         ignore_status=True, debug=True)
+                if result.exit_status != 0:
+                    try:
+                        session = vm.wait_for_login()
+                        cmd = "lscpu | grep \"^CPU(s):\""
+                        operation = "setvcpus"
+                        prev_output = -1
+                        while True:
```
dzhengfy: Is it better to add a timeout checking? I am wondering if the vcpu hotplug can not reach the target number, your loop will be infinite.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -438,11 +440,40 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+                            ret, output = session.cmd_status_output(cmd)
+                            if ret:
+                                test.fail("CPU %s failed - %s" % (operation, output))
```
kylazhang: Here should be test error as it is not libvirt command fail and the failure log should be change to get the real reason

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
balamuruhans: it wont, because if the lscpu returns same value twice the loop will break.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -439,10 +441,40 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+                # may not mean it is complete, a poll checking of guest vcpu numbers
+                # need to be executed.
+                # So for case of unpluging vcpus from max vcpu number to 1, when
+                # setvcpus return timeout, need continue to obverse if vcpu number is
+                # continually to be unplugged to 1 gradually.
                 result = virsh.setvcpus(vm_name, vcpu_unplug_num,
                                         setvcpu_option,
                                         readonly=setvcpu_readonly,
                                         ignore_status=True, debug=True)
+                if result.exit_status != 0:
```
balamuruhans: In my case, exit_status would return 0 but vcpu takes time to reflect from inside guest, but here you would get error from exit_status ?

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
dzhengfy: Sorry, it is my misunderstanding.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
czhang0: I just retry this case and can't re-happen this timeout on my local environment.
So I'd like to modify logic to check lscpu directly regardless the return of setvcpus.
The condition of lscpu returning same value twice between 5 seconds is regarded as setvcpus complete.

libvirt/tests/src/libvirt_vcpu_plug_unplug.py
```diff
@@ -439,10 +441,39 @@ def check_setvcpus_result(cmd_result, expect_error):
...
+                # need to be executed.
+                # So for case of unpluging vcpus from max vcpu number to 1, when
+                # setvcpus return, need continue to obverse if vcpu number is
+                # continually to be unplugged to 1 gradually.
                 result = virsh.setvcpus(vm_name, vcpu_unplug_num,
                                         setvcpu_option,
                                         readonly=setvcpu_readonly,
                                         ignore_status=True, debug=True)
+                try:
+                    session = vm.wait_for_login()
```
sathnaga: It is better to keep this out of this file and have the logic implemented in the avocado-vt/virttest/utils_hotplug.py
as we do all the expected values check through `check_vcpu_value()` and the same can be used across the tests, so better to be placed in `utils_hotplug.py`.

czhang0: I noticed this issue about at same time with you, I just sent out patch again.

PR #1406, opened 2018-02-12
virtual_disks_multivms: Fix the error:Fail to start domain vm2 

ZongqingQi:
At the time of testing, the same virtual hard disk is loaded
because of the two VM, which causes vm2 to not start.
Add a piece of code to the virtual_disks_multivms.py
file so that the XML of the second virtual disks added
during the test has a shareable attribute,
which solves the problem of the vm2 start failure

Signed-off-by: zonqi <zonqi@redhat.com>

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
@@ -14,6 +14,8 @@ from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml.devices.disk import Disk
 from virttest.libvirt_xml.devices.controller import Controller
 
+from provider import libvirt_version
+
 
 def run(test, params, env):
     """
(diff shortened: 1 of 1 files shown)


balamuruhans:
@ZongqingQi please fix the travis failure,

Have commit msg,
virtual_disks_multivms.py: <Subject line of fix>
<Explaining the code changes>

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
```diff
@@ -179,6 +181,11 @@ def get_vm_disk_xml(dev_type, dev_name, **options):
             if len(vms_sgio) > i:
                 disk_sgio = vms_sgio[i]
             shareable = ""
+
+            # Since lock feature is introduced in libvirt 3.9.0 afterwards, disk shareable attribute
+            # need be set if both of VMs need be started successfully in case they
```
balamuruhans: the comment is not completed, do we miss some words.

PR #1414, opened 2018-02-20
virsh_domblkerror: check to install nfs packages

balamuruhans:
unspecified_error testcase uses nfs setup to simulate the error,
check for nfs package available before starting the test.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -9,6 +9,7 @@ from avocado.utils import distro
 from virttest import virsh
 from virttest import data_dir
 from virttest import utils_misc
+from virttest import utils_package
 from virttest.utils_test import libvirt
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -39,11 +40,18 @@ def run(test, params, env):
...
     nfs_service = None
+    session = None
+    selinux_bak = ""
 
     vm = env.get_vm(vm_name)
-    # backup /etc/exports
-    shutil.copyfile(export_file, "%s.bak" % export_file)
-    selinux_bak = ""
+    if error_type == "unspecified error":
+        session = vm.wait_for_login()
```
chunfuwen: It sound like session need to be closed gracefully

balamuruhans: yes, you are right! I will make the necessary change

PR #1422, opened 2018-02-27
Fixup: Handle cgroup path for different distros

sathnaga:
Looks like libvirt cgroup path is different for
ubuntu distro, so handle multiple paths to resolve
libvirt cgroup folders.

### DIFF
--- a/libvirt/tests/src/cpu/guestpin.py
+++ b/libvirt/tests/src/cpu/guestpin.py
@@ -109,13 +109,21 @@ def run(test, params, env):
                 # Work around due to known cgroup issue after cpu hot(un)plug
                 # sequence
                 root_cpuset_path = utils_cgroup.get_cgroup_mountpoint("cpuset")
-                machine_cpuset_path = os.path.join(root_cpuset_path,
-                                                   "machine.slice")
-                if os.path.isdir(machine_cpuset_path):
-                    root_cpuset_cpus = os.path.join(root_cpuset_path,
-                                                    "cpuset.cpus")
-                    machine_cpuset_cpus = os.path.join(machine_cpuset_path,
-                                                       "cpuset.cpus")
+                machine_cpuset_paths = []
+                if os.path.isdir(os.path.join(root_cpuset_path,
+                                 "machine.slice")):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/cpu/guestpin.py
```diff
@@ -109,24 +109,28 @@ def set_condition(vm_name, condn, reset=False, guestbt=None):
...
-                    machine_cpuset_cpus = os.path.join(machine_cpuset_path,
-                                                       "cpuset.cpus")
+                machine_cpuset_paths = []
+                if os.path.isdir(os.path.join(root_cpuset_path,
+                                 "machine.slice")):
+                    machine_cpuset_paths.append(os.path.join(root_cpuset_path,
+                                                             "machine.slice"))
+                if os.path.isdir(os.path.join(root_cpuset_path, "machine")):
+                    machine_cpuset_paths.append(os.path.join(root_cpuset_path,
+                                                             "machine"))
```
balamuruhans: instead of handling multiple path can we handle root_cpu_path w.r.t to distro

sathnaga: @balamuruhans I am not sure if it changes w.r.t distro solid, may be in future some configs can set it back to machine.slice, do not want to take a chance, no harm to set to all available paths in that level and we set only if it dfiffers.

libvirt/tests/src/cpu/guestpin.py
```diff
@@ -109,24 +109,28 @@ def set_condition(vm_name, condn, reset=False, guestbt=None):
...
+                if not machine_cpuset_paths:
+                    logging.warning("cgroup cpuset might not recover properly "
+                                    "for guests after host smt changes, "
+                                    "restore it manually")
+                root_cpuset_cpus = os.path.join(root_cpuset_path, "cpuset.cpus")
```
chunfuwen: it sound like the variable root_cpuset_cpus was never used

balamuruhans: it is used at L128, L131

PR #1425, opened 2018-03-02
Minorfix: The unspecified_error cases checking only support running vm

hsmj1412:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -45,6 +45,8 @@ def run(test, params, env):
 
     vm = env.get_vm(vm_name)
     if error_type == "unspecified error":
+        if not vm.is_alive():
+            vm.start()
         session = vm.wait_for_login()
         if not ubuntu:
             nfs_service_package = "nfs"


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -44,7 +44,7 @@ def run(test, params, env):
     selinux_bak = ""
 
     vm = env.get_vm(vm_name)
-    if error_type == "unspecified error":
+    if error_type == "unspecified error" and vm.state() == 'running':
```
balamuruhans: vm is not started before this check, can it be included as part of this fix ?

hsmj1412: @balamuruhans Sorry, I do not understand what you mean. The vm is not start before this line, but there is a vm.wait_for_login() followed by.

hsmj1412: Do you want to start the guest before this check?

balamuruhans: correct, that is my concern.
1. based on start_vm param set for the testcase, start the vm or destroy the vm this would make your check more appropriate.
2. By doing this virsh_domblkerror would work properly even if test environment is not clean or polluted by previous error'ed testcases and makes your fix more robust

hsmj1412: @balamuruhans According to the config:
- virsh.domblkerror:
    type = virsh_domblkerror
    start_vm = "no"
Now, the "start_vm" param is destroyed . Whatever, I have updated the case according to my understanding of what you say

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -45,6 +45,8 @@ def run(test, params, env):
 
     vm = env.get_vm(vm_name)
     if error_type == "unspecified error":
+        if not vm.is_alive():
```
chunfuwen: I think just rudely start vm here regardless checking whether test case itself need start it or not. Better we can run some cases to  validate the changes to avoid further regressions.

balamuruhans: https://github.com/autotest/tp-libvirt/blob/04ad3120d71f259f5913bf940161d27ec1bc1646/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py#L48-L54

@hsmj1412 issue is here, we have to do the nfs package check on host not on vm.
so fix would be just remove the session and vm.wait_for_login()

balamuruhans: @hsmj1412 I am sorry and this bug was introduced by my patch in commit 4f78942

hsmj1412: @balamuruhans I will close this pr because I trust you will fix it with a better way.

PR #1429, opened 2018-03-05
fix twice_diff_with_shareable error in virsh_attach_detach_disk

ZongqingQi:
Signed-off-by: zonqi <zonqi@redhat.com>


ZongqingQi:
fix twice_diff_with_shareable error in attach_detach_disk
    
    Modify the judgment condition,add one more level judgment condition,
    so that the case twice_diff executes the line that adds the
    shareable attribute to disks, and the make case
    twice_diff_with_shareable does not execute the line.
    So there is no error in the case twice_diff_with_shareable.

PR #1442, opened 2018-03-14
Python 3: Port others py files of acceptance case to PY3

czhang0:
Python 3: Port non virsh cmd py files to PY3
    1) Remove autotest dependency
    2) Port to PY3
    
### DIFF
--- a/libvirt/tests/src/libvirt_hugepage.py
+++ b/libvirt/tests/src/libvirt_hugepage.py
@@ -4,8 +4,7 @@ import time
 
 from aexpect import ShellError
 
-from autotest.client import utils
-from autotest.client.shared import error
+from avocado.utils import process
 
 from virttest import virsh
 from virttest import utils_libvirtd
@@ -20,7 +19,6 @@ from virttest.staging import utils_memory
 
 
 def prepare_c_file():
-    output = file("/tmp/test.c", 'w')
     outputlines = """
     #include<stdio.h>
     #include<malloc.h>
@@ -40,8 +38,8 @@ def prepare_c_file():
         return 0;
     }
     """
-    output.writelines(outputlines)
-    output.close()
+    with open("/tmp/test.c", 'w') as output:
+        output.writelines(outputlines)
     return "/tmp/test.c"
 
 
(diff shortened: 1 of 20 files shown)


libvirt/tests/src/libvirt_hugepage.py
```diff
@@ -40,8 +39,8 @@ def prepare_c_file():
         return 0;
     }
     """
-    output.writelines(outputlines)
-    output.close()
+    with open("/tmp/test.c", 'wb') as output:
```
chunfuwen: it sound like it not work on python3:
```
>>> outputlines ='hello'
>>> with open("/tmp/test.c", 'w+b') as output:
...     output.writelines(outputlines.encode('utf-8'))
... 
  File "<stdin>", line 2, in <module>
TypeError: 'int' does not support the buffer interface
```

libvirt/tests/src/libvirt_mem.py
```diff
@@ -375,7 +375,7 @@ def modify_domain_xml():
         if add_mem_device:
             at_times = int(params.get("attach_times", 1))
             dev_xml = create_mem_xml()
-            for x in xrange(at_times):
```
chunfuwen: use xrange from six module

libvirt/tests/src/svirt/dac_nfs_disk.py
```diff
@@ -23,7 +23,8 @@ def check_ownership(file_path):
     :return: ownership string user:group or false when file not exist
     """
     try:
-        f = os.open(file_path, 0)
+        with open(file_path, 'rb') as f:
+            pass
     except OSError:
```
chunfuwen: remove exception part

libvirt/tests/src/svirt/svirt_attach_disk.py
```diff
@@ -167,7 +167,7 @@ def run(test, params, env):
                 cap_dict = {}
                 pid = vm.get_pid()
                 pid_status_path = "/proc/%s/status" % pid
-                with open(pid_status_path) as f:
+                with open(pid_status_path, 'rb') as f:
```
chunfuwen: need do a test whether  under mode 'rb' , for line in f: can work normally

PR #1452, opened 2018-03-19
Python 3: Port another 6 directories

hahakiki2010:
1. Port Python3 for tp-libvirt 6 directories as:
lxc, npiv, numa, nwfilter, remote_access and svirt
2. Remove autotest dependencies

### DIFF
--- a/libvirt/tests/src/nwfilter/nwfilter_update_vm_running.py
+++ b/libvirt/tests/src/nwfilter/nwfilter_update_vm_running.py
@@ -1,8 +1,7 @@
 import re
 import logging
 
-from autotest.client.shared import utils
-from autotest.client.shared import error
+from avocado.utils import process
 
 from virttest import virsh
 from virttest import libvirt_xml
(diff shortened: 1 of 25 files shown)


libvirt/tests/src/nwfilter/nwfilter_update_vm_running.py
```diff
@@ -68,22 +67,22 @@ def run(test, params, env):
...
             check_cmd = check_cmd.replace("DEVNAME", iface_target)
         ret = utils_misc.wait_for(lambda: not
-                                  utils.system(check_cmd, ignore_status=True),
+                                  process.system(check_cmd, ignore_status=True, shell=True),
                                   timeout=30)
         if not ret:
-            raise error.TestFail("Rum command '%s' failed" % check_cmd)
-        out = utils.system_output(check_cmd, ignore_status=False)
+            test.fail("Rum command '%s' failed" % check_cmd)
+        out = process.system_output(check_cmd, ignore_status=False, shell=True)
```
chunfuwen: there are minor difference between utils.system_out and process.system_output. Utils will strip off string with tag '\n',however process.system_out() will strip with tag  b'\n\r'.
Please do more testing whether those minor changes can impact test result.

chunfuwen:
@hahakiki2010 ,please do more testing based on those changes.

libvirt/tests/src/nwfilter/nwfilter_update_vm_running.py
hahakiki2010: I checked system_output in avocado latest 52lts, seemingly like the code you listed is a bit out-dated, where  b'\n\r' has updated to '\n\r' already. So I guess there would be no problem then. 
```
def system_output(cmd, timeout=None, verbose=True, ignore_status=False,
                  allow_output_check='all', shell=False, env=None, sudo=False,
                  ignore_bg_processes=False, strip_trail_nl=True):
    """
    Run a subprocess, returning its output.
    :param cmd: Command line to run.
    :type cmd: str
    :param timeout: Time limit in seconds before attempting to kill the
                    running process. This function will take a few seconds
                    longer than 'timeout' to complete if it has to kill the
                    process.
    :type timeout: float
    :param verbose: Whether to log the command run and stdout/stderr.
    :type verbose: bool
    :param ignore_status: Whether to raise an exception when command returns
                          =! 0 (False), or not (True).
    :param allow_output_check: Whether to log the command stream outputs
                               (stdout and stderr) in the test stream
                               files. Valid values: 'stdout', for
                               allowing only standard output, 'stderr',
                               to allow only standard error, 'all',
                               to allow both standard output and error
                               (default), and 'none', to allow
                               none to be recorded.
    :type allow_output_check: str
    :param shell: Whether to run the command on a subshell
    :type shell: bool
    :param env: Use extra environment variables
    :type env: dict
    :param sudo: Whether the command requires admin privileges to run,
                 so that sudo will be prepended to the command.
                 The assumption here is that the user running the command
                 has a sudo configuration such that a password won't be
                 prompted. If that's not the case, the command will
                 straight out fail.
    :type sudo: bool
    :param ignore_bg_processes: Whether to ignore background processes
    :type ignore_bg_processes: bool
    :param strip_trail_nl: Whether to strip the trailing newline
    :type strip_trail_nl: bool
    :return: Command output.
    :rtype: str
    :raise: :class:`CmdError`, if ``ignore_status=False``.
    """
    cmd_result = run(cmd=cmd, timeout=timeout, verbose=verbose, ignore_status=ignore_status,
                     allow_output_check=allow_output_check, shell=shell, env=env,
                     sudo=sudo, ignore_bg_processes=ignore_bg_processes)
    if strip_trail_nl:
        return cmd_result.stdout.rstrip('\n\r')
return cmd_result.stdout
```

PR #1463, opened 2018-03-22
numa_memory_migrate: Fix the list error

hsmj1412:

### DIFF
--- a/libvirt/tests/src/numa/numa_memory_migrate.py
+++ b/libvirt/tests/src/numa/numa_memory_migrate.py
@@ -147,8 +147,12 @@ def run(test, params, env):
             # If there are inconsistent node numbers on host,
             # convert it into sequence number so that it can be used
             # in mem_compare
-            left_node_new = [node_list.index(i) for i in node_list if i != node]
-            used_node = [node_list.index(node)]
+            # memory_status is a total numa list. node_list could not
+            # match the count of nodes
+            total_online_node_list = host_numa_node.online_nodes
+            left_node_new = [total_online_node_list.index(i)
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/numa/numa_memory_migrate.py
```diff
@@ -52,6 +52,7 @@ def mem_compare(used_node, left_node):
     # Get host numa node list
     host_numa_node = utils_misc.NumaInfo()
     node_list = host_numa_node.online_nodes_withmem
+    total_online_node_list = host_numa_node.online_nodes
```
balamuruhans: It would be better to use online_nodes_withmem than online_nodes, how does this fix list error

hsmj1412: @balamuruhans we have used online_nodes_withmem in line 54, but in line 144, the get_numa_status will report all of the nodes status. e.g. there are 4 nodes, [0,1,4,5]. Only node 0 and node 4 could be used(with memory). So, online_nodes_withmem will return[0,4], but get_numa_status will return the memory list [100,0,100,0]

hsmj1412:
@chunfuwen Updated

libvirt/tests/src/numa/numa_memory_migrate.py
balamuruhans: @hsmj1412 Thanks for the explanation, fix looks good to me.

balamuruhans:
ACK

balamuruhans:
commit msg have a typo, %s/memroy/memory.

hsmj1412:
@balamuruhans Fixed

PR #1464, opened 2018-03-22
virsh_iface_trans.py: tests were cancelling because of missing depend…

MalleshKoti:
…ency "mlocate" package.

This patch will check for mlocate and install it.

Signed-off-by: Mallesh Koti <mallesh@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/interface/virsh_iface_trans.py
+++ b/libvirt/tests/src/virsh_cmd/interface/virsh_iface_trans.py
@@ -8,6 +8,7 @@ from avocado.core import exceptions
 
 from virttest import virsh
 from virttest import utils_libvirtd
+from virttest import utils_package
 
 
 def netcf_trans_control(test, command="status"):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/interface/virsh_iface_trans.py
```diff
@@ -190,11 +191,14 @@ def run(test, params, env):
...
-    try:
-        utils_path.find_command("locate")
-    except utils_path.CmdNotFoundError:
-        test.cancel("Command 'locate' is missing. You must "
-                    "install it.")
+    if not utils_package.package_install(["mlocate"]):
+        test.cancel("Failed to install dependency package mlocate"
+                    " on host")
+    cmd = "updatedb"
+    cmd_status = process.system_output(cmd, shell=True)
```
balamuruhans: `process.system(cmd, shell=True)` would return you the `cmd_status`

MalleshKoti: Have modified accordingly, thanks for the suggestion.

PR #1473, opened 2018-03-28
Python 3: Remove autotest dependency for lvsb py files

czhang0:


lvsb/tests/src/lvsb_complex_options.py
```diff
@@ -29,7 +27,7 @@ def run(test, params, env):
     # positive and negative testing #########
     if not status_error:
         if status != 0:
-            raise error.TestFail("%d not a expected command "
-                                 "return value", status)
+            test.fail("%d not a expected command "
+                      "return value", status)
```
balamuruhans:
```
test.fail("%d not a expected command "
          "return value" % status)
```

balamuruhans: please use modulus instead of comma elsewhere

lvsb/tests/src/lvsb_network_options.py
```diff
@@ -134,13 +132,13 @@ def run(test, params, env):
     # positive and negative testing #########
     if status_error:
         if status == 0:
-            raise error.TestFail("%d not a expected command "
-                                 "return value", status)
+            test.fail("%d not a expected command "
+                      "return value", status)
```
kylazhang: same as above

lvsb/tests/src/lvsb_network_options.py
```diff
@@ -134,13 +132,13 @@ def run(test, params, env):
...
         else:
             logging.info("It's an expected error: %s", _params['lvsb_result'])
     else:
         if status != 0:
-            raise error.TestFail("%d not a expected command "
-                                 "return value", status)
+            test.fail("%d not a expected command "
+                      "return value", status)
```
kylazhang: same as above

lvsb/tests/src/lvsb_complex_options.py
czhang0: Thanks for your careful. I find some other same issues in the code, and will modify them together.

lvsb/tests/src/lvsb_network_options.py
czhang0: Thanks. Yes, here need to use modulus.

lvsb/tests/src/lvsb_network_options.py
czhang0: Thanks. Yes, here need to use modulus.

PR #1445, opened 2018-03-14
Python 3: Port to Python3 for 10 pys

dzhengfy:
- Remote autotest reference
- Replace exceptions.Test* with test.*
- Other updates according to rules

Signed-off-by: Dan Zheng <dzheng@redhat.com>

PR #1478, opened 2018-03-29
Python3 : v2v cases

chloerh:

### DIFF
--- a/v2v/tests/src/convert_from_file.py
+++ b/v2v/tests/src/convert_from_file.py
@@ -19,7 +19,7 @@ def run(test, params, env):
     """
     convert specific kvm guest to rhev
     """
-    for v in params.itervalues():
+    for v in list(params.values()):
         if "V2V_EXAMPLE" in v:
             test.cancel("Please set real value for %s" % v)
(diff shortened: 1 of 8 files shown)


v2v/tests/src/convert_from_file.py
```diff
@@ -19,7 +19,7 @@ def run(test, params, env):
     """
     convert specific kvm guest to rhev
     """
-    for v in params.itervalues():
+    for v in list(params.values()):
```
chunfuwen: use six module, see below example:
https://github.com/autotest/tp-libvirt/pull/1445/files

chloerh: This way works fine for v2v cases, I don't see any reason to introduce an unnecessary module.

chunfuwen: This is our unified way to deal with this issue in whole tp-libvirt project

chloerh: In v2v test cases, this is simply a parameters existence check. There's no need to make the params iterable. Still, I don't see the necessity even with the unified way.

PR #1479, opened 2018-03-29
Add info of full xml for testing

chloerh:
All new added cases should at least run pass the 'full xml'


kylazhang:
I think we should not add version related information here, maybe we can change to fedora related image

iccaszhulili:
rh-caselink retest please

PR #1489, opened 2018-04-09
Iface_hotplug: Fixup the serial bug

hsmj1412:
Clean up the serial console and reuse it after that


libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -169,6 +169,7 @@ def create_iface_xml(mac):
             if restart_libvirtd:
                 libvirtd.restart()
                 vm.cleanup_serial_console()
+                vm.create_serial_console()
```
chunfuwen: This logic is to tear down resources,why we need create serial console rather thatn cleanup

hsmj1412: @chunfuwen Updated

PR #1491, opened 2018-04-09
Python 3: Simply replace stdout to stdout_text

hahakiki2010:
Replace stdout to stdout_text for following cases:
1. grep result.stdout -rn | grep -v "Binary file"
2. grep ret.stdout -rn | grep -v "Binary file"
3. grep res.stdout -rn | grep -v "Binary file"
4. grep stdout -rn | grep -v  stdout_text| grep -v "Binary file" | grep
"process.run"

Then use sed -i -e "$lineno s/stdout/stdout_text/" "$file_to_replace"
to replace in bundle


libvirt/tests/src/virsh_cmd/domain/virsh_dump.py
```diff
@@ -163,8 +163,13 @@ def check_dump_format(dump_image_format, dump_file):
             return True
         else:
             file_cmd = "file %s" % dump_file
+<<<<<<< HEAD
```
hsmj1412: Please fix the confict

hahakiki2010: fixed

libvirt/tests/src/libvirt_rng.py
```diff
@@ -144,7 +144,7 @@ def check_snapshot(bgjob=None):
             vm_name, ("%s %s" % (snapshot_name, options)))
         libvirt.check_exit_status(ret)
         ret = virsh.dumpxml(vm_name)
-        if ret.stdout.count("<rng model="):
+        if ret.stdout_text.count("<rng model="):
```
chunfuwen: For virsh stdout or stderr, I suggest virsh.py under avocado-vt address this, which is the unified interface, rather that individually addressed, see https://github.com/avocado-framework/avocado-vt/pull/1480

hahakiki2010: I removed change for virsh.cmd, and kept change for process.run only. You can have a check~

libvirt/tests/src/virsh_cmd/volume/virsh_vol_resize.py
```diff
@@ -145,7 +145,7 @@ def create_luks_vol(vol_name, sec_uuid, params):
         if (key.startswith('vol_') and not key.startswith('vol_new')):
             if key[4:] in ['capacity', 'allocation']:
                 vol_arg[key[4:]] = int(float(utils_misc.normalize_data_size(params[key],
-                                       "B", 1024)))
+                                                                            "B", 1024)))
```
chunfuwen: I don't see any reason to change it here

PR #1493, opened 2018-04-09
virsh_migrate: use virsh uri to handle guest cmd from destination machine

balamuruhans:
Earlier ssh was being used to send cmds to guest from remote machine after
migrating the guest, but going further using virsh uri makes it robust.


balamuruhans:
@hsmj1412 Thank you

PR #1507, opened 2018-04-12
Add more plug options for interface

yalzhang:
Add more plug optinos "--config" "--current" and the combination of
them for attach-device and attach-interface. And test the options both
when vm is running or inactive.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
@@ -2,7 +2,6 @@
     type = iface_hotplug
     take_regular_screendumps = "no"
     start_vm = "yes"
-    status_error = "no"
     poll_timeout = 10
     variants:
         - iface_attach:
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
```diff
@@ -27,12 +26,38 @@
...
+                        - inactive:
+                            start_vm = "no"
                     variants:
                         - default:
-                            status_error = "yes"
+                            attach_option = ""
                         - persistent:
                             attach_option = "--persistent"
+                        - current:
+                            attach_option = "--current"
```
chunfuwen: suggest  '--current',and '--permissment', those strings can be defined as variable

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -97,6 +100,10 @@ def create_iface_xml(mac):
             iface_list = []
             err_msgs = ("No more available PCI slots",
                         "No more available PCI addresses")
+            err_msgs1 = ("domain is not running")
```
chunfuwen: suggest define into cfg file

libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
yalzhang: I understand your solution. But I think it will be more complex. There are 5 options: 
1. null
2. persistent
3. current
4. live
5. config
after combination, come out 5x4x3x2  and we need to select the ones, that will be too much work.

libvirt/tests/src/virtual_network/iface_hotplug.py
yalzhang: updated

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -117,8 +124,19 @@ def create_iface_xml(mac):
                         elif (ret.stderr.count("doesn't support option %s"
                                                % attach_option)):
                             test.cancel(ret.stderr)
-                        elif status_error:
-                            continue
+                        elif err_msgs1 in ret.stderr:
```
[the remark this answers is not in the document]
yalzhang: I have refine the structure and reuse the code

yalzhang:
run all the case locally and pass

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -172,35 +203,107 @@ def create_iface_xml(mac):
                 # so we need create a new serial console
                 vm.cleanup_serial_console()
                 vm.create_serial_console()
+            # in options test, check if the interface is attached
+            # in current state when attach return true
+
+            def get_iface():
```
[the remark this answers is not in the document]
yalzhang: updated

PR #1512, opened 2018-04-16
Fix issue "Can't find any filter policy"

yalzhang:
virtual_network/iface_network.py: the unit of the filter policy can be
'Mb' in addition to 'b', so add the check.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_network.py
+++ b/libvirt/tests/src/virtual_network/iface_network.py
@@ -217,19 +217,23 @@ TIMEOUT 3"""
         logging.debug("Bandwidth filter output: %s", filter_output)
         if not filter_output.count("filter protocol all pref"):
             test.fail("Can't find 'protocol all' settings in filter rules")
-        filter_pattern = ".*police.*rate (\d+)Kbit burst (\d+)(K?M?)b.*"
+        filter_pattern = ".*police.*rate (\d+)(K?M?)bit burst (\d+)(K?M?)b.*"
         se = re.search(r"%s" % filter_pattern, filter_output, re.M)
         if not se:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -217,19 +217,23 @@ def check_filter_rules(ifname, bandwidth):
...
+        filter_pattern = ".*police.*rate (\d+)(K?M?)bit burst (\d+)(K?M?)b.*"
         se = re.search(r"%s" % filter_pattern, filter_output, re.M)
         if not se:
             test.fail("Can't find any filter policy")
         logging.debug("bandwidth from tc output:%s" % str(se.groups()))
         logging.debug("bandwidth from setting:%s" % str(bandwidth))
         if "average" in bandwidth:
-            assert int(se.group(1)) == int(bandwidth["average"]) * 8
+            if se.group(2) == 'M':
+                tc_average = int(se.group(1)) * 1000
```
chunfuwen: line 228 result is *1000,but line 234 is *.1024, is any reason different between them?

yalzhang: Good question, it makes me to search the reason why it is designed like this. It's something about tc. 
For average(line 228), it indicates the transfer rate. Its unit in xml is kilobytes/s(kbps), while transfer to tc(unit is Kbit / Mbit), for linux distributions, network bandwidth uses base-10.
For  burst(line 234), it indicates the amount of bytes that can be burst at peak speed. Its unit in xml is kilobytes, since memory modules and disks came in sizes that were powers of two, so for such size, it uses base-2

PR #1518, opened 2018-04-18
Fix 2 issues in the libvirt-guests cases:

iccaszhulili:
1) Now, there will be some selinux logs during restart libvirt-guests.service.
So the msg about guests parallel shutdown are not always stick togther.
Hence, modified the old checking method.
2) One checking point is missed, when on_shutdown = shutdown, on_boot=ignore,
the guest should be in shutdown status after restarting libvirt-guests.service.
So, add a checking point named check_on_shutdown_vm_status

### DIFF
--- a/libvirt/tests/src/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
+++ b/libvirt/tests/src/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
@@ -59,7 +59,7 @@ def run(test, params, env):
                     test.fail("Shutdown of guest should be failed to "
                               "complete in time")
 
-        if (on_shutdown == "shutdown") and len(vms) > 1:
+        if (on_shutdown == "shutdown") and int(parallel_shutdown) > 0:
             chk_parallel_shutdown(output, parallel_shutdown)
 
     def chk_on_boot(status_error, on_boot):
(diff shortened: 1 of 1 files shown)


chunfuwen:
please paste test result after change

libvirt/tests/src/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
```diff
@@ -85,37 +85,45 @@ def chk_on_boot(status_error, on_boot):
                                   "guest should be shut off after "
                                   "restarting libvirt-guests, ")
 
+    def check_on_shutdown_vm_status():
```
chunfuwen: rename to check_vm_shutdown

libvirt/tests/src/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
```diff
@@ -85,37 +85,45 @@ def chk_on_boot(status_error, on_boot):
...
+
+        pattern = r".+ libvirt-guests.sh: Shutdown of guest.+complete"
+        for line_num, line in enumerate(output.splitlines()):
+            if re.search(pattern, line):
+                shut_complete_first_line = line_num
+                break
+        logging.debug("the first line contains shutdown complete messages is: %s ",
+                      shut_complete_first_line)
+
+        para_shut = int(parallel_shutdown)
```
chunfuwen: 119 and 126 can be merged if shut_start_line_nums[para_shut-1]  !=  shut_complete_first_line:xx

PR #1520, opened 2018-04-19
Add param value for 'vcpu'

chloerh:
Set vcpu = 4


libvirt/tests/cfg/libvirt_mem.cfg
```diff
@@ -7,6 +7,7 @@
     # If align_to_value is not given then '65536' will be taken as default.
     align_mem_values = "no"
     known_unplug_errors = "'pseries-hotplug-mem: Memory indexed-count-remove failed, adding any removed LMBs'"
+    vcpu = 4
```
chunfuwen: I don't see any code are changed to get this parameter value accordingly.

chloerh: We could easily replace with new vcpu number if there is a value there. For ppc

PR #1528, opened 2018-04-25
Add test for only the first console can be a serial

jinqi7:

### DIFF
--- a/libvirt/tests/src/serial/serial_functional.py
+++ b/libvirt/tests/src/serial/serial_functional.py
@@ -299,6 +299,7 @@ def run(test, params, env):
         """
         console = librarian.get('console')(serial_type)
         console.target_type = console_target_type
+        console.target_port = console_target_port
 
         sources = []
         logging.debug(sources_str)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/serial/serial_functional.py
```diff
@@ -625,12 +628,24 @@ def get_console_type():
...
-        if console_target_type == 'serial':
+        if console_target_type == 'serial' or second_serial_console:
             logging.debug('Serial device:\n%s', serial_dev)
             vm_xml.add_device(serial_dev)
         console_dev = prepare_console_device()
         logging.debug('Console device:\n%s', console_dev)
         vm_xml.add_device(console_dev)
+        if second_serial_console:
+            console_target_type = 'serial'
+            console_target_port = '1'
```
chunfuwen: we may not hard code console_target_port to '1', probably calculate it from current console_target_port=params.get('console_target_port', '0')

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -625,12 +628,24 @@ def get_console_type():
...
+            console_dev = prepare_console_device()
+            logging.debug('Console device:\n%s', console_dev)
+            vm_xml.add_device(console_dev)
+            res = virsh.define(vm_xml.xml)
```
chunfuwen: we need do undefine firstly before do define ,I think.

libvirt/tests/src/serial/serial_functional.py
jinqi7: The hard code  "1"  is used since it's to test the second serial console test case.

libvirt/tests/src/serial/serial_functional.py
jinqi7: @done.

PR #1543, opened 2018-05-02
Python 3: Type error fix for base64

hahakiki2010:
This PR is to fix below issue when with base64 encode and decode:
TypeError: a bytes-like object is required, not 'str'

### DIFF
--- a/libvirt/tests/src/virsh_cmd/secret/virsh_secret_set_get.py
+++ b/libvirt/tests/src/virsh_cmd/secret/virsh_secret_set_get.py
@@ -2,6 +2,7 @@ import os
 import re
 import base64
 import logging
+import locale
 from tempfile import mktemp
 
 from avocado.utils import process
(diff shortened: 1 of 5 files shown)


libvirt/tests/src/virsh_cmd/secret/virsh_secret_set_get.py
```diff
@@ -27,8 +28,8 @@ def check_secret(params):
 
     if os.access(base64_file, os.R_OK):
         with open(base64_file, 'rb') as base64file:
-            base64_encoded_string = base64file.read().strip().decode('utf-8')
-        secret_decoded_string = base64.b64decode(base64_encoded_string)
+            base64_encoded_string = base64file.read().strip()
+        secret_decoded_string = base64.b64decode(base64_encoded_string).decode(locale.getpreferredencoding())
```
chunfuwen: cuzhang had ever updated this file ,I assume that restoring original one can fix this issue.
```
     if os.access(base64_file, os.R_OK):
-        base64_encoded_string = open(base64_file, 'r').read().strip()
+        with open(base64_file, 'rb') as base64file:
+            base64_encoded_string = base64file.read().strip().decode('utf-8')
         secret_decoded_string = base64.b64decode(base64_encoded_string)

```

PR #1566, opened 2018-05-09
controller.functional: Add case for PPC usb

dzhengfy:
The default usb controller on PPC is 'qemu-xhci'. This is to test this
feature.

### DIFF
--- a/libvirt/tests/cfg/controller/controller_functional.cfg
+++ b/libvirt/tests/cfg/controller/controller_functional.cfg
@@ -60,13 +60,14 @@
                     controller_type = virtio-serial
                     controller_vectors = '0'
                 - usb_controller:
-                    no machine_pseries
                     controller_type = usb
                     variants:
                         # ehci and ich9-ehci1 are USB2.0 controller model
                         - ehci_model:
+                            no machine_pseries
                             controller_model = ehci
                         - ich9_ehci1_model:
+                            no machine_pseries
                             controller_model = ich9-ehci1
                             companion_controller_model = ich9-uhci
                             companion_controller_num = 3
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/controller/controller_functional.cfg
```diff
@@ -60,23 +60,31 @@
...
                         # nec-xhci and qemu-xhci are USB3.0 controller model
                         - nec_xhci_model:
+                            no machine_pseries
                             controller_model = nec-xhci
                         - qemu_xhci_model:
-                            controller_model = qemu-xhci 
+                            variants:
+                                - non_pseries_machine:
+                                    no machine_pseries
+                                    controller_model = qemu-xhci
```
chunfuwen: duplicate initialize the value

libvirt/tests/src/controller/controller_functional.py
```diff
@@ -388,20 +388,24 @@ def check_qemu_cmdline(search_pattern=None):
...
-                    pattern = r"-device.usb-ehci"
-                elif item == "qemu-xhci":
-                    pattern = r"-device.qemu-xhci"
-                else:
-                    name = item.split('-')
-                    pattern = pattern + r"-device.%s-usb-%s.*" % (name[0], name[1])
+            if not model and os_machine == 'pseries':
+                # On pSeries, the default usb model is qemu-xhci
+                pattern = r"-device.qemu-xhci"
+            else:
```
chunfuwen: I assume whether  model = None or "" , the logic will also enter this block.
Do you need filter this situation?

libvirt/tests/src/controller/controller_functional.py
```diff
@@ -388,20 +388,24 @@ def check_qemu_cmdline(search_pattern=None):
...
+                model_list = [model]
+                if cmpnn_cntlr_num is not None:
+                    for num in range(int(cmpnn_cntlr_num)):
+                        model_list.append(cmpnn_cntlr_model+str(num+1))
```
chunfuwen: I assume you can remove model_list since you just use it to specify pattern type afterwards,therefore you can move line 401 and 408 here and actually item value is equals to cmpnn_cntlr_model+str(num+1).
In this way , you can save one loop.

dzhengfy: Done

libvirt/tests/src/controller/controller_functional.py
dzhengfy: Done

PR #1567, opened 2018-05-09
Fix vm-clone risk

yanqzhan:
For negative scenario: rename vm to an exist vm2, if clone fail,
the rename will succeed, and remove the vm2 will remove the
original one. This patch is to fix this.

Signed-off-by: Yanqiu Zhang <yanqzhan@redhat.com>


chloerh:
Problem fixed

PR #1571, opened 2018-05-10
Update set_mem wait time

hsmj1412:
Need more wait for set_mem take effect on ppc

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
@@ -76,7 +76,7 @@ def run(test, params, env):
     4) Check result.
     """
 
-    def vm_usable_mem(session):
+    def get_vm_usable_mem(session):
         """
         Get total usable RAM from /proc/meminfo
         """
@@ -95,7 +95,7 @@ def run(test, params, env):
         dmi_mem = session.cmd_output(cmd)
         total_physical_mem = reduce(lambda x, y: int(x) + int(y),
                                     re.findall(r'Size:\s(\d+)\sMB', dmi_mem))
-        return int(total_physical_mem) * 1024 - vm_usable_mem(session)
+        return int(total_physical_mem) * 1024 - get_vm_usable_mem(session)
 
     def make_domref(domarg, vm_ref, domid, vm_name, domuuid):
         """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
```diff
@@ -384,27 +369,66 @@ def print_debug_stats(original_inside_mem, original_outside_mem,
...
+                else:
+                    if vm.state() == "shut off":
+                        vm.start()
+                    elif vm.state() == "paused":
+                        # Make sure it's never paused
+                        vm.resume()
+                    session = vm.wait_for_login()
+
+                    # Actual results
+                    test_inside_mem = vm_usable_mem(session)
```
chunfuwen: rename vm_usable_mem to get_vm_usable_mem , method naming conventions:verb+n.

libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
```diff
@@ -384,27 +369,66 @@ def print_debug_stats(original_inside_mem, original_outside_mem,
...
+                    session.close()
+                    test_outside_mem = vm.get_used_mem()
+                return (test_inside_mem, test_outside_mem)
+
+            # Don't care about memory comparison on error test
+            def verify_outside_result():
+                _, test_outside_mem = get_vm_mem()
+                outside_pass = cal_deviation(test_outside_mem,
+                                             expected_outside_mem) <= delta_percentage
+                if outside_pass:
```
chunfuwen: I assume  outside_pass is boolean value ,and can return directly

libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
```diff
@@ -384,27 +369,66 @@ def print_debug_stats(original_inside_mem, original_outside_mem,
...
+                    return True
+                else:
+                    return False
+
+            def verify_inside_result():
```
chunfuwen: merged with verify_outside_result ,and using parameter to diff them

hsmj1412:
```
def wait_for(func, timeout, first=0.0, step=1.0, text=None):
    """
    Wait until func() evaluates to True.

    If func() evaluates to True before timeout expires, return the
    value of func(). Otherwise return None.

    :param timeout: Timeout in seconds
    :param first: Time to sleep before first attempt
    :param steps: Time to sleep between attempts in seconds
    :param text: Text to print while waiting, for debug purposes
    """
    start_time = time.time()
    end_time = time.time() + float(timeout)

    time.sleep(first)

    while time.time() < end_time:
        if text:
            logging.debug("%s (%f secs)", text, (time.time() - start_time))

        output = func()
        if output:
            return output

        time.sleep(step)

    return None
```

This function could not pass parameters

libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
hsmj1412: Updated

libvirt/tests/src/virsh_cmd/domain/virsh_setmem.py
hsmj1412: Updated

chloerh:
All cases passed on p9!

PR #1595, opened 2018-05-18
Add tls related test cases for serial device

jinqi7:

### DIFF
--- a/libvirt/tests/cfg/serial/serial_functional.cfg
+++ b/libvirt/tests/cfg/serial/serial_functional.cfg
@@ -1,10 +1,12 @@
 - serial.functional:
     type = serial_functional
     start_vm = no
+    local_ip_address = "127.0.0.1"
     variants:
         - type_dev:
             serial_type = dev
             serial_sources = path:/dev/ttyS0
+            target_type = isa-serial
         - type_file:
             serial_type = file
             serial_sources = path:/var/log/libvirt/virt-test
@@ -31,6 +33,7 @@
             serial_type = vc
         - type_pty:
             serial_type = pty
+            target_type = isa-serial
         - type_spicevmc:
             serial_type = spicevmc
         - type_spiceport:
@@ -46,3 +49,35 @@
             serial_type = pty
             console_target_type = virtio
             second_serial_console = yes
+        - type_tls_server:
+            serial_type = tls
+            serial_sources = mode:bind,host:${local_ip_address},service:5556,tls:yes
+            server_ip = ${local_ip_address}
+            server_user = "root"
+            server_pwd = "iforgot"
+            client_ip = ${local_ip_address}
+            client_user = "root"
+            client_pwd = "iforgot"
+            server_cn = ${local_ip_address}
+            client_cn = ${local_ip_address}
+            listen_addr = "0.0.0.0"
+            custom_pki_path = "/etc/pki/libvirt-chardev"
+            qemu_tls = "yes"
+            auto_recover = "yes"
+            restart_libvirtd = "yes"
+        - type_tls_client:
+            serial_type = tls
+            serial_sources = mode:connect,host:${local_ip_address},service:5556,tls:yes
+            server_ip = ${local_ip_address}
+            server_user = "root"
+            server_pwd = "iforgot"
+            client_ip = ${local_ip_address}
+            client_user = "root"
+            client_pwd = "iforgot"
+            server_cn = ${local_ip_address}
+            client_cn = ${local_ip_address}
+            listen_addr = "0.0.0.0"
+            custom_pki_path = "/etc/pki/libvirt-chardev"
+            qemu_tls = "yes"
+            auto_recover = "yes"
+            restart_libvirtd = "yes"
--- a/libvirt/tests/src/serial/serial_functional.py
+++ b/libvirt/tests/src/serial/serial_functional.py
@@ -6,12 +6,14 @@ import socket
 import threading
 import select
 import platform
+import subprocess
 
 import aexpect
 
 from virttest import remote
 from virttest import virsh
 from virttest.utils_test import libvirt
+from virttest.utils_conn import TLSConnection
 from virttest.libvirt_xml.vm_xml import VMXML
 from virttest.libvirt_xml.devices import librarian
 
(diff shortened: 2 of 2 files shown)


libvirt/tests/cfg/serial/serial_functional.cfg
```diff
@@ -46,3 +48,35 @@
...
+        - type_tls_server:
+            serial_type = tls
+            serial_sources = mode:bind,host:127.0.0.1,service:5556,tls:yes
+            server_ip = "127.0.0.1"
+            server_user = "root"
+            server_pwd = "ENTER.YOUR.REMOTE.PASSWORD"
+            client_ip = "127.0.0.1"
+            client_user = "root"
+            client_pwd = "ENTER.YOUR.CLIENT.PASSWORD"
+            server_cn = "127.0.0.1"
```
chunfuwen: you may define one variable :local_ip_address = "127.0.0.1",then refer to the variable value by ${local_ip_address }

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -45,18 +48,45 @@ def __init__(self, console_type, address, is_server=False):
         self.read_fd = None
         self.write_fd = None
         self._poll_thread = None
+        self.pki_path = '/etc/pki/libvirt-chardev'
```
chunfuwen: I assume self.pki_path = pki_path since the latter is parameter passing in

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -45,18 +48,45 @@ def __init__(self, console_type, address, is_server=False):
...
         self.prompt = r"^\[.*\][\#\$]\s*$"
         self.status_test_command = "echo $?"
 
         if console_type == 'unix':
             self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             self.socket.connect(address)
+        elif console_type == 'tls':
+            # Tls client type will be implemented later
+            outfilename = '/tmp/gutil.out'
+            self.read_fd = os.open(outfilename, os.O_RDWR
```
chunfuwen: use with ..open to open one file

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -45,18 +48,45 @@ def __init__(self, console_type, address, is_server=False):
...
+                                   | os.O_CREAT | os.O_NONBLOCK)
+            if is_server:
+                cmd = ('cd %s' % self.pki_path
+                       + ';gnutls-serv --echo --x509cafile ca-cert.pem '
+                       + '--x509keyfile server-key.pem --x509certfile '
+                       + 'server-cert.pem')
+                obj = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE,
```
chunfuwen: on python3 , subproces.Popen will return bytes mode instead of text mode,line 71 will throw error, see ,https://docs.python.org/3.4/library/subprocess.html#subprocess.Popen.stdin

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -45,18 +48,45 @@ def __init__(self, console_type, address, is_server=False):
...
+                time.sleep(50)
+                obj.stdin.write('\n')
+                obj.stdin.close()
+            else:
+                cmd = ('gnutls-cli --priority=NORMAL -p5556 --x509cafile=%s'
+                       % self.pki_path + '/ca-cert.pem ' + '127.0.0.1 '
+                       + '--x509certfile=%s' % self.pki_path
+                       + '/client-cert.pem --x509keyfile=%s' % self.pki_path
+                       + '/client-key.pem')
+                obj = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE,
```
chunfuwen: suggest line 81 to 86 to wrap into one function, and use parameter cmd to diff that

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -584,27 +633,40 @@ def check_serial_console(console, username, password):
...
-        Clean up residue file.
+        Clean up test environment
         """
+
         if serial_type == 'file':
             if os.path.exists('/var/log/libvirt/virt-test'):
                 os.remove('/var/log/libvirt/virt-test')
 
+        # recovery test environment
+        for obj in objs_list:
```
chunfuwen: use del objs_list[:] to delete all from list

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -584,27 +633,40 @@ def check_serial_console(console, username, password):
...
+            obj.auto_recover = True
+            del obj
+
+        if serial_type == 'tls' and console_type == 'server':
+            cmd = "kill -SIGTERM `pidof gnutls-serv`"
+            ret = subprocess.call(cmd)
```
chunfuwen: we normally use process.run instead of calliing subprocess.call directly

libvirt/tests/src/serial/serial_functional.py
```diff
@@ -678,11 +761,12 @@ def get_console_type():
...
-        if console_type is not None:
+        if (console_type is not None and serial_type != 'tls' and
+           console_type != 'server'):
             check_serial_console(console, username, password)
 
         vm.destroy()
 
-        check_cleanup()
     finally:
+        cleanup(objs_list)
```
chunfuwen: I assume you need clean up some certificates as well.

libvirt/tests/cfg/serial/serial_functional.cfg
jinqi7: Changed

libvirt/tests/src/serial/serial_functional.py
jinqi7: It's to reuse some old scripts. I changed the old scripts also as you suggested.

libvirt/tests/src/serial/serial_functional.py
jinqi7: Changed

libvirt/tests/src/serial/serial_functional.py
jinqi7: Code changed.

libvirt/tests/src/serial/serial_functional.py
jinqi7: Used the "process.terminate()" instead of it.

libvirt/tests/src/serial/serial_functional.py
jinqi7: The certificates are deleted in utils_conn module.

jinqi7:
@chunfuwen  Can you please help to review again?

PR #1596, opened 2018-05-18
Add secret and volume cleanup before testing started

chunfuwen:
In some case, dirty secret and volume will lead to next case execution failure
The way to fix them is to clean up dirty secret and volumes if they have

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
@@ -12,6 +12,7 @@ from virttest import remote
 from virttest import virt_vm
 from virttest import virsh
 from virttest import utils_disk
+from virttest import libvirt_storage
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml import vol_xml
@@ -69,6 +70,12 @@ def run(test, params, env):
         :param vol_params. Volume parameters dict.
         :return: True if create successfully.
         """
+        # Clean up dirty volumes if pool has.
+        pv = libvirt_storage.PoolVolume(p_name)
+        vol_name_list = pv.list_volumes()
+        for vol_name in vol_name_list:
+            pv.delete_volume(vol_name)
+
         volxml = vol_xml.VolXML()
         v_xml = volxml.new_vol(**vol_params)
         v_xml.encryption = volxml.new_encryption(**target_encrypt_params)
@@ -108,6 +115,30 @@ def run(test, params, env):
                                      **virsh_dargs)
         libvirt.check_exit_status(ret)
         return encryption_uuid
+
+    def get_secret_list():
+        """
+        Get secret list.
+
+        :return: secret list
+        """
+        logging.info("Get secret list ...")
+        secret_list = virsh.secret_list().stdout.strip().splitlines()
+        # First two lines contain table header followed by entries
+        # for each secret, such as:
+        #
+        # UUID                                  Usage
+        # --------------------------------------------------------------------------------
+        # b4e8f6d3-100c-4e71-9f91-069f89742273  ceph client.libvirt secret
+        secret_list = secret_list[2:]
+        result = []
+        # If secret list is not empty.
+        if secret_list:
+            for line in secret_list:
+                # Split on whitespace, assume 1 column
+                linesplit = line.split(None, 1)
+                result.append(linesplit[0])
+        return result
 
     def check_in_vm(vm, target, old_parts):
         """
@@ -208,6 +239,11 @@ def run(test, params, env):
     try:
         # Prepare the disk.
         sec_uuids = []
+        # Clean up dirty secrets in test environments if there are.
+        dirty_secret_list = get_secret_list()
+        if dirty_secret_list:
+            for dirty_secret_uuid in dirty_secret_list:
+                virsh.secret_undefine(dirty_secret_uuid)
         create_pool(pool_name, pool_type, pool_target)
         vol_params = {"name": volume_name, "capacity": int(volume_cap),
                       "allocation": int(volume_alloc), "format":


libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -68,6 +70,12 @@ def create_vol(p_name, target_encrypt_params, vol_params):
         :param vol_params. Volume parameters dict.
         :return: True if create successfully.
         """
+        # Clean up dirty volumes if pool has.
```
balamuruhans: I feel this have to be taken care during post process. The clean up should be taken care in each testcase in finally section before exiting the testcase or you can make the setup/cleanup in pre and postprocess.

The reason is, for this testcase we can cleanup before starting but the unclean environment can still affect other than this testcase

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -68,6 +70,12 @@ def create_vol(p_name, target_encrypt_params, vol_params):
...
+        vol_name_list = get_vol_list(p_name)
```
hahakiki2010: I think there's already existing func to provide this function. See avocado-vt virttest.libvirt_storage
```
class PoolVolume(object):

    """Volume Manager for libvirt storage pool."""

    def __init__(self, pool_name, virsh_instance=virsh):
        self.pool_name = pool_name
        self.virsh_instance = virsh_instance

    def list_volumes(self):
        """
        Return a dict include volumes' name(key) and path(value).
        """
```

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -108,6 +116,49 @@ def create_secret(vol_path):
         libvirt.check_exit_status(ret)
         return encryption_uuid
 
+    def get_vol_list(pool_name):
```
hahakiki2010: same comments as above. Use pv.list_volumes() directly

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -108,6 +116,49 @@ def create_secret(vol_path):
...
+
+    def get_secret_list():
+        """
+        Get secret list.
+
+        :return secret list
+        """
+        logging.info("Get secret list ...")
+        secret_list_result = virsh.secret_list()
+        secret_list = results_stdout_52lts(secret_list_result).strip().splitlines()
```
hahakiki2010: I think results_stdout_52lts is not necessary since the result has converted already. See virsh.py def command():
```
        ret.stdout = results_stdout_52lts(ret)
        ret.stderr = results_stderr_52lts(ret)
        # Return CmdResult instance when ignore_status is True
        return ret
```

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
hahakiki2010: same comments above

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: volume  setup/cleanup is specifics to some cases, if vol cleanup apply widely to all cases, such as networking related, it sound like weird

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
balamuruhans: It is not about specific testcases, it is how we have framework designed and cleanups are expected to be handled. IMHO this is not right way to do.

chunfuwen: That is good initiative if framework designed can handle cleanup issue. But regarding framework designed, it also need categorize it, such as storage, network, computing,  not just so-called "generic framework". Otherwise it look strange that you see volume cleanup in some network related cases.

chunfuwen:
result pasted here.
'''
+ avocado run --vt-type libvirt --vt-machine-type i440fx hotplug.raw_luks.normal_test.encryption_in_source --vt-connect-uri qemu:///system
JOB ID     : 6813a1f3aee6aac05bf9619ee73f19ffc3de52fe
JOB LOG    : /root/avocado/job-results/job-2019-11-04T01.35-6813a1f/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_disks.encryption.hotplug.raw_luks.normal_test.encryption_in_source: PASS (45.99 s)
JOB TIME   : 47.30 s

'''

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -109,6 +116,31 @@ def create_secret(vol_path):
         libvirt.check_exit_status(ret)
         return encryption_uuid
 
+    def get_secret_list():
+        """
+        Get secret list.
+
+        :return secret list
```
kylazhang: :return: secret list

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -109,6 +116,31 @@ def create_secret(vol_path):
...
+        secret_list = secret_list_result.stdout.strip().splitlines()
+        # First two lines contain table header followed by entries
+        # for each secret, such as:
+        #
+        # UUID                                  Usage
+        # --------------------------------------------------------------------------------
+        # b4e8f6d3-100c-4e71-9f91-069f89742273  ceph client.libvirt secret
+        secret_list = secret_list[2:]
+        result = []
+        # If secret list is empty.
```
kylazhang: For the comments do you mean "Check whether secret list is empty"? If yes, please update for better understand

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
```diff
@@ -109,6 +116,31 @@ def create_secret(vol_path):
...
 
+    def get_secret_list():
+        """
+        Get secret list.
+
+        :return secret list
+        """
+        logging.info("Get secret list ...")
+        secret_list_result = virsh.secret_list()
+        secret_list = secret_list_result.stdout.strip().splitlines()
```
kylazhang: Why not merge above 2 lines into 1?

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_encryption.py
chunfuwen: updated

PR #1602, opened 2018-05-22
Add libvirt build test config

sathnaga:
Add libvirt build test config to support virttest.libvirt_installer
enablement.

comment `no build` line from this config and can be used as below
to build, install libvirt package from git source
Usage: avocado run --vt-config build.cfg

### DIFF
--- a/libvirt/cfg/build.cfg
+++ b/libvirt/cfg/build.cfg
@@ -1,1 +1,33 @@
-# TODO: Support libvirt building
+# For other variables in build configuration look at build.cfg in qemu test provider
+
+vm_type = libvirt
+
+variants:
+    - build:
+	type = build
+
+	# Load modules built/installed by the build test?
+	load_modules = no
+	# Save the results of this build on test.resultsdir?
+	save_results = no
+	# Preserve the source code directory between tests?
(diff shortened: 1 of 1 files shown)


libvirt/cfg/build.cfg
```diff
@@ -1 +1,33 @@
...
+	installers = git_repo_libvirt
+
+	# Choose wether you want to include debug information/symbols
+	install_debug_info = yes
+
+	start_vm = no
+
+# Comment out the 'no build' line to enable the build test
+# and this cfg can be used as input of --vt-config
+no build
```
balamuruhans: anyway if we need to run this testcase we would use config file, why do we need "no build" filter ?

sathnaga: 1) This is not getting listed as a test even we remove `no build`, (may be placing it in a different path can make it).
2) This needs a change to be run in the specific env, and not a regular test anyways.
3) Finally I followed the qemu/build.cfg way...
not sure of exact reason why it is kept removed from Cartesian in first place...

iccaszhulili:
rh-caselink retest please

PR #1614, opened 2018-05-29
Fix No such file or directory when attaching disk xml

chunfuwen:
By copying one disk xml for additional guest VM.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_ceph.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_ceph.py
@@ -2,7 +2,9 @@ import os
 import re
 import time
 import logging
+import shutil
 import aexpect
+
 from collections import OrderedDict
 
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_disks/virtual_disks_ceph.py
```diff
@@ -761,7 +767,7 @@ def get_secret_list():
                 # Make sure the additional VM is running
                 if additional_vm.is_dead():
                     additional_vm.start()
-                ret = virsh.attach_device(guest_name, xml_file,
+                ret = virsh.attach_device(guest_name, additional_xml_file,
```
balamuruhans: The xml configuration is not changing for the attaching device for additional vm, then why would we copy xml to different name and attach ? I think existing code seems to be correct.

chunfuwen: I noticed that original xml was deleted in real testing environment after it was once used as attachment parameter,but not quite why it happen, so just copy another xml for reattach.

balamuruhans: IMHO can you help debugging more on why it got deleted, ideally it should work. If there is some inconsistency in framework then we have to address the root cause rather working around in test case

chunfuwen: True, it is occasional issue,and take time. I will track it by creating one jira

PR #1624, opened 2018-06-04
virsh_cpu_stats: Use total host cpu number

dzhengfy:
The case needs the number of SC_NPROCESSORS_CONF for total host cpu
number instead of SC_NPROCESSORS_ONLN for online cpu number

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_cpu_stats.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_cpu_stats.py
@@ -1,7 +1,8 @@
 import re
 import os.path
 import logging
-import multiprocessing
+
+from avocado.utils import cpu
 
 from virttest import virsh
 from virttest.staging import utils_cgroup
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_cpu_stats.py
```diff
@@ -32,7 +33,7 @@ def run(test, params, env):
         vm_ref = vm_name
 
     # get host cpus num
-    cpus = multiprocessing.cpu_count()
+    cpus = cpu.total_cpus_count()
```
chunfuwen: I am cautions about those change,please run related cases

dzhengfy: @chunfuwen , updated the code to narrow the scope. The fix is only for one case which need use total cpu number.

libvirt/tests/src/virsh_cmd/domain/virsh_cpu_stats.py
```diff
@@ -33,7 +35,7 @@ def run(test, params, env):
 
     # get host cpus num
     cpus = multiprocessing.cpu_count()
```
sathnaga: This can also be changed to `cpu.online_cpus_count()` ,

dzhengfy: @sathnaga sure, I 'v updated it to ensure the consistency.

PR #1631, opened 2018-06-05
virsh.boot: Adaption to qemu-kvm update

hahakiki2010:
Install qemu-kvm according to the os version. For RHEL-7, install qemu-kvm-rhev,
otherwise install qemu-kvm

### DIFF
--- a/libvirt/tests/cfg/bios/virsh_boot.cfg
+++ b/libvirt/tests/cfg/bios/virsh_boot.cfg
@@ -13,6 +13,7 @@
     status_error = "no"
     disk_format = "qcow2"
     driver_type = "qcow2"
+    os_version = "EXAMPLE_OS_VERSION"
     variants:
         - by_ovmf:
             boot_type = "ovmf"
--- a/libvirt/tests/src/bios/virsh_boot.py
+++ b/libvirt/tests/src/bios/virsh_boot.py
@@ -217,6 +217,10 @@ def setup_test_env(params, test):
     global cleanup_iso_file
     global cleanup_image_file
 
+    os_version = params.get("os_version")
+    if not os_version.count("EXAMPLE"):
+        os_version = os_version.split(".")[0]
+
     if boot_type == "ovmf":
         if not libvirt_version.version_compare(2, 0, 0):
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/bios/virsh_boot.py
```diff
@@ -217,6 +217,13 @@ def setup_test_env(params, test):
     global cleanup_iso_file
     global cleanup_image_file
 
+    os_version = params.get("os_version")
```
kylazhang: I think we can get it from current os

hahakiki2010: Yes, it's better. But current method shall work enough as the version is dynamically got from CI run.

libvirt/tests/cfg/bios/virsh_boot.cfg
```diff
@@ -13,6 +13,8 @@
     status_error = "no"
     disk_format = "qcow2"
     driver_type = "qcow2"
+    # In format of 'RHEL-7.5'
```
chunfuwen: I suggest remove specific RHEL information here

hahakiki2010: removed

libvirt/tests/src/bios/virsh_boot.py
```diff
@@ -225,8 +229,11 @@ def setup_test_env(params, test):
         if not utils_package.package_install('OVMF'):
             test.error("OVMF package install failed")
 
-        if not utils_package.package_install('qemu-kvm-rhev'):
+        if os_version == "RHEL-7" and not \
+                utils_package.package_install('qemu-kvm-rhev'):
```
chunfuwen: suggest just check whether any qemu-kvm-rhev and qemu-kvm exist , if not , just skip this case. 
No need do additional work to install them.

hahakiki2010: change to check whether qemu-kvm is installed, if not raise test error instead.  Here still keep test error as previous script do. And I think this is part of test env preparation, so raise test error  is also reasonable.

PR #1634, opened 2018-06-06
python3: fix send str to bytes type method issue

chunfuwen:
In serial_functional.py, one string is sent to socket.sendall(cont),which need bytes,
So fix it by transforming cont into bytes before sending.


libvirt/tests/src/serial/serial_functional.py
```diff
@@ -240,6 +241,8 @@ def send(self, cont=""):
 
         :param cont: String to send to the child process.
         """
+        if astring.is_text(cont):
+            cont = cont.encode()
```
hahakiki2010: Need specify the encoding format here otherwise it might have risk encoding error. 
```
import locale
cont = cont.encode(locale.getpreferredencoding())
```

chunfuwen: locale.getpreferredencoding() target some special character and not suit in anywhere, and if not given, utf-8 is default. Here utf-8 is preferred one

hahakiki2010: locale.getpreferredencoding() will return utf-8 also, but it would be definitely much better and more  compatible than the default one. Meanwhile it's avocado suggested method, all previous encoding/decoding issues were resolved in this way

chunfuwen: As previously pointed, avocado use local.getxx to resolve special characters for special function(method), not generic suggest for any functions. Moreover, locale.getdefaultlocale() may return None in some cases.

kylazhang: I agree with hahakiki2010

kylazhang: Approved after it fixed

chunfuwen: @kylazhang 
@hahakiki2010 
Whether use local.getxx() is controversial, see https://github.com/avocado-framework/avocado/pull/2611 for more details.
But 'utf-8' is the most commonly used one, and also that's the reason why python3 choose it as default encoding.

hahakiki2010: Pls check https://github.com/avocado-framework/avocado/pull/2623/files#diff-5e4b1f1e19fa36f2d92b7e4e9d8164a5. In astring.py, ```ENCODING = locale.getpreferredencoding()``` is set by default.

iccaszhulili:
rh-caselink retest please

chunfuwen:
close it since it is not needed now

PR #1650, opened 2018-06-08
Python3: Fixup error a bytes-like object is required

hahakiki2010:
File opened with wb, so need encode the writing data
to bytes to get it compatible with py2 and py3

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
@@ -6,9 +6,11 @@ import os
 import os.path
 import logging
 import aexpect
+from string import ascii_lowercase
+
 from six import iteritems
 
-from string import ascii_lowercase
+from avocado.utils import astring
 
 from virttest import virt_vm, virsh, remote, utils_misc, data_dir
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
```diff
@@ -610,7 +611,8 @@ def make_image_file(self, index):
             # Write simple unique data to file before end
             image_file.seek(byte_size - len(image_file_path) - 1)
             # newline required by aexpect in function()
-            image_file.write(image_file_path + '\n')
+            encoding = locale.getpreferredencoding()
```
chunfuwen: if prefer use locale, you can use astring.encoding constants instead

hahakiki2010: I checked astring.py, what specific function do you mean? only to_text is found

chunfuwen: use encoding=astring.ENCODING instead,not get from locale.getxx()

hahakiki2010: thanks, fixed

PR #1659, opened 2018-06-13
auto_dump: Add exception handling to kill process

hahakiki2010:
Avoid to raise CmdError "that process can not found" when trying
to kill the dump process when dump already accomplished

### DIFF
--- a/libvirt/tests/src/conf_file/qemu_conf/auto_dump.py
+++ b/libvirt/tests/src/conf_file/qemu_conf/auto_dump.py
@@ -107,7 +107,10 @@ def run(test, params, env):
             logging.debug(cmdline.split())
 
         # Kill core dump process to speed up test
-        process.run('kill %s' % iohelper_pid)
+        try:
+            process.run('kill %s' % iohelper_pid)
+        except process.CmdError as detail:
+            logging.debug("Dump already done:\n%s", detail)
 
         arch = platform.machine()
 


libvirt/tests/src/conf_file/qemu_conf/auto_dump.py
```diff
@@ -107,7 +107,10 @@ def run(test, params, env):
             logging.debug(cmdline.split())
 
         # Kill core dump process to speed up test
-        process.run('kill %s' % iohelper_pid)
+        try:
+            process.run('kill %s' % iohelper_pid)
+        except process.CmdError as detail:
+            logging.debug("Dump already done:\n%s", detail)
```
chunfuwen: may change detail to str(detail)

balamuruhans: I doubt this will fix the check of already killed dump process,
`process.CmdError` would be raised if the command is invalid / not available. here when we try to kill the already killed process it would not raise.
The better solution would be check for the process id from ps to check is already done or not.

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/conf_file/qemu_conf/auto_dump.py
hahakiki2010: @balamuruhans , it will raise CmdError:
```
>>> process.run('kill 123123')
  File "<stdin>", line 1, in <module>
  File "/home/code/virtenv_py3/lib/python3.6/site-packages/avocado_framework-61.0-py3.6.egg/avocado/utils/process.py", line 1275, in run
    raise CmdError(cmd, sp.result)
avocado.utils.process.CmdError: ('kill 123123', <avocado.utils.process.CmdResult object at 0x7fec0720fc88>)

```

hahakiki2010: @chunfuwen ,"%s"  will transform to str automatically.

PR #1686, opened 2018-06-23
Add no_option test for save-image-edit

kylazhang:
1. Add no_option test and add xml back/recover step
2. Also update cfg format for save-image-define

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_save_image_edit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_save_image_edit.py
@@ -5,6 +5,7 @@ import aexpect
 
 from virttest import data_dir
 from virttest import virsh
+from virttest.libvirt_xml import vm_xml
 
 
 def run(test, params, env):
@@ -26,6 +27,8 @@ def run(test, params, env):
             option = "--running"
         elif restore_state == "paused":
             option = "--paused"
+        elif restore_state == "":
+            option = ""
         else:
             test.fail("Unknown save-image-define option")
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virsh_cmd/domain/virsh_save_image_define.cfg
```diff
@@ -3,7 +3,7 @@
...
     kill_vm_on_error = "no"
     variants:
-       - running:
-           restore_state = "running"
-       - paused:
-           restore_state = "paused"
+        - running:
+            restore_state = "running"
+        - paused:
+            restore_state = "paused"
```
balamuruhans: we can have this fix as a separate commit.

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/virsh_cmd/domain/virsh_save_image_edit.py
```diff
@@ -97,6 +102,8 @@ def vm_state_check():
 
     finally:
         # cleanup
+        vm_backup.sync()
```
chunfuwen: Better destroy VM before do vm backup

libvirt/tests/cfg/virsh_cmd/domain/virsh_save_image_define.cfg
kylazhang: As I need add another option, so change them at same time.

libvirt/tests/src/virsh_cmd/domain/virsh_save_image_edit.py
kylazhang: no needed, sync will update the inactive xml.

PR #1689, opened 2018-06-26
migrate_options_shared: Fix xbzrle_cache_size issue

dzhengfy:
Qemu-kvm's new behaviour will raise the error for --comp-xbzrle-cache:

Parameter 'xbzrle_cache_size' expects is invalid, it should be bigger than target page size
and a power of two

So this is to use valid number for --comp-xbzrle-cache.

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -5,6 +5,7 @@ import re
 import threading
 
 from avocado.utils import process
+from avocado.utils import memory
 from avocado.core import exceptions
 
 from virttest import libvirt_vm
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -202,6 +203,29 @@ def check_timeout_postcopy(params):
...
+    def get_usable_compress_cache(pagesize):
+        """
+        Get a number which is bigger than pagesize and is power of two.
+
+        :param pagesize: the given integer
+        :return: an integer satisfying the criteria
+        """
+        def calculate(num):
+            result = num & (num - 1)
+            if result == 0:
```
chunfuwen: return (result == 0)

dzhengfy: @chunfuwen updated, thanks.

PR #1693, opened 2018-06-27
Add memory cases: memory should align with 256

chloerh:
- cold-plug memory device
- hot-plug memory device

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -2,6 +2,7 @@ import os
 import re
 import ast
 import logging
+import platform
 import tempfile
 import random
 
@@ -95,9 +96,12 @@ def run(test, params, env):
         config.restore()
         utils_libvirtd.libvirtd_restart()
 
-    def check_qemu_cmd():
+    def check_qemu_cmd(max_mem_rt, tg_size):
         """
         Check qemu command line options.
+        :param max_mem_rt: size of max memory
+        :param tg_size: Target hotplug memory size
+        :return: None
         """
         cmd = ("ps -ef | grep %s | grep -v grep " % vm_name)
         if max_mem_rt:
@@ -121,7 +125,9 @@ def run(test, params, env):
                         (mem_addr['slot'], int(mem_addr['base'], 16)))
             cmd += "'"
         # Run the command
-        process.run(cmd, shell=True)
+        result = process.run(cmd, shell=True, verbose=True, ignore_status=True)
+        if result.exit_status:
+            test.fail('Qemu command check fail.')
 
     def check_guest_meminfo(old_mem, check_option):
         """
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/libvirt_mem.py
```diff
@@ -348,6 +389,11 @@ def modify_domain_xml():
     if not libvirt_version.version_compare(1, 2, 14):
         test.cancel("Memory hotplug not supported in current libvirt version.")
 
+    if 'align_256m' in params.get('name', ''):
```
balamuruhans: I could not see the param `name` defined in config file.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -417,13 +463,29 @@ def modify_domain_xml():
                                   flagstr=max_mem_option)
             libvirt.check_exit_status(ret, maxmem_error)
 
+        # Hotplug memory device
+        if add_mem_device and hot_plug:
```
balamuruhans: we do the same earlier in L377, how does the change in L423 is different.
can we have the logic squashed together avoiding redundancy

libvirt/tests/src/libvirt_mem.py
```diff
@@ -177,6 +180,38 @@ def check_dom_xml(at_mem=False, dt_mem=False):
...
+            logging.info('%-20s:%15d', key, dom_mem[key])
+
+        if dom_mem['memory'] == sum_numa_mem + dom_mem['attached_mem']:
+            logging.info('Check Pass: Memory is equal to (all numa memory + memory device)')
+        else:
+            test.fail('Memory is not equal to (all numa memory + memory device)')
+
+        for key in dom_mem:
+            if dom_mem[key] % 256:
+                test.fail('%s not align to 256' % key)
```
balamuruhans: this can be moved to L202

libvirt/tests/src/libvirt_mem.py
chloerh: I tried, but didn't find a good way without affecting existing cases.

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/libvirt_mem.py
chloerh: I don't understand, is it ok or not?

chloerh: I mean current state...

libvirt/tests/src/libvirt_mem.py
```diff
@@ -95,7 +96,7 @@ def restore_hugepages(page_size=4):
         config.restore()
         utils_libvirtd.libvirtd_restart()
 
-    def check_qemu_cmd():
+    def check_qemu_cmd(max_mem_rt, tg_size):
```
chunfuwen: miss doc string

chloerh: Updated

PR #1696, opened 2018-06-28
virsh_net_define_undefine: Add readonly test

hahakiki2010:


libvirt/tests/cfg/virsh_cmd/network/virsh_net_define_undefine.cfg
```diff
@@ -39,7 +39,7 @@
                             net_active = "no"
                 - default_network:
                     net_define_undefine_net_name = "default"
-                    net_define_undefine_net_uuid = ""
+                    net_define_undefine_net_uuid = ""transaction
```
chunfuwen: string initialization error

hahakiki2010: fixed

libvirt/tests/src/virsh_cmd/network/virsh_net_define_undefine.py
```diff
@@ -122,6 +122,9 @@ def run(test, params, env):
         cmd = "chmod 666 %s" % testnet_xml.xml
         process.run(cmd, shell=True)
 
+    if params.get('net_define_undefine_readonly', 'no') == 'yes':
+        virsh_dargs = {'uri': uri, 'debug': False, 'ignore_status': True,
```
chunfuwen: virsh_dargs are used  by a few functions in this python file, please make this change also apply to all functions when net_define_undefine_readonly = "yes"

hahakiki2010: From my view, only net-define and net-undefine require 'readonly' test

chunfuwen: virsh_dargs is initialized in line 63, no need to reassign one whole value.
Maybe use dict.update({xx})

iccaszhulili:
rh-caselink retest please

PR #1698, opened 2018-06-29
Fix cpu-baseline error

kylazhang:
Fix error as before as new version of libvirt have some update
error: File '/var/tmp/avocado_rCvZEW/cpu.xml' does not contain any <cpu> element or valid domain XML, host capabilities XML, or domain capabilities XML


kylazhang:
The list of host CPUs is built by extracting all <cpu> elements from the <file>. Thus, the <file> can contain either a set of <cpu> elements separated by new lines or even a set of complete <capabilities> elements printed by capabilities command.

kylazhang:
The travis ci failure on python2.6 seems not my patch issue, so ignore it.

PR #1699, opened 2018-06-29
add case to cover attach hostdev with --persistent option

chunfuwen:
device attached by persistent option should be there after VM reboot

### DIFF
--- a/libvirt/tests/src/sriov/sriov.py
+++ b/libvirt/tests/src/sriov/sriov.py
@@ -369,6 +369,11 @@ def run(test, params, env):
         if option == "--config":
             result = virsh.start(vm_name)
             utils_test.libvirt.check_exit_status(result, expect_error=False)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/sriov/sriov.py
```diff
@@ -369,6 +369,11 @@ def attach_interface():
         if option == "--config":
             result = virsh.start(vm_name)
             utils_test.libvirt.check_exit_status(result, expect_error=False)
+        # For option == "--persistent", after VM stop and restart, the device should still be there.
```
balamuruhans: `after VM destroyed and restart` would be more appropriate.

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/sriov/sriov.py
chunfuwen: fixed

PR #1705, opened 2018-07-02
add not existing block job test case

chunfuwen:
do operations for a non-existing block job,and
expected result should fail


iccaszhulili:
rh-caselink retest please

PR #1706, opened 2018-07-02
virsh_nodedev_detach_reattach: Add readonly test for detach

hahakiki2010:
Support detach nodedevice from its device driver when readonly
is enabled

### DIFF
--- a/libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_detach_reattach.py
+++ b/libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_detach_reattach.py
@@ -50,6 +50,7 @@ def run(test, params, env):
         # Libvirt acl polkit related params
         uri = params.get("virsh_uri")
         unprivileged_user = params.get('unprivileged_user')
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/nodedev/virsh_nodedev_detach_reattach.py
```diff
@@ -58,7 +59,7 @@ def detach_reattach_nodedev(device_address, params, options=""):
         logging.debug('Node device name is %s.', device_address)
         CmdResult = virsh.nodedev_detach(device_address, options,
                                          unprivileged_user=unprivileged_user,
-                                         uri=uri)
+                                         uri=uri, readonly=readonly)
```
chunfuwen: I assume readonly is boolean variable, rather than string

hahakiki2010: fixed.

iccaszhulili:
rh-caselink retest please

PR #1708, opened 2018-07-02
[WIP]virsh_migrate: Enable P8 <-> P9 migration

dzhengfy:
The existing configuration only supports P9 <-> P9 migration. So this is
to add P8 <-> P9 migration.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
@@ -58,11 +58,14 @@
                     power9_compat = "yes"
                     power9_compat_remote = "yes"
                     restore_smt = "yes"
-                    # Test P8 compat mode guest on P9 host only
-                    host_arch = POWER9
-                    # Migrating Power8 guest between P9 <-> P9 hosts
-                    # and P8 <-> P9 hosts
                     cpu_model = "power8"
+                    variants:
+                        - from_power8:
+                            # Test P8 compat mode guest between P8 <-> P9 host
+                            host_arch = POWER8
+                        - from_power9:
+                            # Test P8 compat mode guest between P9 <-> P9 host
+                            host_arch = POWER9
                 - non_compat_migration:
     variants:
         - with_cpu_hotplug:


libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate.cfg
```diff
@@ -58,11 +58,14 @@
...
                     restore_smt = "yes"
-                    # Test P8 compat mode guest on P9 host only
-                    host_arch = POWER9
-                    # Migrating Power8 guest between P9 <-> P9 hosts
-                    # and P8 <-> P9 hosts
                     cpu_model = "power8"
+                    variants:
+                        - from_power8:
+                            # Test P8 compat mode guest between P8 <-> P9 host
+                            host_arch = POWER8
```
balamuruhans: This change would boot Power8 compat guest on Power8 host,
1. It is not required to boot Power8 guest on a Power8 host in compat mode.
2. Currently with latest Qemu even for Power9 host it doesn't need this explicit compat mode setting because during guest boot, there is a CAS handshake where Qemu knows that guest kernel is Power8 and branching happen.

dzhengfy: @balamuruhans , 
For 1, do you mean which the following?
1. We need not test P8 guest migration from P8 to P9?
2. P8 guest migration from P8 to P9 do not need compat mode, that is to say,  guest xml does not require below setting?
  <cpu check="partial" mode="host-model">
    <model fallback="allow">power8</model>
  </cpu>

Thanks for your clarification.

balamuruhans: > For 1, do you mean which the following?

>    We need not test P8 guest migration from P8 to P9?

we should be testing P8 <-> P9

>    P8 guest migration from P8 to P9 do not need compat mode, that is to say, guest xml does not require below setting?
>
>  <cpu check="partial" mode="host-model">
>  <model fallback="allow">power8</model>
>  </cpu>

correct, latest qemu works in architected mode and doesn't not require it.

> In our test plan, we still need test cross migration with P8 guest from RHEL7.5 to RHEL-ALT 7.6.

we should test it, as it is good scenario and it is customer usecase too.

> For 2. with latest qemu, migration with P8 guest between P9 hosts will be same with P9 guest, that > is to say, guest configurations have no difference. Am I right?

right, with latest qemu P8 <-> P9 or P9 <-> P9 or P8 <-> P8 we do not need any guest xml changes with respect to cpu compat.

dzhengfy: @balamuruhans thanks for your information. So I close this PR.

dzhengfy:
@balamuruhans , per discussion with dev, we are suggested to test with compat mode explicitly. See https://bugzilla.redhat.com/show_bug.cgi?id=1599167#c26 and https://bugzilla.redhat.com/show_bug.cgi?id=1599167#c8. 
Based on the comments, I reopened this PR in order to able to run these tests in our test environment. 
Please consider.  Thanks.

balamuruhans:
@dzhengfy Okay, I have mistaken it. sorry for it.

balamuruhans:
@dzhengfy,  this test case was running only on POWER9 host as `host_arch` was POWER9, 
```python
    compat_guest_migrate = (params.get("host_arch", "all_arch") in
                            utils_misc.get_cpu_info()['Model name'])
```
but with this change it will start enabled for POWER8 host too. so if both source host and target host are POWER8 then the test is not necessary right ?
 
I believe this test would be appropriate if POWER8 guest migrates to and fro between 
POWER9 host <-> POWER9 host or 
POWER8 host <-> POWER9 host.

Still without this change it is possible to be achieved by having destination as POWER8 host
This existing test would bring up POWER8 compat guest on POWER9 host, if we configure the remote params with POWER8 host then it is taken care.

please share your comments, it will be hard for us to maintain the reason is that we get increased no of testcases by adding this and again we need to have check in script for checking whether this compat migration is between POWER8 < -> POWER9 or POWER9 <-> POWER9 and skip the test if it is POWER8 <-> POWER8.

I am okay to merge it, if you are still interested. Thank you

dzhengfy:
@balamuruhans , 
Yes, I agree your thoughts. I also only need the migration test from P8 to P9 with P8 compat mode guest and P9 to P9 with P8 compat mode guest.  So I think I should ignore the test when the target host is Power8. What do you think of?

balamuruhans:
@dzhengfy I think we can run the compat test on POWER9 Host as source that boots POWER8 compat guest and provide remote params for POWER8 host as destination will serve our purpose even without this PR.

dzhengfy:
@balamuruhans , so your preferred test scenarios are below, right?
1. migration from Power 9 to Power 9 with compat P8 guest
2. migration from Power 9 to Power 8 with compat P8 guest

FYI. My test scenarios are in https://bugzilla.redhat.com/show_bug.cgi?id=1557058#c0.

dzhengfy:
I will be on vacation in following two weeks, so I will revisit this PR after that. Thanks.

PR #1710, opened 2018-07-03
Fix v2v scripts errors on rhel8

xiaodwan:
1) Replace "--network ovirtmgmt" with "ovirtmgmt"
   In virt-v2v-1.38.2-2.el8+7.x86_64, it is "--bridge ovirtmgmt" instead
   of "--network ovirtmgmt". Let's just follow the description in bz1429491,
   and only check string "ovirtmgmt".
2) Python3: Decode bytes-like object to string
   Decode bytest-like object to string in check_result to avoid
   exception in underlying functions if they do not hanlde bytes-like
   problems correctly.
3) Python3: Replace string.replace with str.replace

### DIFF
--- a/v2v/tests/cfg/v2v_options.cfg
+++ b/v2v/tests/cfg/v2v_options.cfg
@@ -142,7 +142,7 @@
                     only input_mode.none
                     only output_mode.none
                     v2v_options = --help
-                    in_man = '--network ovirtmgmt'
+                    in_man = 'ovirtmgmt'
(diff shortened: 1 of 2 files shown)


chloerh:
Looks good to me :+1:

xiaodwan:
@chloerh  I did some changes for this patch and please help to review it again. Thanks in advance!

chloerh:
:+1:

PR #1712, opened 2018-07-04
Add test for schedinfo

kylazhang:
1. Add multiple set at one time
2. Add readonly mode
3. Add --current info show
4. Add shutdown guest testing

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
@@ -72,19 +72,30 @@ def run(test, params, env):
             test.fail("Output is not standard:\n%s" % output)
 
         result_lines = output.splitlines()
-        set_value = None
-        for line in result_lines:
-            key_value = line.split(":")
-            key = key_value[0].strip()
-            value = key_value[1].strip()
-            if key == "Scheduler":
-                if value != scheduler:
-                    test.cancel("This test do not support"
-                                " %s scheduler." % scheduler)
-            elif key == set_ref:
-                set_value = value
-                break
-        return set_value
+        set_value_list = []
+        for set_ref_node in set_ref.split(","):
+            for line in result_lines:
+                key_value = line.split(":")
+                key = key_value[0].strip()
+                value = key_value[1].strip()
+                if key == "Scheduler":
+                    if value != scheduler:
+                        test.cancel("This test do not support"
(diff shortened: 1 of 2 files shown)


kylazhang:
Need more test for this update, will paste result later, thanks

iccaszhulili:
rh-caselink retest please

libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
```diff
@@ -109,7 +123,13 @@ def schedinfo_output_analyse(result, set_ref, scheduler="posix"):
     elif set_ref:
         if set_method == 'cmd':
             if set_value:
-                options_ref = "--set %s=%s" % (set_ref, set_value)
+                set_ref_list = set_ref.split(",")
+                set_value_list = set_value.split(",")
+                for i in range(0, len(set_ref_list)):
+                    if "--set" in options_ref:
```
chunfuwen: options_ref is declared under the condition if set_ref == "none", therefore it may throw variable not defined error in this statement.

libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
```diff
@@ -149,55 +169,81 @@ def schedinfo_output_analyse(result, set_ref, scheduler="posix"):
...
-    else:
-        if not status:
-            test.fail("Run successfully with wrong command.")
+            if not status:
+                test.fail("Run successfully with wrong command.")
+            if readonly:
+                if not re.search(expect_msg, result.stderr.strip()):
+                    test.fail("Fail to get expect err msg: %s" % expect_msg)
+    finally:
+        if set_ref:
```
chunfuwen: suggest add one vm restore function here since it may impact other cases execution if any point fail in this case

libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
kylazhang: We have declared in both if/else so no such issue

libvirt/tests/src/virsh_cmd/domain/virsh_schedinfo_qemu_posix.py
kylazhang: virsh.schedinfo(...) in finally is the restore setting.

PR #1715, opened 2018-07-05
Add ppc-only cases

chloerh:
- Assign address to panic device
- Passthrough non-available PCI device in guest

### DIFF
--- a/libvirt/tests/src/ppc_device.py
+++ b/libvirt/tests/src/ppc_device.py
@@ -1,0 +1,80 @@
+import logging
+import platform
+
+from avocado.utils import process
+
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices.panic import Panic
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    Test only ppc hosts
+    """
+    if 'ppc64le' not in platform.machine().lower():
+        test.cancel('This case is for ppc only.')
+    vm_name = params.get('main_vm', 'EXAMPLE')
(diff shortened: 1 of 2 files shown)


iccaszhulili:
rh-caselink retest please

chloerh:
Test result on power9
---

测试名称 | 运行时间 | 结果
-- | -- | --
panic_address | 28 秒 | Fixed
unavail_pci_device | 15 秒 | Passed

kylazhang:
Please remove chinese in comments as some reviewers may not know what's the mean.

chloerh:
@kylazhang Done

libvirt/tests/src/ppc_device.py
```diff
@@ -0,0 +1,81 @@
...
+            for i in range(5):
+                max_id += 1
+                # function must be <= 7
+                if max_id > 7:
+                    break
+                new_pci_id = '.'.join([prefix, str(max_id)])
+                new_pci_xml = libvirt.create_hostdev_xml(new_pci_id, xmlfile=False)
+                vmxml.add_device(new_pci_xml)
+            vmxml.sync()
+            logging.debug('Vm xml after adding unavailable pci devices:')
```
dzhengfy: Why not use one logging statement?

chloerh: You're right! Updated.

chunfuwen:
@chloerh , suggest use more specific cfg file name:ppc_pci_device.cfg since there are large ranges of devices

chloerh:
@chunfuwen , yes, I have considered that before. But this case file is not just about pci device, therefore the name of file will be either general or inaccurate.

PR #1717, opened 2018-07-06
Fix checking image issue with qemu-img

xiaodwan:
when vm is running, then checking the image of the vm by
qemu-img, '-U' argument is needed.

Signed-off-by: xiaodwan <xiaodwan@redhat.com>

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -19,6 +19,7 @@ from virttest import utils_misc
 from virttest import utils_v2v
 from virttest import virsh
 from virttest import remote
+from virttest import libvirt_storage
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt as utlv
 from virttest.compat_52lts import decode_to_text as to_text
@@ -469,8 +470,13 @@ def run(test, params, env):
             if checkpoint == 'compress':
                 img_path = get_img_path(output)
                 logging.info('Image path: %s', img_path)
-                disk_check = process.run(
-                    'qemu-img check %s' % img_path).stdout_text
+
+                qemu_img_cmd = 'qemu-img check %s' % img_path
+                qemu_img_locking_feature_support = libvirt_storage.check_qemu_image_lock_support()
+                if qemu_img_locking_feature_support:
+                    qemu_img_cmd = 'qemu-img check %s -U' % img_path
+
+                disk_check = process.run(qemu_img_cmd).stdout_text
                 logging.info(disk_check)
                 compress_info = disk_check.split(',')[-1].split('%')[0].strip()
                 compress_rate = float(compress_info)


xiaodwan:
@chloerh please help to review it. Thanks!

v2v/tests/src/v2v_options.py
```diff
@@ -468,7 +468,7 @@ def check_alloc():
                 img_path = get_img_path(output)
                 logging.info('Image path: %s', img_path)
                 disk_check = process.run(
-                    'qemu-img check %s' % img_path).stdout_text
+                    'qemu-img check %s -U' % img_path).stdout_text
                 logging.info(disk_check)
```
chloerh: The output will be stored in var 'disk_check' which will be split and abstract info from. Will the outputs have the same structure with '-U' or without it?

xiaodwan: --force-share (-U)
           If specified, "qemu-img" will open the image in shared mode, allowing other QEMU processes to open it in write mode. For example, this can be used to get the image
           information (with 'info' subcommand) when the image is used by a running guest.  Note that this could produce inconsistent results because of concurrent metadata changes,
           etc. This option is only allowed when opening images in read-only mode.

'-U' will not change any output format, without it, you qemu-img will report error when the image is used by another process.

BTW, the case have passed.

chloerh: Great! Then maybe next time, we could paste the passed test result like xxx.xxx.xxx passed(xx secs) to be more persuasive.
I'm OK with this one.

v2v/tests/src/v2v_options.py
```diff
@@ -468,7 +468,7 @@ def check_alloc():
                 img_path = get_img_path(output)
                 logging.info('Image path: %s', img_path)
                 disk_check = process.run(
-                    'qemu-img check %s' % img_path).stdout_text
+                    'qemu-img check %s -U' % img_path).stdout_text
```
chunfuwen: I assume '-U' is supported since RHEL7.5 onward, so you may need add some version assertion here

xiaodwan: ok

xiaodwan:
# bin/avocado run --vt-type v2v --vt-machine-type i440fx v2v_options.positive_test.compress.input_mode.libvirt.kvm.default.output_mode.libvirt --vt-connect-uri qemu:///system
JOB ID     : cd1d51e425c38a6303a63a7b48f0db595e42d1f3
JOB LOG    : /root/avocado/job-results/job-2018-07-06T02.23-cd1d51e/job.log
 (1/1) type_specific.io-github-autotest-libvirt.v2v_options.positive_test.compress.input_mode.libvirt.kvm.default.output_mode.libvirt: PASS (81.69 s)
JOB TIME   : 166.00 s

xiaodwan:
@chunfuwen please help to review again, thanks!

PR #1724, opened 2018-07-11
Update alias testing

kylazhang:
According to new change, when alias is not given, it will be generated
automatically, but changing device alias is not allowed, so do update
to get running vm's disks' alias and add into xml

depends on https://github.com/avocado-framework/avocado-vt/pull/1660


juteman:
it seems omit a default parameter in fun update_device.
and i give a pr.

PR #1739, opened 2018-07-19
virsh_change_media.py: add default parameter to fun update_device

juteman:
while test, find that var disk_alias is not
reference before use.So add a default parameter
to fix it

Signed-off-by: juteman <qiulzhan@redhat.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_change_media.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_change_media.py
@@ -34,8 +34,10 @@ def run(test, params, env):
         :param new_iso: sourse file for update
         """
         error_context.context("Preparing ISO images")
-        process.run("dd if=/dev/urandom of=%s/old bs=1M count=1" % iso_dir, shell=True)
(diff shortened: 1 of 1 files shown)


juteman:
while run 
```
sudo avocado run --vt-type libvirt  virsh.change_media.cdrom_test.scsi_.positive_test.eject.none.running_guest     --show-job-log
```
will  throw error
```
  File "//home/juteman/code/tp-libvirt/libvirt/tests/src/virsh_cmd/domain/virsh_change_media.py", line 283, in run
    update_device(vm_name, "", options, start_vm)
  File "//home/juteman/code/tp-libvirt/libvirt/tests/src/virsh_cmd/domain/virsh_change_media.py", line 107, in update_device
    if disk_alias is None:
NameError: free variable 'disk_alias' referenced before assignment in enclosing scope

```
so i add a default parameter

iccaszhulili:
rh-caselink retest please

kylazhang:
disk_alias should have default value "None" in libvirt.get_disk_alias() so please check again for the test error as it may because of avocado-vt patch not merge during your previous test

PR #1741, opened 2018-07-20
virsh.migrate: Adjust param for calling vm.migrate

dzhengfy:
Due to avocado-vt commit b43df1ff, all functions using migrate() may need adjust the parameters.

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -102,7 +102,7 @@ def run(test, params, env):
         logging.info("Sleeping 10 seconds before migration")
         time.sleep(10)
         # Migrate the guest.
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -102,7 +102,7 @@ def do_migration(vm, dest_uri, options, extra):
         logging.info("Sleeping 10 seconds before migration")
         time.sleep(10)
         # Migrate the guest.
-        migration_res = vm.migrate(dest_uri, options, extra, True, True)
+        migration_res = vm.migrate(dest_uri, options, extra, ignore_status=True, debug=True)
```
chunfuwen: you may pack those parameters as **virsh_args{}

dzhengfy: @chunfuwen updated.

PR #1749, opened 2018-07-24
virsh_hostname: Update judgement for remote_ip

hahakiki2010:
The old condition will cause other cases skipped when with remote ip not configured.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
@@ -43,8 +43,8 @@ def run(test, params, env):
 
     # Run test case
     option = params.get("virsh_hostname_options")
-    remote_ip = params.get("remote_ip", None)
-    if not remote_ip or remote_ip.count("EXAMPLE"):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
```diff
@@ -43,8 +43,8 @@ def run(test, params, env):
 
     # Run test case
     option = params.get("virsh_hostname_options")
-    remote_ip = params.get("remote_ip", None)
-    if not remote_ip or remote_ip.count("EXAMPLE"):
+    remote_ip = params.get("remote_ip")
```
kylazhang: No need to set None for remote_ip as default is None

hahakiki2010: Seemingly you comment on the old fix. I didn't set None for remote_ip in new fix

kylazhang: OK, thanks

PR #1750, opened 2018-07-24
migrate_vm: Compatible with old behavior

dzhengfy:
There is a migration different result from libvirt 4.3.0-1 when
migrating without shared storage and --copy-storage-all
Before: migration succeeds with image preallocation on target
host.
After: migration is forbidden.


iccaszhulili:
rh-caselink retest please

PR #1751, opened 2018-07-25
Add test for uri test

kylazhang:
1. Add invalid uri test
2. Add non_root_user test

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_uri.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_uri.py
@@ -1,7 +1,6 @@
 import logging
 
 from avocado.utils import process
-from avocado.core import exceptions
 
 from virttest import libvirt_vm
 from virttest import virsh
@@ -25,6 +24,7 @@ def run(test, params, env):
                                                               "default"))
 
     option = params.get("virsh_uri_options")
+    unprivileged_user = params.get('unprivileged_user')
 
     remote_ip = params.get("remote_ip", "REMOTE.EXAMPLE.COM")
     remote_pwd = params.get("remote_pwd", None)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/host/virsh_uri.py
```diff
@@ -55,33 +55,40 @@ def run(test, params, env):
...
 
+    if unprivileged_user:
+        if process.run("id %s" % unprivileged_user, ignore_status=True).exit_status != 0:
+            process.run("useradd %s" % unprivileged_user)
     try:
-        if remote_ref == "remote":
+        if remote_ref == "remote" or unprivileged_user:
             connect_uri = target_uri
-        uri_test = virsh.canonical_uri(option, uri=connect_uri,
+        uri_test = virsh.canonical_uri(option,
```
chunfuwen: suggest not change virsh.connxx_uri () for remote_ref == remote, just modify unprivei_user situation:
```
virsh_args = {}
if unprevious_user:
   virssh_args.update{'unprivileged_user':unprivileged_user, 'debug': True, 'ignore_status': True}
uri_test = virsh.canonical_uri(option, uri=connect_uri, **virsh_args
```

kylazhang: no, I need to update the uri as for remote it is different

libvirt/tests/src/virsh_cmd/host/virsh_uri.py
```diff
@@ -57,37 +57,44 @@ def run(test, params, env):
...
+                                       unprivileged_user=unprivileged_user,
+                                       uri=connect_uri,
                                        ignore_status=False,
                                        debug=True)
         status = 0  # good
     except process.CmdError:
         status = 1  # bad
         uri_test = ''
+        if unprivileged_user:
```
dzhengfy: When no exception, 'userdel' is not executed. So is it better to add a finally clause?

PR #1760, opened 2018-07-27
add blockjob bytes options cases

chunfuwen:
blockjob --info and --bandwidth can combine with bytes options
and expected output unit is bytes


chunfuwen:
minor change

PR #1762, opened 2018-07-29
Add --xml test for restore

kylazhang:

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
@@ -1,9 +1,11 @@
 import re
 import os
+import logging
 
 from virttest import virsh
 from virttest import data_dir
 from virttest import utils_libvirtd
+from virttest.libvirt_xml import vm_xml
 
 from provider import libvirt_version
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
```diff
@@ -41,6 +43,18 @@ def run(test, params, env):
             test.cancel("API acl test not supported in current"
                         " libvirt version.")
 
+    if "--xml" in extra_param:
+        vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name, options="--migratable")
```
chunfuwen: Since vm_backup generally apply to any usecase, so suggest move backup code outside this block,and do it generally
```
    backup_xml = vmxml.copy()
```

libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
```diff
@@ -98,11 +112,19 @@ def run(test, params, env):
                     test.fail("Guest state should be"
                               " paused after restore"
                               " due to the option --paused")
-            if extra_param.count("running"):
+            if (extra_param.count("running") or
```
chunfuwen: here, 'or' is used,so not extra_param is superset, the previous two conditional are not necessary

libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
kylazhang: Generally no xml setting for this case, so only with "--xml" will need backup/restore.

libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
kylazhang: We deal with those 3 kinds of situation with same way, so it is needed.

PR #1763, opened 2018-07-30
Add holdtime test for sendkey

kylazhang:

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_sendkey.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_sendkey.cfg
@@ -19,6 +19,11 @@
                     only Linux
                     codeset = "linux"
                     sendkey = "20 24 22 46 35 57 30 48 46 28"
+                    variants:
+                        - without_holdtime:
+                        - with_holdtime:
+                            holdtime = "1000"
+                            hold_timeout = "20"
                 - os-x_name:
                     no Linux, Windows
                     codeset = "os_x"
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
@@ -9,6 +9,8 @@ from virttest.libvirt_xml import vm_xml
 
 from provider import libvirt_version
 
+from avocado.utils import wait
+
 
 def run(test, params, env):
     """
@@ -28,6 +30,7 @@ def run(test, params, env):
     keystrokes = params.get("sendkey", "")
     codeset = params.get("codeset", "")
     holdtime = params.get("holdtime", "")
+    hold_timeout = eval(params.get("hold_timeout", "1"))
     sysrq_test = ("yes" == params.get("sendkey_sysrq", "no"))
     sleep_time = int(params.get("sendkey_sleeptime", 5))
     readonly = params.get("readonly", False)
@@ -164,12 +167,10 @@ def run(test, params, env):
         if create_file is not None:
             # check if created file exist
             cmd_ls = "ls %s" % create_file
-            sec_status, sec_output = session.get_command_status_output(cmd_ls)
-            if sec_status == 0:
-                logging.info("Succeed to create file with send key")
-            else:
-                test.fail("Fail to create file with send key, Error:%s" %
-                          sec_output)
+            if not wait.wait_for(lambda: session.get_command_status_output(cmd_ls),
+                                 hold_timeout, step=5):
+                test.fail("Fail to create file with send key")
+            logging.info("Succeed to create file with send key")
         elif sysrq_test:
             # check LOG_FILE info according to different key
 


libvirt/tests/cfg/virsh_cmd/domain/virsh_sendkey.cfg
```diff
@@ -19,6 +19,12 @@
                     only Linux
                     codeset = "linux"
                     sendkey = "20 24 22 46 35 57 30 48 46 28"
+                    variants:
+                        - without_holdtime:
+                        - with_holdtime:
+                            only acl_test
```
balamuruhans: why do we test holdtime only for acl_test ? is there any specific reason

balamuruhans: Also, when keystrokes are held for 1 sec (1000ms) the no of keystrokes sent will not be in the control, so the file 
IIUC, name created and keys sent will be different and random which cause the test to fail.
so for `touch abc\n" it can be like tttooouuccch abbc as each keystokes would take 1sec as holdtime.

libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
```diff
@@ -140,11 +141,16 @@ def send_line(send_str):
 
         if create_file is not None:
             # check if created file exist
-            cmd_ls = "ls %s" % create_file
-            sec_status, sec_output = session.get_command_status_output(cmd_ls)
-            if sec_status == 0:
-                logging.info("Succeed to create file with send key")
-            else:
+            while hold_timeout > 0:
```
chunfuwen: suppose utils_misc.wait_for(function, timeout=600, step=10.0) can satisfy your request

libvirt/tests/cfg/virsh_cmd/domain/virsh_sendkey.cfg
kylazhang: I set acl_test as it do not need to double test for all as they have no difference.
For my test, sendkey will not have unpredictable situation as you show, so you test them on ppc with the random result?

libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
kylazhang: no, they are different func. Thanks

libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
```diff
@@ -140,11 +141,16 @@ def send_line(send_str):
...
+                cmd_ls = "ls %s" % create_file
+                sec_status, sec_output = session.get_command_status_output(cmd_ls)
+                if sec_status == 0:
+                    logging.info("Succeed to create file with send key")
+                    break
+                else:
+                    time.sleep(5)
+                    hold_timeout -= 5
```
balamuruhans: I am inclined to use,
```python
from avocado.utils import wait
::
if not wait.wait_for(lambda: session.get_command_status_output(cmd_ls),
                     hold_timeout):
    test.fail()
```

libvirt/tests/cfg/virsh_cmd/domain/virsh_sendkey.cfg
balamuruhans: > I set acl_test as it do not need to double test for all as they have no difference.

we can't say it as no difference, as it basically tests libvirt polkit permission for the user who executes this command

> For my test, sendkey will not have unpredictable situation as you show, so you test them on ppc with the random result?

I tried them once with `holdtime > 0` and observed the behaviour of keystrokes to be random and learnt from the man page as well,
```
:::
 If --holdtime is given, each keystroke will be held for that many milliseconds. The default codeset is linux, but use of the --codeset option allows other codesets to be chosen.

If multiple keycodes are specified, they are all sent simultaneously to the guest, and they may be received in random order. If you need distinct keypresses, you must use multiple send-key invocations. 
:::
```

kylazhang: Thanks and I saw now the test for holdtime is sendkey_simultaneous = "no" so they will be send one by one and no simultaneous issues

kylazhang: For the acl test if you think it is needed for your test, I can add it.

libvirt/tests/src/virsh_cmd/domain/virsh_sendkey.py
kylazhang: Thanks and will update soon

kylazhang:
```
Passed after updated
# avocado run --vt-type libvirt virsh.sendkey.params_test.non_acl.linux_keycode.with_holdtime
JOB ID     : f2624789661d2b1863dd526f6f2bbf5d6de92871
JOB LOG    : /root/avocado/job-results/job-2020-02-13T18.07-f262478/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.sendkey.params_test.non_acl.linux_keycode.with_holdtime: PASS (64.78 s)
JOB TIME   : 69.55 s
```

kylazhang:
@balamuruhans Also thanks for your detailed review, comments and recheck in time. I will merge it.

PR #1766, opened 2018-07-31
Add negative test for net-dhcp-leases and net-define/create

kylazhang:


balamuruhans:
@kylazhang, request to have commit msg for the negative testcases added to different files

PR #1784, opened 2018-08-09
add block job checking after executing blockcommit

chunfuwen:
If with shadow option,add additional check after blockcommit

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
@@ -459,6 +459,9 @@ def run(test, params, env):
                             if not ret:
                                 test.fail("Domain image backing "
                                           "chain check failed")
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
```diff
@@ -459,6 +459,9 @@ def check_chain_backing_files(disk_src_file, expect_backing_file=False):
                             if not ret:
                                 test.fail("Domain image backing "
                                           "chain check failed")
+                            cmd_result = virsh.blockjob(vm_name, blk_target, '',
+                                                        ignore_status=True, debug=True)
```
balamuruhans: this will print the job information of the block commit progress, we do not assert here, please help to explain the purpose of having it here to check

chunfuwen: the check is specific to shadow option, where ever one bug raised when do virsh blockjob

PR #1786, opened 2018-08-10
add qemu-img info validation after block resize

chunfuwen:
it need validate the size returned by qemu info is equals to primitive expected size

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
@@ -122,6 +122,12 @@ def run(test, params, env):
         elif resize_value[-1] == "g":
             value = int(resize_value[:-1])
             expected_size = value * 1024 * 1024 * 1024
+            cmd = "qemu-img info %s" % image_path
+            ret = process.run(cmd, allow_output_check='combined', shell=True)
+            status, output = (ret.exit_status, ret.stdout_text.strip())
+            value_return_by_qemu_img = re.search(r'virtual size:\s+(\d+(\.\d+)?)+G', output).group(1)
+            if value != int(float(value_return_by_qemu_img)):
+                test.fail("initial image size in config is not equals to value returned by qemu-img info")
         else:
             test.error("Unknown scale value")
 


chunfuwen:
paste result here
```
PASS 1-type_specific.io-github-autotest-libvirt.virsh.blockresize.normal_test.gigabyte.qcow2_type
```

libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
```diff
@@ -122,6 +122,12 @@ def run(test, params, env):
         elif resize_value[-1] == "g":
             value = int(resize_value[:-1])
             expected_size = value * 1024 * 1024 * 1024
+            cmd = "qemu-img info %s" % image_path
+            ret = process.run(cmd, allow_output_check='combined', shell=True)
+            status, output = (ret.exit_status, ret.stdout_text.strip())
```
balamuruhans: using `getstatusoutput` will be more appropriate
```python
status, output = process.getstatusoutput(cmd)
```

chunfuwen: process.run enable redirect std.out and std.err to together.

balamuruhans: getstatusoutput() still uses process.run and we can pass the `allow_output_check` option to enable std.out and std.err

libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
```diff
@@ -122,6 +122,12 @@ def run(test, params, env):
...
+            value_return_by_qemu_img = re.search(r'virtual size:\s+(\d+\.\d*)+G', output).group(1)
```
hahakiki2010: If the virtual size might return integer and float both, then the regex match should be like ```(\d+\.?\d*)+G'```

libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
chunfuwen: getstatusoutput() is added later for compatability purpose between python2 and 3 ,and directly copied from previous python2 command module. process.run is more native ,I assume.

libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
chunfuwen: updated regexp to virtual size:\s+(\d+(\.\d+)?)+G

libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
balamuruhans: okay, we can reuse API `get_image_info()` from utils_misc.py and your fix if required in API could help ?

PR #1788, opened 2018-08-13
The error message check should be applicable for all the scenario's of "iface_attach".

santwana:
The error message check should be applicable for all the scenario's of "iface_attach".

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_hotplug.py
+++ b/libvirt/tests/src/virtual_network/iface_hotplug.py
@@ -139,7 +139,7 @@ def run(test, params, env):
                         if not ("--current" in sep_options and len(sep_options) > 1):
                             test.fail("return mutualy exclusive, but it is unexpected")
                     else:
-                        logging.error("Command output %s" % ret.stderr.strip())
+                        test.fail("Failed to attach-interface: %s" % ret.stderr.strip())
                 elif stress_test:
                     if attach_device:
                         # Detach the device immediately for stress test
--- a/libvirt/tests/src/virtual_network/iface_hotplug.py
+++ b/libvirt/tests/src/virtual_network/iface_hotplug.py
@@ -139,8 +139,7 @@ def run(test, params, env):
                         if not ("--current" in sep_options and len(sep_options) > 1):
                             test.fail("return mutualy exclusive, but it is unexpected")
                     else:
-                        logging.error("Command output %s" % ret.stdout.strip())
-                        test.fail("Failed to attach-interface")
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -139,7 +139,7 @@ def create_iface_xml(mac):
                         if not ("--current" in sep_options and len(sep_options) > 1):
                             test.fail("return mutualy exclusive, but it is unexpected")
                     else:
-                        logging.error("Command output %s" % ret.stdout.strip())
+                        libvirt.check_exit_status(ret)
                         test.fail("Failed to attach-interface")
```
balamuruhans: `libvirt.check_exit_status()` will take care of raising the test failure, so it can be removed.

santwana: @balamuruhans Addressed the changes, kindly review, Thanks!!

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -139,8 +139,7 @@ def create_iface_xml(mac):
                         if not ("--current" in sep_options and len(sep_options) > 1):
                             test.fail("return mutualy exclusive, but it is unexpected")
                     else:
-                        logging.error("Command output %s" % ret.stdout.strip())
-                        test.fail("Failed to attach-interface")
+                        libvirt.check_exit_status(ret)
```
balamuruhans: I would suggest to use `libvirt.check_result()` to give expected fails or skip regex which can refactor you code from L127 to L142

santwana: @balamuruhans I modified the existing, ret.stdout.strip() to ret.stderr.strip(), since it wasn't returning the command output. Now it solves the purpose. We can skip the libvirt API. Please review.

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -139,8 +139,7 @@ def create_iface_xml(mac):
                         if not ("--current" in sep_options and len(sep_options) > 1):
                             test.fail("return mutualy exclusive, but it is unexpected")
                     else:
-                        logging.error("Command output %s" % ret.stdout.strip())
-                        test.fail("Failed to attach-interface")
+                        logging.error("Command output %s" % ret.stderr.strip())
```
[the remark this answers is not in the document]
santwana: @balamuruhans had missed the failure part, thanks , now have incorporated the changes. Kindly review.

PR #1792, opened 2018-08-14
new case :verify the rpm pacakge after install

juteman:
verify package libvirt-daemon-config-nwfilter

Signed-off-by: qiulzhan <qiulzhan@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/nwfilter/verify_package.cfg
+++ b/libvirt/tests/cfg/nwfilter/verify_package.cfg
@@ -1,0 +1,8 @@
+- verify_nwfilter_package:
+    type = verify_nwfilter_package
+    start_vm = "no"
+    kill_vm = "no"
+    status_error = "no"
+    variants:
+        - libvirt_daemon_config_nwfilter:
+            check_cmd = "rpm -V libvirt-daemon-config-nwfilter"
--- a/libvirt/tests/src/nwfilter/verify_nwfilter_package.py
+++ b/libvirt/tests/src/nwfilter/verify_nwfilter_package.py
@@ -1,0 +1,24 @@
+import logging
+
+from avocado.utils import process
+
+from virttest.compat_52lts import decode_to_text as to_text
+
+
+def run(test, params, env):
+    """
+    verify the rpm package after install
+    1) run test cmd
+    2) check result
+    """
+    check_cmd = params.get("check_cmd")
+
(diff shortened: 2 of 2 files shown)


chloerh:
Looks good to me

libvirt/tests/src/nwfilter/verify_nwfilter_package.py
```diff
@@ -0,0 +1,24 @@
+import logging
+
+from avocado.utils import process
+
+from virttest.compat_52lts import decode_to_text as to_text
+
+
+def run(test, params, env):
+    """
+    verify the rpm package after install
```
balamuruhans: How does this testcase to be used ?
if the package is not installed, the testcase would fail why dont we try installing the package and then check if it is installed ?

libvirt/tests/src/nwfilter/verify_nwfilter_package.py
```diff
@@ -0,0 +1,24 @@
...
+    1) run test cmd
+    2) check result
+    """
+    check_cmd = params.get("check_cmd")
+
+    try:
+        out = to_text(process.system_output(check_cmd,
```
balamuruhans: `virttest.utils_package()` can be used to try installing and check for the package.

libvirt/tests/cfg/nwfilter/verify_package.cfg
```diff
@@ -0,0 +1,8 @@
+- verify_nwfilter_package:
+    type = verify_nwfilter_package
+    start_vm = "no"
+    kill_vm = "no"
+    status_error = "no"
+    variants:
+        - libvirt_daemon_config_nwfilter:
+            check_cmd = "rpm -V libvirt-daemon-config-nwfilter"
```
balamuruhans: package name can be used,
```
    variants:
        - libvirt_daemon_config_nwfilter:
            package_name = "libvirt-daemon-config-nwfilter"
         - package_name2:
            package_name = "package_name2"
```

libvirt/tests/src/nwfilter/verify_nwfilter_package.py
juteman: This case just check the package after install.The case test that package has been installed , and  use rpm -V to check package and should validate successfully and no output. If get something output may be a bug here

PR #1793, opened 2018-08-14
Drop the package virttest.staging.backports

luckyh:
Since legacy python (2.4-2.6) is no longer supported by avocado-vt,
now it is the time to drop the package `virttest.staging.backports`
as it is used for backward compatibility.

Signed-off-by: Xu Han <xuhan@redhat.com>


luckyh:
@kylazhang @balamuruhans thank you!

PR #1801, opened 2018-08-21
virsh_migrate_copy_storage: Add tls option tests

qiankehan:
Since libvirt-4.4, --copy-storage-{all,inc} can work together with
--tls. Add variants for tls.

Libvirt bug reference: https://bugzilla.redhat.com/show_bug.cgi?id=1300772

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate_copy_storage.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate_copy_storage.cfg
@@ -6,6 +6,10 @@
     precreation_pool_type = "dir"
     precreation_pool_name = "precreation_pool"
     variants:
+        - with_tls:
+            tls_option = "--tls"
+        - without_tls:
+    variants:
         - file_image:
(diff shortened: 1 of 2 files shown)


balamuruhans:
@qiankehan copy-storage migration is deprecated, do we continue testing it ?
https://access.redhat.com/solutions/60034

qiankehan:
@balamuruhans Sorry for late reply. The deprecated feature mentioned in https://access.redhat.com/solutions/60034 is non-shared storage without storage copy.
Migration with **--copy-storage-{all,inc}** is the migration with storage copying via NBD and drive_mirror. That is supported in libvirt and is planned for a future release of RHEV as that article said. :)

dzhengfy:
@qiankehan It is better to paste your local test results.

qiankehan:
@dzhengfy  Could you please add this PR to CI and run it first? I cannot test it by local env.

dzhengfy:
@qiankehan could you add --tls section in the first place of variant, otherwise all cases' name have to be changed.

PR #1807, opened 2018-08-27
add detach disk with print-xml case

chunfuwen:
when detach disk with print-xml, it print xml info only, not actual disk detachment

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
@@ -140,6 +140,7 @@ def run(test, params, env):
     test_block_dev = "yes" == params.get("at_dt_disk_iscsi_device", "no")
     test_logcial_dev = "yes" == params.get("at_dt_disk_logical_device", "no")
     restart_libvirtd = "yes" == params.get("at_dt_disk_restart_libvirtd", "no")
+    detach_disk_with_print_xml = "yes" == params.get("detach_disk_with_print_xml", "no")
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
```diff
@@ -285,6 +286,14 @@ def acpiphp_module_modprobe(vm, os_type):
         status = virsh.attach_disk(vm_ref, device_source, device_target,
                                    at_options, debug=True).exit_status
     elif test_cmd == "detach-disk":
+        # For detach disk with print-xml option, it only print information,and not actual disk detachment.
+        if detach_disk_with_print_xml and libvirt_version.version_compare(4, 5, 0):
+            ret = virsh.detach_disk(vm_ref, device_target, at_options)
+            libvirt.check_exit_status(ret)
+            cmd = ("echo \"%s\" | grep -A 16 %s"
+                   % (ret.stdout.strip(), device_source_name))
```
hahakiki2010: except check for the cmd output, suggest also to add the disk check to make sure it is not detached actually

kylazhang: she means need to check the disk status, but not only the stdout

chunfuwen: actually line 297 validate it further since  virsh.detach_disk will fail if previous operation is wrong

kylazhang:
@hahakiki2010 also checked without more concern. So merge it now.

PR #1818, opened 2018-09-06
add nowait option for virsh domstats commands

chunfuwen:
nowait option enable domstats result can returned quickly although
the other thread may concurrently operate on the same virtual machine

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
@@ -3,9 +3,28 @@ import logging
 from aexpect import ShellTimeoutError
 from aexpect import ShellProcessTerminatedError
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/monitor/virsh_domstats.py
```diff
@@ -147,6 +212,8 @@ def run(test, params, env):
     enforce_command = "yes" == params.get("enforce_command", "no")
     status_error = (params.get("status_error", "no") == "yes")
 
+    if "--nowait" in domstats_option and not libvirt_version.version_compare(4, 5, 0):
+        test.skip("--nowait option is supported until libvirt 4.5.0 version")
```
kylazhang: test.cancel

chunfuwen: fixed

PR #1824, opened 2018-09-10
Modify the regular expression in line 235 of virsh_boot.cfg

buaaweaker:
Signed-off-by: buaaweaker <2923040052@qq.com>

Because in the line 235 of file tp-libvirt/libvirt/tests/cfg/bios/virsh_boot.cfg, the outputs are more than one
sort, so we should modify the regression formula so that we can deal with more kinds of outputs.

### DIFF
--- a/libvirt/tests/cfg/bios/virsh_boot.cfg
+++ b/libvirt/tests/cfg/bios/virsh_boot.cfg
@@ -232,7 +232,7 @@
                     variants:
                         - not_existing_loader:
                             loader = "noexist"
-                            checkpoint = "error: Path.+is not accessible: No such file or directory"
+                            checkpoint = "error:(.|\n)*: No such file or directory"
                         - not_existing_loader_type:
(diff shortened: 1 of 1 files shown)


libvirt/tests/cfg/bios/virsh_boot.cfg
```diff
@@ -232,7 +232,7 @@
                     variants:
                         - not_existing_loader:
                             loader = "noexist"
-                            checkpoint = "error: Path.+is not accessible: No such file or directory"
+                            checkpoint = "error:(.|\n)*: No such file or directory"
```
hahakiki2010: no need to match ```\n``` here, it's not usual to have \n in the cmd output. ```error:(.*):\s+No such file or directory``` is enough

hahakiki2010:
Pls update the commit message, normally we explain what and why instead of how when writing a commit message

dzhengfy:

PR #1826, opened 2018-09-12
Add func for set nwfilter env

juteman:
the autocase about nwfilter which set the init xml
can be set by a utilities function

Signed-off-by: qiulzhan <qiulzhan@redhat.com>

### DIFF
--- a/libvirt/tests/src/nwfilter/utilities.py
+++ b/libvirt/tests/src/nwfilter/utilities.py
@@ -1,0 +1,28 @@
+import logging
+
+
+def set_nwfilter_iface(new_iface,
+                       type_name="network",
+                       source={'network': "default"},
+                       filterref_dict={}):
+    """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/nwfilter/utilities.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Params new_iface: instance of Interface,
+    Params type_name: the type name which bind to interface
+    Params source: source of new_iface, set as default
+    Params filterref_dict: dict of network filter, which is not
+    blank will bind to new_iface
+
+    return:  new network interface
+    """
+    # set iface type name as network and binding source type
```
chunfuwen: capitalize the first letter

chunfuwen:
suggest change file name to nwfilter_utility.py to avoid overlapping name with other utilities.py

PR #1829, opened 2018-09-14
RFC: Demonstrate benefit of config re-order through cpu hotplug tests

sathnaga:
With reference to this mail, https://www.redhat.com/archives/avocado-devel/2018-May/msg00011.html
I have come up with cpuhotplug testcase which is copy of
libvirt_vcpu_plug_unplug testcase with changes to incorporate
the features described in mail.

Here is what I have changed:
1. Re-used global config params for vcpus and other params used in test.
2. Removed xml changes inside the test as we define,boot
the guest at every subtest in the way we want.
3. Dependancy it needs subtests.cfg to be placed at last in tests-shared.cfg
as mentioned in the above mail for this to work.
4. Added new params in test config to define and delete guest as part of test.

What is the benefit for us:
1. Reducing unwanted multiple params to represent same vm attributes e:g:- vcpus (simplicity)
2. We can reuse the library function not with much changes in other tests (modularity and reduce maintenance overhead)
3. Test execution time is reduced by around ~66%

libvirt_vcpu_plug_unplug.positive_test.vcpu_set.live.vm_operate.no_operation: PASS (128.45 s)
libvirt_vcpu_plug_unplug: JOB TIME   : 11356.14 s

cpuhotplug.positive_test.vcpu_set.live.vm_operate.no_operation: PASS (87.46 s)
cpuhotplug: JOB TIME   : 3894.54 s

I have not faced any other obvious framework issue with this change.
I agree this will require quite a change to our old testcases, but given
the benefit its worth doing it :-), we can discuss further on this and
we can get to common grounds.

### DIFF
--- a/libvirt/tests/cfg/cpu/cpuhotplug.cfg
+++ b/libvirt/tests/cfg/cpu/cpuhotplug.cfg
@@ -1,0 +1,114 @@
+- cpuhotplug:
+    type = cpuhotplug
+    start_vm = yes
+    create_vm_libvirt = yes
+    kill_vm_libvirt =  yes
+    virtinstall_extra_args = " --channel name='org.qemu.guest_agent.0',mode=bind,target_type=virtio,char_type=unix"
+    cores = 4
+    sockets = 1
+    smp = 2
+    threads = 1
+    vcpu_maxcpus = 4
+    mem = 8192
+    env_cleanup = yes
+    vcpu_plug = "yes"
+    vcpu_unplug = "yes"
+    variants:
+        - positive_test:
+            status_error = "no"
+            variants:
+                - vcpu_set:
+                    variants:
+                        - vm_operate:
+                            variants:
+                                - no_operation:
+                                - save:
+                                    vm_operation = "save"
+                                    vcpu_unplug = "no"
+                                - managedsave:
+                                    vm_operation = "managedsave"
+                                    vcpu_unplug = "no"
+                                - suspend:
+                                    vm_operation = "suspend"
+                                    vcpu_unplug = "no"
+                                - suspend_to_mem:
+                                    no ppc64le
+                                    vm_operation = "s3"
+                                    install_qemuga = "yes"
+                                - suspend_to_disk:
+                                    no ppc64le
+                                    vm_operation = "s4"
+                                    install_qemuga = "yes"
+                                - save_with_unplug:
+                                    vm_operation = "save"
+                                - managedsave_with_unplug:
+                                    vm_operation = "managedsave"
+                                - suspend_with_unplug:
+                                    vm_operation = "suspend"
+                                - reboot:
+                                    vm_operation = "reboot"
+                        - libvirtd_restart:
+                                restart_libvirtd = "yes"
+                        - vcpu_pin:
+                            only live
+                            variants:
+                                - pin_plug_unplug:
+                                    pin_before_plug = "yes"
+                                    pin_vcpu = "0"
+                                    pin_cpu_list = "x,y"
+                                - pin_unplug:
+                                    vcpu_plug = "no"
+                                    pin_before_unplug = "yes"
+                                    pin_vcpu = "2"
+                                    pin_cpu_list = "x-y,^z"
+                                - plug_pin:
+                                    vcpu_unplug = "no"
+                                    pin_after_plug = "yes"
+                                    pin_vcpu = "2"
+                                    pin_cpu_list = "x-y"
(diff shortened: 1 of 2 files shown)


sathnaga:
@ldoktor @luckyh @clebergnu @balamuruhans @lmr @hereischen @chunfuwen needs your inputs...

ldoktor:
@sathnaga I don't have a strong opinion here as far as everything works and doesn't complicate things too much. Cartesian is fragile, but scaleable so reorganization can bring some benefits, but also can require some tweaks that could result in quite the opposite. Anyway if you have enough time to play with it and gain some useful flexibility in subtests, go ahead.

Anyway I'm not hardcore-user so @luckyh, would you please provide a statement for RH?

ldoktor:
Btw I'd suggest adding `cartesian_config.py -c` selftest with generated outputs using the current setup with several `filter-only` and `filter-outs` to limit the amount of variants (and time to execute the selftest). Ideally after the change this test would pass without any modification (although there might be some understandable little changes, but at least we'd get an overview of what changed)

chunfuwen:
@sathnaga  I am open to reorder the config file order since subtest.cfg is specific parameter to specific test case, and it is also reasonable to be put in last place where it may override some global variable config.

luckyh:
Well it is a nice improvement, by doing this, we can abandon some bad structure of the params description. However, the idea is we should do this carefully since there are many params referenced across config files, tweaking the order may break the correctness of value assignment.

I wrote a script to help give a rough check.
```bash
#!/bin/bash
OUTDIR="./result"
mkdir -p $OUTDIR

CFG_LIST="$(sed 's/.*subtests //' <<< $(grep -E "^include" tests-shared.cfg | sed -e 's/^include\s\+//' -e 's/\.cfg$//'))"
for target in $CFG_LIST; do
    grep -oE "^\s+[A-Za-z0-9][A-Za-z0-9_-]*\s*\??[<+]?=" ${target}.cfg | sed -e 's/\s\+//' -e 's/[ ?<+=]\+$//' | sort -u > $OUTDIR/${target}-params
    while read word; do
        if grep -qE "\b$word\b\s*\??[<+]?=" subtests.cfg; then
            grep -nE "\b$word\b\s*\??[<+]?=" subtests.cfg ${target}.cfg | sed -e 's/\s\+/\t/' -e "s/^/[$word]\t/"
        fi
    done < $OUTDIR/${target}-params > $OUTDIR/${target}-vs-subtests
done

echo "Total defined params:"
wc -l $OUTDIR/*-params
echo
echo "Total reference count of shared params:"
wc -l $OUTDIR/*-vs-*
```
here is the result of tp-libvirt (running the script under `/var/lib/avocado/data/avocado-vt/backends/libvirt/cfg`):
```
Total defined params:
   1 ./result/cdkeys-params
   5 ./result/drivers-params
  77 ./result/guest-hw-params
 257 ./result/guest-os-params
   0 ./result/host-os-params
  18 ./result/machines-params
  10 ./result/virtio-win-params
 368 total

Total reference count of shared params:
    0 ./result/cdkeys-vs-subtests
  148 ./result/drivers-vs-subtests
  188 ./result/guest-hw-vs-subtests
  212 ./result/guest-os-vs-subtests
    0 ./result/host-os-vs-subtests
  141 ./result/machines-vs-subtests
    0 ./result/virtio-win-vs-subtests
  689 total
```
So, once we decide to have the improvement right now, we'd better to ensure that there is nothing broken by this change, you can use the method provided by @ldoktor .

Btw, the result looks much better than tp-qemu's.
```
Total defined params:
   1 ./result/cdkeys-params
  77 ./result/guest-hw-params
 257 ./result/guest-os-params
   0 ./result/host-os-params
  10 ./result/virtio-win-params
 345 total

Total reference count of shared params:
     0 ./result/cdkeys-vs-subtests
   333 ./result/guest-hw-vs-subtests
  3270 ./result/guest-os-vs-subtests
     0 ./result/host-os-vs-subtests
    37 ./result/virtio-win-vs-subtests
  3640 total
```
So, (at least to me) having this improvement for tp-qemu is a long term work.

sathnaga:
> Btw I'd suggest adding `cartesian_config.py -c` selftest with generated outputs using the current setup with several `filter-only` and `filter-outs` to limit the amount of variants (and time to execute the selftest). Ideally after the change this test would pass without any modification (although there might be some understandable little changes, but at least we'd get an overview of what changed)

@ldoktor I would have to explore on this selftest part, will do it..

dzhengfy:

PR #1830, opened 2018-09-14
Fix cleanup vm problem and update converted OS

xiaodwan:
1) Fix vm is not cleaned up if start failed
2) Update windows version list
3) Update converted OS for multi system


xiaodwan:
@chloerh @chunfuwen Please help to review it, thanks!

PR #1835, opened 2018-09-20
Rename case *.no_create_target_image.*

fangge1212:
The original name "no_create_target_image" for the negative case
in live storage migration is not proper any more and is ambiguous,
as libvirt has been able to create target image automatically as long 
as storage pool exists on target host.


fangge1212:
This is minor change

fangge1212:
I moved commit 2&&3 to another pull request

PR #1839, opened 2018-09-21
Fixed the distro specific issues in "nwfilter_update_vm_running".

santwana:
The test-case was failing in Ubuntu for package installation and
error handling.

### DIFF
--- a/libvirt/tests/cfg/nwfilter/nwfilter_update_vm_running.cfg
+++ b/libvirt/tests/cfg/nwfilter/nwfilter_update_vm_running.cfg
@@ -10,4 +10,4 @@
             check_cmd = "ebtables -t nat -L O-DEVNAME-arp"
             expect_match = "-j DROP"
             check_vm_cmd = "arping -c 3 192.168.122.1"
-            vm_expect_match = "Received\s0\sresponse"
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/nwfilter/nwfilter_update_vm_running.py
```diff
@@ -1,10 +1,11 @@
 import re
 import logging
-
+import ast
```
balamuruhans: I think this change would not be necessary.

libvirt/tests/cfg/nwfilter/nwfilter_update_vm_running.cfg
```diff
@@ -10,4 +10,4 @@
             check_cmd = "ebtables -t nat -L O-DEVNAME-arp"
             expect_match = "-j DROP"
             check_vm_cmd = "arping -c 3 192.168.122.1"
-            vm_expect_match = "Received\s0\sresponse"
+            vm_expect_match = {"default": "Received\s0\sresponse", "ubuntu": "Timeout"}
```
balamuruhans: why do we assign `Timeout` for Ubuntu in vm_expect_match ? instead this change can be removed. we can always override the param value during run.

dzhengfy:

PR #1847, opened 2018-09-24
Rename case *.no_create_target_image.*

fangge1212:
The original name "no_create_target_image" for the negative case
in live storage migration is not proper any more and is ambiguous,
as libvirt has been able to create target image automatically as long
as storage pool exists on target host.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_vm.cfg
+++ b/libvirt/tests/cfg/migration/migrate_vm.cfg
@@ -677,8 +677,9 @@
                     target_pool_name = "precreation_pool"
                     target_pool_type = "dir"
                     variants:
-                        - no_create_target_image:
+                        - create_neither_target_pool_nor_image:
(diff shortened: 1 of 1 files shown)


fangge1212:
This is minor change

libvirt/tests/cfg/migration/migrate_vm.cfg
```diff
@@ -677,8 +677,9 @@
                     target_pool_name = "precreation_pool"
                     target_pool_type = "dir"
                     variants:
-                        - no_create_target_image:
+                        - no_create_target_pool_nor_image:
```
chunfuwen: update no_create_target_pool_nor_image to neither_create_target_pool_nor_image ?

fangge1212: How about create_neither_target_pool_nor_image?

PR #1852, opened 2018-09-27
virsh_change_media: Minor fix

chloerh:
In case 'disk_alias' referenced before assignment


chunfuwen:
minor fix

PR #1853, opened 2018-09-27
Add variable unsupported_conf

fangge1212:
Test should not just cancel when "unsupported conf" error happens 
during migration, because it is expected for some test cases. Add
a variable to distinguish when error "unsupported conf" is expected


chunfuwen:
minor change

PR #1855, opened 2018-09-28
Add test for rom enable updating

kylazhang:
Add a new attribute in rom element "enabled" and it can be "yes" or "no", default if no set is "yes". Set to "no" to disable network boot.
1. Test for xml and related combination with <rom enabled='no'>, <rom enabled='yes'>
2. Test for update-device, hotplug, define, start

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
@@ -12,6 +12,28 @@
             err_msgs1 = "domain is not running"
             err_msgs2 = "are mutually exclusive"
             variants:
+                - rom_bar_off:
+                    only at_device
+                    iface_rom = "{'bar':'off'}"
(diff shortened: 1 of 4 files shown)


libvirt/tests/cfg/virtual_network/iface_hotplug.cfg
```diff
@@ -12,6 +12,28 @@
             err_msgs1 = "domain is not running"
             err_msgs2 = "are mutually exclusive"
             variants:
+                - rom_bar_off:
+                    only at_device
```
chunfuwen: may recontruct  config to make it more layered
--rom
       variants:
                 --enable
                 --disable
       variables:
                --bar
                --off-bar

kylazhang: Can not use Cartesian format as they all have different configuration

chloerh:
LGTM

PR #1859, opened 2018-09-30
Changed the timeout for all events test

iccaszhulili:
Since the timeout is not enough for all events test
in RHEL8 tests, increase the timeout


libvirt/tests/cfg/virsh_cmd/domain/virsh_event.cfg
```diff
@@ -12,6 +12,7 @@
                 - no_timeout:
                 - timeout:
                     event_timeout = "60"
+                    all_events_timeout = "180"
```
sathnaga: `all_events_timeout` looks redundant, `event_timeout` can be used to increase the timeout based on Host filter for different distro....

PR #1861, opened 2018-10-05
virsh_migrate_stress: Add Transparent Hugepage enabled stress migration

balamuruhans:
Enable Transparent Hugepages with enabled, never and madvise to trigger `stress-ng`
commands accordingly to exercise THP in guest/host/remote in the testcase. For
host and remote host THP stress, we align guest boot memory with Huge Page aligned.
If the hugepages are different in host and remote host then we make sure guest boot
memory is aligned to both hugepages. For the `stress-ng` --vm-bytes argument is also
taken care to be aligned with respective hugepage size of host/remote host/guest.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
@@ -1,12 +1,20 @@
 import logging
+import os
+from fractions import gcd
 
 from virttest import libvirt_vm
 from virttest import utils_test
 from virttest import utils_misc
+from virttest import kernel_interface
 from virttest import utils_package
-from virttest import remote
+from virttest import test_setup
+from virttest import utils_numeric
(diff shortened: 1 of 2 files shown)


balamuruhans:
Depends on:
utils_memory - https://github.com/avocado-framework/avocado-vt/pull/1777
sysfs and procfs APIs - https://github.com/avocado-framework/avocado-vt/pull/1729

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_stress.py
```diff
@@ -26,6 +33,29 @@ def set_cpu_memory(vm_name, cpu, memory):
     vmxml.sync()
 
 
+def thp_enable(test, params, sysfs_path, value=None, session=None, name="host"):
```
chunfuwen: method name rule:verb+n.

dzhengfy:

PR #1863, opened 2018-10-10
Skip luks encryption vol clone for disk pool type

bssrikanth:
Currently we do not have support for luks encryption for disk pool.
This patch excludes this test scenario as we cover non encrypted
in non_encrypt.[acl_test/non_acl].disk_part.pool_type.disk

Signed-off-by: Srikanth Aithal <sraithal@linux.vnet.ibm.com>


bssrikanth:
discussion on supporting same are pursued at https://bugzilla.redhat.com/show_bug.cgi?id=1560946

bssrikanth:
Just saw that backend is implemented with libvirt-4.4.0-2. I am in process of verifying same. If that's the case I might have to send PR for testcase to include disk type for encryption. Anyway will close this PR for time being, will reopen/send new PR soon based in finding/testing

PR #1865, opened 2018-10-19
migrate_options_shared: Add case for htm migration

dzhengfy:
Libvirt supports <htm state='xxx'/> feature on PPC now. So this is to
test the migration with htm feature setting.

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -69,15 +69,18 @@ def run(test, params, env):
     Test virsh migrate command.
     """
 
-    def set_hpt(resizing, vmxml):
+    def set_feature(vmxml, feature, value):
         """
-        Set the hpt feature for PPC
+        Set guest features for PPC
 
-        :param resizing: The value needed
+        :param state: the htm status
         :param vmxml: guest xml
         """
         features_xml = vm_xml.VMFeaturesXML()
-        features_xml.hpt_resizing = resizing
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -355,11 +367,12 @@ def calculate(num):
                           'ssh_remote_auth': True}
 
     hpt_resize = params.get("hpt_resize", None)
+    htm_state = params.get("htm_state", None)
     qemu_check = params.get("qemu_check", None)
     xml_check_after_mig = params.get("guest_xml_check_after_mig", None)
 
     arch = platform.machine()
-    if (hpt_resize or contrl_index) and 'ppc64' not in arch:
+    if (hpt_resize or contrl_index or htm_state) and 'ppc64' not in arch:
```
chunfuwen: suggest use if any to wrap multiple boolean value

dzhengfy: updated

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -440,6 +455,8 @@ def calculate(num):
             check_content = qemu_check
             if hpt_resize:
                 check_content = "%s%s" % (qemu_check, hpt_resize)
+            if htm_state:
```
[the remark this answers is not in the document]
dzhengfy: update

PR #1866, opened 2018-10-19
Add hotplug/unplug test for rng_device for backend rdm,tcp,udp

kylazhang:

### DIFF
--- a/libvirt/tests/src/libvirt_rng.py
+++ b/libvirt/tests/src/libvirt_rng.py
@@ -4,17 +4,20 @@ import ast
 import shutil
 import logging
 import uuid
-import time
+import aexpect
 
 from six.moves import xrange
 
 from avocado.utils import process
+
 from virttest import virt_vm, virsh
 from virttest import utils_package
 from virttest import utils_misc
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml import xcepts
 from virttest.libvirt_xml.devices import rng
+
 from provider import libvirt_version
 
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/libvirt_rng.py
```diff
@@ -61,6 +107,8 @@ def modify_rng_xml(dparams, sync=True):
         rng_xml.backend = backend
 
         logging.debug("Rng xml: %s", rng_xml)
+        if get_xml:
+            return rng_xml
```
chloerh: If 'get_xml' is True, 'sync' could still be True. Maybe line:110-111 should be moved after the 'sync' part

libvirt/tests/src/libvirt_rng.py
```diff
@@ -29,9 +30,54 @@ def run(test, params, env):
...
+            xml_get = xml_devices[rng_index]
+
+            if not exists:
+                test.fail("Rng device still exists after detach!")
+        except IndexError:
+            if exists:
+                test.fail("Can not find rng device in xml")
+            else:
+                logging.info("Can not find rng xml as expected")
+                return
```
chunfuwen: no need

libvirt/tests/src/libvirt_rng.py
```diff
@@ -29,9 +30,54 @@ def run(test, params, env):
...
+
+        match = 1
```
chunfuwen: match = True?

libvirt/tests/src/libvirt_rng.py
kylazhang: Here the logic is when get_xml is True, will not sync to guest. I have already update the function docstring.

libvirt/tests/src/libvirt_rng.py
kylazhang: fixed, thanks

PR #1868, opened 2018-10-22
Add test for shared physical network

hahakiki2010:
Create a new bridge with a physical nic,
and connect it to multiple vms, test if
the vms, host, and remote ip can ping
each other succeed

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
@@ -1,0 +1,17 @@
+- virtual_network.iface_bridge:
+    type = iface_bridge
+    start_vm = no
+    vms = avocado-vt-vm1 vm2
+    bridge_name = "test_br0"
+    netdst_nic1 = ${bridge_name}
+    # Replace remote_ip by a actual IP address
+    remote_ip = "www.baidu.com"
+    ping_count = 3
+    pint_timeout = 10
+    filter_uuid = "11111111-b071-6127-b4ec-111111111111"
+    variants:
+        - shared_physical_network:
+
+
+
+
--- a/libvirt/tests/src/virtual_network/iface_bridge.py
+++ b/libvirt/tests/src/virtual_network/iface_bridge.py
@@ -1,0 +1,208 @@
+import os
+import logging
+
+from avocado.utils import process
+
+from virttest import utils_net
+from virttest import data_dir
+from virttest import virsh
+from virttest.utils_test import libvirt
+from virttest.staging import service
+from virttest.libvirt_xml import vm_xml
+from virttest import utils_package
+
+NETWORK_SCRIPT = "/etc/sysconfig/network-scripts/ifcfg-"
+
+
+def run(test, params, env):
+    """
+    Test bridge support from network
+
+    1) create a linux bridge and connect a physical interface to it
+    2) define nwfilter with "vdsm-no-mac-spoofing"
+    3) redefine the vm with the new create bridge and filter
+    4) check if guest can get public ip after vm start
+    5) check if guest and host can ping each other
+    6) check if guest and host can ping outside
+    7) start another vm connected to the same bridge
+    8) check if the 2 guests can ping each other
+    """
+
+    def create_bridge(br_name, iface_name):
+        """
+        Create a linux bridge by virsh cmd:
+        1. Stop NetworkManager and Start network service
+        2. virsh iface-bridge <iface> <name> [--no-stp]
+
+        :param br_name: bridge name
+        :param iface_name: physical interface name
+        :return: bridge created or raise exception
+        """
+        # Make sure the bridge not exist
+        if libvirt.check_iface(br_name, "exists", "--all"):
+            test.cancel("The bridge %s already exist" % br_name)
+
+        # Create bridge
+        # This cmd run a long time, so set timeout=240
+        result = virsh.iface_bridge(iface_name, br_name, "--no-stp", debug=True, timeout=240)
+        if result.exit_status:
+            test.fail("Failed to create bridge:\n%s" % result.stderr)
+
+    def define_nwfilter(filter_name):
+        """
+        Define nwfilter vdsm-no-mac-spoofing with content like:
+        <filter name='vdsm-no-mac-spoofing' chain='root'>
+            <filterref filter='no-mac-spoofing'/>
+            <filterref filter='no-arp-mac-spoofing'/>
+        </filter>
+
+        :param filter_name: the name of nwfilter
+        :return: filter created or raise exception
+        """
+        filter_uuid = params.get("filter_uuid", "11111111-b071-6127-b4ec-111111111111")
+        filter_params = {"filter_name": "vdsm-no-mac-spoofing",
+                         "filter_chain": "root",
+                         "filter_uuid": filter_uuid,
+                         "filterref_name_1": "no-mac-spoofing",
+                         "filterref_name_2": "no-arp-mac-spoofing"}
+        filter_xml = libvirt.create_nwfilter_xml(filter_params).xml
+        # Run command
+        result = virsh.nwfilter_define(filter_xml, ignore_status=True, debug=True)
+        if result.exit_status:
+            test.fail("Failed to define nwfilter with %s" % filter_xml)
+
+    def modify_iface_xml(br_name, nwfilter, vm_name):
+        """
+        Modify interface xml with the new bridge and the nwfilter
+
+        :param br_name: bridge name
+        :param nwfilter: nwfilter name
+        :param vm_name: vm name
+        """
+        vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name)
+        iface_xml = vmxml.get_devices('interface')[0]
+        vmxml.del_device(iface_xml)
+        iface_xml.source = {'bridge': br_name}
+        iface_xml.filterref = iface_xml.new_filterref(name=nwfilter)
+        logging.debug("new interface xml is: %s" % iface_xml)
+        vmxml.add_device(iface_xml)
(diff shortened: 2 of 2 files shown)


hahakiki2010:
```
(virtenv_py3) [root@localhost libvirt_ci]# avocado run --vt-type libvirt iface_bridge.shared_physical_network 
JOB ID     : 60afe8c1ddf9fd4d1dd9a061ea0a8ebfede699ec
JOB LOG    : /root/avocado/job-results/job-2018-10-22T16.49-60afe8c/job.log
 (1/1) type_specific.io-github-autotest-libvirt.iface_bridge.shared_physical_network: PASS (73.90 s)
JOB TIME   : 96.11 s
```

libvirt/tests/cfg/virtual_network/iface_bridge.cfg
```diff
@@ -0,0 +1,18 @@
+- virtual_network.iface_bridge:
+    type = iface_bridge
+    start_vm = no
+    vms = avocado-vt-vm1 vm2
+    bridge_name = "test_br0"
+    # The netdst must be the same with the bridge_name
+    netdst_nic1 = "test_br0"
```
chunfuwen: use ${bridge_name} instead

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,200 @@
...
+        if libvirt.check_iface(bridge_name, "exists", "--all"):
+            virsh.iface_unbridge(bridge_name)
+        if os.path.exists(iface_script_bk):
+            process.run("mv %s %s" % (iface_script_bk, iface_script), shell=True)
+            # Need reload script
+            process.run("ifdown %s" % iface_name, shell=True)
+            process.run("ifup %s" % iface_name, shell=True)
+        if NM_is_running:
+            NM_service.restart()
+        if NW_is_running:
```
chunfuwen: duplicate code

libvirt/tests/cfg/virtual_network/iface_bridge.cfg
hahakiki2010: accepted

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: they are different, NM is for NetworkManager, NW is for network service

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,195 @@
...
+    filter_name = params.get("filter_name", "vdsm-no-mac-spoofing")
+    ping_count = params.get("ping_count", "5")
+    ping_timeout = float(params.get("ping_timeout", "10"))
+    iface_name = utils_net.get_net_if(state="UP")[0]
+    iface_script = NETWORK_SCRIPT + iface_name
+    iface_script_bk = os.path.join(data_dir.get_tmp_dir(), "iface-%s.bk" % iface_name)
+
+    vms = params.get("vms").split()
+    if len(vms) <= 1:
+        test.cancel("Need more than two VMs to test")
```
sathnaga: handle cases for more than 2 VMs, or reword this to mention just two VMs.

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,195 @@
...
+        NM_service.stop()
+
+    # Start network service
+    # Sometimes network.service not installed by default
+    NW_res = process.run("systemctl status network", shell=True, ignore_status=True)
+    if NW_res.exit_status:
+        if re.search(r'.*(network.service)\s.*not\s.*found', NW_res.stderr_text):
+            if not utils_package.package_install('network-scripts') or \
+                    not utils_package.package_install('initscripts'):
+                test.cancel("Failed to install network service")
```
sathnaga: we should be able to use `service.Factory` itself for this check aswell....can you please figure out, looks like code repeat here...

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if vm1.is_alive():
+            vm1.destroy()
+
+        vm1.start()
+        # Check if vm can get ip with the new create bridge
+        vm1_ip = vm2_ip = None
+        session1 = session2 = None
+        try:
+            utils_net.update_mac_ip_address(vm1, timeout=240)
```
sathnaga: `vm1.wait_for_login()` should suffice here? as it does all these inside...we can have that on a try block and failout appropriately...

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,195 @@
...
+            ping(vm1_ip, remote_ip, ping_count, ping_timeout, session=session1)
+
+        # Start vm2 connect to the same bridge
+        modify_iface_xml(bridge_name, filter_name, vm2_name)
+        if vm2.is_alive():
+            vm2.destroy()
+        vm2.start()
+        # Check if vm1 and vm2 can ping each other
+        try:
+            utils_net.update_mac_ip_address(vm2, timeout=240)
```
sathnaga: same as above comment

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,195 @@
...
+        vm2_xml_bak.sync()
+        virsh.nwfilter_undefine(filter_name, ignore_status=True)
+        if libvirt.check_iface(bridge_name, "exists", "--all"):
+            virsh.iface_unbridge(bridge_name)
+        if os.path.exists(iface_script_bk):
+            process.run("mv %s %s" % (iface_script_bk, iface_script), shell=True)
+            # Need reload script
+            process.run("ifdown %s" % iface_name, shell=True)
+            process.run("ifup %s" % iface_name, shell=True)
+        if NM_is_running:
```
sathnaga: it has been stopped in the teststep, so ideally it should be checked for stopped and invoke start instead restart?

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: updated

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: yes, service.Factory return None when service is not found.  So I guess I can use ```None``` for check here. Thanks for your comments

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: Well, It is just to get the vm ip and do no need to wait vm really get started. But now I think you're right, as vm.wait_for_login is more than the code I picked above. So updated, thanks for your comments

hahakiki2010:
@sathnaga @chunfuwen  all your comments updated. Pls help review again, thanks~

kylazhang:
@hahakiki2010 Please add parameters explanation for all the functions you defined.

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,193 @@
...
+        vm2.start()
+
+        # Check if vm1 and vm2 can ping each other
+        try:
+            utils_net.update_mac_ip_address(vm2, timeout=120)
+            vm2_ip = vm2.get_address()
+        except Exception as errs:
+            test.fail("vm2 can't get IP with the new create bridge: %s" % errs)
+        session2 = vm2.wait_for_login()
+        if session1 and session2 and vm1_ip and vm2_ip:
```
kylazhang: what to do for else? Or no else then no need to do judgement?

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,193 @@
...
+        NM_service.stop()
+
+    # Start network service
+    NW_service = service.Factory.create_service("network")
+    NW_status = NW_service.status()
+    if NW_status is None:
+        logging.debug("network service not found")
+        if not utils_package.package_install('network-scripts') or \
+                not utils_package.package_install('initscripts'):
+            test.cancel("Failed to install network service")
```
kylazhang: should be test.error?

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def create_bridge(br_name, iface_name):
+        """
+        Create a linux bridge by virsh cmd:
+        1. Stop NetworkManager and Start network service
+        2. virsh iface-bridge <iface> <name> [--no-stp]
+        """
+        # Make sure the bridge not exist
+        if libvirt.check_iface(br_name, "exists", "--all"):
+            test.cancel("The bridge %s already exist" % br_name)
```
kylazhang: if exist, why not just use it?

hahakiki2010: Sometimes when bridge already exists, the test env might not be clean, for example the bridge failed to be deleted in last run. So manual recovery would be better here as the scenario is complicated

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: It's a precondition but not a test run failure. So I choose test.cancel. And checked previous cases, half for test.cancel, half for test.error, seemingly both they are accepted.

libvirt/tests/src/virtual_network/iface_bridge.py
hahakiki2010: accepted, remove the judgement

PR #1870, opened 2018-10-24
Increase the consumed memory for testing memory binding

jinqi7:

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -467,7 +467,8 @@ def run(test, params, env):
             pid = vm.get_pid()
             old_numastat = read_from_numastat(pid, "Total")
             logging.debug("Numastat: %s", old_numastat)
-            consume_vm_mem()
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_mem.py
```diff
@@ -467,7 +467,8 @@ def modify_domain_xml():
             pid = vm.get_pid()
             old_numastat = read_from_numastat(pid, "Total")
             logging.debug("Numastat: %s", old_numastat)
-            consume_vm_mem()
+            # Increase the memory consumed to  1500
+            consume_vm_mem(1500)
```
chunfuwen: suggest culculate the memory based on free memory of the host, rather than hardcode 1500

jinqi7: The fix maybe not need if the failure is occassional. The memory size in the case is fixed as 2G,
and here just change hard code from 1000 to 1500 to fix the occassional issue.

PR #1871, opened 2018-10-24
Add disk cache mode "directsync" to unsafe migration check

fangge1212:
Disk cache mode can be either "none" or "directsync" for a safe migration

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
@@ -800,7 +800,7 @@ def run(test, params, env):
     disk_cache = params.get("virsh_migrate_disk_cache", "none")
     params["driver_cache"] = disk_cache
     unsafe_test = False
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -800,7 +800,7 @@ def check_vm_state_dest():
     disk_cache = params.get("virsh_migrate_disk_cache", "none")
     params["driver_cache"] = disk_cache
     unsafe_test = False
-    if options.count("unsafe") and disk_cache != "none":
+    if options.count("unsafe") and disk_cache not in ["none", "directsync"]:
```
chunfuwen: I assume when disk_cache is directsync, you prefer to  go to the statement: unsafe_test = True

fangge1212: No,live migration is safe when disk_cache is either none or directsync

PR #1876, opened 2018-10-31
virsh_migrate: check hotplug vcpu using wait_for()

balamuruhans:
virsh console session is used to take vcpu count after guest is migrated
to target host, so other dmesg printed on console due to cpu hotplug event
is taken along with lscpu value cause validation to fail. use wait_for()
to avoid while loop and check the vcpu count after 5 sec of hotplug event
so that appropriate value is considered for validation.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
@@ -394,26 +394,10 @@ def run(test, params, env):
 
             logging.debug("Checking CPU number gets reflected from inside "
                           "guest")
-            cmd = "lscpu | grep \"^CPU(s):\""
-            # vcpu count gets reflected step by step gradually, so we check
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_migrate.py
```diff
@@ -360,6 +360,24 @@ def cpu_hotplug_hotunplug(vm, cpu_count, operation, uri=None):
 
         :raise test.fail if hotplug or hotunplug doesn't work
         """
+        prev_output = -1
+
+        def check_vcpu():
```
sathnaga: @balamuruhans you could try this ..
```
 if not utils_misc.wait_for(lambda: utils_misc.check_if_vm_vcpu_match(vcpu_plug_num, vm),
                                                   vcpu_max_timeout, text="wait for vcpu online"):
                            test.fail("Fail to enable new added cpu")
```

PR #1882, opened 2018-11-09
Only create or destroy target pool when target_pool_name is not None

fangge1212:
It's meaningless to create or destroy target storage pool when
target_pool_name is None, and it will fail if target libvirtd is
stopped in some negative test cases


dzhengfy:
@fangge1212 
pool_name = params.get("target_pool_name", "tmp_pool_1")
’target_pool_name‘ has a default value. Please consider more.

fangge1212:
@dzhengfy 
Updated again, please review
Need to determine whether pool name is '' before dumpxml

PR #1889, opened 2018-11-16
migrate_vm: Enable port on target host

dzhengfy:
Migration should be done on both hosts running firewalld if any. So this
is to enable special port on target host which is not done before.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -2338,7 +2338,7 @@ def run(test, params, env):
 
                 # Permit iptables to permit special port to libvirt for
                 # migration on local machine
-                migrate_setup.migrate_pre_setup(src_uri, params, ports=uri_port)
+                migrate_setup.migrate_pre_setup(src_uri, params, ports=uri_port[1:])
 
             except (process.CmdError, remote.SCPError) as e:
                 logging.debug(e)
@@ -2350,7 +2350,7 @@ def run(test, params, env):
                 remote_virsh_session.close_session()
 
             uri = "%s%s%s://%s:%s%s" % (driver, plus, transport,
-                                        client_cn, uri_port, uri_path)
+                                        client_cn, uri_port[1:], uri_path)
             test_dict["desuri"] = uri
             test_dict["vm_name_to_migrate"] = target_vm_name
 
@@ -2366,6 +2366,9 @@ def run(test, params, env):
             status_error = "no"
             test_dict['err_msg'] = None
 
+        if transport in ('tcp', 'tls') and uri_port:
+            migrate_setup.migrate_pre_setup("//%s/" % server_ip, test_dict,
+                                            cleanup=False, ports=uri_port[1:])
         if run_migr_front:
             migrate_vm(test_dict)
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -974,6 +975,47 @@ def check_domjobinfo_on_complete(test, source_jobinfo, target_jobinfo):
...
+    Enable or disable a port on certain host
+
+    :param test: test object
+    :param params: dict for the test
+    :param port: the port to be operated
+    :param isRemote: True for remote host
+    :param disable: True if you want to disable the port
+
+    """
+    server_ip = params.get("server_ip")
```
chunfuwen: move those 3 variables into isRemote block since it apply this situation only.

dzhengfy: @chunfuwen update.

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2366,6 +2405,8 @@ def run(test, params, env):
             status_error = "no"
             test_dict['err_msg'] = None
 
+        if transport in ('tcp', 'tls') and uri_port:
+            enable_or_disable_port(test, test_dict, uri_port[1:], True, False)
```
fangge1212: It seems that the first number in uri_port is lost:

2018-11-28 03:08:09,088 client           L1126 DEBUG| Sending command: firewall-cmd --add-port=**6514**/tcp --permanent --zone=public

Here, tls port should be 16514, not 6514.

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2780,3 +2821,5 @@ def run(test, params, env):
             n_client_s.package.env_cleanup(True)
 
         cleanup(objs_list)
+        if transport in ('tcp', 'tls') and uri_port:
+            enable_or_disable_port(test, test_dict, uri_port[1:], True, True)
```
fangge1212: Same as above

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -973,6 +974,44 @@ def check_domjobinfo_on_complete(test, source_jobinfo, target_jobinfo):
...
+    try:
+        if params.get('firewalld_service', "yes") == "yes":
+            session = None
+            if isRemote:
+                server_ip = params.get("server_ip")
+                server_user = params.get("server_user")
+                server_pwd = params.get("server_pwd")
+                session = remote.wait_for_login('ssh', server_ip, '22',
+                                                server_user, server_pwd,
+                                                r"[\#\$]\s*$")
```
balamuruhans: this piece of code from L991 to L996 can be avoided using `test_setup.remote_session()`

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: fixed

libvirt/tests/src/migration/migrate_vm.py
dzhengfy: fixed

dzhengfy:
Hold on, I want to update more.

PR #1891, opened 2018-11-16
virsh_iface_bridge: Fix Interface 'HOST_INTERFACE' not exists

hahakiki2010:
The script required to manually configure the iface_name in cfg file,
but it's hard when CI running, cos the iface_name is varied from servers,
so it's better to get the iface_name by script instead of manual configuration.
besides fix some other issues during this PR testing, as this script never
been successfully executed

### DIFF
--- a/libvirt/tests/src/virsh_cmd/interface/virsh_iface_bridge.py
+++ b/libvirt/tests/src/virsh_cmd/interface/virsh_iface_bridge.py
@@ -10,6 +10,7 @@ from virttest import data_dir
 from virttest import virsh
 from virttest.utils_test import libvirt
 from virttest.staging import service
+from virttest import utils_package
 
 NETWORK_SCRIPT = "/etc/sysconfig/network-scripts/ifcfg-"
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/interface/virsh_iface_bridge.py
```diff
@@ -85,7 +104,8 @@ def unbridge_check():
     try:
         if create_bridge:
             # Create bridge
-            result = virsh.iface_bridge(iface_name, bridge_name, bridge_option)
+            # Set timeout as the cmd run a long time
+            result = virsh.iface_bridge(iface_name, bridge_name, bridge_option, timeout=240, debug=True)
```
chunfuwen: suggest add below clean up statement ahead virsh.iface_bridge to avoid bridge creation failure
```
            if os.path.exists(bridge_script):
                process.run("rm -rf %s" % bridge_script, shell=True)
```

hahakiki2010: Any difference? I think they are the same.

PR #1894, opened 2018-11-19
virsh_remote: Use server session instead of vm session

hahakiki2010:
The old script use vm session for remote test, but it's not
applicable when testing on remote server. So change to
remote.wait_for_login to compat all cases

### DIFF
--- a/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
+++ b/libvirt/tests/src/virsh_cmd/host/virsh_hostname.py
@@ -1,10 +1,10 @@
 from avocado.utils import process
 
 from virttest import virsh
-from virttest import utils_libvirtd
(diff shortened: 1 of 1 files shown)


libvirt/tests/cfg/virsh_cmd/host/virsh_hostname.cfg
```diff
@@ -15,7 +15,6 @@
             status_error = "yes"
             libvirtd = "off"
         - remote_connect:
-            start_vm = yes
```
chunfuwen: I assume we still need start vm firstly

hahakiki2010: accepted, in case test on vm on the host

PR #1898, opened 2018-11-23
Add prepare and recover multipath conf for npiv caese

lento-sun:
If /etc/multipath.conf doesn't exist, multipath command will be
failed, add steps to avoid this problem.


lento-sun:
This pr depends on: https://github.com/avocado-framework/avocado-vt/pull/1839

libvirt/tests/src/npiv/npiv_concurrent.py
```diff
@@ -308,3 +310,6 @@ def run(test, params, env):
         for new_vhba in new_vhbas:
             virsh.nodedev_destroy(new_vhba)
         process.system('service multipathd restart', verbose=True)
+        if old_mpath_conf:
+            utils_npiv.prepare_multipath_conf(conf_content=old_mpath_conf,
+                                              replace_existing=True)
```
kylazhang: for else, do we need to remove the config file?

lento-sun: actually no need to remove, if it's not existing, multipath cmd doesn't work well.

lento-sun:
@kylazhang add logic to tell if original mpath conf file existing, if now, will remove the file in "finally".

PR #1900, opened 2018-11-26
Check the target dev set in the interface xml

yalzhang:
Target dev can set in the guest interface xml including direct type and
network type interface. The name with prefix as "vnet" or "macvtap" may
be override.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_options.py
+++ b/libvirt/tests/src/virtual_network/iface_options.py
@@ -64,6 +64,8 @@ def run(test, params, env):
         iface.driver = iface.new_driver(driver_attr=driver_dict,
                                         driver_host=driver_host,
                                         driver_guest=driver_guest)
+        if test_target:
+            iface.target = {"dev": target_dev}
         logging.debug("Create new interface xml: %s", iface)
         return iface
 
@@ -112,6 +114,9 @@ def run(test, params, env):
         iface.driver = iface.new_driver(driver_attr=driver_dict,
                                         driver_host=driver_host,
                                         driver_guest=driver_guest)
+        if test_target:
+            logging.debug("iface.target is %s" % target_dev)
+            iface.target = {"dev": target_dev}
         if iface.address:
             del iface.address
         if set_ip:
@@ -220,6 +225,22 @@ def run(test, params, env):
                     mode = "passthru"
                 if not re.search("macvtap\s+mode %s" % mode, output):
                     test.fail("Failed to verify macvtap mode")
+        # Check if the "target dev" is set successfully
+        # 1. Target dev name with prefix as "vnet" will always be override;
+        # 2. Target dev name with prefix as "macvtap" or "macvlan" with direct
+        # type interface will be override;
+        # 3. Other scenarios, the target dev should be set successfully.
+        if test_target:
+            if target_dev != iface.target["dev"]:
+                if target_dev.startswith("vnet") or \
+                        (iface_type == "direct" and
+                         (target_dev.startswith("macvtap") or
+                          target_dev.startswith("macvlan"))):
+                    logging.debug("target dev %s is override" % target_dev)
+                else:
+                    test.fail("Failed to set target dev to %s", target_dev)
+            else:
+                logging.debug("target dev set successfully to %s", iface.target["dev"])
 
     def run_cmdline_test(iface_mac):
         """
@@ -430,6 +451,8 @@ def run(test, params, env):
     expect_ip = params.get("expect_ip")
     expect_gw = params.get("expect_gw")
     expect_ns = params.get("expect_ns")
+    test_target = "yes" == params.get("test_target", "no")
+    target_dev = params.get("target_dev", None)
 
     if iface_driver_host or iface_driver_guest or test_backend:
         if not libvirt_version.version_compare(1, 2, 8):
@@ -489,6 +512,10 @@ def run(test, params, env):
                 modify_iface_xml(update=False)
                 if define_error:
                     return
+
+            if test_target:
+                logging.debug("Setting target device name to %s", target_dev)
+                modify_iface_xml(update=False)
 
             if rm_vhost_driver:
                 # remove vhost driver on host and
@@ -614,6 +641,9 @@ def run(test, params, env):
                             logging.info("Find TX settint TX:%s by ethtool", driver_dict['tx_queue_size'])
                         else:
                             test.fail("Cannot find matching tx setting")
+            if test_target:
+                logging.debug("Check if the target dev is set")
+                run_xml_test(iface_mac)
 
             session.close()
             # Restart libvirtd and guest, then test again
--- a/libvirt/tests/cfg/virtual_network/iface_options.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_options.cfg
@@ -221,3 +221,28 @@
                 - error_test:
                     start_error = "yes"
                     iface_backend = "{'tap':'/dev/net/mytap','vhost':'/dev/myvhost-net'}"
+        - iface_target_prefix:
+            test_target = "yes"
+            variants:
+                - network:
+                    iface_type = "network"
+                    iface_source = "{'network':'default'}"
+                - direct:
+                    iface_type = "direct"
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -112,6 +114,9 @@ def modify_iface_xml(update, status_error=False):
         iface.driver = iface.new_driver(driver_attr=driver_dict,
                                         driver_host=driver_host,
                                         driver_guest=driver_guest)
+        if test_target:
+            logging.debug("iface.target is %s" % target_dev)
+            iface.target = {"dev": target_dev}
```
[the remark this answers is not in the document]
yalzhang: No, 67~68 is in function "create_iface_xml()" and 117~118 is in function "modify_iface_xml()"
modify_iface_xml() will be called in each case to edit the target dev before  define or create the vm.
But create_iface_xml() will only be called when test the duplicate target name which need to attach a new iface with the created xml.

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -217,6 +222,17 @@ def run_xml_test(iface_mac):
                     mode = "passthru"
                 if not re.search("macvtap\s+mode %s" % mode, output):
                     test.fail("Failed to verify macvtap mode")
+        if test_target:
+            if target_dev != iface.target["dev"]:
+                if target_dev.startswith("vnet") or \
```
chunfuwen: suggest add comments to explain in what situation the target dev will be overrided.

libvirt/tests/src/virtual_network/iface_options.py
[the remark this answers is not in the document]
sathnaga: this check not needed as you have already checking for `test_target` before calling `modify_iface_xml` ?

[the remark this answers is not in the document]
yalzhang: "modify_iface_xml()" is called before I start a vm with the single modified interface, this is step 1;
"create_iface_xml()" only needed when I test the case "duplicate_target", before start the vm, will attach a device with such xml with target dev set as duplicate, this is step 2, so both functions need to modify the target dev.

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -423,6 +439,8 @@ def check_mcast_network(session):
     test_guest_ip = "yes" == params.get("test_guest_ip", "no")
     test_backend = "yes" == params.get("test_backend", "no")
     check_guest_trans = "yes" == params.get("check_guest_trans", "no")
+    test_target = "yes" == params.get("test_target", "no")
+    target_dev = params.get("target_dev")
```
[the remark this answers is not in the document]
yalzhang: Thank you for your review, update to:
target_dev = params.get("target_dev", None)

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: done

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -423,6 +444,8 @@ def check_mcast_network(session):
     test_guest_ip = "yes" == params.get("test_guest_ip", "no")
     test_backend = "yes" == params.get("test_backend", "no")
     check_guest_trans = "yes" == params.get("check_guest_trans", "no")
+    test_target = "yes" == params.get("test_target", "no")
+    target_dev = params.get("target_dev", None)
```
chloerh: target_dev = params.get("target_dev")

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -217,6 +222,22 @@ def run_xml_test(iface_mac):
...
+            if target_dev != iface.target["dev"]:
+                if target_dev.startswith("vnet") or \
+                        (iface_type == "direct" and
+                         (target_dev.startswith("macvtap") or
+                          target_dev.startswith("macvlan"))):
+                    logging.debug("target dev %s is override" % target_dev)
+                else:
+                    test.fail("target dev is not set successfully to %s" % target_dev)
+            else:
+                logging.debug("target dev set successfully to %s" % iface.target["dev"])
```
chloerh: 'logging' could use ',' instead of '%', it's actually recommended

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: updated

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: updated and Thank you!

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -217,6 +222,22 @@ def run_xml_test(iface_mac):
...
+        # 3. Other scenarios, the target dev should be set successfully.
+        if test_target:
+            if target_dev != iface.target["dev"]:
+                if target_dev.startswith("vnet") or \
+                        (iface_type == "direct" and
+                         (target_dev.startswith("macvtap") or
+                          target_dev.startswith("macvlan"))):
+                    logging.debug("target dev %s is override" % target_dev)
+                else:
+                    test.fail("target dev is not set successfully to %s" % target_dev)
```
sathnaga: `failed to set target dev to %s` seems better,

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -217,6 +222,22 @@ def run_xml_test(iface_mac):
...
+        # 2. Target dev name with prefix as "macvtap" or "macvlan" with direct
+        # type interface will be override;
+        # 3. Other scenarios, the target dev should be set successfully.
+        if test_target:
+            if target_dev != iface.target["dev"]:
+                if target_dev.startswith("vnet") or \
+                        (iface_type == "direct" and
+                         (target_dev.startswith("macvtap") or
+                          target_dev.startswith("macvlan"))):
+                    logging.debug("target dev %s is override" % target_dev)
```
sathnaga: unable to set target dev due to type `vnet` or `macvtap` something like that would be better explain user

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -483,6 +506,10 @@ def check_mcast_network(session):
                 if define_error:
                     return
 
+            if test_target:
+                logging.debug("Test the target name as %s" % target_dev)
```
sathnaga: `Setting target device name to %s`

yalzhang: updated

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: updated

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: When we set the target dev name with prefix like "macvtap" "macvlan", it will be override as expected. This is by design, so I think if we update to "unable to...", it will cause misunderstanding.

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: sorry, I must forgot it...

libvirt/tests/src/virtual_network/iface_options.py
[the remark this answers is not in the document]
yalzhang: sorry, I forgot this one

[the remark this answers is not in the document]
yalzhang: For the next comment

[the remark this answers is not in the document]
yalzhang: For the next comment 
"this check not needed as you have already checking for test_target before calling modify_iface_xml ?" 
it is not a check step, it is a setting step. modify the iface xml to be with the target dev.

yalzhang:
@sathnaga , I have finished the update, please check again, Thank you very much!

PR #1923, opened 2018-12-11
remove fsid=0 for nfs prepare

lento-sun:
fsid=0 misused here, in nfsv4, the path to the export dir will be
missing with current code. Cause a access denied error.


lento-sun:
seems these script also misused fsid=0, so related cases failed on rhel8

PR #1869, opened 2018-10-22
Remove --unsafe from virsh options of live migration

fangge1212:
Using "--unsafe" in live migration is not safe for vm, and may result in
leak of bugs. Remove "--unsafe" from virsh options, and set disk cache
to "none" by default for a successful safe migration(except for
storage migration where disk cache can be anything for safe migration or
--unsafe is the test point of the case).

Signed-off-by: Fangge Jin <fjin@redhat.com>

PR #1929, opened 2018-12-17
migrate_vm: Remove remote disk cache setting

dzhengfy:
Existing logic has issues which make case result not statble. The cache
setting can be done when installing the remote guest.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -2291,7 +2291,7 @@ def run(test, params, env):
                 cmd = "virsh dumpxml %s > %s" % (target_vm_name, xml_path)
                 status, output = run_remote_cmd(cmd, server_ip,
                                                 server_user, server_pwd)
-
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2306,12 +2306,6 @@ def run(test, params, env):
                                 % (nfs_mount_dir, image_name)}
                 guest_config.sub(pattern2repl)
 
-                logging.debug("Modify remote guest xml's disk cache mode")
-                for driver_type in ["qcow2", "raw"]:
```
chunfuwen: I assume this work on version before, so suggest add one version condition here to skip cache mode modification

dzhengfy: @chunfuwen , line 2309, 2310 was added in https://github.com/autotest/tp-libvirt/pull/1869. With 1869, the case is always failed. So it is safe to remove it and use other way, like adding cache when installing the guest.

PR #1931, opened 2018-12-18
Fix the no ip issue on new libvirt version

yalzhang:
As new release use nftables as backend of firewalld, the guest can not get
dhcp ip address from nat network. Current workaround is to add the
bridge interface into the libvirt zone.


yalzhang:
Run related cases on rhel7 and rhel8, both passed.

PR #1944, opened 2018-12-25
Install tar before install netperf

fangge1212:
Package tar is used to untar package netperf and need to be
installed before netperf.


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2003,6 +2004,18 @@ def run(test, params, env):
 
         netperf_version = test_dict.get("netperf_version")
         if netperf_version:
+            # Install tar on client
+            # Note: tar is used to untar netperf package later.
+            if not utils_package.package_install(["tar"]):
+                test.error("Failed to install tar on client")
```
chunfuwen: suggest skip this case ,rather than fail this case

PR #1946, opened 2018-12-26
Wait 5 seconds to fully detach interface

chloerh:


libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_interface.py
```diff
@@ -397,6 +398,7 @@ def run(test, params, env):
 
         if detach_status == 0 and status_error == 0:
             # Check the xml after detach and clean up if needed.
+            time.sleep(5)
```
chunfuwen: I suggest use utils_mis.waitfor(xx), rather than hardcode timeout value

PR #1962, opened 2019-01-03
Fix setvcpus error_test shut_off error option with topology case fail

chunfuwen:
The precondition for this case is to have one topology element in vmxml,
Therefore, by adding one topology element before executing any further action

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
@@ -5,7 +5,8 @@ import logging
 from virttest import ssh_key
 from virttest import data_dir
 from virttest import virsh
-from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml import vm_xml, xcepts
+from virttest.libvirt_xml.vm_xml import VMCPUXML
 from virttest import utils_hotplug
 
 
@@ -53,6 +54,7 @@ def run(test, params, env):
     remote_uri = params.get("remote_uri")
     tmpxml = os.path.join(data_dir.get_tmp_dir(), 'tmp.xml')
     topology_correction = "yes" == params.get("topology_correction", "yes")
+    with_topology = "yes" == params.get("with_topology", "yes")
     result = True
 
     # Early death 1.1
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
```diff
@@ -149,6 +151,22 @@ def set_expected(vm, options):
...
+            except xcepts.LibvirtXMLNotFoundError:
+                logging.debug("Can not find any cpu tag, now create one.")
+                vmcpu_xml = VMCPUXML()
+            cores = vmxml['vcpu']
+            vmcpu_xml['topology'] = {'sockets': 1,
+                                     'cores': cores,
+                                     'threads': 1}
+            vmxml['cpu'] = vmcpu_xml
+            vmxml.sync()
+
```
sathnaga: this can done by passing a param to below function `add_topology=True`, we need to just make sure `sockets, threads, cores` params are defined,
```
vmxml.set_vm_vcpus(vm_name, max_vcpu, current_vcpu, topology_correction=topology_correction, add_topology=with_topology)
```

chunfuwen: the function need specify max_vcpu, I don't think here we have this.

libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
```diff
@@ -53,6 +54,7 @@ def run(test, params, env):
     remote_uri = params.get("remote_uri")
     tmpxml = os.path.join(data_dir.get_tmp_dir(), 'tmp.xml')
     topology_correction = "yes" == params.get("topology_correction", "yes")
+    with_topology = "yes" == params.get("with_topology", "yes")
```
chloerh: So 'with_topology' is True by default?

chunfuwen: Yes, that's basic statement in https://bugzilla.redhat.com/show_bug.cgi?id=1426220

PR #1963, opened 2019-01-03
Fix cpu_stats error message failure

chunfuwen:
Error message is changed on new release


libvirt/tests/cfg/virsh_cmd/domain/virsh_cpu_stats.cfg
```diff
@@ -39,7 +39,7 @@
                     error_msg = "command 'cpu-stats' requires <domain> option"
                 - unexpect_domain:
                     cpu_stats_vm_ref = "\#"
-                    error_msg = "Domain not found"
+                    error_msg = "failed to get domain"
```
sathnaga: better to handle this by libvirt version check in testcase, to have backward compatibility of testcase.

PR #1965, opened 2019-01-04
migrate_vm: adjust the stress installation

dzhengfy:
The stress package installation is moved to the avocado-vt.


dzhengfy:
Depends on https://github.com/avocado-framework/avocado-vt/pull/1889

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1985,8 +1985,7 @@ def run(test, params, env):
             elif s_list and s_list[-1] == "host":
                 logging.info("Run '%s %s' in %s", s_list[0],
                              stress_args, s_list[-1])
-                pkg_list = ['stress']
-                libvirt.yum_install(pkg_list)
+                test_dict['stress_package'] = 'stress'
```
balamuruhans: can we add `stress_install_from_repo = yes` in migrate_vm.cfg ?

dzhengfy: @balamuruhans , this param was originally used to be set directly in avocado-vt/shared/cfg/base.cfg. But it seems it is better to be enabled as default. I am fine with this. Updated. Thanks.

PR #1983, opened 2019-01-18
Fix virsh event case failures due to unsupported IDE controller on q35

chunfuwen:
On q35 machine type, IDE controller type is not supported, and
scsi or sata are preferred.
Moreover, some hardcarded string match is removed with self adapted parameter


chunfuwen:
paste test result
```
PASS 1-type_specific.io-github-autotest-libvirt.virsh.event.positive_test.virsh_event.test_events.tray-change_event.timeout.loop

Not logging /proc/pci (file does not exist)
```

PR #1991, opened 2019-01-23
Add new test cases for migration.

fangge1212:
Add virsh_migrate_vm_cfg.cfg and virsh_migrate_vm_cfg.py, which tests
--persistent, --persistent-xml, --xml, --dname, --undefinesource and
with/without graphic passwd, multiplied by p2p/tunnelled/normal
migraiton, precopy/postcopy, src vm status and src vm persistency.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_vm_cfg.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_vm_cfg.py
@@ -1,0 +1,526 @@
+import logging
+import time
+import re
+
+from six import itervalues, string_types
+from avocado.utils import process
+
+from virttest import utils_selinux
+from virttest import virsh
+from virttest import utils_package
+from virttest.utils_conn import TLSConnection
+from virttest.utils_test import libvirt
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices import graphics
+from virttest.libvirt_xml.xcepts import LibvirtXMLNotFoundError
+
+
+def run(test, params, env):
+    """
+    Test virsh migrate command.
+    """
+
+    def cleanup_vm(vm, vm_name='', uri=''):
+        """
+        Clean up vm in the src or destination host environment
+        when doing the uni-direction migration.
+        """
+        # Backup vm name and uri
+        uri_bak = vm.connect_uri
+        vm_name_bak = vm.name
+
+        # Destroy and undefine vm
+        vm.connect_uri = uri if uri else uri_bak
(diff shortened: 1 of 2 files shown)


balamuruhans:
@fangge1212, can you take a look at virsh_migration.py,
I have brought down the line of code as well with multivm support. can we add new testcases in that ?

fangge1212:
> @fangge1212, can you take a look at virsh_migration.py,
> I have brought down the line of code as well with multivm support. can we add new testcases in that ?

Hi @balamuruhans , I had a look at virsh_migration.py, it's design is not same as my original thoughts.
My thought is to include relevant test conditions in a cfg,  treat each virsh option or other condition as a variant, so we can get more flexible combinations of test conditions.
What's your opinion?

balamuruhans:
@fangge1212, most of the scenarios in this patch are covered as part of 
```
migrate_vm.py,
virsh_migrate.py,
virsh_migration.py,
migrate_options_shared.py,
virsh_migrate_*.py,```
the intention is to contain the new scenarios with existing file.

@dzhengfy request to share your thoughts on this ?

File: libvirt/tests/src/virsh_cmd/domain/virsh_migrate_vm_cfg.py [hunk 1]
```diff
@@ -0,0 +1,512 @@
...
+        # Recover source vm defination (just in case).
+        logging.info("Recover vm defination on source")
+        if vm_xml_backup:
+            vm_xml_backup.define()
+
+        # Check test result.
+        if not result_check_pass:
+            test.fail("Migration succeed, but some check points didn't pass."
+                      "Please check the error log for details")
+        """
```
dzhengfy: Remove this line and after if we do not need them.

dzhengfy:
@balamuruhans , fangge1212 is the migration feature owner. After discussion with her, I got below understanding:
1. She need design a new matrix to test with multiple options. And she plan to add more options in future if any to test as many options combined as we can.
2. Current virsh_migrate.py does cover some options same with this PR, but virsh_migrate.py is already so big that we does not recommend to add new tests any more. migrate_vm.py is also already too big.
3. Other existing py files also have their own focus, like set_get_speed, set_get_maxdowntime and etc, but they do not have a big matrix testing. 

Based on above understanding, could you help reconsider it?

balamuruhans:
@dzhengfy okay, sure.

fangge1212:
Depends on https://github.com/avocado-framework/avocado-vt/pull/2114

fangge1212:
Updates:
1. Deleted useless comments
2. Set default remote access transport to 'tls'

PR #1992, opened 2019-01-24
Difference the expected result of storage migration per qemu version

fangge1212:
For qemu-kvm>=2.10.0, storage migration with shared storage can succeed;
For qemu-kvm<2.10.0, storage migration with shared storage will fail.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -147,9 +147,13 @@ def check_output(test, output_msg, params):
                           with some conditions satisfied
     """
     err_msg = params.get("err_msg", None)
-    if err_msg and err_msg in output_msg:
-        logging.debug("Expected error '%s' was found", err_msg)
-        return
+    status_error = params.get("status_error", "no")
+    if status_error == "yes" and err_msg:
+        if err_msg in output_msg:
+            logging.debug("Expected error '%s' was found", err_msg)
+            return
+        else:
+            test.fail("The expected error '%s' was not found in output '%s'" % (err_msg, output_msg))
 
     ERR_MSGDICT = {"Bug 1249587": "error: Operation not supported: " +
                    "pre-creation of storage targets for incremental " +
@@ -1057,7 +1061,6 @@ def run(test, params, env):
     vm_name = test_dict.get("main_vm")
     vm = env.get_vm(vm_name)
     start_vm = test_dict.get("start_vm", "no")
-    status_error = test_dict.get("status_error", "no")
     transport = test_dict.get("transport")
     plus = test_dict.get("conn_plus", "+")
     config_ipv6 = test_dict.get("config_ipv6", "no")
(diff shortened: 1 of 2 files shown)


dzhengfy:
@fangge1212 , please double check the PR description because I see a conflict between your description and the comments in code.

> For qemu-kvm>=2.10.0, storage migration with shared storage can succeed;

vs.
```
        # Check points for qemu-kvm >=2.10.0
        # 1. Migration fails with error "Block node is read-only"
```

fangge1212:
> @fangge1212 , please double check the PR description because I see a conflict between your description and the comments in code.
> 
> > For qemu-kvm>=2.10.0, storage migration with shared storage can succeed;
> 
> vs.
> 
> ```
>         # Check points for qemu-kvm >=2.10.0
>         # 1. Migration fails with error "Block node is read-only"
> ```

Oh, you are right. Updated.

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -160,11 +160,28 @@ def check_output(output_msg, params):
...
+                   "without shared storage is unsafe",
+                   "ERROR 5": "error: internal error: unable to " +
+                   "execute QEMU command 'nbd-server-add': " +
+                   "Block node is read-only"}
 
     # Check for special case firstly
     migrate_disks = "yes" == params.get("migrate_disks")
     status_error = "yes" == params.get("status_error")
+    error_index = params.get("error_index")
+    if error_index == "ERROR 5":
```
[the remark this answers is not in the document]
fangge1212: I just had a try, there is no exception when error_index is None:
`>>> a = None
>>> if a == "ddd":
...  print(a)
... 
`

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -160,11 +160,28 @@ def check_output(output_msg, params):
...
+            if output_msg.find(expect_msg) >= 0:
+                logging.debug("The expected error '%s' was found in output '%s'",
+                              expect_msg, output_msg)
+                return
+            else:
+                raise exceptions.TestFail("The expected error '%s' was not found in output '%s'",
+                                          expect_msg, output_msg)
+        else:
+            return
+
```
dzhengfy: I suggest we use 'err_msg' for your case to detect if the expected error message happens.
And feel free to add the required else branch.

fangge1212: Updated

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1057,7 +1061,6 @@ def run(test, params, env):
     vm_name = test_dict.get("main_vm")
     vm = env.get_vm(vm_name)
     start_vm = test_dict.get("start_vm", "no")
-    status_error = test_dict.get("status_error", "no")
```
kylazhang: no need to change the place of get("status_error", ''), just change the following 
```test_dict["status_error"] = "yes"  to status_error = "yes" ```

File: libvirt/tests/src/migration/migrate_vm.py [hunk 4]
```diff
@@ -1138,6 +1141,15 @@ def run(test, params, env):
     restart_vm = "yes" == test_dict.get("restart_vm", "no")
     diff_cpu_vendor = "yes" == test_dict.get("diff_cpu_vendor", "no")
 
+    # Get err_msg and status_error parameters
+    err_msg = test_dict.get("err_msg", None)
+    if err_msg == "error: internal error: unable to execute QEMU command 'nbd-server-add': Block node is read-only":
+        if utils_misc.compare_qemu_version(2, 10, 0, is_rhev=True):
+            test_dict["status_error"] = "yes"
+        else:
+            test_dict["status_error"] = "no"
```
kylazhang: Change here 2 lines

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1138,6 +1141,15 @@ def run(test, params, env):
...
+    status_error = test_dict.get("status_error", "no")
```
kylazhang: And remove it

libvirt/tests/src/migration/migrate_vm.py
fangge1212: I need to change the value in test_dict, because test_dict will be used in migrate_vm():

 2428         if run_migr_front:
 2429             migrate_vm(test, test_dict)

kylazhang: ok

PR #1998, opened 2019-01-28
Fix Attach and Detach failure in 'network_hook' test

yanjing1993:
Attach failure is caused by 'insufficient' pci controller
Add corresponding pci controller for both 'q35' and 'i440fx' VM
Detach failure is caused by 'timeout' issue
Add 'timeout' for both Attach and Detach operation

Signed-off-by: Jing Yan <jiyan@redhat.com>

### DIFF
--- a/libvirt/tests/src/libvirt_hooks.py
+++ b/libvirt/tests/src/libvirt_hooks.py
@@ -12,6 +12,8 @@ from virttest import utils_libvirtd
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
 from virttest import data_dir
+from virttest.libvirt_xml.devices.controller import Controller
+from virttest import utils_misc
 
 
 def run(test, params, env):
@@ -28,6 +30,7 @@ def run(test, params, env):
     virsh_dargs = {'debug': True, 'ignore_status': False}
     hook_file = params.get("hook_file", "/etc/libvirt/hooks/qemu")
     hook_log = params.get("hook_log", "/tmp/qemu.log")
+    machine_type = params.get("machine_type", "")
 
     def prepare_hook_file(hook_op):
         """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_hooks.py
```diff
@@ -12,6 +12,8 @@
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
 from virttest import data_dir
+from virttest.libvirt_xml.devices.controller import Controller
+from virttest import utils_misc
```
chloerh: We could organize there a little, like:

from xxx import xx
from yyy import yy
from xxx.xx.xxx import x
from yyy.yy.yyy import y

looks better

libvirt/tests/src/libvirt_hooks.py
```diff
@@ -298,21 +301,93 @@ def network_hook():
...
+                vm_inactive_xml.add_device(controller_pcie_root_port)
+                controller_pcie_to_pci_bridge = Controller("controller")
+                controller_pcie_to_pci_bridge.model = "pcie-to-pci-bridge"
+                controller_pcie_to_pci_bridge.type = "pci"
+                controller_pcie_to_pci_bridge.index = 2
+                vm_inactive_xml.add_device(controller_pcie_to_pci_bridge)
+            vm_inactive_xml.sync()
+
+            # Plug a interface and Unplug the interface
+            vm.define(vm_inactive_xml.xml)
```
chloerh: in line:350, vm_inactive_xml was synced, why define it here again?

libvirt/tests/src/libvirt_hooks.py
```diff
@@ -298,21 +301,93 @@ def network_hook():
...
+                while True:
+                    iface_features = vm_active_xml.get_iface_by_mac(vm_name, mac_addr)
+                    if iface_features is None:
+                        continue
+                    else:
+                        break
+                    return True
+                return False
+            if utils_misc.wait_for(is_attached_interface, 20) is not True:
+                logging.debug("Attach failed actually")
```
chloerh: If we must use 'while True', at least sleep 1 sec between each loop.
Or we could remove while and  line 373 could be 
if not utils_misc.wait_for(is_attached_interface, 20):

yanjing1993: Hi, submit another PR and close this one.
https://github.com/autotest/tp-libvirt/pull/2039
Thanks for your suggestion.

PR #2002, opened 2019-01-29
Enable nfs setup for RDMA migration

fangge1212:
If nfs setup is not enable for RDMA migration, it will reports:
"Unsafe migration: Migration without shared storage is unsafe"


dzhengfy:
@fangge1212 ，
1. Please help update your PR description and commit message with more positive information as we discussed.
2. Please add an error message matched for this negative case. This negative case is intended to fail with some specified error message.

fangge1212:
> @fangge1212 ，
> 
> Please help update your PR description and commit message with more positive information as we discussed.
> Please add an error message matched for this negative case. This negative case is intended to fail with some specified error message.

Updated

PR #2005, opened 2019-01-29
gluster: Change the gluster vol name

hsmj1412:
It will be used in different cases running at the same time


balamuruhans:
Looks good to me, request @dzhengfy @chunfuwen review.

PR #2009, opened 2019-01-30
virsh_boot.cfg: Update pattern of error message

meinaLi:
virsh_boot.cfg: Update pattern of error message

### DIFF
--- a/libvirt/tests/cfg/bios/virsh_boot.cfg
+++ b/libvirt/tests/cfg/bios/virsh_boot.cfg
@@ -232,7 +232,7 @@
                     variants:
                         - not_existing_loader:
                             loader = "noexist"
-                            checkpoint = "error: Path.+is not accessible: No such file or directory"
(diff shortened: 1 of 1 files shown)


libvirt/tests/cfg/bios/virsh_boot.cfg
```diff
@@ -232,7 +232,7 @@
                     variants:
                         - not_existing_loader:
                             loader = "noexist"
-                            checkpoint = "error: Path.+is not accessible: No such file or directory"
+                            checkpoint = "error:.+No such file or directory"
```
balamuruhans: why do we change the error msg ? is there any change in boot log/error msg ?

meinaLi: Yes, the error message has already changed:
error: Failed to start domain avocado-vt-vm1error: internal error: Process exited prior to exec: libvirt: QEMU Driver error : Unable to stat noexist: No such file or directory

PR #2011, opened 2019-01-30
virsh_event.py: dont ignore action failure

yanqzhan:
Set ignore_status=False for event actions, to show detail errors
as failure msgs, since no event will be caught if action fails.

Signed-off-by: Yanqiu Zhang <yanqzhan@redhat.com>


kylazhang:
@yanqzhan all the virsh_dargs will be changed include line 325 and line 330, are you sure it also need to set ignore_status = False?

harihare:
The code works fine with out mentioned change too. The patch does not change anything. One suggestion use **virsh_dargs in line 325 and 330 also after modifying 'ignore_status'

yanqzhan:
@harihare Thank you for suggestion. Code is modified. Pls help review again.

PR #2012, opened 2019-01-30
Should use secret driver in ACL for the encrypted volume creation

meinaLi:
Should use secret driver in ACL for the encrypted volume creation

### DIFF
--- a/libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
+++ b/libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
@@ -284,9 +284,18 @@ def run(test, params, env):
                         try:
                             polkit = test_setup.LibvirtPolkitConfig(params)
                             polkit_rules_path = polkit.polkit_rules_path
-                            cmd = "sed -i '/.secret./{n;s/QEMU/secret/g;s/pool_name/secret_uuid/;s/virt-dir-pool/%s/;}' %s" % (luks_secret_uuid, polkit_rules_path)
-                            process.system(cmd, ignore_status=True)
-                            rule = open(polkit_rules_path, 'r+').read()
+                            with open(polkit_rules_path, 'r+') as f:
+                                rule = f.readlines()
+                                for index, v in enumerate(rule):
+                                    if v.find("secret") >= 0:
+                                        nextline = rule[index + 1]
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
```diff
@@ -279,6 +280,17 @@ def check_vol(pool_name, vol_name, expect_exist=True):
                 if params.get('setup_libvirt_polkit') == 'yes':
                     process.run("chmod 666 %s" % vol_xml, ignore_status=True,
                                 shell=True)
+                    if luks_encrypted and libvirt_version.version_compare(4, 5, 0):
+                        try:
+                            polkit = test_setup.LibvirtPolkitConfig(params)
+                            polkit_rules_path = polkit.polkit_rules_path
+                            cmd = "sed -i '/.secret./{n;s/QEMU/secret/g;s/pool_name/secret_uuid/;s/virt-dir-pool/%s/;}' %s" % (luks_secret_uuid, polkit_rules_path)
```
chloerh: We should try to avoid using shell command lines in test scripts. It's better to use python lines instead.

libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
```diff
@@ -279,6 +280,17 @@ def check_vol(pool_name, vol_name, expect_exist=True):
...
+                            process.system(cmd, ignore_status=True)
```
balamuruhans: we should be backing up and revert the changes of polkit config once the test is done. please handle it as part of clean up.

libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
meinaLi: Finish to update.

libvirt/tests/src/virsh_cmd/volume/virsh_vol_create.py
meinaLi: The polkit config file changed in script is newly created before test, and it will automatically clean up after test. So I think there's no need to back up it.

chloerh:
Looks good to me

PR #2029, opened 2019-02-13
Fix the network issue in iface_bridge.py

yalzhang:
Use ip command to create bridge to replace the original iface-bridge
command as it is not stable. Execute the commands in tmux to avoid
breaking the host network. And newtork.service is not
recommended, replace it with NetworkManager service.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
@@ -5,7 +5,7 @@
     bridge_name = "test_br0"
     netdst_nic1 = ${bridge_name}
     # Replace remote_ip by a actual IP address
-    remote_ip = "www.baidu.com"
+    remote_ip = "www.google.com"
     ping_count = 3
     pint_timeout = 10
     filter_uuid = "11111111-b071-6127-b4ec-111111111111"
--- a/libvirt/tests/src/virtual_network/iface_bridge.py
+++ b/libvirt/tests/src/virtual_network/iface_bridge.py
@@ -10,6 +10,7 @@ from virttest.utils_test import libvirt
 from virttest.staging import service
 from virttest.libvirt_xml import vm_xml
 from virttest import utils_package
+from virttest.libvirt_xml.devices.interface import Interface
 
 NETWORK_SCRIPT = "/etc/sysconfig/network-scripts/ifcfg-"
 
@@ -43,10 +44,11 @@ def run(test, params, env):
             test.cancel("The bridge %s already exist" % br_name)
 
         # Create bridge
-        # This cmd run a long time, so set timeout=240
-        result = virsh.iface_bridge(iface_name, br_name, "--no-stp", debug=True, timeout=240)
-        if result.exit_status:
-            test.fail("Failed to create bridge:\n%s" % result.stderr)
+        utils_package.package_install('tmux')
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up;' \
+              ' pkill dhclient; sleep 6; dhclient {0}; ifconfig {1} 0"'.format(br_name, iface_name)
+        process.run(cmd, shell=True, verbose=True)
 
     def define_nwfilter(filter_name):
         """
@@ -82,7 +84,9 @@ def run(test, params, env):
         vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name)
         iface_xml = vmxml.get_devices('interface')[0]
         vmxml.del_device(iface_xml)
+        iface_xml = Interface(type_name='bridge')
         iface_xml.source = {'bridge': br_name}
+        iface_xml.model = 'virtio'
         iface_xml.filterref = iface_xml.new_filterref(name=nwfilter)
         logging.debug("new interface xml is: %s" % iface_xml)
         vmxml.add_device(iface_xml)
@@ -105,8 +109,21 @@ def run(test, params, env):
         if status:
             test.fail("Fail to ping %s from %s" % (dest_ip, src_ip))
 
+    def check_net_functions(guest_ip, ping_count, ping_timeout, guest_session, host_ip, remote_url, endpoint_ip):
+        # make sure host network works well
+        # host ping remote url
+        ping(host_ip, remote_url, ping_count, ping_timeout)
+        # host ping guest
+        ping(host_ip, guest_ip, ping_count, ping_timeout)
+        # guest ping host
+        ping(guest_ip, host_ip, ping_count, ping_timeout, session=guest_session)
+        # guest ping remote url
+        ping(guest_ip, remote_url, ping_count, ping_timeout, session=guest_session)
+        # guest ping endpoint
+        ping(guest_ip, endpoint_ip, ping_count, ping_timeout, session=guest_session)
+
     # Get test params
-    bridge_name = params.get("bridge_name", "br0")
+    bridge_name = params.get("bridge_name", "test_br0")
     filter_name = params.get("filter_name", "vdsm-no-mac-spoofing")
     ping_count = params.get("ping_count", "5")
     ping_timeout = float(params.get("ping_timeout", "10"))
@@ -126,7 +143,8 @@ def run(test, params, env):
     vm2 = env.get_vm(vm2_name)
 
     # Back up the interface script
-    process.run("cp %s %s" % (iface_script, iface_script_bk), shell=True)
+    process.run("cp %s %s" % (iface_script, iface_script_bk),
+                shell=True, verbose=True)
     # Back up vm xml
     vm1_xml_bak = vm_xml.VMXML.new_from_dumpxml(vm1_name)
     vm2_xml_bak = vm_xml.VMXML.new_from_dumpxml(vm2_name)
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -43,10 +44,12 @@ def create_bridge(br_name, iface_name):
...
 
         # Create bridge
-        # This cmd run a long time, so set timeout=240
-        result = virsh.iface_bridge(iface_name, br_name, "--no-stp", debug=True, timeout=240)
-        if result.exit_status:
-            test.fail("Failed to create bridge:\n%s" % result.stderr)
+        utils_package.package_install('tmux')
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up;' \
+              ' pkill dhclient; dhclient {0}"'.format(br_name, iface_name)
```
balamuruhans: dhclient would be performed on interface rather than bridge ?

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -43,10 +44,12 @@ def create_bridge(br_name, iface_name):
...
+        logging.debug("The cmd is %s" % cmd)
+        process.run(cmd)
```
balamuruhans: `shell=True`

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -134,20 +139,8 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
...
-    if NW_status is None:
-        logging.debug("network service not found")
-        if not utils_package.package_install('network-scripts') or \
-                not utils_package.package_install('initscripts'):
-            test.cancel("Failed to install network service")
-    if NW_status is not True:
-        logging.debug("network service is not running")
-        NW_service.start()
+    if NM_status is False:
+        NM_service.start()
```
balamuruhans: is this change backward compatible ? can we fallback with `NetworkManager` if `network` fails

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -169,10 +162,20 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
...
-        ping(host_ip, vm1_ip, ping_count, ping_timeout)
-        ping(host_ip, remote_ip, ping_count, ping_timeout)
+        # make sure the guest has got ip address
         session1 = vm1.wait_for_login()
+        session1.cmd("pkill -9 dhclient", ignore_all_errors=True)
+        session1.cmd("dhclient %s " % iface_name, ignore_all_errors=True)
+        output = session1.cmd_output("ifconfig || ip a")
+        logging.debug("guest ip info %s" % output)
+        # make sure host network works well
+        logging.debug("******** host ping remote baidu ********")
```
balamuruhans: can we remove `baidu` from log msgs here and elsewhere

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -201,8 +211,12 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
             process.run("mv %s %s" % (iface_script_bk, iface_script), shell=True)
         if os.path.exists(bridge_script):
             process.run("rm -rf %s" % bridge_script, shell=True)
+        cmd = 'tmux -c "ip link set {1} nomaster;  ip link delete {0};' \
+              'pkill dhclient; dhclient {1}"'.format(bridge_name, iface_name)
+        logging.debug("delete the bridge cmd is %s" % cmd)
+        process.run(cmd)
```
balamuruhans: same here `shell=True`

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -201,8 +211,12 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
             process.run("mv %s %s" % (iface_script_bk, iface_script), shell=True)
         if os.path.exists(bridge_script):
             process.run("rm -rf %s" % bridge_script, shell=True)
+        cmd = 'tmux -c "ip link set {1} nomaster;  ip link delete {0};' \
```
chloerh: Same here

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: > dhclient would be performed on interface rather than bridge ?

After attaching the interface to the bridge, the interface will change into promiscuous mode, and the bridge will change to forwarding state and share the same mac and ip address as eth0. We must set the default route via bridge not the physical interface, or the network will be broken.

yalzhang: > We should try to avoid using long shell expression

Well, after some attempt, I decide not to make so many changes and not fix the interface configuration issue.
And keep using the iface-bridge/ iface-unbridge command to create/delete a linux bridge.
There is something to do in the future: deprecate the network.service, and replace it with NetworkManager.service;

balamuruhans: got it, Thanks @yalzhang.

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: Yes. In the original script, stop the NetworkManager service is to avoid the host network broken issue caused by iface-bridge/ iface-unbridge.
It is said that "The host network interface management API is based on NET CF library. we try to make everythign work in both cases, but only officially support tie interface APIs when NM is disabled". So it is recommend to use network.service with iface-*. 
I have delete the iface-* command in this PR and use ip command with tmux to create bridge instead. It is more stable. So the purpose of this case change to test the behavior of bridge typ e interface. And the ip command works well both with network service and NetworkManager.

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: Thank you for the suggestion, I will remove it from the log info

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: will update and Thank you

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -43,10 +44,12 @@ def create_bridge(br_name, iface_name):
...
         # Create bridge
-        # This cmd run a long time, so set timeout=240
-        result = virsh.iface_bridge(iface_name, br_name, "--no-stp", debug=True, timeout=240)
-        if result.exit_status:
-            test.fail("Failed to create bridge:\n%s" % result.stderr)
+        utils_package.package_install('tmux')
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up;' \
+              ' pkill dhclient; dhclient {0}"'.format(br_name, iface_name)
+        logging.debug("The cmd is %s" % cmd)
```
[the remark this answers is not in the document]
yalzhang: deleted

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -188,7 +191,14 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
...
             test.fail("vm2 can't get IP with the new create bridge: %s" % errs)
         session2 = vm2.wait_for_login()
+        # make sure guest has got ip address
+        session2.cmd("pkill -9 dhclient", ignore_all_errors=True)
+        session2.cmd("dhclient %s " % iface_name, ignore_all_errors=True)
+        output2 = session2.cmd_output("ifconfig || ip a")
+        logging.debug("guest ip info %s" % output2)
+        logging.debug("******** vm2 ping vm1 ********")
         ping(vm2_ip, vm1_ip, ping_count, ping_timeout, session=session2)
+        logging.debug("******** vm1 ping vm2 ********")
```
kylazhang: We can remove the duplicate code and define a function to implement it.

yalzhang: updated according to all the comments above, please help to review, Thank you!

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -43,10 +44,11 @@ def create_bridge(br_name, iface_name):
...
         # Create bridge
-        # This cmd run a long time, so set timeout=240
-        result = virsh.iface_bridge(iface_name, br_name, "--no-stp", debug=True, timeout=240)
-        if result.exit_status:
-            test.fail("Failed to create bridge:\n%s" % result.stderr)
+        utils_package.package_install('tmux')
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up;' \
+              ' pkill dhclient; sleep 3; dhclient {0}; ifconfig {1} 0"'.format(br_name, iface_name)
+        process.run(cmd, shell=True)
```
chloerh: verbose=True

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -201,8 +215,11 @@ def ping(src_ip, dest_ip, ping_count, timeout, session=None):
             process.run("mv %s %s" % (iface_script_bk, iface_script), shell=True)
         if os.path.exists(bridge_script):
             process.run("rm -rf %s" % bridge_script, shell=True)
+        cmd = 'tmux -c "ip link set {1} nomaster;  ip link delete {0};' \
+              'pkill dhclient; sleep 3; dhclient {1}"'.format(bridge_name, iface_name)
+        process.run(cmd, shell=True)
```
chloerh: verbose=True

yalzhang: updated

libvirt/tests/cfg/virtual_network/iface_bridge.cfg
```diff
@@ -5,7 +5,7 @@
     bridge_name = "test_br0"
     netdst_nic1 = ${bridge_name}
     # Replace remote_ip by a actual IP address
-    remote_ip = "www.baidu.com"
+    remote_ip = "www.google.com"
```
balamuruhans: can we have something generic,

`remote_ip = ENTER.YOUR.REMOTE.EXAMPLE.COM`

PR #2034, opened 2019-02-18
Rename result in case the name being used in other ways

chloerh:
The purpose of using the expression "if 'result' in locals()" is
to check 'result' only if there is a 'result' to be checked. But
the var name 'result' is too common, too much risk to be used in
another way, being another var and not need to be checked.
Therefore, we rename this var to make it unique.


chloerh:
@kylazhang  Would you like to take a look at this?

PR #2035, opened 2019-02-18
Fix local variable 'backup_files' issue

yalzhang:
Fix local variable 'backup_files' referenced before assignment issue.
When there is no enough physical interface on the host, test will be
canceled, but untill then the backup_files is not defined. Another
update is delete the workaround for network issue as the bug is already
fixed.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/network/virsh_net_update.py
+++ b/libvirt/tests/src/virsh_cmd/network/virsh_net_update.py
@@ -110,21 +110,6 @@ def run(test, params, env):
     guest_iface_num = int(params.get("guest_iface_num", 1))
     newxml = ""
 
-    # As new release use nftables as backend of firewalld, the guest can not get
-    # dhcp ip address from nat network. Current workaround is to add the
-    # bridge interface into the libvirt zone.
-    check_zone_cmd = "firewall-cmd --get-active-zones"
-    outputs = process.run(check_zone_cmd, ignore_status=True).stdout_text
-    if "libvirt" in outputs:
-        iface_cmd = "firewall-cmd --list-interface --zone=libvirt"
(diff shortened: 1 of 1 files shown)


yalzhang:
after the fix, if there is no enough interface, the test will be canceled.
$ run virsh.net_update.normal-test.add_delete.forward_interface.macvtap.net_active.options_no.default_opt.without-guest
JOB ID     : 0755f4c90f3be3a6ba0798193d2abc8ed73795d2
JOB LOG    : /root/avocado/job-results/job-2019-02-18T15.48-0755f4c/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.net_update.normal-test.add_delete.forward_interface.macvtap.net_active.options_no.default_opt.without-guest: CANCEL: Can not find another physical interface to attach (5.97 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 1

libvirt/tests/src/virsh_cmd/network/virsh_net_update.py
```diff
@@ -110,21 +110,6 @@ def run(test, params, env):
     guest_iface_num = int(params.get("guest_iface_num", 1))
     newxml = ""
 
-    # As new release use nftables as backend of firewalld, the guest can not get
```
chloerh: Is this workaround necessary on any older releases?

yalzhang: No, the bug occur in rhel8 and exists for a period of time, but now it is fixed. So the workaround is not necessary for rhel8, and for any other older rhel versions.

chloerh:
Looks good to me

PR #2040, opened 2019-02-19
Install gcc before check isa feature

chloerh:
gcc is necessary since command 'cc' is used in the checking command.
Print out some debug info about vm xml, and command output for
future debugging.


chloerh:
@sathnaga Could you take a look? It's about ppc cpu compat cases

PR #2059, opened 2019-03-01
Fix TypoError while printing numa nodes.

harihare:
Numa nodes list has node numbers in int, while printing the same
it has TypoError, Hence before printing coverting it to string.


harihare:
With out fix

balamuruhans:
Thanks for the patch @harihare

PR #2070, opened 2019-03-11
Fix "failed to remove tty related kernel cmd line" issue

yanjing1993:
Adding 'guest kernel option console=ttyS0,115200' is not necessary
when these options are already in VM image file
Deleting these parameters also causes problems
Failing to remove guest kernel option occurs occasionally
Besides, the later auto jobs demanded 'virsh console' usage
can not succeed once removing guest kernel option successfully.

Signed-off-by: Jing Yan <jiyan@redhat.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_console.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_console.py
@@ -187,6 +187,7 @@ def run(test, params, env):
     if 'ppc64' in platform.machine():
         params["console_device"] = 'hvc0'
 
+    update_console = params.get("update_console", "yes")
     console_dev = params.get("console_device", "ttyS0")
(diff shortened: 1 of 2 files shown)


dzhengfy:
@yanjing1993 Please provide meaningful message in the description field, for example, why do you do the changes. Thank.

libvirt/tests/src/virsh_cmd/domain/virsh_console.py
```diff
@@ -187,6 +187,7 @@ def run(test, params, env):
     if 'ppc64' in platform.machine():
         params["console_device"] = 'hvc0'
 
+    update_console = params.get("update_console", "yes")
```
chunfuwen: this will lead to update_console is always 'yes'

PR #2083, opened 2019-03-20
Fix Video driver checking in Xorg log for rhel8 guest

xiaodwan:
In rhel8 guest, when it boots up, no 'qxl' in /var/log/Xorg.0.log.
We just check if it contains 'qxl' by lspci command.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -232,10 +232,15 @@ class VMChecker(object):
         vm_xorg_log = self.checker.get_vm_xorg()
         if vm_xorg_log:
             if expect_video not in vm_xorg_log:
-                err_msg = "Not find %s in Xorg log", expect_video
-                self.log_err(err_msg)
-            else:
-                logging.info("PASS")
+                err_msg = "Not find %s in Xorg log" % expect_video
+                logging.info(err_msg)
+                # RHEL8 desn't include any qxl string in xorg log.
+                # If expect_video is in lspci output, we think it passed.
+                if re.search(expect_video, pci_devs, re.IGNORECASE) is None:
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.display.spice.kvm.output_mode.rhev: PASS (1150.72 s)

xiaodwan:
@chloerh @chunfuwen please help to review this pr, thx.

provider/v2v_vmcheck_helper.py
```diff
@@ -232,10 +232,15 @@ def check_linux_vm(self):
...
-            else:
-                logging.info("PASS")
+                err_msg = "Not find %s in Xorg log" % expect_video
+                logging.info(err_msg)
+                # RHEL8 desn't include any qxl string in xorg log.
+                # If expect_video is in lspci output, we think it passed.
+                if re.search(expect_video, pci_devs, re.IGNORECASE) is None:
+                    err_msg += " And Not find %s device by lspci" % expect_video
+                    self.log_err(err_msg)
+                    return
```
chunfuwen: I assume you don't need return

xiaodwan: @chunfuwen if no return, the logging.info('pass') will be printed when checkpoint is failed.  It can avoid confusing logs.

chunfuwen: why need logging.info('pass') here. By default, if not error or test.fail, framework itself will display 'pass' result

xiaodwan: @chunfuwen I think it means this checkpoint is pass, not the whole case. The log info can show some code running paths when debugging.

PR #2092, opened 2019-03-25
Update checkpoint for device.map if converting from KVM

xiaodwan:
If the image's OS system is installed with bus=IDE, the device.map
will be '(hd0) /dev/sda'. Then if users create a new vm by this image,
and specify the bus to virtio, and after converting by virt-v2v,
v2v will give a warning and doesn't update /dev/sda to /dev/vda in
device.map.
This is a scenario only occurs when converting from KVM. Let's update
the checkpoint and just give a warning instead of fail the case for
this scenario.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -218,7 +218,17 @@ class VMChecker(object):
             logging.info("The guest is uefi mode,skip the checkpoint")
         elif not self.checker.get_grub_device():
             err_msg = "Not find vd? in disk partition"
-            self.log_err(err_msg)
+            if self.hypervisor != 'kvm':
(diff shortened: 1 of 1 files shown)


xiaodwan:
KVM to RHV is not a valid scenario and those related checks have already been covered with ESX cases, It's fine even we skip the who checking for device.map.  This patch doesn't affect any other hypervisor( xen, esx) testing.

xiaodwan:
@chloerh @chunfuwen Please help to review this patch, thanks.

xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.firewalld.host_start.kvm.output_mode.rhev: PASS (1144.74 s)

PR #2096, opened 2019-03-27
add subcase to test case vcpu_affinity

brucepatricklee:
set vcpu affinity for none exists vcpu and check xml result

### DIFF
--- a/libvirt/tests/src/cpu/vcpu_affinity.py
+++ b/libvirt/tests/src/cpu/vcpu_affinity.py
@@ -5,6 +5,7 @@ from avocado.utils import process
 
 from virttest import virsh
 from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml import xcepts
 from virttest.utils_test import libvirt
 from virttest import utils_hotplug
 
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/cpu/vcpu_affinity.py
```diff
@@ -168,6 +178,9 @@ def check_vcpu_affinity():
                 err_msg = err_msg.split(";")
             libvirt.check_result(result_to_check, err_msg)
 
+    except xcepts.LibvirtXMLError:
+        pass
+
```
kylazhang: Please check this part for exception catch as we discussed.

brucepatricklee:
@kylazhang code updated with trying except xcepts.LibvirtXMLError in  check_vcpu_affinity(). And I want to add some explanation about exception xcepts.LibvirtXMLError. We can the definition below:
class LibvirtXMLError(Exception):

    """
    Error originating within libvirt_xml module
    """

Therefore, it's safe to use xcepts.LibvirtXMLError here as only libvirt_xml module will send this kind of exception.

brucepatricklee:
@chloerh Please help to review, thanks

kylazhang:
OK, got it. Thanks

PR #2109, opened 2019-04-02
fix mem_align test scenario issue.

harihare:
mem_align does hotplug of memory and trys to check the qemu_cmd_line with out
setting the parameter  ``test_qemu_cmd`` in the config file. More over
test_qemu_cmd should be enbled only for the coldplug not for hotplug, hotplug
will not update the qemu command line.

Removed the check itself, because if  ``test_qemu_cmd`` is set the qemu command line
is verified in the generic flow. 

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -522,7 +522,6 @@ def run(test, params, env):
 
         if mem_align:
             dom_mem = check_mem_align()
-            check_qemu_cmd(dom_mem['maxMemory'], dom_mem['attached_mem'])
             if hot_plug and params['delta'] != dom_mem['attached_mem']:
                 test.fail('Memory after attach not equal to original mem + attached mem')
 


harihare:
**### Before Patch :** 
(2/2) guest_memhotplug.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.Ubuntu.genesis-18.04.ppc64le.powerkvm-libvirt.libvirt_mem.positive_test.align_256m.hot_plug: \^[  FAIL: Qemu command check fail.

**### After Patch :**
(2/2) guest_memhotplug.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.Ubuntu.genesis-18.04.ppc64le.powerkvm-libvirt.libvirt_mem.positive_test.align_256m.hot_plug: PASS (53.49 s)

chloerh:
@dzhengfy  This pr will remove the qemu command line check step, would you like to review it?  

bz 1249006

libvirt/tests/src/libvirt_mem.py
```diff
@@ -522,7 +522,6 @@ def modify_domain_xml():
 
         if mem_align:
             dom_mem = check_mem_align()
-            check_qemu_cmd(dom_mem['maxMemory'], dom_mem['attached_mem'])
```
dzhengfy: This line is necessary for other non-hotlug cases which need verify the memory 256 alignment.  And the generic checking using  check_qemu_cmd in L530 which you mentioned will not be run for those align_256m cases. So simply removing this line will have side effect for other existing align_256m cases, like cold plug.

So here is the suggestion:
1.  Change L525 to:
if not hot_plug:
    check_qemu_cmd(dom_mem['maxMemory'], dom_mem['attached_mem'])   
or
2. Close this PR and then the last author of this part codes would like to fix this issue.

Any comment?

dzhengfy:

PR #2122, opened 2019-04-11
Add new case of cpu management

chloerh:
- Verify hv_frequencies, hv_reenlightenment and hv-tlbflush
 enlightenments works

depends on https://github.com/avocado-framework/avocado-vt/pull/2037

### DIFF
--- a/libvirt/tests/cfg/cpu/vm_features.cfg
+++ b/libvirt/tests/cfg/cpu/vm_features.cfg
@@ -1,0 +1,15 @@
+- vm_features:
+    type = vm_features
+    start_vm = 'no'
+    no pseries
+    variants:
+        - positive_test:
+            variants:
+                - hyperv:
+                    variants:
+                        - tlbflush:
+                            hyperv_attr = {'tlbflush': 'on'}
+                        - frequencies:
+                            hyperv_attr = {'relaxed': 'on', 'vapic': 'on', 'vpindex': 'on', 'frequencies': 'on'}
+                        - reenlightenment:
+                            hyperv_attr = {'reenlightenment': 'on'}
--- a/libvirt/tests/src/cpu/vm_features.py
+++ b/libvirt/tests/src/cpu/vm_features.py
@@ -1,0 +1,49 @@
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_test import libvirt
+
+from provider import libvirt_version
+
+
+def run(test, params, env):
+    """
+    Test vm features
+    """
+    vm_name = params.get('main_vm')
+    vm = env.get_vm(vm_name)
+
+    hyperv_attr = eval(params.get('hyperv_attr', '{}'))
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    try:
+
+        # Set hyperv attribs if there're attribs to set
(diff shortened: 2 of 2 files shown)


chloerh:
Test result on q35 vm:
---
 (1/3) type_specific.io-github-autotest-libvirt.vm_features.positive_test.hyperv.tlbflush: PASS (19.46 s)
 (2/3) type_specific.io-github-autotest-libvirt.vm_features.positive_test.hyperv.frequencies: PASS (21.25 s)
 (3/3) type_specific.io-github-autotest-libvirt.vm_features.positive_test.hyperv.reenlightenment: PASS (21.31 s)
JOB TIME   : 63.98 s

libvirt/tests/src/cpu/vm_features.py
```diff
@@ -0,0 +1,52 @@
...
+                if not libvirt_version.version_compare(5, 0, 0):
+                    test.cancel('This version of libvirt does\'nt support '
+                                'the setting: %r' % hyperv_attr)
+
+            features_xml = vm_xml.VMFeaturesXML()
+            for attr, value in hyperv_attr.items():
+                setattr(features_xml, 'hyperv_%s_state' % attr, value)
+            logging.debug('New features_xml: %s', features_xml)
+            vmxml.features = features_xml
+            vmxml.sync()
```
kylazhang: It should be more common in VMFeaturesXML as class function, add a new one, or have existing similar function?

libvirt/tests/cfg/cpu/vm_features.cfg
```diff
@@ -0,0 +1,16 @@
+- vm_features:
+    type = vm_features
+    start_vm = 'no'
+    no pseries
+
```
kylazhang: do not need the blank line here

libvirt/tests/src/cpu/vm_features.py
chloerh: add_feature(self, name, attr_name='', attr_value=''):  
could do part, but it can set only 1 pair of attr:value in one feature since it will remove the existing feature first before setting if we try to set a second attrib

libvirt/tests/cfg/cpu/vm_features.cfg
chloerh: Updated

libvirt/tests/src/cpu/vm_features.py
```diff
@@ -0,0 +1,52 @@
...
+
+def run(test, params, env):
+    """
+    Test vm features
+    """
+    vm_name = params.get('main_vm')
+    vm = env.get_vm(vm_name)
+
+    hyperv_attr = eval(params.get('hyperv_attr', '{}'))
+
```
brucepatricklee: ast.literal_eval is recommended even if the string is not got from input

chloerh: I don't think that's necessary since we're not doing any damage to system with the var 'hyperv_attr', and the value of this var is completely under control. I guess that's one choice.

libvirt/tests/src/cpu/vm_features.py
chloerh: @kylazhang is it ok?

libvirt/tests/src/cpu/vm_features.py
brucepatricklee: good

PR #2123, opened 2019-04-11
virsh_event: Add wait time

dzhengfy:
Some virsh options need more seconds to finish completely. So this is to
add some seconds to make the script more stable.


kylazhang:
Can we use wait_for + timeout instead of sleep?

dzhengfy:
I decide to close this PR because this fix is not necessary for all architectures.

PR #2131, opened 2019-04-16
Fix a npiv scsi hostdev hotplug issue

lento-sun:
Sometimes the script cannot get the attached disk in vm just after
the hotpluging, the disk will show up after a short time. So change
to use a wait_for() to make sure attached disk can be captured by
script.

### DIFF
--- a/libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
+++ b/libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
@@ -37,9 +37,13 @@ def run(test, params, env):
         :return: True if check successfully.
        """
         try:
-            session = vm.wait_for_login()
-            new_parts = utils_disk.get_parts_list(session)
-            added_parts = list(set(new_parts).difference(set(old_parts)))
+            def get_attached_disk():
+                session = vm.wait_for_login()
+                new_parts = utils_disk.get_parts_list(session)
+                session.close()
+                added_parts = list(set(new_parts).difference(set(old_parts)))
+                return added_parts
+            added_parts = utils_misc.wait_for(get_attached_disk, _TIMEOUT)
             logging.info("Added parts:%s", added_parts)
             if len(added_parts) != 1:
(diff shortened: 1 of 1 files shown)


lento-sun:
test with latest rhel7 and rhel8, 100% passed.
The original code will have a 20% failure rate.

libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
```diff
@@ -36,9 +36,13 @@ def check_in_vm(vm, target, old_parts):
...
-            session = vm.wait_for_login()
-            new_parts = libvirt.get_parts_list(session)
-            added_parts = list(set(new_parts).difference(set(old_parts)))
+            def get_attached_disk():
+                session = vm.wait_for_login()
+                new_parts = libvirt.get_parts_list(session)
+                session.close()
+                added_parts = list(set(new_parts).difference(set(old_parts)))
+                return added_parts
+            added_parts = utils_misc.wait_for(get_attached_disk, _TIMEOUT)
```
chloerh: Timeout is 5s. Should it be longer, or 5s is already more than enough? Since it is a 'timeout', it should be a bit longer than the average value.

chloerh: Timeout is 5s. Should it be longer, or 5s is already more than enough? Since it is a 'timeout', it should be a bit longer than the average value.

lento-sun: the _TIMEOUT used in this case is to wait for: attach disk / create vhba
The normal time spent by these actions should be measured by 'ms' but not 's'. So even a _TIMEOUT=1s should work. 5s is long enough, and if action does not succeed within 5s, we should checkout what's wrong with the test.

chloerh: Great!

PR #2137, opened 2019-04-18
Modify nfv cases to support both q35 and pc machine_type

yafu-1:

### DIFF
--- a/libvirt/tests/cfg/sriov/sriov.cfg
+++ b/libvirt/tests/cfg/sriov/sriov.cfg
@@ -56,7 +56,7 @@
                     no scalability_test, vf_pool
                 - nop:
                 - nfv:
-                    only q35
+                    only q35, i440fx
                     only normal_test..vf.managed_yes
                     nfv = "yes"
                     controller_index = 12
--- a/libvirt/tests/src/sriov/sriov.py
+++ b/libvirt/tests/src/sriov/sriov.py
@@ -717,12 +717,14 @@ def run(test, params, env):
             result = virsh.net_create(netxml.xml, ignore_status=True, debug=True)
             utils_test.libvirt.check_exit_status(result, expected_error)
         if nfv:
-            for os_machine_type in (machine_type, vmxml.os.machine):
-                'q35' in os_machine_type or test.cancel("nfv only run with q35 machine type")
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/sriov/sriov.py
```diff
@@ -717,12 +717,14 @@ def remove_devices(vmxml, device_type):
...
             utils_test.libvirt.check_exit_status(result, expected_error)
         if nfv:
-            for os_machine_type in (machine_type, vmxml.os.machine):
-                'q35' in os_machine_type or test.cancel("nfv only run with q35 machine type")
             vf_driver = os.readlink(os.path.join(pci_device_dir, vf_list[0], "driver")).split('/')[-1]
             vmxml.remove_all_device_by_type('controller')
             remove_devices(vmxml, 'address')
             remove_devices(vmxml, 'usb')
+            osxml = vmxml.os
+            if "i440fx" in vmxml.os.machine:
```
kylazhang: what about "pc" ?

libvirt/tests/cfg/sriov/sriov.cfg
```diff
@@ -56,7 +56,6 @@
                     no scalability_test, vf_pool
                 - nop:
                 - nfv:
-                    only q35
```
kylazhang: so all the architecture will support? what about only i440fx, q35?

chloerh:
LGTM

libvirt/tests/src/sriov/sriov.py
yafu-1: q35 machine type also contains 'pc', such as: pc-q35-rhel7.6.0

libvirt/tests/cfg/sriov/sriov.cfg
yafu-1: No, only support i440fx and q35. I will modify the cfg file.

libvirt/tests/src/sriov/sriov.py
kylazhang: I mean the machine_type='pc' which means default/alias of i440fx

kylazhang: I know it. Thanks

PR #2144, opened 2019-04-23
Start guest with direct type interface

yalzhang:
Start guest with direct type interface when there are existing macvtap
devices created by other than libvirt.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_network.py
+++ b/libvirt/tests/src/virtual_network/iface_network.py
@@ -555,6 +555,7 @@ TIMEOUT 3"""
     boot_failure = "yes" == params.get("boot_failure", "no")
     ipt_rules = []
     ipt6_rules = []
+    define_macvtap = "yes" == params.get("define_macvtap", "no")
 
     # Destroy VM first
     if vm.is_alive():
@@ -606,6 +607,11 @@ TIMEOUT 3"""
                              forward['dev'], net_ifs[0])
                 forward['dev'] = net_ifs[0]
                 params["net_forward"] = str(forward)
+            if define_macvtap:
+                for i in [0, 2, 4]:
+                    cmd = "ip l add li %s macvtap%s type macvtap" % (forward['dev'], i)
+                    process.run(cmd, shell=True, verbose=True)
+                process.run("ip l", shell=True, verbose=True)
             forward_iface = params.get("forward_iface")
             if forward_iface:
                 interface = [x for x in forward_iface.split()]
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -496,6 +496,10 @@ def run_guest_libvirt(session):
         if not utils_package.package_remove("libvirt*", session):
             test.error("Failed to remove libvirt packages on guest")
 
+    def expected_macvtaps():
+        return {'macvtap0', 'macvtap1', 'macvtap2', 'macvtap3', 'macvtap4',
+                'macvtap5', 'macvtap6', 'macvtap7'}
```
chloerh: First, this could be a var.
Second, set(['macvtap' + str(x) for x in range(8)])

yalzhang: awesome! updated.

libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -606,6 +607,11 @@ def run_guest_libvirt(session):
                              forward['dev'], net_ifs[0])
                 forward['dev'] = net_ifs[0]
                 params["net_forward"] = str(forward)
+            if define_macvtap:
+                for i in [0, 2, 4]:
+                    cmd = "ip l add li %s macvtap%s type macvtap" % (forward['dev'], i)
+                    process.run(cmd, shell=True, verbose=True)
+                process.run("ip l", shell=True, verbose=True)
```
dzhengfy: What's purpose for this line? Do you need to check the list result? 
If you don't intend to check the result, this line can be removed.

libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -798,6 +804,15 @@ def run_guest_libvirt(session):
...
                 test.fail("VM started unexpectedly")
+            if define_macvtap:
+                cmd = "ls /sys/devices/virtual/net"
+                output = process.run(cmd, shell=True, verbose=True).stdout_text
+                macvtap_list = re.findall(r'macvtap0|macvtap1|macvtap2|macvtap3'
+                                          r'|macvtap4|macvtap5|macvtap6|macvtap7',
+                                          output)
+                logging.debug("The macvtap_list is %s" % macvtap_list)
+                if set(macvtap_list) != set(['macvtap' + str(x) for x in range(8)]):
+                    test.fail("The macvtap_list is %s" % macvtap_list)
```
dzhengfy: This log is not informative enough. It is better to describe more about the error.

libvirt/tests/src/virtual_network/iface_network.py
yalzhang: The purpose is to pre-define some macvtap device out of libvirt and name it like macvtap0, when start the guest, we can test if libvirt can skip the pre-defined macvtap0 and create device named like macvtap1. It is a test case from a bug that guest fail to start as the name "macvtap0" is occupied. There is check in line 811, and guest can start successfully is also a check point.

libvirt/tests/src/virtual_network/iface_network.py
yalzhang: Thank you for the suggestion, update to "test.fail("Existing macvtap device %s is not expected! should be macvtap(0-7)" % macvtap_list)"

libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -798,6 +804,15 @@ def run_guest_libvirt(session):
             vm.start()
             if start_error:
                 test.fail("VM started unexpectedly")
+            if define_macvtap:
+                cmd = "ls /sys/devices/virtual/net"
+                output = process.run(cmd, shell=True, verbose=True).stdout_text
+                macvtap_list = re.findall(r'macvtap0|macvtap1|macvtap2|macvtap3'
```
chunfuwen: Can we use general match characters ,rather than listing all here?

libvirt/tests/src/virtual_network/iface_network.py
dzhengfy: Thanks

PR #2168, opened 2019-05-15
[timer] Added the case of timer migration

iccaszhulili:
Migrate the timer device, and the check the vm and
source host time after migration

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -874,6 +874,25 @@ def run(test, params, env):
             if int(result.exit_status) != 0:
                 test.fail(result.stderr_text.strip())
 
+    def time_diff_between_vm_host(localvm=True):
+        """
+        check the time difference between vm and source host
+        :param localvm: True if vm is not migrated yet
+        """
+        if localvm:
+            vm_time = virsh.domtime(vm_name, debug=True).stdout
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+        else:
+            remote_virsh_session = virsh.VirshPersistent(**remote_virsh_dargs)
+            vm_time = remote_virsh_session.domtime(vm_name, debug=True).stdout
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+            remote_virsh_session.close_session()
+
+        host_time_value = int(time.time())
+        time_diff = host_time_value - vm_time_value
+        logging.debug("Time difference between source host and vm is %s", time_diff)
+        return time_diff
+
     check_parameters(test, params)
 
     # Params for NFS shared storage
@@ -923,6 +942,7 @@ def run(test, params, env):
     stress_in_vm = "yes" == params.get("stress_in_vm", "no")
     low_speed = params.get("low_speed", None)
     migr_vm_back = "yes" == params.get("migr_vm_back", "no")
+    timer_migration = "yes" == params.get("timer_migration")
 
     remote_virsh_dargs = {'remote_ip': server_ip, 'remote_user': server_user,
                           'remote_pwd': server_pwd, 'unprivileged_user': None,
@@ -1215,6 +1235,9 @@ def run(test, params, env):
                                              args=())
             stress_thread.start()
 
+        if timer_migration:
+            source_vm_host_time_diff = time_diff_between_vm_host(localvm=True)
+
         if extra.count("timeout-postcopy"):
             func_name = check_timeout_postcopy
         if params.get("actions_during_migration"):
@@ -1444,6 +1467,11 @@ def run(test, params, env):
                 test.fail("Failed to run '%s' on remote: %s"
                           % (cmd, cmd_result))
 
+        if timer_migration:
+            target_vm_host_time_diff = time_diff_between_vm_host(localvm=False)
+            if abs(target_vm_host_time_diff - source_vm_host_time_diff) > 1:
+                test.fail("The difference of target_vm_host_time_diff and source_vm_host_time_diff"
+                          "should not more than 1 second")
     except exceptions.TestFail as details:
         is_TestFail = True
         test_exception = details
--- a/libvirt/tests/cfg/migration/migrate_options_shared.cfg
+++ b/libvirt/tests/cfg/migration/migrate_options_shared.cfg
@@ -355,6 +355,11 @@
                             src_secret_value = "sec value test"
                             dst_secret_value = ${src_secret_value}
                             cmd_in_vm_after_migration = "tpm2_getrandom 10"                                                   
+                - timer:
+                    timer_migration = "yes"
+                    asynch_migration = "yes"
+                    actions_during_migration = "setmaxdowntime"
+                    migrate_maxdowntime = "30"
         - negative_test:
             virsh_migrate_options = "--live --verbose"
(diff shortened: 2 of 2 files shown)


iccaszhulili:
# avocado run --vt-type libvirt migrate_options_shared..timer
JOB ID     : 9dea6c860939f92b85c4e7764ea8399f6b1afe4d
JOB LOG    : /root/avocado/job-results/job-2019-05-15T14.45-9dea6c8/job.log
 (1/2) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.timer.with_postcopy: PASS (120.76 s)
 (2/2) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.timer.without_postcopy: PASS (119.93 s)
JOB TIME   : 242.62 s

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -491,6 +511,7 @@ def check_migration_res(result):
 
     hpt_resize = params.get("hpt_resize", None)
     htm_state = params.get("htm_state", None)
+    timer_migration = params.get("timer_migration", None) == "yes"
```
chloerh: I don't think the 'None' here is necessary, but it looks good with the codes above. Maybe we could pay attention to it in the future :stuck_out_tongue_winking_eye:

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -772,6 +796,12 @@ def check_migration_res(result):
                                                    r"[\#\$]\s*$")
             check_vm_network_accessed(server_session)
             server_session.close()
+
+        if timer_migration:
+            target_vm_host_time_diff = time_diff_between_vm_host(localvm=False)
+            if abs(target_vm_host_time_diff - source_vm_host_time_diff) > 1:
+                test.fail("The difference of target_vm_host_time_diff and source_vm_host_time_diff"
+                          "should not more than 1 second")
```
chloerh: The difference between, 
should not be more.
Forgive me, but I've been trying to practice English everywhere.

chloerh:
I'll unsubscribe this pr. So... I don't care what will happen :wink:

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -445,6 +445,26 @@ def check_migration_res(result):
             if int(result.exit_status) != 0:
                 test.fail(results_stderr_52lts(result).strip())
 
+    def time_diff_between_vm_host(localvm=True):
+        """
+        check the time difference between vm and source host
+        :param localvm: True if vm is not migrated yet
+        """
+        if localvm:
+            vm_time = virsh.domtime(vm_name, debug=True, ignore_status=True).stdout
```
dzhengfy: As you use 'ignore_status=True', there will be no exception raised when this command execution fails. So we'd better check the exit_status.

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -445,6 +445,26 @@ def check_migration_res(result):
...
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+        else:
+            remote_virsh_session = virsh.VirshPersistent(**remote_virsh_dargs)
+            vm_time = remote_virsh_session.domtime(vm_name, debug=True,
```
dzhengfy: Also please check your exit_status

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -445,6 +445,26 @@ def check_migration_res(result):
...
+                                                   ignore_status=True).stdout
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+            remote_virsh_session.close_session()
+
+        host_time_value = int(time.time())
+        time_diff = host_time_value - vm_time_value
+        logging.debug("Time difference between source host and vm is %s", time_diff)
```
dzhengfy: As we also support vm on target host, 'source host' is not accurate, right?

libvirt/tests/src/migration/migrate_options_shared.py
dzhengfy: Usually we use  `timer_migration = "yes" == params.get("timer_migration", "no")` which is much meaningful. Or you can use `timer_migration = params.get("timer_migration")` which is enough. And yes, here 'None' is useless.

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -445,6 +445,26 @@ def check_migration_res(result):
...
+            vm_time = virsh.domtime(vm_name, debug=True, ignore_status=True).stdout
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+        else:
+            remote_virsh_session = virsh.VirshPersistent(**remote_virsh_dargs)
+            vm_time = remote_virsh_session.domtime(vm_name, debug=True,
+                                                   ignore_status=True).stdout
+            vm_time_value = int(vm_time.strip().split(":")[-1])
+            remote_virsh_session.close_session()
+
+        host_time_value = int(time.time())
```
dzhengfy: Question is:
When you diff the time for source vm and target vm, you both use the time on source host instead of using source host and target host respectively. Is It your intention?

libvirt/tests/src/migration/migrate_options_shared.py
dzhengfy: Should '1' second be a must to follow? Otherwise we can configure it in cfg file for future tuning.

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: not quite clear about this question

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: Yes, the feature is designed like this

libvirt/tests/src/migration/migrate_options_shared.py
dzhengfy: OK, after discussion, I agree with the log message.

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: updated

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: the difference between the 2 kinds of time is rounding. Ideally, we hope the time is the same, but because of rounding, there will be some difference between this 2 kinds of time. It is not a parameter, so I think it is not good to configure it in cfg file.

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: updated

libvirt/tests/src/migration/migrate_options_shared.py
iccaszhulili: updated

dzhengfy:
@iccaszhulili please help fix the conflict.

iccaszhulili:
@dzhengfy conflict resolved, please help to review it again, Thanks

PR #2179, opened 2019-05-24
 libvirt_pci_passthrough: refactor tests to use `host_pf_filter` for device pass-through.

santwana:
The PciAssignable class uses the "host_pf_filter" for the PCI Devices
to be pass-through'ed, and the same can be used in the test-case.
Now the user also have the flexibilty to create the desired no.of Virtual Functions,
for each Physical Function incase of SR-IOV.

Removed the device_name params for comparision of the PCI devices
before/after passthrough.

### DIFF
--- a/libvirt/tests/cfg/libvirt_pci_passthrough.cfg
+++ b/libvirt/tests/cfg/libvirt_pci_passthrough.cfg
@@ -5,6 +5,10 @@
             libvirt_pci_SRIOV = no
         - SRIOV:
             libvirt_pci_SRIOV = yes
+            vf_filter = "Virtual Function"
+            # Enter the no.of Virtual Function's to be created
+            # for each Physical Function.
+            number_vfs = 4
     variants:
         - NIC:
             libvirt_pci_device_type = "NIC"
@@ -12,8 +16,8 @@
             # a network device. We will attach this
             # device to guest. Then this network device
             # will be unavailable on host.
-            # E.g: pci_0000_05_00_0
-            libvirt_pci_net_dev_label = "ENTER.YOUR.PCI.LABEL"
+            # E.g: 0000:05:00.0
+            pf_filter = "ENTER.YOUR.PCI.LABEL"
             # Please enter the ip what is used by the device
             # you are going to attach to guest.
             libvirt_pci_net_ip = "ENTER.YOUR.IP"
@@ -21,7 +25,6 @@
             # We need to ping it after attaching pci device
             # to guest to verify this device works well in guest.
             libvirt_pci_server_ip = "ENTER.YOUR.SERVER.IP"
-            libvirt_pci_net_dev_name = "ENTER.YOUR.DEVICE.NAME"
             # Enter netmask in CIDR notation
             libvirt_pci_net_mask = "ENTER.YOUR.NETMASK"
         - STORAGE:
--- a/libvirt/tests/src/libvirt_pci_passthrough.py
+++ b/libvirt/tests/src/libvirt_pci_passthrough.py
@@ -1,4 +1,3 @@
-import re
 import logging
 import netaddr
 from virttest import utils_test
@@ -15,9 +14,9 @@ def run(test, params, env):
     a). NIC:
         1. Get params.
         2. Get the pci device for specific net_name.
-        3. Attach pci device to guest.
-        4. Start guest and set the ip to all the physical functions.
-        5. Ping to server_ip from each physical function
+        3. Attach Physical Function's/Virtual Function's to single guest
+        4. Start guest and set the ip to all the functions.
+        5. Ping to server_ip from each function
            to verify the new network device.
     b). STORAGE:
         1. Get params.
@@ -34,13 +33,13 @@ def run(test, params, env):
     vm = env.get_vm(vm_name)
     sriov = ('yes' == params.get("libvirt_pci_SRIOV", 'no'))
     device_type = params.get("libvirt_pci_device_type", "NIC")
+    vm_vfs = int(params.get("number_vfs", 2))
     pci_dev = None
-    device_name = None
     pci_address = None
     bus_info = []
     if device_type == "NIC":
-        pci_dev = params.get("libvirt_pci_net_dev_label")
-        device_name = params.get("libvirt_pci_net_dev_name", "None")
+        pf_filter = params.get("pf_filter", "0000:01:00.0")
+        vf_filter = params.get("vf_filter", "Virtual Function")
     else:
         pci_dev = params.get("libvirt_pci_storage_dev_label")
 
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -192,4 +171,6 @@ def run(test, params, env):
         backup_xml.sync()
         # For SR-IOV , VF's should be cleaned up in the post-processing.
         if sriov:
-            sriov_setup.release_devs()
+            if obj.get_vfs_count() != 0:
+                for clean in pci_ids:
+                    obj.set_vf(clean)
```
sathnaga: minor though, for readability, we can pass  `vf_no="0" ` explicitly to say we are reseting/cleaning...

I would prefer having a
```
def  reset_vf(pci_pf):
     set_vf(pci_pf, vf_no="0")
```

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -192,4 +171,6 @@ def run(test, params, env):
         backup_xml.sync()
         # For SR-IOV , VF's should be cleaned up in the post-processing.
         if sriov:
-            sriov_setup.release_devs()
+            if obj.get_vfs_count() != 0:
+                for clean in pci_ids:
```
sathnaga: `for pci_pf in pci_ids:` would be better

libvirt/tests/cfg/libvirt_pci_passthrough.cfg
```diff
@@ -12,8 +12,9 @@
             # a network device. We will attach this
             # device to guest. Then this network device
             # will be unavailable on host.
-            # E.g: pci_0000_05_00_0
-            libvirt_pci_net_dev_label = "ENTER.YOUR.PCI.LABEL"
+            # E.g: 0000:05:00.0
+            pf_filter = "ENTER.YOUR.PCI.LABEL"
+            vf_filter = "Virtual Function"
```
sathnaga: move `vf_filter` and `nvf` param under SRIOV variant, it makes more sense there

libvirt/tests/cfg/libvirt_pci_passthrough.cfg
```diff
@@ -24,6 +25,9 @@
             libvirt_pci_net_dev_name = "ENTER.YOUR.DEVICE.NAME"
             # Enter netmask in CIDR notation
             libvirt_pci_net_mask = "ENTER.YOUR.NETMASK"
+            # Enter the no.of Virtual Function's to be created
+            # for each Physical Function.
+            nvf = 3
```
sathnaga: this param can be little expanded, like, `number_vfs = 3`

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -66,37 +68,23 @@ def run(test, params, env):
...
         nic_list_before = output.splitlines()
+        obj = PciAssignable(pf_filter_re=pf_filter,
+                            vf_filter_re=vf_filter)
+        # get all functions id's
+        pci_ids = obj.get_same_group_devs(pf_filter)
+        pci_devs = []
+        for val in pci_ids:
+            temp = val.replace(":", "_")
+            pci_devs.extend(["pci_"+temp])
         if sriov:
```
sathnaga: change is not part of this patch, though we can handle now choosing sriov by `number_vfs` or `vf_filter` instead of additional param.

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -177,7 +156,7 @@ def run(test, params, env):
                 logging.info(o_ping)
                 if s_ping != 0:
                     err_msg = "Ping test fails, error info: '%s'"
-                    test.fail(err_msg % o_ping)
+                    logging.debug(err_msg % o_ping)
```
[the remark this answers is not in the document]
santwana: @sathnaga  Under certain circumstances if the network setting's are not proper, its better to print the error log, rather than failing the test-case.

libvirt/tests/src/libvirt_pci_passthrough.py
santwana: @sathnaga Addressed the changes suggested, kindly review.

libvirt/tests/cfg/libvirt_pci_passthrough.cfg
santwana: @sathnaga changed as suggested.

libvirt/tests/cfg/libvirt_pci_passthrough.cfg
santwana: @sathnaga changed as suggested.

sathnaga:
On Thu, May 30, 2019 at 02:25:18AM -0700, santwana wrote:
> santwana commented on this pull request.
> 
> 
> 
> > @@ -177,7 +156,7 @@ def run(test, params, env):
>                  logging.info(o_ping)
>                  if s_ping != 0:
>                      err_msg = "Ping test fails, error info: '%s'"
> -                    test.fail(err_msg % o_ping)
> +                    logging.debug(err_msg % o_ping)
> 
> @sathnaga  Under certain circumstances if the network setting's are not proper, its better to print the error log, rather than failing the test-case. 

@santwana in that case, better to control via an param like `ignore_status`, and decide via config, as it is a generic testcase and changing test case failure
statement is not a good idea.

santwana:
@sathnaga reverted back to the test.fail , please review. Thanks!!

sathnaga:
@santwana you might want to squash the commits, if not would expect you to revert the test.fail from first commit, coz that is the one changes to debug in first place.

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -144,24 +123,23 @@ def run(test, params, env):
...
             for val in range(len(nic_list)):
                 bus_info.append(str(nic_list[val]).split(' ', 1)[0])
                 nic_list[val] = str(nic_list[val]).split(' ', 1)[0][:-2]
             bus_info.sort()
             if not sriov:
                 # check all functions get same iommu group
                 if len(set(nic_list)) != 1:
-                    test.fail("Multifunction Device passthroughed but "
-                              "functions are in different iommu group")
+                    logging.debug("Multifunction Device passthroughed but "
```
balamuruhans: please raise the fail()

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -28,19 +27,21 @@ def run(test, params, env):
         5. Compare the result of 'fdisk -l' before and after
             attaching storage pci device to guest.
     """
+    def reset_vf(pci_pf):
```
balamuruhans: this method is not required, please call set_vf() directly.

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -144,24 +123,23 @@ def run(test, params, env):
         session = vm.wait_for_login()
         # The Network configuration is generic irrespective of PF or SRIOV VF
         if device_type == "NIC":
-            output = session.cmd_output("lspci -nn")
+            output = session.cmd_output("lspci -D")
```
balamuruhans: `vm.get_pci_devices()`

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -50,53 +51,40 @@ def run(test, params, env):
...
+        test.cancel("Please enter your Storage Adapter details for test.")
     fdisk_list_before = None
     vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
     backup_xml = vmxml.copy()
     if device_type == "NIC":
         if not vm.is_alive():
             vm.start()
         session = vm.wait_for_login()
-        output = session.cmd_output("lspci -nn")
+        output = session.cmd_output("lspci -D")
```
balamuruhans: same here

libvirt/tests/src/libvirt_pci_passthrough.py
```diff
@@ -50,53 +51,40 @@ def run(test, params, env):
...
         nic_list_before = output.splitlines()
+        obj = PciAssignable(pf_filter_re=pf_filter,
+                            vf_filter_re=vf_filter)
+        # get all functions id's
+        pci_ids = obj.get_same_group_devs(pf_filter)
+        pci_devs = []
+        for val in pci_ids:
+            temp = val.replace(":", "_")
+            pci_devs.extend(["pci_"+temp])
```
balamuruhans: we discussed to avoid these code to convert `:` and `_`, please have a look

santwana: @balamuruhans I have already taken care of the config changes from pci_ to : format, the existing code is required for the NodedevXML .

santwana:
@balamuruhans Implemented the changes as suggested, kindly review. Thanks!

balamuruhans:
```

PR #2191, opened 2019-06-11
virsh_domblkerror: set `virt_use_nfs=on` for selinux to permit libvirt

balamuruhans:
Allow selinux to permit libvirt to access file from NFS else virsh
attach-device fails with permission denied issue as selinux blocks
libvirt and testcase will errors out.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -41,17 +41,19 @@ def run(test, params, env):
     rhel = distro.detect().name == 'rhel'
     nfs_service_package = params.get("nfs_service_package", "nfs-kernel-server")
     nfs_service = None
+    selinux_bool = None
     session = None
     selinux_bak = ""
 
     vm = env.get_vm(vm_name)
     if error_type == "unspecified error":
+        selinux_local = params.get("setup_selinux_local", "yes") == "yes"
         if not ubuntu and not rhel:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -209,6 +217,8 @@ def _check_state():
                                                mount_dir=nfs_dir,
                                                export_dir=img_dir,
                                                restore_selinux=selinux_bak)
+            if selinux_bool:
+                selinux_bool.cleanup(keep_authorized_keys=True)
```
hsmj1412:
```
        if self.server_ip:
            # Setup SSH connection
            from virttest.utils_conn import SSHConnection
            self.ssh_obj = SSHConnection(params)
```
The server_ip is not set, so the ssh_obj will not be created. Then, in cleanup function,
```
    def cleanup(self, keep_authorized_keys=False):
        """
        Cleanup SELinux boolean value.
        """
        self.ssh_obj.auto_recover = True
```
The ssh_obj is not used with check, so it will report an error:
AttributeError: 'SELinuxBoolean' object has no attribute 'ssh_obj'
Please help check how to fix it.

balamuruhans:
@hsmj1412 Thank you for reporting, 
I have fixed in https://github.com/avocado-framework/avocado-vt/pull/2158
Also similar thing I have fixed for nfs as well in https://github.com/avocado-framework/avocado-vt/pull/1337

PR #2197, opened 2019-06-13
Remove unsupported cases from pseries tests

chloerh:
Devices are not supported by pseries


libvirt/tests/cfg/virtual_device/watchdog.cfg
```diff
@@ -4,6 +4,7 @@
     take_regular_screendumps = "no"
     variants:
         - model_i6300esb:
+            no pseries
             model = "i6300esb"
         - model_ib700:
```
dzhengfy: ‘ib700’ is also not supported on pseries. Could you help updated?

chloerh: Updated

PR #2201, opened 2019-06-17
migration: Update to check libvirt version both in source and destina

Yingshun:
…tion

In cross migration, the output of domjobinfo in source and destination
may different which makes script fail. The reason is the item "Postcopy
requests" is introduced in libvirt 4.10. So in some case, it returns different
domjobinfo result in source and dest. So need to check the libvirt version in
both to see if "Postcopy requests" should be in expected_list or not

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -18,6 +18,7 @@ from virttest import utils_misc
 from virttest import defaults
 from virttest import data_dir
 from virttest import virsh
+from virttest import libvirt_version
 from virttest import remote
 from virttest import utils_package
 from virttest import xml_utils
@@ -27,7 +28,6 @@ from virttest.utils_test import libvirt
 from virttest.utils_conn import TLSConnection
 from virttest.compat_52lts import results_stdout_52lts, results_stderr_52lts
 from virttest.libvirt_xml.devices.controller import Controller
-from provider import libvirt_version
 
 
 def check_parameters(test, params):
@@ -323,9 +323,8 @@ def run(test, params, env):
                 if expected_value["all_items"][0] == item_key:
                     del expected_value["all_items"][0]
                 else:
-                    test.fail("The item {} should be {}"
-                              .format(item_key,
-                                      expected_value["all_items"][0]))
+                    test.fail("The item '%s' should be '%s'" %
+                              (item_key, expected_value["all_items"][0]))
 
             if item_key in expected_value:
                 item_value = ':'.join(item.strip().split(':')[1:]).strip()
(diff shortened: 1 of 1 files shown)


Yingshun:
**Test result: migration from RHEL 7.7 to 8.0**
(p3-ve) [root@host193 cfg]# avocado run --vt-type libvirt --vt-machine-type q35 virsh.migrate_options_shared.positive_test.check_domjobinfo.without_postcopy --vt-connect-uri qemu:///system
JOB ID     : 08a37402cf9c36a308366fc00011656da53a6394
JOB LOG    : /root/avocado/job-results/job-2019-06-17T09.53-08a3740/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.check_domjobinfo.without_postcopy: PASS (360.53 s)
JOB TIME   : 362.89 s
(p3-ve) [root@host193 cfg]#

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -384,10 +383,19 @@ def check_domjobinfo_output(option="", is_mig_compelete=False):
...
+                                       + ['Total downtime', 'Downtime w/o network']
+                                       + expected_list_during_mig[-1:])
+        expected_list_after_mig_src.insert(3, 'Time elapsed w/o network')
+        expected_list_after_mig_dest = copy.deepcopy(expected_list_after_mig_src)
+
+        # Check version in remote
+        if not expected_list_after_mig_dest.count("Postcopy requests"):
+            remote_session = remote.remote_login("ssh", server_ip, "22", server_user,
+                                                 server_pwd, "#")
+            if libvirt_version.version_compare(4, 10, 0, session=remote_session):
```
dzhengfy: If we only need run 'virsh version' on target host,  we can reuse the existing variable 'remote_virsh_session' instead of using remote_session.

provider/libvirt_version.py
```diff
@@ -24,16 +24,27 @@ def version_compare(major, minor, update):
     :param major: Major version to compare against
     :param minor: Minor version to compare against
     :param update: Update value to compare against
+    :param session: Sellsession for remote/guest machine
```
dzhengfy: Could you consider to use remote_virsh_session?

provider/libvirt_version.py
```diff
@@ -24,16 +24,27 @@ def version_compare(major, minor, update):
...
     :return: True if running version is greater than or
                   equal to the input libvirt version
     """
     global LIBVIRT_LIB_VERSION
 
+    if session:
+        LIBVIRT_LIB_VERSION = 0
```
dzhengfy: Is it better to backup the value before we make it zero? Then after you are done, you can simply recover it to avoid of the calculation in the next invoking?

provider/libvirt_version.py
```diff
@@ -46,7 +57,11 @@ def version_compare(major, minor, update):
             return False
 
     compare_version = major * 1000000 + minor * 1000 + update
+    # Clear LIBVIRT_LIB_VERSION for remote session
+    LIBVIRT_LIB_VERSION_tmp = LIBVIRT_LIB_VERSION
+    if session:
+        LIBVIRT_LIB_VERSION = 0
```
dzhengfy: Just here to restore the old value.

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -323,9 +323,8 @@ def search_jobinfo_output(jobinfo, items_to_check, postcopy_req=False):
                 if expected_value["all_items"][0] == item_key:
                     del expected_value["all_items"][0]
                 else:
-                    test.fail("The item {} should be {}"
-                              .format(item_key,
-                                      expected_value["all_items"][0]))
+                    test.fail("The item '%s' should be '%s'" %
```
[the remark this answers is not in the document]
Yingshun: Some items of domjobinfo output contain a space, so use quotes there.

libvirt/tests/src/migration/migrate_options_shared.py
Yingshun: @dzhengfy  Updated.

Yingshun: @dzhengfy  have to recover it because libvirt_version.py is moved to avocado-vt and no longer use virsh session to check version

Yingshun:
depends on https://github.com/avocado-framework/avocado-vt/pull/2138

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -27,7 +28,7 @@
 from virttest.utils_conn import TLSConnection
 from virttest.compat_52lts import results_stdout_52lts, results_stderr_52lts
 from virttest.libvirt_xml.devices.controller import Controller
-from provider import libvirt_version
+#from provider import libvirt_version
```
dzhengfy: Remove this comment line

libvirt/tests/src/migration/migrate_options_shared.py
dzhengfy: ok

libvirt/tests/src/migration/migrate_options_shared.py
Yingshun: Removed.

PR #2202, opened 2019-06-18
Add log info for disk attachment in block resize cases

lento-sun:
The disk attachment is a key step here. Without the info, it's
hard for debugger to manually check what happened according to
the log.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
@@ -57,7 +57,8 @@ def run(test, params, env):
 
     # Hotplug the image as disk device
     result = virsh.attach_disk(vm_name, source=image_path, target="vdd",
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockresize.py
```diff
@@ -148,7 +149,4 @@ def run(test, params, env):
                           "different from actual size from "
                           "'qemu-img info'")
     finally:
-        virsh.detach_disk(vm_name, target="vdd")
-
-        if os.path.exists(image_path):
-            os.remove(image_path)
```
dzhengfy: Why is this line removed?

lento-sun: the image is generated into tmpdir, no need to recycle it within case, the framework will remove it. And I found sometimes the image removed before the "detach" actually finished, cause vm destroyed with error.

PR #2204, opened 2019-06-19
Add logs and fix a var not defined error

chloerh:
- Add debug logs for future debugging
- One var named 'error' is not defined
- Fix that fail to add input device when it's pseries vms


libvirt/tests/src/virtual_device/input_devices.py
```diff
@@ -94,7 +99,7 @@ def check_qemu_cmd_line():
         if res.exit_status:
             if not status_error:
                 test.fail("Failed to start vm after adding the %s input "
-                          "device xml. Details: %s " % (input_type, error))
+                          "device xml. Details: %s " % (input_type, res.stderr_text))
```
dzhengfy: 'res.stderr_text' should be ''res.stderr"

chloerh: Updated

PR #2223, opened 2019-07-08
Add a switch to control vm checking for gating cases

xiaodwan:

### DIFF
--- a/v2v/tests/cfg/convert_from_file.cfg
+++ b/v2v/tests/cfg/convert_from_file.cfg
@@ -156,6 +156,8 @@
                         - vmx_default:
                             # Default
                             vmx_nfs_src = NFS_ESX67_VMX_V2V_EXAMPLE
+                            # Do not skip vm checking by default
+                            skip_vm_check = no
                             variants:
                                 - win2019:
                                     only esx_67


xiaodwan:
@kylazhang @chloerh please help to review this pr, and merge it if it has no problem.

v2v/tests/cfg/convert_from_file.cfg
```diff
@@ -138,6 +138,9 @@
                     input_mode = "vmx"
                     only source_esx
                     no output_mode.local
+                    # For gating, skipping vm check in openstack because of IP
+                    # Address problem
+                    skip_vm_check = SKIP_VM_CHECK_OPENSTACK
```
sathnaga: `SKIP_VM_CHECK_OPENSTACK` where is this defined?, if this something internal env variable, avoid using it.

you can use it in your local test config and run this test, no need for this change.

xiaodwan: This is a MARCO which will be replaced with the real value based on the real job or the needs how you run the script. The value is define in our libvirt-ci project. You can also find other similar MARCOs in other cfg files in tp-libvirt.

sathnaga: I guess libvirt-ci is internal, but not tp-libvirt, and it is always good practice, not to have internal code here.

sathnaga: am fine to have `skip_vm_check` as a new param, just that keep it a known default value and override with your needs in your `libvirt-ci` project.

balamuruhans: Agree with @sathnaga, @xiaodwan you can always override the param `skip_vm_check` --vt-extra-params or from your test config in the CI setup. this is not right way to do in upstream.

xiaodwan: @sathnaga @balamuruhans Thanks for your suggestion. I updated the code to set default value for skip_vm_check explicitly. Please help to check again.

PR #2224, opened 2019-07-10
hpt:Fix failed cases on x86

chloerh:
- Fix that passing the wrong params to function:set_hpt.
- Update expected error message since the actual error message is
changed.

### DIFF
--- a/libvirt/tests/cfg/papr_hpt.cfg
+++ b/libvirt/tests/cfg/papr_hpt.cfg
@@ -10,7 +10,7 @@
                 - disabled:
                     hpt_attrs = "{'resizing': 'disabled'}"
                     qemu_check = 'resize-hpt=disabled'
-                    error_msg = 'echo: write error: No such device'
+                    error_msg = ['echo: write error: No such device']
                 - enabled:
                     hpt_attrs = "{'resizing': 'enabled'}"
                     qemu_check = 'resize-hpt=enabled'
@@ -45,7 +45,7 @@
 
         - negative_test:
             status_error = yes
-            error_msg = error: unsupported configuration: HPT tuning is only supported for pSeries guests
+            error_msg = ["error: unsupported configuration: The 'hpt' feature is not supported for architecture", "error: unsupported configuration: HPT tuning is only supported for pSeries guests"]
             variants:
                 - not_ppc:
                     no pseries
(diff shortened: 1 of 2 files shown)


chloerh:
Test result:
---
 (1/3) type_specific.local.papr_hpt.negative_test.not_ppc.required: PASS (3.27 s)
 (2/3) type_specific.local.papr_hpt.negative_test.not_ppc.enabled: PASS (3.11 s)
 (3/3) type_specific.local.papr_hpt.negative_test.not_ppc.disabled: PASS (3.25 s)
JOB TIME   : 12.41 s

libvirt/tests/cfg/papr_hpt.cfg
```diff
@@ -45,7 +45,7 @@
 
         - negative_test:
             status_error = yes
-            error_msg = error: unsupported configuration: HPT tuning is only supported for pSeries guests
+            error_msg = error: unsupported configuration: The 'hpt' feature is not supported for architecture
```
sathnaga: how about converting the error into a list, so that we can support for old versions too, coz it may still fail if we run on older version with the previous error msg?

chloerh: I don't think error message of an old version should be tested since we should all update pkgs to latest versions.

sathnaga: @chloerh different distribution version support different versions of packages, so no issues for us to keep older version of error msgs too, that will break the backward compatibility of these tests run in older distro versions for any regression testing.

sathnaga: if you want to be more accurate use ` libvirt_version.version_compare(0, 6, 0)` and verify the error msg accordingly...

chloerh:
@sathnaga Updated, please review

sathnaga:
@chloerh pls update test results...

chloerh:
(1/3) type_specific.local.papr_hpt.negative_test.not_ppc.required: PASS (3.20 s)
 (2/3) type_specific.local.papr_hpt.negative_test.not_ppc.enabled: PASS (2.94 s)
 (3/3) type_specific.local.papr_hpt.negative_test.not_ppc.disabled: PASS (2.96 s)
JOB TIME   : 11.58 s

@sathnaga  thank you for reviewing my pr

PR #2237, opened 2019-07-18
update test case bios

brucepatricklee:
remove the machine varient from test case cfg file

### DIFF
--- a/libvirt/tests/cfg/bios/boot_integration.cfg
+++ b/libvirt/tests/cfg/bios/boot_integration.cfg
@@ -15,6 +15,7 @@
             boot_ref = "order"
     variants:
         - by_ovmf:
+            only q35
             boot_type = "ovmf"
             loader = "/usr/share/OVMF/OVMF_CODE.secboot.fd"
             loader_type = "pflash"
@@ -30,9 +31,3 @@
             bios_reboot_timeout = "1000"
             disk_target_dev = 'hda'
             disk_target_bus = 'ide'
-    variants:
-        - ma_q35:
-            ma_type = "q35"
-        - ma_i440fx:
-            only by_seabios
-            ma_type = "i440fx"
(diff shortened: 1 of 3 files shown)


libvirt/tests/cfg/bios/boot_integration.cfg
```diff
@@ -30,9 +31,3 @@
             bios_reboot_timeout = "1000"
             disk_target_dev = 'hda'
             disk_target_bus = 'ide'
-    variants:
-        - ma_q35:
-            ma_type = "q35"
-        - ma_i440fx:
-            only by_seabios
-            ma_type = "i440fx"
```
kylazhang: Need to add only i440fx in by_seabios

libvirt/tests/cfg/bios/virsh_boot_tseg.cfg
```diff
@@ -47,14 +47,9 @@
...
+            only q35
             boot_type = "ovmf"
             loader = "/usr/share/OVMF/OVMF_CODE.secboot.fd"
             loader_type = "pflash"
         - by_seabios:
             boot_type = "seabios"
-    variants:
-        - ma_q35:
-            only q35
-        - ma_i440fx:
```
kylazhang: In the file line 29 need to remove and add only i440fx

libvirt/tests/cfg/bios/boot_integration.cfg
kylazhang: Here previous setting is only by_seabios but not sure if it needed to test with q35, need to check with feature owner
If not needed, need to add 'only i440fx' in by_seabios part

libvirt/tests/cfg/bios/virsh_boot_tseg.cfg
brucepatricklee: good point, updated

libvirt/tests/cfg/bios/boot_integration.cfg
brucepatricklee: The previous setting means i440fx only support seabios and confirmed with feature owner, the seabios feature support both i440fx and q35 machine

PR #2261, opened 2019-08-01
domiftune: Add detection of crash after command executed

chloerh:
In case test can't continue after crash

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
@@ -2,6 +2,7 @@ import logging
 import re
 
 from virttest import virsh
+from virttest import utils_libvirtd
 from virttest.libvirt_xml import vm_xml
 from provider import libvirt_version
 from virttest.utils_test import libvirt
@@ -196,6 +197,9 @@ def get_domiftune_parameter(params, test):
 
     if status_error == "yes":
         if status:
+            libvirtd = utils_libvirtd.Libvirtd()
+            if not libvirtd.is_running():
+                test.fail('Libvirtd crashed after getting domiftune!')
             logging.info("It's an expected error: %s", result.stderr)
         else:
             test.fail("%d not a expected command return value" %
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
```diff
@@ -256,6 +260,9 @@ def set_domiftune_parameter(params, test):
 
     if status_error == "yes":
         if status:
+            libvirtd = utils_libvirtd.Libvirtd()
```
dzhengfy: There are several places to create libvirtd object. I think we should use single object which is as expected and detect libvirtd auto-restarted issue potentially.

chloerh: @dzhengfy I understand. But if not this way, then a global var 'libvirtd' will be created,  outside 'run' function. I don't feel right. But if it's ok to you, I'll make the change.

dzhengfy: Accept

libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
```diff
@@ -351,5 +358,9 @@ def run(test, params, env):
         get_domiftune_parameter(test_dict, test)
     else:
         set_domiftune_parameter(test_dict, test)
-        ret = virsh.domiftune(vm_name, interface, 'current', '0', '0')
+
+    libvirtd = utils_libvirtd.Libvirtd()
+    libvirtd.restart()
```
kylazhang: Why adding restart libvirtd here? If libvirtd is not running, should it fail?

chloerh: I couldn't remember exactly. But I think it's because some cases could crash libvirtd, and according to the lines that I added above, these cases could be detected and failed. But we don't want a crashed libvirtd to affect later cleanup or other operations, that's why adding a restart.

PR #2273, opened 2019-08-07
Clean up topology/numatune settings in certain test cases

chloerh:
To avoid influence on test results

### DIFF
--- a/libvirt/tests/src/cpu/vcpu_affinity.py
+++ b/libvirt/tests/src/cpu/vcpu_affinity.py
@@ -131,6 +131,12 @@ def run(test, params, env):
         vmxml.vcpu = int(maxvcpu)
         vmxml.current_vcpu = current_vcpu
 
+        # Remove cpu topology to avoid that it doesn't match vcpu count
+        if vmxml.get_cpu_topology():
+            new_cpu = vmxml.cpu
+            del new_cpu.topology
+            vmxml.cpu = new_cpu
+
         # config vcpu cpuset for cpuset range test
         num = 1 if not status_error else 0
         cpuset_new = "0-{},^{}".format(hostcpu_num-num, cpuset_mask)
@@ -140,6 +146,12 @@ def run(test, params, env):
 
         if vcpu_placement:
             vmxml.placement = vcpu_placement
+
+            # Remove numatune node since it will be automatically set
+            # under 'auto' state
+            if vcpu_placement == 'auto':
+                vmxml.xmltreefile.remove_by_xpath('/numatune', remove_all=True)
+                vmxml.xmltreefile.write()
 
         if config_xml == "cputune":
             cputune = vm_xml.VMCPUTuneXML()


libvirt/tests/src/cpu/vcpu_affinity.py
```diff
@@ -131,6 +131,12 @@ def check_vcpu_affinity():
         vmxml.vcpu = int(maxvcpu)
         vmxml.current_vcpu = current_vcpu
 
+        # Remove cpu topology to avoid that it doesn't match vcpu count
```
dzhengfy: The test strategy is to test guest xml elements as much as we can to ensure the test scope. So is it a must to remove 'topology' configuration? or we can consider how to refine the script to support it?

libvirt/tests/src/cpu/vcpu_affinity.py
```diff
@@ -141,6 +147,12 @@ def check_vcpu_affinity():
         if vcpu_placement:
             vmxml.placement = vcpu_placement
 
+            # Remove numatune node since it will be automatically set
+            # under 'auto' state
+            if vcpu_placement == 'auto':
```
dzhengfy: In vcpu_affinity.cfg, there are configurations as below.  Could you try to remove 'numatune_memory_mode' and 'numatune_memory_placement' to see if same result as you expect?
```
                - vcpu_placement:
                    check = "vcpu_placement"
                    config_xml = "vcpu"
                    vcpu_cpuset = "1"
                    variants:
                        - auto:
                            vcpu_placement = "auto"
                            numatune_memory_mode = "strict"
                            numatune_memory_placement = "auto"
```

chloerh: I believe that's what's to test, it's from the case. They're not supposed to be removed.

libvirt/tests/src/cpu/vcpu_affinity.py
chloerh: Since this is an ppc-only failure and the only reason that other archs don't have this failure is that the vmxml doesn't have <topology> tag, if this modification is not acceptable, then I'll close this one and ask the original author to fix as he knows about the logic better.

dzhengfy: We'd better choose better fix strategy which is to make the scripts stronger and support more guest configurations. So please try to enhance.

chloerh: This way of dealing is much more stronger than the original code since it avoids influence of the xml tag 'topology' which is not necessary to this test. I think this is the most economic way.

libvirt/tests/src/cpu/vcpu_affinity.py
```diff
@@ -131,6 +131,12 @@ def check_vcpu_affinity():
         vmxml.vcpu = int(maxvcpu)
         vmxml.current_vcpu = current_vcpu
 
+        # Remove cpu topology to avoid that it doesn't match vcpu count
+        if vmxml.get_cpu_topology():
```
chunfuwen: suggest add constraint to limit to ppc only since it is specific to ppc failure

chloerh: It's actuallly about <topology> which is not ppc specific, therefore no need to limit to 'ppc only'

PR #2287, opened 2019-08-21
Hotpulg differnet Memory size

harihare:
The patch hotplugs memory of different size as listed
[1, 2, 4, 8, 16, 32, 64] all size in GiB.

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -378,6 +378,7 @@ def run(test, params, env):
     align_to_value = int(params.get("align_to_value", "65536"))
     hot_reboot = "yes" == params.get("hot_reboot", "no")
     rand_reboot = "yes" == params.get("rand_reboot", "no")
+
     guest_known_unplug_errors = []
     guest_known_unplug_errors.append(params.get("guest_known_unplug_errors"))
     host_known_unplug_errors = []
@@ -390,6 +391,7 @@ def run(test, params, env):
     tg_node = params.get("tg_node", 0)
     pg_size = params.get("page_size")
     pg_unit = params.get("page_unit", "KiB")
+    diff_tg_size = eval(params.get("diff_tg_size", None))
     node_mask = params.get("node_mask", "0")
     mem_addr = ast.literal_eval(params.get("memory_addr", "{}"))
     huge_pages = [ast.literal_eval(x)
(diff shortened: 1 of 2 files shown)


harihare:
avocado run libvirt_mem.positive_test.diff_size --vt-type libvirt --vt-guest-os RHEL.8.1.ppc64le --vt-extra-params create_vm_libvirt=yes kill_vm_libvirt=yes env_cleanup=yes backup_image_before_testing=no restore_image_after_testing=no vcpu_cores=2 smp=2 mem=4096 use_os_variant=no display=nographic take_regular_screendumps=no
JOB ID     : d08675ac25b324e96c96ad23d2b19392c89f7b02
JOB LOG    : /home/harihare/avocado_rt/results/job-2019-08-21T15.32-d08675a/job.log
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.diff_size: PASS (243.89 s)
JOB TIME   : 248.54 s
JOB HTML   : /home/harihare/avocado_rt/results/job-2019-08-21T15.32-d08675a/results.html

libvirt/tests/src/libvirt_mem.py
```diff
@@ -443,16 +445,23 @@ def modify_domain_xml():
...
+            if diff_tg_size is None:
+                dev_xml = utils_hotplug.create_mem_xml(tg_size, pg_size,
+                                                       mem_addr, tg_sizeunit,
+                                                       pg_unit, tg_node,
+                                                       node_mask, mem_model)
             randvar = 0
             if rand_reboot:
                 rand_value = random.randint(15, 25)
                 logging.debug("reboots at %s", rand_value)
             for x in xrange(at_times):
```
bssrikanth: better to use meaningful variable name than `x`

bssrikanth: same comment applies to different places in this file. Always better to have context dependent variable names.. avoid using same name in different context..so you can provide more simplicity for understandability and maintainability.

harihare: Change the index variable "x" to dimmno, as it refers to dimm no.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -443,19 +445,26 @@ def modify_domain_xml():
...
             randvar = 0
             if rand_reboot:
                 rand_value = random.randint(15, 25)
                 logging.debug("reboots at %s", rand_value)
-            for x in xrange(at_times):
+            for dimmno in xrange(at_times):
                 # If any error excepted, command error status should be
                 # checked in the last time
+                if diff_tg_size is not None:
+                    dev_xml = utils_hotplug.create_mem_xml(diff_tg_size[dimmno], pg_size,
```
sathnaga: looks like attach_times depends on length of diff_tg_size list. it is not good to have such dependency introduced, coz at_times> length of list would lead to indexError, and both has different purpose.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -443,19 +445,26 @@ def modify_domain_xml():
...
+                                                           mem_addr, tg_sizeunit,
```
sathnaga: Better to introduce a check for host memory availability w.r.t `tg_size `

dzhengfy:

PR #2318, opened 2019-09-10
Fix invalid domxml check to be less platform dependent

smitterl:
s390x don't have 'devices/video'. The test just checks for valid domxml.
Therefore, using disk device instead which is more common. Running the tests
manually passed.


smitterl:
dzhengfy Can you have a look?

PR #2344, opened 2019-09-26
Correct the failure message of block copy case

lento-sun:
The bug mentioned in the script is already fixed in
libvirt-1.3.2-1.el7, but the failure message will always
be "Hit on bug xxx" if some errors captured. This will make
the log confusing if error happened. The fix here is to make
the log describe the problem more pricisely.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
@@ -343,17 +343,21 @@ def run(test, params, env):
         blockcopy hang with state change lock.
         This is a specific bug verify, so ignore status_error here.
         """
-        bug_url_ = "https://bugzilla.redhat.com/show_bug.cgi?id=1197592"
+        failure_msg = ""
         err_msg = "internal error: unable to execute QEMU command"
         err_msg += " 'block-job-complete'"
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcopy.py
```diff
@@ -343,17 +343,21 @@ def _blockjob_and_libvirtd_chk(cmd_result):
...
+        failure_msg = ""
         bug_url_ = "https://bugzilla.redhat.com/show_bug.cgi?id=1197592"
+        if not libvirt_version.version_compare(1, 3, 2):
+            failure_msg += "Hit on bug: %s " % bug_url_
         err_msg = "internal error: unable to execute QEMU command"
         err_msg += " 'block-job-complete'"
         if err_msg in cmd_result.stderr:
-            raise exceptions.TestFail("Hit on bug: %s" % bug_url_)
-
+            failure_msg += "The error msg is: %s" % err_msg
```
chunfuwen: would be better line 346 logic is moved here?

lento-sun: change the logic, use only one "test.fail" in the end to make sure all errors will be captured, even if they happened at the same time.

PR #2351, opened 2019-09-29
Do not set rebootTimeout to -1

qiankehan:
Since qemu v4.0.0(commit ee5d0f89de3), qemu disallows reboot-timeout set
beyond 0~0xffff. Because reboot-timeout is disabled by default, just
remove rebootTimeout for disabling.


qiankehan:
@dzhengfy 
-1 is **valid** for qemu<4.0.0. For  qemu<4.0.0, the default value of reboot-timeout(it is the qemu cmdline parameter which is translated from libvirt argument rebootTimeout) is -1. So it doesn't matter if rebootTimeout is removed. 
Update the check in case the users doesn't know the default value of rebootTimeout

PR #2349, opened 2019-09-27
Fix test expectation if I/O scheduler is not cfq

smitterl:
If I/O scheduler is e.g. bfq blkio device weights are not supported
and virsh will return an error:
https://github.com/libvirt/libvirt/commit/7e8a1a6e21cee67f6fa5bd2126ec17f96e5857d6
https://github.com/libvirt/libvirt/commit/035ebe9390a630964f391816f05c9cc7792212ad

The test now passes logging expected error message.

Test run:
JOB ID     : 66b7b57011e52ff8a2968d9ddbad5fba3538800d
JOB LOG    : /root/avocado/job-results/job-2019-09-27T18.33-66b7b57/job.log
 (01/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.running_guest.options.none: PASS (6.64 s)
 (02/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.running_guest.options.live: PASS (6.32 s)
 (03/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.running_guest.options.current: PASS (6.29 s)
 (04/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.shutoff_guest.options.none: PASS (6.88 s)
 (05/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.shutoff_guest.options.config: PASS (5.45 s)
 (06/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.get_blkio_parameter.shutoff_guest.options.current: PASS (5.62 s)
 (07/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.set_blkio_parameter.shutoff_guest.change_weight.options.config.minimum_boundary: PASS (5.46 s)
 (08/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.set_blkio_parameter.shutoff_guest.change_weight.options.config.inside_boundary: PASS (5.67 s)
 (09/76) type_specific.io-github-autotest-libvirt.virsh.blkiotune.positive.set_blkio_parameter.shutoff_guest.change_weight.options.config.maximum_boundary: PASS (5.96 s)
[...]

PR #2355, opened 2019-10-02
Add kernel hugepagesize support to log

smitterl:
If a test cannot mount due to hugepage pagesize restriction by kernel
the log contained 'wrong fs type, bad option, bad superblock on hugetlbfs,
missing codepage or helper program, or other error'

Now additionally the following message is logged to assist debugging:
2019-10-02 07:03:50,292 process          L0626 INFO | Running 'cat /proc/meminfo |grep Hugepagesize'
2019-10-02 07:03:50,298 process          L0458 DEBUG| [stdout] Hugepagesize:       1024 kB

Tests failed because they tried to mount a page size that wasn't supported
by the host.

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -64,6 +64,7 @@ def run(test, params, env):
 
         :param page_size: unit is kB, it can be 4,2048,1048576,etc
         """
+        log_kernel_hugepagesize(page_size)
         if page_size == 4:
             perm = ""
         else:
@@ -74,6 +75,10 @@ def run(test, params, env):
         if tlbfs_status:
             utils_misc.umount("hugetlbfs", "/dev/hugepages", "hugetlbfs")
         utils_misc.mount("hugetlbfs", "/dev/hugepages", "hugetlbfs", perm)
+
+    def log_kernel_hugepagesize(page_size):
+        cmd = 'cat /proc/meminfo |grep Hugepagesize'
+        process.run(cmd, shell=True, verbose=True, ignore_status=True)
 
     def setup_hugepages(page_size=2048, shp_num=2000):
         """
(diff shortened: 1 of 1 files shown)


smitterl:
@dzhengfy :)

dzhengfy:
@smitterl please help correct the check failure

libvirt/tests/src/libvirt_mem.py
```diff
@@ -109,7 +114,7 @@ def check_qemu_cmd(max_mem_rt, tg_size):
                     % (max_mem_slots, max_mem_rt))
         if tg_size:
             size = int(tg_size) * 1024
-            cmd_str = 'memdimm.\|memory-backend-ram,id=ram-node.'
+            cmd_str = 'memdimm.\\|memory-backend-ram,id=ram-node.'
```
sathnaga: this change is not explained on commit message, typo or mistaken?

smitterl: sathnaga, thanks for asking, inspektor requires this change.

sathnaga: hope that's not altering the functionality...

smitterl: @sathnaga I checked this by doing a regex matching manually with python 2.7 and python 3.6 interactive console. Alternatively, I can check if this will work without any escaping at all as suggested https://github.com/autotest/tp-libvirt/pull/2349 but I also had a system level test run with the "\\" in regex in another occasion: https://github.com/autotest/tp-libvirt/commit/6d1f03161a4a606d1efbe0d963899a4795233f95

smitterl:
@dzhengfy I prefer to issue a new PR for the fix itself which would consist of more additional setup steps. I scoped this very PR more to improve the output.

libvirt/tests/src/libvirt_mem.py
sathnaga: @smitterl make sense, but this change nothing to do with actual fix of this PR, so you can skip this change and update the PR with only add log method, and we can take this change as separate PR as we need.

smitterl: @sathnaga, I agree, but I've found that there's already a function in virttest.staging.utils_memory. I'll close this one and add the log improvement to the script fix. Thank you for reviewing this anyways. Sorry for the churn.

PR #2357, opened 2019-10-08
virtual_disks_multidisks: Do not check ide-cd

qiankehan:
The device ide-cd is introduced very early at qemu v0.15.0(commit 1f56e32a).
And no build configration to disable it. We can think it is defaultly
supported.


chunfuwen:
please upload test result

PR #2358, opened 2019-10-08
Fix cpu xml attrs missing

chloerh:
The original code is to create new cpu xml then set hpt/numa info
which could lead to info missing from original cpu xml. Now modify
to update original cpu xml which should fix the problem.

### DIFF
--- a/libvirt/tests/src/papr_hpt.py
+++ b/libvirt/tests/src/papr_hpt.py
@@ -62,9 +62,12 @@ def run(test, params, env):
         :param vmxml: xml of vm to be manipulated
         :param attrs: attrs to set to cpu xml
         """
-        cpu = vm_xml.VMCPUXML()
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/papr_hpt.py
```diff
@@ -62,9 +62,9 @@ def set_cpu(vmxml, **attrs):
         :param vmxml: xml of vm to be manipulated
         :param attrs: attrs to set to cpu xml
         """
-        cpu = vm_xml.VMCPUXML()
+        cpu = vmxml.cpu
```
sathnaga: try to assign `cpu = vmxml.cpu` if fail then `cpu = vm_xml.VMCPUXML()` can be assigned...
same holds good for below change...

chloerh: Good point! Updated. Thanks @sathnaga

PR #2363, opened 2019-10-10
Remove 1 spice related case from pseries

chloerh:
This is a positive case and pseries does not support spice


libvirt/tests/cfg/serial/serial_functional.cfg
```diff
@@ -40,6 +40,7 @@
         - type_spicevmc:
             serial_dev_type = spicevmc
         - type_spiceport:
+            no pseries
```
dzhengfy: Spice is also not supported by s390x. So could you help also exclude s390-virtio?

PR #2371, opened 2019-10-16
add new auto case for v2v RHEL-170557

musuemu:
Add new auto case for rhel-170557:Virt-v2v can convert guest which has a file without file_architecture
Signed-off-by: liuzi <zili@redhat.com>

### DIFF
--- a/v2v/tests/cfg/function_test_esx.cfg
+++ b/v2v/tests/cfg/function_test_esx.cfg
@@ -247,6 +247,10 @@
         - epoch:
             only esx_67
             main_vm = 'VM_NAME_ESX_EPOCH_V2V_EXAMPLE'
+        - without_file_architecture
+            only esx_67
+            main_vm = 'VM_NAME_NO_FILE_ARCHITECTURE_V2V_EXAMPLE'
+            checkpoint = file_architecture
     variants:
         - positive_test:
             status_error = 'no'
--- a/v2v/tests/src/function_test_esx.py
+++ b/v2v/tests/src/function_test_esx.py
@@ -174,6 +174,18 @@ def run(test, params, env):
             else:
                 log_fail('%s does not exist after convert to rhv' % key)
 
+    def check_file_architecture(vmcheck):
+        """
+        Check the 3rd party module info
+        :param vmcheck: VMCheck object for vm checking
+        """
+        content = vmcheck.session.cmd('uname -r').strip()
+        status = vmcheck.session.cmd_status('rpm -qf /lib/modules/%s/fileaccess/fileaccess_mod.ko ' % content)
+        if status == 0:
+            log_fail('3rd party module info is not correct')
+        else:
+            logging.info('file /lib/modules/%s/fileaccess/fileaccess_mod.ko is not owned by any package' % content)
+
     def check_result(result, status_error):
         """
         Check virt-v2v command result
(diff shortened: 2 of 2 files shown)


musuemu:
#  avocado run --vt-type v2v --vt-machine-type q35 function_test_esx.positive_test.without_file_architecture.esx_67.rhev.rhv_upload.it_vddk.esx_uri --vt-connect-uri qemu:///system
JOB ID     : 4b69f0317a8ca690dbc2661a23013d7e8e73c6bd
JOB LOG    : /root/avocado/job-results/job-2019-10-16T18.19-4b69f03/job.log
 (1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.without_file_architecture.esx_67.rhev.rhv_upload.it_vddk.esx_uri: PASS (2702.54 s)
JOB TIME   : 2703.29 s

v2v/tests/src/function_test_esx.py
```diff
@@ -174,6 +174,16 @@ def check_rhev_file_exist(vmcheck):
             else:
                 log_fail('%s does not exist after convert to rhv' % key)
 
+    def check_file_architecture(vmcheck):
+        """
+        Check the 3rd party module info
+        """
+        status = vmcheck.session.cmd_status('rpm -qf /lib/modules/3.10.0-957.el7.x86_64/fileaccess/fileaccess_mod.ko ')
```
xiaodwan: '3.10.0-957.el7.x86_64' is not general. The code should adapt to any kernel versions.

v2v/tests/src/function_test_esx.py
```diff
@@ -174,6 +174,16 @@ def check_rhev_file_exist(vmcheck):
...
+        if status == 0:
+            log_fail('3rd party module info is not correct')
+        else:
+            logging.info('file /lib/modules/3.10.0-957.el7.x86_64/fileaccess/fileaccess_mod.ko is not owned by any package')
```
xiaodwan: same as above.

musuemu: The case cannot adapt all linux guests,the key point of the case is to convert a guest which has a file without file_architecture,so the guest is a special guest  prepared before.

xiaodwan: I mean you can get the path by 'uname -r' instead of a fixed kernel version.

musuemu: changed

v2v/tests/src/function_test_esx.py
```diff
@@ -174,6 +174,17 @@ def check_rhev_file_exist(vmcheck):
             else:
                 log_fail('%s does not exist after convert to rhv' % key)
 
+    def check_file_architecture(vmcheck):
+        """
```
chunfuwen: miss parameter doc

musuemu: There is no need to add parameter info,because all parameters are described in the function check_vmtools(vmcheck, check)  in this file and there are no new parameter in this function.

xiaodwan: The old code doesn't have params information, but they should have. 
We should add params information in any functions.

v2v/tests/cfg/function_test_esx.cfg
```diff
@@ -247,6 +247,10 @@
         - epoch:
             only esx_67
             main_vm = 'VM_NAME_ESX_EPOCH_V2V_EXAMPLE'
+        - without_file_architecture
```
chloerh: Missing ":"

PR #2372, opened 2019-10-17
Change the destination for v2v function_test_xen cases

mxie91:
Make every functional xen case except with_vdsm and pool_uuid case
has three destinations 'libvirt,rhv rhv-upload'.

Signed-off-by: mxie91 <mxie@redhat.com>

### DIFF
--- a/v2v/tests/cfg/function_test_xen.cfg
+++ b/v2v/tests/cfg/function_test_xen.cfg
@@ -109,7 +109,6 @@
             sasl_server_passwd = SERVER_PASSWORD_V2V_EXAMPLE
             sasl_user = test
             sasl_pwd = redhat
-            only libvirt
         - encryped:
             main_vm = VM_NAME_XEN_ENCRYPED_V2V_EXAMPLE
         - multidisk:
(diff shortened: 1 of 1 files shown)


v2v/tests/cfg/function_test_xen.cfg
```diff
@@ -178,16 +177,13 @@
             no xen_vm_default
             variants:
                 - libvirt:
-                    only pool_uuid, display, sound, virtio_win, with_vdsm
                     only output_mode.libvirt
-                    no encrypt_warning
```
xiaodwan: encrypt_warning is a rhv only scenario.

xiaodwan:
You should test all new cases because we are not sure they will definitely pass for new destinations.

xiaodwan:
It's better to list the diff outputs for the case changes before and after the modification.

xiaodwan:
Please check my latest comments.

display.vnc.encrypt and display.vnc.encrypt_warning correspond to libvirt and rhv respectively in RHEL7-13054.
You cannot extent the output mode for them, or cases will be duplicate.

with_vdsm, the case was changed to 'manual only'. It's a low prority to maintain the script until we have a good solution for autamatation.
Now I sugguest keep it as it is.

display.vnc.autoport and display.vnc.sdl have already had all output methods.

virtio_win, please upload testing result for rhv output.
sound, I doubt if sound to rhv can work because by default sound device is disabled and the codes didn't do any enable operation.
so please also upload sound testing result.

PR #2222, opened 2019-07-05
Fixup: Modify cpu library functions

sathnaga:
Let's modify the cpu library functions used in the
tests to match the re-order and grouping of library functions
addressed by below Patch(PR).
https://github.com/avocado-framework/avocado-vt/pull/2150

Signed-off-by: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>

PR #2378, opened 2019-10-21
Rename imported cpu module from avocado to avoid conflict

chloerh:
A recently patch has added virttest.cpu into this file, which has
the same name as avocado.utils.cpu, caused massive failures.
Rename one module to avoid this kind of conflict.

### DIFF
--- a/libvirt/tests/src/cpu/ppccpucompat.py
+++ b/libvirt/tests/src/cpu/ppccpucompat.py
@@ -2,7 +2,7 @@ import os
 import time
 import logging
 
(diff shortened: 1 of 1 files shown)


chloerh:
@sathnaga this should fix the case failures that https://github.com/autotest/tp-libvirt/pull/2222 has brought in.

sathnaga:
@chloerh Yep, tried to change most of places, looks like missed some, Thanks for the fix :-)

sathnaga:
@chloerh sorry for the inconvenience caused.

PR #2379, opened 2019-10-21
Fixup: python3: TypeError: a bytes-like object is required, not 'str'

sathnaga:
Let's decode the dmesg output as text before using because
python3 complains the same.

### DIFF
--- a/libvirt/tests/src/cpu/powerpc_hmi.py
+++ b/libvirt/tests/src/cpu/powerpc_hmi.py
@@ -10,6 +10,7 @@ from virttest import data_dir
 from virttest import virsh
 from virttest import libvirt_xml
 from virttest import utils_test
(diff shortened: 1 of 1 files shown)


sathnaga:
Before fix:
```
>>> from virttest.compat_52lts import decode_to_text
>>> host_dmesg = process.system_output("dmesg", verbose=False)
>>> if "Unrecovered" in host_dmesg:
...     print("yes")
... 
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
```
After fix:
```
>>> host_dmesg = decode_to_text(process.system_output("dmesg", verbose=False))
>>> if "Unrecovered" in host_dmesg:
...     print("yes")
... 
>>> 
```

PR #2391, opened 2019-10-25
Add testcases for 'vnc_tls_x509_secret_uuid' in qemu.conf

yafu-1:
3 cases added:
1. Set valid and existing 'vnc_tls_x509_secret_uuid' in qemu.conf;
2. Set invalid 'vnc_tls_x509_secret_uuid' in qemu.conf;
3. Set valid but non-existing 'vnc_tls_x509_secret_uuid' in qemu.conf;

### DIFF
--- a/libvirt/tests/src/graphics/graphics_functional.py
+++ b/libvirt/tests/src/graphics/graphics_functional.py
@@ -10,6 +10,8 @@ import logging
 import threading
 import time
 import datetime
+import locale
+import base64
 try:
     import queue as Queue
 except ImportError:
@@ -21,12 +23,15 @@ from avocado.utils import process
 from virttest.libvirt_xml.vm_xml import VMXML
 from virttest import data_dir
 from virttest import virt_vm
+from virttest import virsh
 from virttest import utils_net
 from virttest import utils_misc
 from virttest import utils_config
 from virttest import utils_libvirtd
+from virttest.utils_test import libvirt
 from virttest.utils_test.libvirt import LibvirtNetwork
 from virttest.libvirt_xml.devices.graphics import Graphics
+from virttest.libvirt_xml.secret_xml import SecretXML
 
 from provider import libvirt_version
 
@@ -117,6 +122,8 @@ class EnvState(object):
         port_max = params.get("remote_display_port_max", 'not_set')
         auto_unix_socket = params.get("vnc_auto_unix_socket", 'not_set')
         tls_x509_verify = params.get("vnc_tls_x509_verify", 'not_set')
+        vnc_tls_x509_secret_uuid = params.get("vnc_tls_x509_secret_uuid", 'not_set')
+        vnc_secret_uuid = params.get("vnc_secret_uuid", "not_set")
 
         if spice_x509_dir == 'not_set':
             self.spice_x509_dir_real = '/etc/pki/libvirt-spice'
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -192,6 +199,19 @@ def __init__(self, params, expected_result):
...
 
+        if vnc_tls_x509_secret_uuid == "not_set":
+            del self.qemu_config.vnc_tls_x509_verify
+        else:
+            self.qemu_config.vnc_tls = "1"
+            if vnc_secret_uuid == "invalid":
+                self.qemu_config.vnc_tls_x509_secret_uuid = "00000-11111"
+            else:
+                cmd = "uuidgen"
+                status, uuid = process.getstatusoutput(cmd)
```
chunfuwen: use the method compatible with python 2 and 3 to get output

libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -325,14 +345,15 @@ def qemu_vnc_options(libvirt_vm):
...
-    if libvirt_version.version_compare(5, 0, 0):
-        if re.search(r'(tls-creds-x509.*?\s+)', res.stdout_text):
-            vnc_out = re.search(r'(tls-creds-x509.*?\s+)', res.stdout_text).group(0)
-            if re.match("tls-creds-x509,id=[a-z0-9-]+", vnc_out):
-                vnc_dict['tls'] = "yes"
-                vnc_dict['id'] = re.match("tls-creds-x509,id=([a-z0-9-]+)", vnc_out).group(1)
-            if re.search("dir=[a-z/-]+", vnc_out):
-                vnc_dict['dir'] = re.search("dir=([a-z/-]+)", vnc_out).group(1)
+
+    #if libvirt_version.version_compare(5, 0, 0):
```
chunfuwen: remove commented out code

libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -1320,6 +1387,8 @@ def run(test, params, env):
         if networks:
             for network in list(networks.values()):
                 network.cleanup()
+        if vnc_secret_uuid == 'valid':
```
chunfuwen: suggest check whether secret exist before cleanup since sometime secret may not be created successfully before entering finally block

libvirt/tests/src/graphics/graphics_functional.py
```diff
@@ -192,6 +199,19 @@ def __init__(self, params, expected_result):
         else:
             self.qemu_config.vnc_tls_x509_verify = tls_x509_verify
 
+        if vnc_tls_x509_secret_uuid == "not_set":
+            del self.qemu_config.vnc_tls_x509_verify
```
chunfuwen: I don't see restore operation in your finally block, do you need that?

libvirt/tests/src/graphics/graphics_functional.py
yafu-1: As i need to check both the result and status, and this method also compatible with python 2 and 3.  I decide not change it.

libvirt/tests/src/graphics/graphics_functional.py
yafu-1: Changed. The code was commented while debugging, which no need to be commented actually.

libvirt/tests/src/graphics/graphics_functional.py
yafu-1: Added.

PR #2420, opened 2019-11-08
Fix iface_options unprivileged

smitterl:
This PR requires https://github.com/avocado-framework/avocado-vt/pull/2329

1. del_seclabel:
 a. If original domain xml had dac security driver the test will fail
    for unprivileged user because dac is not available in this case.
    The test now removes the dac security driver from the domain.xml
    of unprivileged user. This is what a real user would have to do,
    too.

2. virsh.remove_domain:
 a. Removed extra option as in this case it doesn't remove the image
    because the image is created below /tmp and not registered as volume.
 b. Remove test image but emit warning if there was an issue.

3. fixed regex as required by inspektor

Test run on s390x with <seclabel model='dac'.../> in domain.xml:
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system virtual_network.iface_options.iface_unprivileged_user
 
JOB ID     : 29fedc5c7ad8f88c04485ffcd0d761769ebd608c
JOB LOG    : /root/avocado/job-results/job-2019-11-08T14.24-29fedc5/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_options.iface_unprivileged_user: PASS (68.89 s)
JOB TIME   : 70.82 s
```

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_options.py
+++ b/libvirt/tests/src/virtual_network/iface_options.py
@@ -141,6 +141,9 @@ def run(test, params, env):
             channels = vmxml.get_devices(device_type="channel")
             for channel in channels:
                 vmxml.del_device(channel)
+            logging.info("Unprivileged users can't use 'dac' security driver,"
+                         " removing from domain xml if present...")
+            vmxml.del_seclabel([('model', 'dac')])
 
             vmxml.xmltreefile.write()
             logging.debug("New VM xml: %s", vmxml)
@@ -225,7 +228,7 @@ def run(test, params, env):
                 mode = iface.source["mode"]
                 if mode == "passthrough":
                     mode = "passthru"
-                if not re.search("macvtap\s+mode %s" % mode, output):
+                if not re.search(r"macvtap\s+mode %s" % mode, output):
                     test.fail("Failed to verify macvtap mode")
         # Check if the "target dev" is set successfully
         # 1. Target dev name with prefix as "vnet" will always be override;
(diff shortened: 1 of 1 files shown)


smitterl:
@dzhengfy

dzhengfy:
@smitterl I have to hold on this reviewing due to the question in jira task. That is I am wondering the root cause of 'del_seclabel'.
Thanks.

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -678,8 +681,12 @@ def check_mcast_network(session):
             # Restore vhost_net driver
             process.system("modprobe vhost_net", shell=True)
         if unprivileged_user:
-            virsh.remove_domain(vm_name, "--remove-all-storage",
-                                **virsh_dargs)
+            virsh.remove_domain(vm_name, **virsh_dargs)
+            result = process.run('rm -f %s' % dst_disk)
+            if result.exit_status:
+                logging.warning("Couldn't remove %s. Please"
```
chunfuwen: it is not so useful that it give out this kind of warning during automation execution , can we try to delete it in other way?

smitterl: @chunfuwen Why isn't this useful? If the image for some unlikely reason couldn't be removed the we know that that might have an impact on next test case's execution. I can't think of a better way trying to delete the file than doing a "rm -f" as root user. Do you prefer the warning to be removed?

chunfuwen: I doesn't mean rm -rf doesn't work, I mean the warning message maybe useless since normally the running case context is destroyed by next case running.

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -678,8 +681,12 @@ def check_mcast_network(session):
             # Restore vhost_net driver
             process.system("modprobe vhost_net", shell=True)
         if unprivileged_user:
-            virsh.remove_domain(vm_name, "--remove-all-storage",
-                                **virsh_dargs)
+            virsh.remove_domain(vm_name, **virsh_dargs)
+            result = process.run('rm -f %s' % dst_disk)
```
chunfuwen: add shell=True parameter for process.run

libvirt/tests/src/virtual_network/iface_options.py
smitterl: @chunfuwen When you say "normally the running case context is destroyed by next case running" I suppose you mean the avocado data_dir? In this test case, however, the path is below /tmp: `dst_disk = "/tmp/%s.img" % unprivileged_user` I don't understand how that would normally be destroyed by the next case running.

smitterl: @chunfuwen Do you prefer the warning to be removed?

libvirt/tests/src/virtual_network/iface_options.py
smitterl: OK, thanks :)

PR #2430, opened 2019-11-15
Avoid xattr left on image blocking all followed up cases

lento-sun:
This is due to bz1741456. When image has xattr left after vm
destroyed, vm cannot be started anymore until we clear these
xattrs. This will block all followed up cases when happened.
This commit is to report the error and clear the xattr within
the problematic case, not blocking followed up cases.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
@@ -20,6 +20,40 @@ from virttest.utils_test import libvirt
 from provider import libvirt_version
 
 
+def get_images_with_xattr(vm):
+    """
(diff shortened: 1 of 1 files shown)


lento-sun:
Test result before the fix:
https://libvirt-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/libvirt/view/RHEL-8.1%20x86_64/job/libvirt-RHEL-8.1-runtest-x86_64-function-block_job_commit_pull/48/testReport/

Test result after the fix (failed cases hitting bz1741456 and won't block other cases):
https://libvirt-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/libvirt/view/RHEL-8.1%20x86_64/job/libvirt-RHEL-8.1-runtest-x86_64-function-block_job_commit_pull/49/testReport/

PR #2435, opened 2019-11-18
virsh_snapshot_disk: modified the test case status from CANCEL to ERROR

KalaiIBM:
when test case is trying to attach the disk to the VM, it got cancelled.
Actually it should throw the error if it is not able to attach it.
So this test case code has modified from CANCEL to ERROR.

Suggested-by: Balamuruhan S <bala24@linux.ibm.com>
Signed-off-by: Kalaiarasan Panneerselvam <panneer1@in.ibm.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_disk.py
+++ b/libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_disk.py
@@ -189,8 +189,8 @@ def run(test, params, env):
             result = virsh.attach_disk(vm_name, source=img_path, target="vdf",
                                        extra=extra, debug=True)
             if result.exit_status:
-                test.cancel("Failed to attach disk %s to VM."
-                            "Detail: %s." % (img_path, result.stderr))
+                test.error("Failed to attach disk %s to VM."
+                           "Detail: %s." % (img_path, result.stderr))
 
         # Create snapshot.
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/snapshot/virsh_snapshot_disk.py
```diff
@@ -189,8 +189,8 @@ def run(test, params, env):
             result = virsh.attach_disk(vm_name, source=img_path, target="vdf",
                                        extra=extra, debug=True)
             if result.exit_status:
-                test.cancel("Failed to attach disk %s to VM."
-                            "Detail: %s." % (img_path, result.stderr))
+                test.error("Failed to attach disk %s to VM."
```
sathnaga: is it the case, `cancel` due to disk attach is kind of pre-requisite to test  ? and not relevant to
the feature under test?,

smitterl: IIUC, disk has to be attached in order to create snapshot with it. So by avocado's exceptions doc - TestError should be used in these cases https://github.com/avocado-framework/avocado/blob/4f0e84548b70c9ed960b6a4d585614f3031107c5/avocado/core/exceptions.py#L69 To me, this makes sense as it's more specific as TestCancel.

PR #2439, opened 2019-11-21
Delete virt-v2v case RHEL7-48427

musuemu:
Signed-off-by: liuzi <zili@redhat.com>


xiaodwan:
@musuemu It's better to add a description about the reason you delete the codes.  And the subject is useless since 'RHEL7-48427' will be an invalid link when we delete it.

xiaodwan:
and fix below errors please.
W0611:  5,0: : Unused process imported from avocado.utils

W0611:  7,0: : Unused data_dir imported from virttest

PR #2440, opened 2019-11-21
Delete v2v cases about parameter copy_to_local

musuemu:
Because v2v code changed on 2018.11.01
https://github.com/libguestfs/virt-v2v/commit/c1c4ed4c16e83a83d6d2835e05d6adf97836db70#diff-31c097437346a8f33915e8de88c011d2
we need to delete related cases which use parameter copy_to_local to convert guest from ESXI.


xiaodwan:
I just notice your subject is '-s', please double check it. @musuemu

PR #2443, opened 2019-11-22
virtio_transitional: Add test cases for memory balloon

vivianQizhu:
Signed-off-by: Qianqian Zhu <qizhu@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
+++ b/libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
@@ -1,0 +1,20 @@
+- virtio_transitional_mem_balloon:
+    type = virtio_transitional_mem_balloon
+    only q35
+    start_vm = no
+    variants:
+        - virtio:
+            virtio_model = "virtio"
+        - virtio_transitional:
+            virtio_model = "virtio-transitional"
+        - virtio_non_transitional:
+            virtio_model = "virtio-non-transitional"
+    variants:
+        - @default:
+            no virtio_transitional
+        - rhel6_guest:
+            os_variant = rhel6
+            image_path = images/rhel6-x86_64-latest.qcow2
+            guest_src_url = "http://download.libvirt.redhat.com/libvirt-CI-resources/RHEL-6.10-x86_64-latest.qcow2"
+            only virtio_transitional
+    disk_model = ${virtio_model}
--- a/libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
+++ b/libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
@@ -1,0 +1,88 @@
+import os
+import logging
+
+from avocado.utils import download
+
+from virttest import virsh
+from virttest import data_dir
+from virttest import utils_misc
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    Test virtio/virtio-transitional/virtio-non-transitional model of
+    memory balloon
+
+    :param test: Test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    vm_name = params.get("main_vm", "avocado-vt-vm1")
+    vm = env.get_vm(params["main_vm"])
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    backup_xml = vmxml.copy()
+    guest_src_url = params.get("guest_src_url")
+    virtio_model = params['virtio_model']
+
+    # Download and replace image when guest_src_url provided
+    if guest_src_url:
+        image_name = params['image_path']
+        target_path = utils_misc.get_path(data_dir.get_data_dir(), image_name)
+        if not os.path.exists(target_path):
+            download.get_file(guest_src_url, target_path)
+        params["blk_source_name"] = target_path
+
+    try:
+        # Update disk and interface to correct model
+        if (params["os_variant"] == 'rhel6' or
+                'rhel6' in params.get("shortname")):
+            iface_params = {'model': 'virtio-transitional'}
+            libvirt.modify_vm_iface(vm_name, "update_iface", iface_params)
+        libvirt.set_vm_disk(vm, params)
+        # vmxml will not be updated since set_vm_disk
+        # sync with another dumped xml inside the function
+        vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+        # Update memory balloon device to correct model
+        membal_dict = {'membal_model': virtio_model,
+                       'membal_stats_period': '10'}
+        libvirt.update_memballoon_xml(vmxml, membal_dict)
+        if not vm.is_alive():
+            vm.start()
+        is_windows_guest = (params['os_type'] == 'Windows')
+        session = vm.wait_for_login()
+        # Finish test for Windows guest
+        if is_windows_guest:
+            return
+        # Check if memory balloon device exists on guest
+        status = session.cmd_status_output('lspci |grep balloon')[0]
+        if status != 0:
+            test.fail('Not detect memory balloon device on guest.')
+        # Save and restore guest
+        sn_path = os.path.join(data_dir.get_tmp_dir(), params['os_variant'])
+        session.close()
+        virsh.save(vm_name, sn_path)
+        virsh.restore(sn_path)
+        session = vm.wait_for_login()
+        # Get original memory for later balloon function check
+        ori_outside_mem = vm.get_max_mem()
+        ori_guest_mem = vm.get_current_memory_size()
+        # balloon half of the memory
+        ballooned_mem = ori_outside_mem // 2
+        # Set memory to test balloon function
+        virsh.setmem(vm_name, ballooned_mem)
+        # Check if memory is ballooned successfully
+        logging.info("Check memory status")
+        unusable_mem = ori_outside_mem - ori_guest_mem
+        gcompare_threshold = int(
+            params.get("guest_compare_threshold", unusable_mem))
+        after_mem = vm.get_current_memory_size()
+        act_threshold = ballooned_mem - after_mem
+        if (after_mem > ballooned_mem) or (
+                abs(act_threshold) > gcompare_threshold):
+            test.fail("Balloon test failed")
+    finally:
+        vm.destroy()
+        backup_xml.sync()


libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Test virtio/virtio-transitional/virtio-non-transitional model of
+    memory balloon
+
+    :param test: Test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    vm_name = params.get("main_vm", "avocado-vt-vm1")
```
smitterl: I expect main_vm to always be present, if not the setup should be broken. You might want to consider removing default value to avoid test breaking later.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+        params["blk_source_name"] = target_path
+
+    try:
+        # Update disk and interface to correct model
+        if (params["os_variant"] == 'rhel6' or
+                'rhel6' in params.get("shortname")):
+            iface_params = {'model': 'virtio-transitional'}
+            libvirt.modify_vm_iface(vm_name, "update_iface", iface_params)
+        libvirt.set_vm_disk(vm, params)
+        # vmxml will not be updated since set_vm_disk
```
smitterl: I don't understand this comment line and the next. Do you mean that vmxml needs to be refreshed and that's why you call ```vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)```?

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+        membal_dict = {'membal_model': virtio_model,
+                       'membal_stats_period': '10'}
+        libvirt.update_memballoon_xml(vmxml, membal_dict)
+        if not vm.is_alive():
+            vm.start()
+        is_windows_guest = (params['os_type'] == 'Windows')
+        session = vm.wait_for_login()
+        # Finish test for Windows guest
+        if is_windows_guest:
+            return
```
smitterl: Does this mean the test will always pass if the guest is Windows and the vm shows the login prompt?

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+        # Check if memory balloon device exists on guest
+        status = session.cmd_status_output('lspci |grep balloon')[0]
+        if status != 0:
+            test.fail('Not detect memory balloon device on guest.')
+        # Save and restore guest
+        sn_path = os.path.join(data_dir.get_tmp_dir(), params['os_variant'])
```
smitterl: s. above

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+    if guest_src_url:
+        image_name = params['image_path']
+        target_path = utils_misc.get_path(data_dir.get_data_dir(), image_name)
+        if not os.path.exists(target_path):
+            download.get_file(guest_src_url, target_path)
+        params["blk_source_name"] = target_path
+
+    try:
+        # Update disk and interface to correct model
+        if (params["os_variant"] == 'rhel6' or
```
smitterl: ```params["os_variant"]``` is used more than once in the script. You might consider defining it into a variable ```os_variant``` at the top of the run function.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+            vm.start()
+        is_windows_guest = (params['os_type'] == 'Windows')
+        session = vm.wait_for_login()
+        # Finish test for Windows guest
+        if is_windows_guest:
+            return
+        # Check if memory balloon device exists on guest
+        status = session.cmd_status_output('lspci |grep balloon')[0]
+        if status != 0:
+            test.fail('Not detect memory balloon device on guest.')
```
smitterl: Better say "Didn't detect memory balloon device on guest."

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+from virttest import virsh
+from virttest import data_dir
+from virttest import utils_misc
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    Test virtio/virtio-transitional/virtio-non-transitional model of
```
smitterl: Could you add the test step overview to the comment? See, for example, documentation of run function here: https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virtual_device/vsock.py

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
```diff
@@ -0,0 +1,88 @@
...
+        if status != 0:
+            test.fail('Not detect memory balloon device on guest.')
+        # Save and restore guest
+        sn_path = os.path.join(data_dir.get_tmp_dir(), params['os_variant'])
+        session.close()
+        virsh.save(vm_name, sn_path)
+        virsh.restore(sn_path)
+        session = vm.wait_for_login()
+        # Get original memory for later balloon function check
+        ori_outside_mem = vm.get_max_mem()
```
smitterl: You might want to consider putting your calculations and checks into methods to improve readability/checkability of the formulae.

libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
```diff
@@ -0,0 +1,20 @@
...
+            virtio_model = "virtio-non-transitional"
+    variants:
+        - @default:
+            no virtio_transitional
+        - rhel6_guest:
+            os_variant = rhel6
+            image_path = images/rhel6-x86_64-latest.qcow2
+            guest_src_url = "http://download.libvirt.redhat.com/libvirt-CI-resources/RHEL-6.10-x86_64-latest.qcow2"
+            only virtio_transitional
+    disk_model = ${virtio_model}
```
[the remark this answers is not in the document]
vivianQizhu: @smitterl Yes `disk_model` is for `libvirt.set_vm_disk()`, I have to specify it because the variants is named `disk_model`, while `virtio_model` is a concept for all virtio devices, I think it does not make sense to have memory balloon device use a variable named `disk_model`, so I would like to keep `virtio_model` there. Thanks.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: I have no opinion on this, @kylazhang what do you think about this one?

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl Okay, just I thought tp-libvirt does not really care it since many test cases does not have that, will add it.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl Yeap.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl Yes, I will say `The local variable "vmxml" will not be updated...` if it seems clearer.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl Yes, currently tp-libvirt has many trouble on Windows guest support, further checking cost too much and we don't really need it since this feature virtio transitional is mainly designed for RHEL6 guest, so after discussed with feature owner we decide to only test this much for Windows guest.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl Yeap.

libvirt/tests/src/virtio_transitional/virtio_transitional_mem_balloon.py
vivianQizhu: @smitterl I don't think there is much difference, the function won't be called more that once anyway.

libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
```diff
@@ -0,0 +1,20 @@
...
+            virtio_model = "virtio-transitional"
+        - virtio_non_transitional:
+            virtio_model = "virtio-non-transitional"
+    variants:
+        - @default:
+            no virtio_transitional
+        - rhel6_guest:
+            os_variant = rhel6
+            image_path = images/rhel6-x86_64-latest.qcow2
+            guest_src_url = "http://download.libvirt.redhat.com/libvirt-CI-resources/RHEL-6.10-x86_64-latest.qcow2"
```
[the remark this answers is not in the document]
vivianQizhu: @smitterl That's both okay to me, I agree that internal link is not proper here, but depends on the user's need, @kylazhang do you want that way, you will need replace the link before you run it if so.

libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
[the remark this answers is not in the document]
smitterl: @vivianQizhu Got it :) Do you think it makes sense to add ```params['disk_model'] = virtio_model``` in the code directly before using it in ```libvirt.set_vm_disk(vm, params)``` and removing it from the cfg file; this way I believe it might be less confusing as what it is meant to be used for.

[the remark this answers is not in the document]
vivianQizhu: @smitterl Yep, will update all together. Thanks.

libvirt/tests/cfg/virtio_transitional/virtio_transitional_mem_balloon.cfg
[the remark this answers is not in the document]
kylazhang: It is OK for us as we can use it directly in gating test (do not suggest to do replace for any case) and it meaningless for upstream, when others want to use it just change it.

kylazhang:
Thanks for all the comments, as vivianQizhu will be offline for more than 4 month, I suggest to merge the new file first and then submit the fix by others. Thanks

PR #2467, opened 2019-12-11
Add new error message check to "iface_hotplug".

santwana:
The existing error message list wasn't containing the new messages.
Specified the new expected error message to the existing list.


balamuruhans:
@santwana could you please post the logs for the error msg ?

dzhengfy:

PR #2470, opened 2019-12-12
enable specified gluster server in config file

chunfuwen:
if third party gluster server is provisioned, this can be specified
in cfg file.

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_gluster.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_gluster.py
@@ -27,6 +27,10 @@ def run(test, params, env):
     vm_name = params.get("main_vm")
     vm = env.get_vm(vm_name)
     virsh_dargs = {'debug': True, 'ignore_status': True}
+    gluster_server_name = params.get("gluster_server_name")
+    # If gluster_server is specified from config file, just use this gluster server.
+    if 'EXAMPLE' not in gluster_server_name:
+        params.update({'gluster_server_ip': gluster_server_name})
 
     def prepare_gluster_disk(disk_img, disk_format):
         """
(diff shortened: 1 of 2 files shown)


chunfuwen:
result posted here
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.gluster.hot_plug.disk_qcow2.attach-device.transport_rdma

libvirt/tests/src/virtual_disks/virtual_disks_gluster.py
```diff
@@ -27,6 +27,10 @@ def run(test, params, env):
     vm_name = params.get("main_vm")
     vm = env.get_vm(vm_name)
     virsh_dargs = {'debug': True, 'ignore_status': True}
+    gluster_server_ip = params.get("gluster_server_ip")
+    # If gluster_server_ip is specified with gluster server, just use this gluster server.
+    if 'EXAMPLE' not in gluster_server_ip:
```
Yingshun: So need to skip test if 'EXAMPLE' in cluster_server_ip?

chloerh: If gluster_server_ip is not None, then there's no need to update params['gluster_server_ip'] since gluster_server_ip comes from params.
Or if gluster_server_ip IS None, then this line will throw out an error.
Is that right?

chunfuwen: 1)If EXAMPLE in gluster server ip ,it means that in CI environment, it doesn't replace real gluster server. Then it will naturally use temp glusetr server created by CI.  
2)And gluster_server_ip would never be None since it get from params.get("gluster_server_ip")
3)gluster_server_ip is need to be set explicitly since the following method gluster.setup_or_cleanup_gluster() need it

chunfuwen: update variable gluster_server_ip  to gluster_server_server to avoid confusion.

chunfuwen:
new test result upload:
'''
[stdout] 
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.gluster.cold_plug.disk_raw.attach-device.iothread
'''

PR #2479, opened 2019-12-20
Add blockcopy cases

chloerh:
- with option 'reuse-external'

### DIFF
--- a/libvirt/tests/src/backingchain/blockcopy.py
+++ b/libvirt/tests/src/backingchain/blockcopy.py
@@ -1,0 +1,81 @@
+import os
+import logging
+
+from avocado.utils import process
+
+from virttest import virsh
+from virttest import utils_misc
+from virttest.libvirt_xml import vm_xml
+
+
+def run(test, params, env):
+    """
+    Test vm backingchain, blockcopy
+    """
+    vm_name = params.get('main_vm')
+    case = params.get('case', '')
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    file_to_del = []
+
+    try:
+        if case:
+            if case == 'reuse_external':
+                # Create a transient vm for test
+                virsh.undefine(vm_name, debug=True, ignore_status=False)
+                virsh.create(vmxml.xml)
+
+                all_disks = vmxml.get_disk_source(vm_name)
+                if not all_disks:
+                    test.error('Not found any disk file in vm.')
+                image_file = all_disks[0].find('source').get('file')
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/backingchain/blockcopy.py
```diff
@@ -0,0 +1,79 @@
...
+    vm_name = params.get('main_vm')
+    case = params.get('case', '')
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    file_to_del = []
+
+    try:
+        if case:
```
chunfuwen: I assume 'if case' already return True, why need it?

libvirt/tests/src/backingchain/blockcopy.py
```diff
@@ -0,0 +1,79 @@
...
+                cur_sfile = cur_disks[0].find('source').get('file')
+                logging.debug('Now disk source file is: %s', cur_sfile)
+                if cur_sfile.strip() != new_image_path:
+                    test.fail('Disk source file is not updated.')
+
+    finally:
+        if case == 'reuse_external':
+            # Recover vm and remove the transient vm
+            virsh.destroy(vm_name, debug=True)
+            virsh.define(bkxml.xml, debug=True)
```
chunfuwen: this is duplicate with line 75

libvirt/tests/src/backingchain/blockcopy.py
```diff
@@ -0,0 +1,79 @@
...
+        bkxml.sync()
+
+        # Remove files to be deleted
+        if file_to_del:
```
chunfuwen: How can you make use file really exists , which is gotten from file_do_del

libvirt/tests/src/backingchain/blockcopy.py
chloerh: This var is prepared for special cases. There might be other cases that fit in this file in the future, but do not need this var. Maybe some cases with some shared steps that could fit for all cases.

libvirt/tests/src/backingchain/blockcopy.py
chloerh: No, it is not. We're going to remove a transient vm by destroy it(line 73), and recover the original vm(line 74), which are all for case 'reuse_external'.
bkxml.sync() will undefine and redefine, and it's for all cases, not just 'reuse_external'

libvirt/tests/src/backingchain/blockcopy.py
chloerh: Updated. Check if file exists before deleting it.

PR #2501, opened 2020-01-09
Remove vcpu_cache test for s390x

smitterl:
CPU cache specification is not supported for 's390x' architecture


smitterl:
@dzhengfy

PR #2533, opened 2020-02-04
Set start_time to 240s for pseries to fix ip missing issue

chloerh:
The original start_time was set to 60s which might be sufficient
for x86 vms but not pseries vms. In 60s ppc vms might not finish
booting up yet. Setting timeout to 240s could easily fix this
problem

### DIFF
--- a/libvirt/tests/cfg/cpu/setvcpu.cfg
+++ b/libvirt/tests/cfg/cpu/setvcpu.cfg
@@ -5,6 +5,8 @@
     vcpu_current = "1"
     maxvcpu = "8"
     start_timeout = "60"
+    pseries:
+        start_timeout = '240'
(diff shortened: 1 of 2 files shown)


chunfuwen:
please upload test result

libvirt/tests/cfg/cpu/setvcpu.cfg
```diff
@@ -5,6 +5,8 @@
     vcpu_current = "1"
     maxvcpu = "8"
     start_timeout = "60"
+    machine_type == 'pseries':
```
dzhengfy: Use 'pseries:' instead of 'machine_type == 'pseries':'
'machine_type == 'pseries':' will also apply to other machine types.

PR #2537, opened 2020-02-06
Skip check when source tag has no 'file' attr

chloerh:
Currently this only happens to ppc. Xml usually does not have 'source'
tag if the source file is not specified. But now a tag like
<source index='3' /> is being created, which is affecting test.
Therefore skip check if this happens.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
@@ -1,6 +1,7 @@
 import os
 import time
 import logging
+import platform
 from virttest import virsh
 from virttest import data_dir
 from virttest.libvirt_xml import vm_xml
@@ -36,7 +37,9 @@ def run(test, params, env):
                 continue
             if disk.target['dev'] != target_dev:
                 continue
-            if disk.xmltreefile.find('source') is not None:
+            if disk.xmltreefile.find('source') is not None and \
+                    not ('ppc64le' in platform.machine()
+                         and 'file' not in disk.source.attrs):
                 if disk.source.attrs['file'] != source_file:
                     continue
             else:
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_update_device_matrix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_update_device_matrix.py
@@ -2,6 +2,7 @@ import os
 import time
 import shutil
 import logging
+import platform
 from avocado.core import exceptions
 from avocado.utils import process
 from virttest import virsh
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
```diff
@@ -36,7 +37,9 @@ def is_attached(vmxml_devices, disk_type, source_file, target_dev):
                 continue
             if disk.target['dev'] != target_dev:
                 continue
-            if disk.xmltreefile.find('source') is not None:
+            if disk.xmltreefile.find('source') is not None and \
+                    not ('ppc64le' in platform.machine()
+                         and 'file' not in disk.source.attrs):
```
chunfuwen: this logic will apply platform except ppc64le

chloerh: not (A and B) could only be True when (A and B) is False which means A is False or B is False.
A is False: not ppc64le, nothing's changed just like before. Go on to check disk.source.attrs['file']
B is False: disk.source.attrs DOES have 'file' in the attrs.  Go on to check disk.source.attrs['file']

chloerh: There's ONLY one situation that will be skipped, which is 

ppc64le and disk.source.attrs['file'] is None

chunfuwen: Assume :
A = disk.xmltreefile.find('source') is not None
B = 'ppc64le' in platform.machine()
C= 'file' not in disk.source.attrs
your expression A and not (B and C) = A and ( not B  or  not C) = (A and not B ) or (A and not C)
so the latter part  (A and not C) is confused since it is generic,not specific to ppc64le

chloerh: By your logic which **I don't think is right**, since I have put '()' there to clarify the priority of the expression,  I can explain this way:
You know that 'OR' expression can be True even there's only 1 parts is True, right?
Therefore, 

1. when  (A and not B ) is True, which means not ppc, then there's no need to calculate
(A and not C), since the 'OR' expression is already True, it doesn't matter whether the latter part is True of False.
2. when (A and not B ) is False, which means ppc, then the latter part matters, that happens to be what I want to fix with this commit.

I think this is very clear, although I don't think it should be explained this way, since the '()' are already there.

chloerh: @chunfuwen if it's ok with you, please help merge it, the error is causing 200+ failures of attach_detach_media job of ppc

chunfuwen: As I mentioned before, it may function correctly,but from the human-understandable point view, suggest create separate if statement

dzhengfy:
Due to PR 2558, I close this one.

PR #2538, opened 2020-02-06
Add pkg dependency list to powerpc_hmi and guestpin stress tests

chloerh:
Stress tests requires stress test tool installed which requies
pkgs 'gcc, make' installed inside vms.


sathnaga:
LGTM, minor comment, just add little detailed commit msg...

chloerh:
@sathnaga commit message updated, thanks for reviewing

PR #2544, opened 2020-02-11
Fix qemu-img error

dzhengfy:
The error is "TestError: Image size provided is not in valid format,
specify proper units [K|M|G".
setup_or_cleanup_iscsi call iscsi.py which needs image size is ended
with an unit. So this fix is to provide a virtual size with an unit.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1059,15 +1059,24 @@ def read_domjobinfo(test, domjobinfo):
     return jobinfo_dict
 
 
-def get_virtual_size(disk_source):
+def get_virtual_size(test, disk_source):
     """
     Get the virtual size of image.
+
     :param disk_source: path to the image
     :return: virtual size, e.g. 10G
+    :raise: test.error if checking fails
     """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1066,8 +1066,11 @@ def get_virtual_size(disk_source):
     :return: virtual size, e.g. 10G
     """
     result = process.run("qemu-img info %s" % disk_source, ignore_status=False)
-    size_pattern = r"virtual size: (\d+\w)"
-    return re.findall(size_pattern, to_text(result.stdout))[0]
+    size_pattern = r"virtual size: (\d+) (.*) \("
+    match = re.findall(size_pattern, to_text(result.stdout))[0]
+    size = match[0]
```
chunfuwen: Better do pre-check for match before call match[0]

chloerh: `if match:
    xxxx
`

dzhengfy: updated @chloerh @chunfuwen

PR #2558, opened 2020-02-21
Update conf of disk dev and bus to support q35 and pseries

chloerh:
IDE is not supported by q35, update bus to SCSI which is supported.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_change_media_matrix.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_change_media_matrix.cfg
@@ -6,6 +6,14 @@
     change_media_new_iso = "change_media_new.iso"
     change_media_source =
     change_media_init_iso = "change_media_old.iso"
+    change_media_target_bus = "ide"
+    change_media_target_device = "hdc"
+    q35:
+        change_media_target_bus = "scsi"
+        change_media_target_device = "sdc"
+    pseries:
+        change_media_target_bus = "scsi"
+        change_media_target_device = "sdc"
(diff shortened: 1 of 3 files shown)


chunfuwen:
@chloerh , please upload test result

chloerh:
@chunfuwen Test result would be all fail, since this commit is to expose problems that have been covered. All these 700+ cases have been blacked for 'not supported by q35'. We could accept failures, or keep them blacked and overlook the problems.
And @qiankehan should be the feature owner who will review this when he has the time.

dzhengfy:
> @chunfuwen Test result would be all fail, since this commit is to expose problems that have been covered. All these 700+ cases have been blacked for 'not supported by q35'. We could accept failures, or keep them blacked and overlook the problems.
> And @qiankehan should be the feature owner who will review this when he has the time.

@chloerh , maybe you can paste the failed error message and say  this error will not be exposed without your PR. So you PR is an improvement.

chloerh:
> > @chunfuwen Test result would be all fail, since this commit is to expose problems that have been covered. All these 700+ cases have been blacked for 'not supported by q35'. We could accept failures, or keep them blacked and overlook the problems.
> > And @qiankehan should be the feature owner who will review this when he has the time.
> 
> @chloerh , maybe you can paste the failed error message and say this error will not be exposed without your PR. So you PR is an improvement.

Sure.
In ppc test result, error message is like
`KeyError: 'file'&#10;`
But, in x86 result, tests were blacked with
`Skip Message
[BLACKLISTED] Feature 'IDE controllers' is not supported in guest with q35`

500+ cases blacked with this reason.

PR #2562, opened 2020-02-24
Fixed Get wrong disk0 name sda issue

kumudag:
disk_only_spec, compress_default and multi_disk_external tests creates VM with sda as the disk and snapshot is created with vda. Later check_snapslist() is called to compare dname that fails due to mismatch(FAIL: Get wrong disk0 name sda).

So, config & test code is changed to take this disk name correctly.

Signed-off-by : Kumuda G <kumuda@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/snapshot/virsh_snapshot_create_as.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/snapshot/virsh_snapshot_create_as.cfg
@@ -91,7 +91,7 @@
                     snap_createas_opts = "a--a a--a--snap1"
                 - disk_only_spec:
                     snap_createas_opts = "--disk-only"
-                    diskspec_opts = "vda,snapshot=external,driver=qcow2,file=disk-snap.img"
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virsh_cmd/snapshot/virsh_snapshot_create_as.cfg
```diff
@@ -174,7 +174,7 @@
                     only disk_only_spec, with_diskspec, compress_default, multi_disk_external
                     replace_vm_disk = "yes"
                     disk_type = "network"
-                    disk_target = "vda"
+                    disk_target = "sda"
```
sathnaga: do not change the disk target to `sda`, in default test config, other params are aligned with virtio itself,
bydefault virtio disks named as `vda`..., and scsi as `sda`...

 let's handle inside the test for which ever disk option given in the config.

chunfuwen:
no update in long run, so close it firstly, but it can be reopened once needed again

PR #2570, opened 2020-03-03
Skip check when source tag has no 'file' attr

chloerh:
Xml usually does not have 'source' tag if the source file is not specified.
But now a tag like <source index='3' /> is being created, which is affecting test.
Therefore skip check if this happens.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
@@ -36,7 +36,8 @@ def run(test, params, env):
                 continue
             if disk.target['dev'] != target_dev:
                 continue
-            if disk.xmltreefile.find('source') is not None:
+            if disk.xmltreefile.find('source') is not None and \
+                    'file' in disk.source.attrs:
                 if disk.source.attrs['file'] != source_file:
                     continue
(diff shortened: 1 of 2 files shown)


chloerh:
@qiankehan would you like to review this?

libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
```diff
@@ -1,6 +1,7 @@
 import os
 import time
 import logging
+import platform
```
Yingshun: why import platform?

libvirt/tests/src/virsh_cmd/domain/virsh_update_device_matrix.py
```diff
@@ -2,6 +2,7 @@
 import time
 import shutil
 import logging
+import platform
```
Yingshun: same as above

libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
chloerh: Right, updated.

chloerh:
This could hold till the issue involved confirmed by @qiankehan

PR #2589, opened 2020-03-13
Block rhv output for special_name case

mxie91:
The guest whose nanme contains special charecter can't be power on
rhv so skip rhv output for the case.

Signed-off-by: mxie91 <mxie@redhat.com>

### DIFF
--- a/v2v/tests/cfg/function_test_esx.cfg
+++ b/v2v/tests/cfg/function_test_esx.cfg
@@ -121,6 +121,8 @@
             main_vm = 'VM_NAME_ESX_MULCPUS_V2V_EXAMPLE'
         - special_name:
             only esx_55
+            #rhv env does not allowed guest to have speical characters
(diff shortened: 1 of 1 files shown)


v2v/tests/cfg/function_test_esx.cfg
```diff
@@ -121,6 +121,7 @@
             main_vm = 'VM_NAME_ESX_MULCPUS_V2V_EXAMPLE'
         - special_name:
             only esx_55
+            only libvirt
```
xiaodwan: This should be added into blacklist instead of deleting it.

mxie91: This change is not deleting the case, just make sure the output is only libvirt.

xiaodwan:
The special characters doesn't support by rhv server, so this case should be removed from esx cases.

PR #2590, opened 2020-03-13
add install ceph-common on remote migration host

chunfuwen:

### DIFF
--- a/libvirt/tests/src/migration/migrate_ceph.py
+++ b/libvirt/tests/src/migration/migrate_ceph.py
@@ -499,6 +499,13 @@ def run(test, params, env):
     # Setup migration context
     migrate_setup = libvirt.MigrationTest()
     migrate_setup.migrate_pre_setup(test_dict["desuri"], params)
+
+    # Install ceph-common on remote host machine.
+    remote_ssh_session = remote.remote_login("ssh", server_ip, "22", server_user,
+                                             server_pwd, r"[\#\$]\s*$")
+    if not utils_package.package_install(["ceph-common"], remote_ssh_session):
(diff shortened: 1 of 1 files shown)


chunfuwen:
''' paste test result here"
'''
Attempting to log into 10.66.85.245:22 using ssh (timeout 240s)
Sending command: firewall-cmd --remove-port=49152-49216/tcp --permanent --zone=public
Sending command: echo $?
Sending command: firewall-cmd --reload
Sending command: echo $?
Sending command: grep cpu /proc/cpuinfo | awk '{print $3}' | head -n 1
Sending command: echo $?
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_ceph.auth.with_postcopy

'''

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -499,6 +499,13 @@ def run(test, params, env):
     # Setup migration context
     migrate_setup = libvirt.MigrationTest()
     migrate_setup.migrate_pre_setup(test_dict["desuri"], params)
+
+    # Install ceph-common on remote host machine.
+    remote_ssh_session = remote.remote_login("ssh", server_ip, "22", server_user,
+                                             server_pwd, r"[\#\$]\s*$")
+    if not utils_package.package_install(["ceph-common"], remote_ssh_session):
```
Yingshun: better to check if the package is installed first

chunfuwen: that will be handled by utils_package.package_install itself , I believe.

PR #2603, opened 2020-03-20
Wait some time before doing managedsave/save for pseries vms

chloerh:
In some cases, doing managedsave/save too soon will cause failures
after restoring.


chloerh:
> IMHO this should be platform independent (managedsave/save to soon can cause error in restore).
> Therefore, I think it makes more sense to define a global constant or config param so that the test can be tuned depending if it's necessary or not, e.g.
> `time.sleep(WAIT_BEFORE_SAVE_SECS)`
> 
> @chloerh What do you think?

Great idea! :+1:

PR #2620, opened 2020-03-27
Fix domiftune teardown

smitterl:
1. Restart service only if `libvirtd` crashed: the original code created a race
   condition where the restart happened while the test also tried to log into
   the guest on serial console for dmesg verification; this caused the console
   to be disconnected. The login never succeeded but timed out and dmesg could
   not be verified.
   Don't catch exception from `start`: if this fails all other test steps won't
   be executed correctly, so letting error pass at this point will avoid late
   test failure.
2. `check_libvirtd`: create method to check and handle if `libvirtd` crashed.
   Use the same error message because test name already determines which case
   was run.
3. Move service restart to crash checker: the restart is only necessary
   if later code requires it to be running, e.g. when reestablishing
   the bandwidth restrictions.
4. Make `libvirtd` global to `run` and pass as param; `libvirtd` instance
   was created several times unnecessarily.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domiftune.py
@@ -180,10 +180,26 @@ def check_domiftune(params, test_clear):
     return True
 
 
-def get_domiftune_parameter(params, test):
+def check_libvirtd(test, libvirtd):
+    """
+    Check if libvirtd is running and restart to ensure available for cleanup.
+
+    :param test: test instance
+    :param libvirtd: libvirt daemon instance
+    """
+    if not libvirtd.is_running():
+        try:
+            libvirtd.start()
+        finally:
+            test.fail('Libvirtd crashed after running `virsh domiftune`')
+
+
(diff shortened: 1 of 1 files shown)


smitterl:
```bash
# avocado run --vt-type libvirt --vt-connect-uri qemu:///system virsh.domiftune.positive_testing.set_domif_parameter.running_guest.change_outbound.options.live.minimum_boundary,virsh.domiftune.positive_testing.get_domif_parameter.running_guest.options.live --vt-extra-params verify_guest_dmesg=yes
JOB ID     : 21b81bc31b14f915b745f9315159858be0f219e5
JOB LOG    : /root/avocado/job-results/job-2020-03-27T11.02-21b81bc/job.log
 (1/2) type_specific.io-github-autotest-libvirt.virsh.domiftune.positive_testing.get_domif_parameter.running_guest.options.live: PASS (26.28 s)
 (2/2) type_specific.io-github-autotest-libvirt.virsh.domiftune.positive_testing.set_domif_parameter.running_guest.change_outbound.options.live.minimum_boundary: PASS (32.14 s)
JOB TIME   : 59.76 s
```

PR #2627, opened 2020-03-31
Skip gluster rdma cases in blockcommit feature

Yingshun:
Transport protocol 'rdma' is not yet supported for now.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
@@ -17,7 +17,7 @@ from virttest import gluster
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
 
-from provider import libvirt_version
+from virttest import libvirt_version
 
 
 def get_images_with_xattr(vm):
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
```diff
@@ -271,6 +271,11 @@ def check_chain_backing_files(disk_src_file, expect_backing_file=False):
             test.cancel("live active block commit is not supported"
                         " in current libvirt version.")
 
+    # This is brought by new feature:block-dev
+    if (libvirt_version.version_compare(6, 0, 0) and
+       params.get("transport", "") == "rdma"):
+        test.cancel("Transport protocol 'rdma' is not yet supported.")
```
chunfuwen: maybe change to "Transport protocol 'rdma' is not yet supported in block-dev"

Yingshun: the error message comes from virsh.start, so will keep the original changing.

chunfuwen: it is confused. Actually before  block-dev replace block driver as new feature, Transport protocol 'rdma' is supported.

Yingshun: updated.

PR #2635, opened 2020-04-01
Add version check before running the cases of print-estimate

xiaodwan:


xiaodwan:
It's just a simple configuration change and has no harm to other cases. I think it's not worth wasting time setting an environment just for a silly test.

xiaodwan:
This can avoid the failure on rhel8 slow which has  a very low virt-v2v version that doesn't support 'print-estimate' option.

xiaodwan:
@chloerh @chunfuwen Please help review this PR, thanks.

PR #2641, opened 2020-04-03
src: Drop domain_opt and file_opt

qiankehan:
These two arugments are duplicated with the arugments domainarg and
filearg.

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -692,7 +692,7 @@ def run(test, params, env):
                   'target_type': target_type,
                   'target_name': target_name}
         channel_xml = libvirt.create_channel_xml(params)
-        virsh.attach_device(domain_opt=vm_name, file_opt=channel_xml.xml,
+        virsh.attach_device(vm_name, channel_xml.xml,
                             flagstr="--config", ignore_status=False)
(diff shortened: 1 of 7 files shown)


chunfuwen:
@qiankehan , previous code experiences are good practices with key=value as input parameter
The benefit for these have at least items below :
1)More clear about each parameter meaning(human readable)
2)Even put parameter in wrong order for method, it still can work normally since it is key=value.
Anyway, welcome your thoughts as well.

chunfuwen:
@kylazhang 
@dzhengfy 
@Yingshun 
Please provide your comments.

qiankehan:
@chunfuwen Let me try to explain it more clearly:  
The key problem I tried to fix is that **domain_opt** and **file_opt** of method **attach_device** are dup to **domainarg** and **filearg**. However, direct remove in avocado-vt will cause the case failed at where the codes using the duplicated arguments. So I first update that in tp-libvirt. But I agree that key=value form is more readable.  
BTW, the dup agruments should be remove from **detach_device update_device**, too  
They should be like:
```py
def attach_device(domainarg=None, filearg=None, flagstr=None, **dargs)
```

qiankehan:
@chunfuwen  Any more review?

chloerh:
Code check haven't completed for some reason.

qiankehan:
> Code check haven't completed for some reason.

Passed now

PR #2647, opened 2020-04-03
Modify the default perms of socket file after 5.6.0

yafu-1:

### DIFF
--- a/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
+++ b/libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
@@ -3,12 +3,15 @@ import stat
 import logging
 import os
 from avocado.utils import distro
+from avocado.utils import process
 from avocado.core import exceptions
 
 from virttest import virsh
 from virttest import data_dir
 from virttest import utils_config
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/conf_file/libvirtd_conf/unix_sock.py
```diff
@@ -168,3 +175,4 @@ def check_all_unix_sock(group, ro_perms, rw_perms, root_path):
     finally:
         config.restore()
         libvirtd.restart()
+        process.system("systemctl restart libvirtd.socket", ignore_status=True)
```
chunfuwen, line 178: We need simulate libvirtd.restart() to controll libvirtd.socket , where libvirtd operation is packaged in one method. Otherwise we can not easily disable function after split libvird daemons feature is enabled in near coming feature.

yafu-1: libvirtd.socket restart in this part just used to avoid the connection refused if the libvirtd starting in the foreground before. libvirtd.restart() does not need to restart the socket file in normal use.

PR #2648, opened 2020-04-03
Fix issue of adding pci-bridge incorrectly

chloerh:
When there's pcie-root-port, no longer add pci-bridge

### DIFF
--- a/libvirt/tests/src/sriov/sriov.py
+++ b/libvirt/tests/src/sriov/sriov.py
@@ -608,7 +608,8 @@ def run(test, params, env):
         pci_bridge_controllers = []
         for device in controller_devices:
             logging.debug(device)
-            if device.type == 'pci' and device.model == "pci-bridge":
+            if device.type == 'pci' and device.model in (
+                    "pci-bridge", "pcie-root-port"):
                 pci_bridge_controllers.append(device)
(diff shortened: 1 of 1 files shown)


chunfuwen:
is there any validation result  available?

chloerh:
> is there any validation result available?

No, we need to test and see

libvirt/tests/src/sriov/sriov.py
```diff
@@ -608,7 +608,8 @@ def remove_devices(vmxml, device_type):
         pci_bridge_controllers = []
         for device in controller_devices:
             logging.debug(device)
-            if device.type == 'pci' and device.model == "pci-bridge":
+            if device.type == 'pci' and device.model in (
+                    "pci-bridge", "pcie-root-port"):
```
dzhengfy, line 612: I can not understand this code clearly. For example, when device.model == pcie-root-port and device.type == pci,  the device will be added to pci_bridge_controllers, but it is a pcie-root-port controller instead of pci-bridge. Is that expected by you?

chloerh: We should not add pci-bridge to vm when there's already pcie-root-port there.

PR #2653, opened 2020-04-07
Fix blockcommand block device

smitterl:
Introduction of `-blockdev` has unmasked a failure in test script:
it validates the `filename` parameter and doesn't allow for non-block
devices for `driver=host_device`.

The test script created originally
1. a logical volume as a block device (`/dev/HostVG/vol1`)
2. an image file using the logical volume as backingchain, that is,
   it created `/dev/HostVG/vol2` which is a file, not a block device
3. attached the image file to the vm

Libvirt doesn't check if the path for `@dev` is really a blockdevice in
`<source dev='/dev/HostVG/vol2'/>` so that the file was correctly
attached before `-blockdev` was introduced, `-drive file=/dev/HostVG/vol2,format=qcow2,if=none,id=drive-virtio-disk1 -device virtio-blk-ccw,scsi=off,devno=fe.0.0009,drive=drive-virtio-disk1,id=virtio-disk1`.

Related bz https://bugzilla.redhat.com/show_bug.cgi?id=1821663

If using a volume as block device we can only use it as bottom
in a backingchain. Adapt the code accordingly:

1. Use consistently `vol1` for volume and block device
2. Remove unused `cmd_create_img`: this command was never used but
   overwritten later in the code https://github.com/autotest/tp-libvirt/blob/cd030eef4852e8eff31459895ca30ef4b08aa04a/libvirt/tests/src/backingchain/blockcommand.py#L192
3. Allow for last element in snapshot image list not to have a backing file

Although this change is related to introduction of `-blockdev` for
qemu-kvm 4.2 and libvirt 5.10 the test script doesn't need to switch
over these versions as the expected behavior hasn't changed but rather
this change also applies to older versions.

Also, change clean up command to make sure vg path is removed.

### DIFF
--- a/libvirt/tests/src/backingchain/blockcommand.py
+++ b/libvirt/tests/src/backingchain/blockcommand.py
@@ -28,7 +28,8 @@ def check_backingchain(img_list, img_info):
     pattern = ''
     for i in range(len(img_list) - 1):
         pattern += 'image: ' + img_list[i]
-        pattern += '.*backing file: ' + img_list[i + 1]
+        if img_list[i + 1]:
+            pattern += '.*backing file: ' + img_list[i + 1]
         pattern += '.*'
 
     logging.debug(pattern)
@@ -138,6 +139,7 @@ def run(test, params, env):
         # Setup lvm
         elif disk_src == 'lvm':
             # Stop multipathd to avoid vgcreate fail
+            lvm_vol_size = '100M'
             multipathd = service.Factory.create_service("multipathd")
             multipathd_status = multipathd.status()
             if multipathd_status:
@@ -156,13 +158,12 @@ def run(test, params, env):
 
             # Create logical volume as backing store
             vol_bk, vol_disk = 'vol1', 'vol2'
-            lv_utils.lv_create(vg_name, vol_bk, '100M')
+            if disk_type == "block":
+                vol_bk, vol_disk = 'vol1', 'vol1'
+            lv_utils.lv_create(vg_name, vol_bk, lvm_vol_size)
 
             disk_target = '/dev/%s/%s' % (vg_name, vol_bk)
             src_vol = '/dev/%s/%s' % (vg_name, vol_disk)
-
-            # Command of creating block disk
-            cmd_create_img = 'qemu-img create -f qcow2 -b %s %s' % (disk_target, src_vol)
 
         # Setup gluster
         elif disk_src == 'gluster':
@@ -190,6 +191,8 @@ def run(test, params, env):
             file_to_del.append(new_image)
 
         cmd_create_img = 'qemu-img create -f qcow2 -b %s %s' % (disk_target, new_image)
+        if disk_type == "block":
+            cmd_create_img = 'qemu-img create -f qcow2 %s %s' % (disk_target, lvm_vol_size)
         process.run(cmd_create_img, verbose=True, shell=True)
         info_new = utils_misc.get_image_info(new_image)
         logging.debug(info_new)
@@ -234,7 +237,10 @@ def run(test, params, env):
             qemu_img_cmd += " -U"
         bc_info = process.run(qemu_img_cmd, verbose=True, shell=True).stdout_text
 
-        bc_chain = snapshot_image_list[::-1] + [new_image, disk_target]
+        if not disk_type == "block":
+            bc_chain = snapshot_image_list[::-1] + [new_image, disk_target]
+        else:
+            bc_chain = snapshot_image_list[::-1] + [disk_target, None]
         bc_result = check_backingchain(bc_chain, bc_info)
         if not bc_result:
             test.fail('qemu-img info output of backing chain is not correct: %s'
@@ -287,7 +293,6 @@ def run(test, params, env):
         if disk_src == 'iscsi':
             libvirt.setup_or_cleanup_iscsi(is_setup=False)
         elif disk_src == 'lvm':
-            process.run('rm -rf /dev/%s/%s' % (vg_name, vol_disk), ignore_status=True)
             if 'vol_bk' in locals():
                 lv_utils.lv_remove(vg_name, vol_bk)
             if 'vg_created' in locals() and vg_created:
@@ -296,6 +301,7 @@ def run(test, params, env):
                 pv_name = process.system_output(cmd, shell=True, verbose=True).strip()
                 if pv_name:
                     process.run("pvremove %s" % pv_name, verbose=True, ignore_status=True)
+            process.run('rm -rf /dev/%s' % vg_name, ignore_status=True)
             libvirt.setup_or_cleanup_iscsi(is_setup=False)
         elif disk_src == 'gluster':
             gluster.setup_or_cleanup_gluster(is_setup=False, vol_name=vol_name, brick_path=brick_path, pool_name=pool_name)


smitterl:
Test execution with `blockdev` usage:
```bash
# rpm -q libvirt && rpm -q qemu-kvm
libvirt-6.0.0-13.module+el8.2.1+6089+d30049aa.s390x
qemu-kvm-4.2.0-15.module+el8.2.0+6029+618ef2ec.s390x

# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk
JOB ID     : 100a5950e073d27bda5541beba0c0e7783e778e7
JOB LOG    : /root/avocado/job-results/job-2020-04-07T05.41-100a595/job.log
 (1/1) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk: PASS (220.88 s)
JOB TIME   : 223.25 s
```

smitterl:
Test execution without `blockdev`:
```bash
# rpm -q libvirt && rpm -q qemu-kvm
libvirt-5.6.0-10.module+el8.1.1+5309+6d656f05.s390x
qemu-kvm-4.1.0-23.module+el8.1.1+5938+f5e53076.2.s390x

# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk
JOB ID     : 286bed8003a08c3cb72c50a96075842115ccb2d6
JOB LOG    : /root/avocado/job-results/job-2020-04-07T06.17-286bed8/job.log
 (1/1) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk: PASS (89.61 s)
JOB TIME   : 91.45 s
```

libvirt/tests/src/backingchain/blockcommand.py
```diff
@@ -190,6 +191,8 @@ def run(test, params, env):
             file_to_del.append(new_image)
 
         cmd_create_img = 'qemu-img create -f qcow2 -b %s %s' % (disk_target, new_image)
+        if disk_type == "block":
```
chunfuwen, line 194: I assume previous logic is to create one file backend new_image for disk_target, and attach backend new_image to VM.
If here you remove backend creation, it will change the scenario intended for this case.

libvirt/tests/src/backingchain/blockcommand.py
```diff
@@ -190,6 +191,8 @@ def run(test, params, env):
...
+            cmd_create_img = 'qemu-img create -f qcow2 %s %s' % (disk_target, lvm_vol_size)
```
chunfuwen, line 195: Here,it sound like it doesn't make sense that disk_target was created again since it was already available by previous lvm_create.

chunfuwen:
@smitterl 
I observe another solution for your reference:
Since backing file is file,so we can change disk type from 'block'
 to 'file' when creating disk xml, and the case can pass. Some piece of code follow as:
        # Create xml of new disk and add it to vmxml
        if disk_type:
            new_disk = Disk()
            new_disk.xml = libvirt.create_disk_xml({
                'type_name': 'file',
                'driver_type': 'qcow2',
                'target_dev': new_dev,
                'source_file': new_image
            })

********************************************
PASS 1-type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk

smitterl:
> @smitterl
> I observe another solution for your reference:
> Since backing file is file,so we can change disk type from 'block'
> to 'file' when creating disk xml, and nothing else need to be changed more.
> Finally the case can pass. Some piece of code follow as:
> # Create xml of new disk and add it to vmxml
> if disk_type:
> new_disk = Disk()
> new_disk.xml = libvirt.create_disk_xml({
> 'type_name': 'file',
> 'driver_type': 'qcow2',
> 'target_dev': new_dev,
> 'source_file': new_image
> })
> 
> DATA (filename=output.expected) => NOT FOUND (data sources: variant, test, file)
> DATA (filename=stdout.expected) => NOT FOUND (data sources: variant, test, file)
> DATA (filename=stderr.expected) => NOT FOUND (data sources: variant, test, file)
> PASS 1-type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk

@chunfuwen 
In this case you don't test a block device IIUC? I mean, the test case is called 'block_disk' and the intention is to use a valid block device configuration 'https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/cfg/backingchain/blockcommand.cfg'
In your proposed setup, what is the final disk xml? Is the first disk to be attached really a block device /dev/HostVG/volX, or just the first snapshot file with backingfile the block device?

libvirt/tests/src/backingchain/blockcommand.py
smitterl: I don't think this to be the case: 'lv_create' creates the volume, however, 'qemu-img' creates an image of format 'qcow2' on the volume for it to be attached with 'driver@format='qcow2'". I added this line because the original code created a file '/dev/HostVG/vol2' and attached it erroneously as block device which is only possible due to the known issue https://bugzilla.redhat.com/show_bug.cgi?id=1821663
However, the test intention is to create and block(pull|commit) starting from a block device. Therefore the original block device '/dev/HostVG/vol1' is formatted as qcow2 here.
Does this make sense?

chunfuwen: In my understanding,here by your change, qemu-img will  only override the volume previously created by lvm_create.
Furthermore, backing file scenarios will be lost.

chunfuwen:
@smitterl , in most cases the backing file we used is one image file created by qemu-img, but the backend storage type can be gluster, lvm, ceph ,iscsi ,etc. So here it make sense that we change backing file disk type from block to file since backing file is itself file-based image file.

smitterl:
@qiankehan @lento-sun Could you have a look? I'm unsure what the actual test intention is. The status of the current discussion IIUC is that there are two possibilities to fix the current test failure, that is, when libvirt uses 'blockdev' disk@type=block will use blockdev with driver 'host_device' but that doesn't work with a file on //source@dev.

1. The variant 'block_disk' sets `disk_type=block`- https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/cfg/backingchain/blockcommand.cfg. Therefore, I expect the test script to attach the volume /dev/HostVG/vol1 to the machine as such, /devices/disk@type=block/source@dev=/dev/HostVG/vol1 . Currently, a file is created (/dev/HostVG/vol2) and attached - which is wrong https://bugzilla.redhat.com/show_bug.cgi?id=1821663. If this is the expected test case, should the block device still be formated with qemu-img?
2. If I understand https://github.com/autotest/tp-libvirt/pull/2653#issuecomment-611314530 correctly, the attached device should rather be a file that's placed _on_ a block device, therefore solution proposal https://github.com/autotest/tp-libvirt/pull/2653#issuecomment-610719238

lento-sun:
> IIUC

The failed cases should be in commit 62429c1c as "- **when guest image with "block" disk has "block" backing file**"
And the original script only created a **FILE** named "/dev/HostVG/vol2", not even used lvcreate to create /dev/HostVG/vol2 as lvm volume. So I think it's a script issue but PASSED since qemu allow a file to be used as type=block before --blockdev enabled.
From my opinion, the cases with pattern = "backingchain.blockcommand..block_disk.block_bk" should actually having block devices used as active layer and backing layer. It should be something like:
/dev/sde based on /dev/sdd, and used /dev/sde as <disk/> source.  Or, if type=block still supports lvm volume, we can use /dev/HostVG/vol2 based on /dev/HostVG/vol1. But vol1 and vol2 should be real lvm volumes.

qiankehan:
> > IIUC
> 
> The failed cases should be in commit [62429c1](https://github.com/autotest/tp-libvirt/commit/62429c1c3146b256358880438b911dcbd44bd021) as "- **when guest image with "block" disk has "block" backing file**"
> And the original script only created a **FILE** named "/dev/HostVG/vol2", not even used lvcreate to create /dev/HostVG/vol2 as lvm volume. So I think it's a script issue but PASSED since qemu allow a file to be used as type=block before --blockdev enabled.
> From my opinion, the cases with pattern = "backingchain.blockcommand..block_disk.block_bk" should actually having block devices used as active layer and backing layer. It should be something like:
Agree. From the case name, the uplayer image and the backing image should be all block devices. They can be create by **lvcreate**(created as LV) OR **fdisk**(created as partition) OR **iscsiadm/scsi_debug**(created as scsi block)

The cmd `qemu-img create -f qcow2 /dev/...` is used to format the host_device. So the blk file should  exist before `qemu-img create`. Otherwise you will create a common file.   
However  `qemu-img create` is not necessary if it be used as **raw** format  

One more notification that, considering the qcow2 metadata, the formatted virtual size of qcow2 should be smaller than the host_device size. In other words, qcow2 metadata size + virtual size <= host_device size. Otherwise you may get I/O error in guest if the guest try to read/write the data beyond the actual host_device size.

> /dev/sde based on /dev/sdd, and used /dev/sde as source. Or, if type=block still supports lvm volume, we can use /dev/HostVG/vol2 based on /dev/HostVG/vol1. But vol1 and vol2 should be real lvm volumes.

libvirt/tests/src/backingchain/blockcommand.py
qiankehan: > Here,it sound like it doesn't make sense that disk_target was created again since it was already available by previous lvm_create.

It make sense here if you want to use the host block device as qcow2. It will **format the block device with qcow2 header**. But the formatted size should be **smaller than lv_size** to reserve some space for qcow2 header.
pseudo-code:
```
if disk_target is not blk_file:
    # create blk dev by lvcreate
    create_blk(disk_target)
if disk_format == "qcow2":
    # virtual size is smaller than lvm_vol_size
    cmd_create_img = 'qemu-img create -f qcow2 %s %s' % (disk_target, virtual_size)
```

smitterl:
Thank you everybody. After this discussion and on IRC, I'll close this and create a new PR with the change we agreed upon.

PR #2594, opened 2020-03-16
Remove 52lts compatiblity layer dependences

chunfuwen:
Remove any reference to results_stdout_52lts,results_stderr_52lts,decode_to_text from
compat_52lts.py under avocado-vt

Signed-off-by: chunfuwen <chwen@redhat.com>

PR #2583, opened 2020-03-11
Fix virtual network pxe boot failure

kylazhang:
1. Add missing step for ldlinux.c32 copy
2. Remove the failure of debug output
from get_stripped_output to avoid error
ERROR: [1S is not included in the known console codes list
which is caused in aexpect and not fixed now. please check
https://github.com/avocado-framework/aexpect/pull/48 for detailed info

Signed-off-by: Kylazhang <weizhan@redhat.com>

PR #2659, opened 2020-04-09
Fix virtual network pxe boot failure

kylazhang:
Add this patch again as the previous code is change back by another pr by mistake(should be not rebase before submit)
Approved and merged pr before: https://github.com/autotest/tp-libvirt/pull/2583
Removed pr: https://github.com/autotest/tp-libvirt/pull/2594

1. Add missing step for ldlinux.c32 copy
2. Remove the failure of debug output
from get_stripped_output to avoid error
ERROR: [1S is not included in the known console codes list
which is caused in aexpect and not fixed now. please check
https://github.com/avocado-framework/aexpect/pull/48 for detailed info


kylazhang:
Test

pevogam:
@kylazhang What about the issue on the aexpect side then? Do you think it should be closed? Do you think https://github.com/avocado-framework/aexpect/pull/57 could solve it too in a more universal way and would like to give a thumps up for merging it instead?

PR #2675, opened 2020-04-14
Fix incorrect param name

dzhengfy:
When stress_install_from_repo = yes, the stress package should be installed from repo. But it also requires 'stress_package' parameter. This parameter is missed in the case.


dzhengfy:
Before fix:
Installation failed with output: checking for a BSD-compatible install... /usr/bin/..
After fix:
2020-04-14 17:50:18,901 client           L1126 DEBUG| Sending command: rpm -q stress
2020-04-14 17:50:19,063 client           L1126 DEBUG| Sending command: echo $?
2020-04-14 17:50:19,165 client           L1126 DEBUG| Sending command: yum install -y stress
2020-04-14 17:50:24,933 client           L1126 DEBUG| Sending command: echo $?
2020-04-14 17:50:26,936 __init__         L2267 DEBUG| Successful to install stress tool via repo

PR #2676, opened 2020-04-14
Fix blockcommand test for block device

smitterl:
Introduction of `-blockdev` has unmasked a failure in the test script.
The backing chain that was created originally for `disk_type=block` consisted
only of files instead of block devices. This didn't lead to issues before
introduction of `-blockdev` because of https://bugzilla.redhat.com/show_bug.cgi?id=1821663

The test intention is to use block devices (volumes) for all snapshots.

Therefore, create volumes for all files in the chain beforehand to make
sure snapshots are created to block device with backing file another block
device, that is
`/dev/HostVG/vol1 <-- /dev/HostVG/vol2 <-- ... <-- /dev/HostVG/vol5`
for 4 snapshots, attaching `vol1` as base.

The test case should use `vol1` as `raw`.

`.py`:
1. `vol_size`: introduce variable for reuse
2. `driver_type`: introduce variable in order to allow for raw block
   device to be attached, use previous value as default to maintain
   legacy behavior
3. `check_backingchain`: in case of a block device we start the backing
   chain directly with `/dev/HostVG/vol1`; therefore, the chain is shorter
   than in the case of `file_disk` variants because `vol1` won't have a
   backing file, e.g. there is no element `img[i + 1]` for the last
   element `i` in the list
4. `cmd_create_img`: remove unused definition; it's overwritten further
   down
5. create all volumes: in case of lvm volumes, create all volumes
   for snapshots beforehand and use the `vol1` as first image; don't
   remove `src_vol` from code as this can be used to configure a test
   case of type `file` with `lvm` source volume as first backing file
6. `cmd_create_img`: skip device formatting for raw block devices
7. `bc_chain`: there's no backing file for `vol1`

`.cfg`:
1. `driver_type`: set `raw` for `vol1` to be attached as raw block device

### DIFF
--- a/libvirt/tests/cfg/backingchain/blockcommand.cfg
+++ b/libvirt/tests/cfg/backingchain/blockcommand.cfg
@@ -15,6 +15,7 @@
                     pool_name = 'pool_blockpull'
         - block_disk:
             disk_type = 'block'
+            driver_type = 'raw'
             variants:
                 - block_bk:
                     disk_src = 'lvm'
--- a/libvirt/tests/src/backingchain/blockcommand.py
+++ b/libvirt/tests/src/backingchain/blockcommand.py
@@ -26,9 +26,11 @@ def check_backingchain(img_list, img_info):
     :return: bool, meets expectation or not
     """
     pattern = ''
-    for i in range(len(img_list) - 1):
+    chain_length = len(img_list)
+    for i in range(chain_length):
         pattern += 'image: ' + img_list[i]
-        pattern += '.*backing file: ' + img_list[i + 1]
+        if i + 1 < chain_length:
+            pattern += '.*backing file: ' + img_list[i + 1]
         pattern += '.*'
 
     logging.debug(pattern)
@@ -104,10 +106,12 @@ def run(test, params, env):
     check_func = params.get('check_func', '')
     disk_type = params.get('disk_type', '')
     disk_src = params.get('disk_src', '')
+    driver_type = params.get('driver_type', 'qcow2')
     vol_name = params.get('vol_name', 'vol_blockpull')
     pool_name = params.get('pool_name', '')
     brick_path = os.path.join(data_dir.get_tmp_dir(), pool_name)
     vg_name = params.get('vg_name', 'HostVG')
+    vol_size = params.get('vol_size', '10M')
 
     vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     bkxml = vmxml.copy()
@@ -156,13 +160,10 @@ def run(test, params, env):
 
             # Create logical volume as backing store
             vol_bk, vol_disk = 'vol1', 'vol2'
-            lv_utils.lv_create(vg_name, vol_bk, '100M')
+            lv_utils.lv_create(vg_name, vol_bk, vol_size)
 
             disk_target = '/dev/%s/%s' % (vg_name, vol_bk)
             src_vol = '/dev/%s/%s' % (vg_name, vol_disk)
-
-            # Command of creating block disk
-            cmd_create_img = 'qemu-img create -f qcow2 -b %s %s' % (disk_target, src_vol)
 
         # Setup gluster
         elif disk_src == 'gluster':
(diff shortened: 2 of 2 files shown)


smitterl:
@chunfuwen @qiankehan @lento-sun This is the fix as result of discussion on https://github.com/autotest/tp-libvirt/pull/2653

smitterl:
```bash
# avocado run --vt-type libvirt backingchain.blockcommand..base_top.positive_test..block_bk
JOB ID     : ee5e9d45d6cf8390e1e6be24f35b8da2b56a0f5d
JOB LOG    : /root/avocado/job-results/job-2020-04-14T23.59-ee5e9d4/job.log
 (1/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.file_disk.block_bk: PASS (36.81 s)
 (2/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockpull.base_top.positive_test.block_disk.block_bk: PASS (41.50 s)
 (3/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockcommit.base_top.positive_test.file_disk.block_bk: PASS (38.74 s)
 (4/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommand.blockcommit.base_top.positive_test.block_disk.block_bk: PASS (40.58 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 159.20 s
```

libvirt/tests/cfg/backingchain/blockcommand.cfg
```diff
@@ -15,6 +15,7 @@
                     pool_name = 'pool_blockpull'
         - block_disk:
             disk_type = 'block'
+            driver_type = 'raw'
```
chloerh, line 18: According to the case description, the driver type is 'qcow2'. Looks like this does fix the problem, then does it mean that we need to update case description too?

libvirt/tests/src/backingchain/blockcommand.py
```diff
@@ -182,15 +183,23 @@ def run(test, params, env):
 
         new_image = os.path.join(os.path.split(image_file)[0], 'test.img')
         params['snapshot_list'] = ['s%d' % i for i in range(1, 5)]
-        snapshot_image_list = [new_image.replace('img', i) for i in params['snapshot_list']]
 
         if disk_src == 'lvm':
             new_image = src_vol
+            if disk_type == 'block':
+                new_image = disk_target
+                for i in range(2, 6):
```
chunfuwen, line 191: can those volume creation operations be combined with line 163? since they both are volume creation actions

smitterl: I didn't want to combine it with line 163 because this way another test with 'disk_type==file' and 'disk_src=lvm' can be defined and setup will still work. Let me know if this should be changed, thanks.

libvirt/tests/cfg/backingchain/blockcommand.cfg
smitterl: @qiankehan Can you confirm the test case should start with a raw block device instead of one that's formatted qcow2 so we can update the case description? Thank you.

qiankehan: @smitterl  For base image **lv1**, I think it's OK to use **raw** format.
@chloerh Where is the description that says qcow2 base image should be used?

chloerh: @qiankehan RHEL-114408 I think it's this one.

qiankehan: @chloerh  Well. Please update that manual case @meinaLi

meinaLi: @qiankehan We don't need to update that manual case because the base image has actually used raw format.

PR #2690, opened 2020-04-17
Add cases of iface operation with unprivileged user

chloerh:
- [bridge] Connect <interface> to a pre-created host tap device
in unprivileged libvirt
- [direct] connect <interface> to a pre-created host macvtap device
in unprivileged libvirt

depend on https://github.com/avocado-framework/avocado-vt/pull/2515

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_unprivileged.py
+++ b/libvirt/tests/src/virtual_network/iface_unprivileged.py
@@ -1,0 +1,213 @@
+import logging
+import shutil
+import aexpect
+
+from avocado.utils import process
+
+from virttest import libvirt_version
+from virttest import remote
+from virttest import utils_misc
+from virttest import utils_net
+from virttest import utils_package
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    Test interface with unprivileged user
+    """
+
+    def create_bridge(br_name, iface_name):
+        """
+        Create bridge attached to physical interface
+        """
+        # Make sure the bridge not exist
+        if libvirt.check_iface(br_name, "exists", "--all"):
+            test.cancel("The bridge %s already exist" % br_name)
+
+        # Create bridge
+        utils_package.package_install('tmux')
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up; pkill dhclient; ' \
+              'sleep 6; dhclient {0}; ifconfig {1} 0"'.format(br_name, iface_name)
+        process.run(cmd, shell=True, verbose=True)
+
+    def check_ping(dest_ip, ping_count, timeout, src_ip=None, session=None,
+                   expect_success=True):
+        """
+        Check if ping result meets expectaion
+        """
+        status, output = utils_net.ping(dest=dest_ip, count=ping_count,
+                                        interface=src_ip, timeout=timeout,
+                                        session=session, force_ipv4=True)
+        success = True if status == 0 else False
+
+        if success != expect_success:
+            test.fail('Ping result not met expectation, '
+                      'actual result is {}'.format(success))
+
+    if not libvirt_version.version_compare(5, 6, 0):
+        test.cancel('Libvirt version is too low for this test.')
+
+    vm_name = params.get('main_vm')
+    rand_id = '_' + utils_misc.generate_random_string(3)
+
+    upu_vm_name = 'upu_vm' + rand_id
+    user_vm_name = params.get('user_vm_name', 'non_root_vm')
+    bridge_name = params.get('bridge_name', 'test_br0') + rand_id
+    device_type = params.get('device_type', '')
+    iface_name = utils_net.get_net_if(state="UP")[0]
+    tap_name = params.get('tap_name', 'mytap0') + rand_id
+    macvtap_name = params.get('macvtap_name', 'mymacvtap0') + rand_id
+    remote_ip = params.get('remote_ip')
+    up_user = params.get('up_user', 'test_upu') + rand_id
+    case = params.get('case', '')
+
+    # Create unprivileged user
+    logging.info('Create unprivileged user %s', up_user)
+    process.run('useradd %s' % up_user, shell=True, verbose=True)
+    root_vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+
+    try:
+        # Create vm as unprivileged user
+        upu_virsh = virsh.VirshPersistent(unprivileged_user=up_user)
+        logging.info('Create vm as unprivileged user')
+        upu_vmxml = root_vmxml.copy()
+
+        # Prepare vm for unprivileged user
+        xml_devices = upu_vmxml.devices
+        disks = xml_devices.by_device_tag("disk")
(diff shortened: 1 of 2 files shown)


chloerh:
(1/2) type_specific.io-github-autotest-libvirt.virtual_network.iface_unprivileged.precreated.host_tap: PASS (78.50 s)
 (2/2) type_specific.io-github-autotest-libvirt.virtual_network.iface_unprivileged.precreated.host_macvtap: PASS (65.30 s)
JOB TIME   : 145.62 s

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+            if device_type == 'macvtap':
+                iface.mac_address = mac_addr
+            logging.debug(iface)
+
+            upu_vmxml.devices = all_devices
+            logging.debug(upu_vmxml)
+
+            # Define updated xml
+            shutil.copyfile(upu_vmxml.xml, new_xml_path)
+            upu_vmxml.xml = new_xml_path
```
dzhengfy, line 182: What is the reason for us to change the path of the xml? Couldn't we use the original path of upu_vmxml.xml to define the new guest?

dzhengfy:
Others LGTM

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: I couldn't remember exactly, but it seems when defining vm as unprivileged user, the path of xml should be at user's home directory, so that there won't be permission issues.

chloerh:
@dzhengfy updated

chloerh:
@Yingshun

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    Test interface with unprivileged user
+    """
+
+    def create_bridge(br_name, iface_name):
+        """
```
Yingshun, line 23: please update docstring for parameters.

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+        Create bridge attached to physical interface
+        """
+        # Make sure the bridge not exist
+        if libvirt.check_iface(br_name, "exists", "--all"):
+            test.cancel("The bridge %s already exist" % br_name)
```
Yingshun, line 28: can we remove the bridge if it exists?

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up; pkill dhclient; ' \
+              'sleep 6; dhclient {0}; ifconfig {1} 0"'.format(br_name, iface_name)
+        process.run(cmd, shell=True, verbose=True)
+
+    def check_ping(dest_ip, ping_count, timeout, src_ip=None, session=None,
+                   expect_success=True):
+        """
+        Check if ping result meets expectaion
+        """
```
Yingshun, line 41: please update docstring.

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+            xml_devices[disk_index].source = source
+            logging.debug(xml_devices[disk_index].source)
+
+        upu_vmxml.devices = xml_devices
+
+        new_xml_path = '/home/{}/upu.xml'.format(up_user)
+        shutil.copyfile(upu_vmxml.xml, new_xml_path)
+
+        # Define vm for unprivileged user
+        upu_virsh.define(new_xml_path)
```
Yingshun, line 111: what is the vm name in new_xml_path?  what will happen if it contains same name with main_vm?

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -0,0 +1,213 @@
...
+            else:
+                test.error('Invalid device type: {}'.format(device_type))
+
+            if_index = all_devices.index(iface)
+            iface = all_devices[if_index]
+            iface.type_name = 'ethernet'
+            iface.target = {
+                'dev': dev_name,
+                'managed': 'no'
+            }
```
Yingshun, line 171: I think you can remove all interfaces and then reuse libvirt.modify_vm_iface

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: This is a vm created by unprivileged user, therefore it doesn't matter if it's name is same with main vm

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: I don't think so, this is just an exception catching code. The scenario you talk about should not happen easily.

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: Sure, but I don't think docstring here is very important, at least not as important as they are in avocado-vt, since we are not actually generating docs here.

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: As you can see, modify_vm_iface is designed to serve vms that being created by root user. If we try to use it here, it would need modification on the function itself and lots of tests crossing many other test cases including the ones with unprivileged users and privileged users, which is too much work to the necessity of modifying just one case, not to mention more risk there.

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: Actually, I intentionally removed the docstring of parameters, since I found that when people modifying functions, they tend not to modify docstrings, especially when without touching parameters, which makes the docstring really hard to understand, in some cases, even misleading.

libvirt/tests/src/virtual_network/iface_unprivileged.py
Yingshun: Okay, got it.

libvirt/tests/src/virtual_network/iface_unprivileged.py
Yingshun: haha... It's a common issue in tp-libvirt_review_comment_summary.rst.
I think it's really a good habit to add the right docstring.Let's do it. 😃

libvirt/tests/src/virtual_network/iface_unprivileged.py
Yingshun: Okay, see.

libvirt/tests/src/virtual_network/iface_unprivileged.py
chloerh: I hope you could understand, but this is just a helper function. I really don't want to add docstring, especially the one down there, to check ping result.

PR #2706, opened 2020-04-23
virtual_disks_gluster: Cancel all gluster rdma tests

qiankehan:
In QEMU[1], gluster+rdma scheme will be regarded as tcp tranport. That
has been duplicated to gluster tcp cases. The cancel will be revoked after
gluster rdma are really supported in QEMU.

[1]:
https://github.com/qemu/qemu/blob/762fa6d79aa30e1a713444da0399739423f8d00e/block/gluster.c#L375


qiankehan:
@Yingshun  A minor fix. Please help to review it

PR #2722, opened 2020-04-29
Fix detach disk test audit check failure

chunfuwen:
Add wait time until match string is found on log file or timeout exit

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
@@ -31,6 +31,18 @@ def run(test, params, env):
     3.Recover test environment.
     4.Confirm the test result.
     """
+    def check_info_in_audit_log_file(test_cmd, device_source):
+        """
+        Check if information can be found in audit.log.
+
+        :params test_cmd: test command
+        :params device_source: device source
+        """
+        grep_audit = ('grep "%s" /var/log/audit/audit.log'
+                      % test_cmd.split("-")[0])
+        cmd = (grep_audit + ' | ' + 'grep "%s" | tail -n1 | grep "res=success"'
+               % device_source)
+        return process.run(cmd, ignore_status=True, shell=True).exit_status == 0
 
     def check_vm_partition(vm, device, os_type, target_name, old_parts):
         """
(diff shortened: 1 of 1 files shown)


chunfuwen:
PASS 1-type_specific.io-github-autotest-libvirt.virsh.attach_detach_disk.detach_disk.normal_test.audit_check

libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
```diff
@@ -31,6 +31,18 @@ def run(test, params, env):
     3.Recover test environment.
     4.Confirm the test result.
     """
+    def check_info_in_audit_log_file(test_cmd, device_source):
```
dzhengfy, line 34: This function can be abstracted as a function which can search in a specified file for a specified pattern and return True or False.  There is a similar function libvirt.check_logfile() which may meet your needs after you kindly make minor changes. Could you help have a look?

chunfuwen: this is used by utils_misc.wait_for(xx), which may that this method can not be broken(test.fail or exception thrown) even after condition is not met .
Check the function you mentioned, it will exit once condition not met:
****************************************************************************
    if str_in_log == bool(int(cmdRes.exit_status)):
        raise exceptions.TestFail("The string '{}' {} included in {}"
                                  .format(search_str, "is not" if str_in_log else "is",
                                          log_file))

dzhengfy: I think it may be used after your minor changes.
Personally thinking, there is no need to write this local function everywhere in tp-libvirt.

chunfuwen: As you may know, utils_misc.wait_for(func, timeout), the wrapped func can return True or False,rather than exit, otherwise there is no effect in using utils_misc.wait_for()

libvirt/tests/src/virsh_cmd/domain/virsh_attach_detach_disk.py
```diff
@@ -378,7 +390,8 @@ def check_shareable(at_with_shareable, test_twice):
                       % test_cmd.split("-")[0])
         cmd = (grep_audit + ' | ' + 'grep "%s" | tail -n1 | grep "res=success"'
                % device_source)
```
kylazhang, line 392: So the above 2 param setting should be removed?

chunfuwen: updated

chunfuwen:
[stdout] systemd
legacy start libvirtd daemon IGNORED with function name: restart
PASS 1-type_specific.io-github-autotest-libvirt.virsh.attach_detach_disk.detach_disk.normal_test.audit_check

PR #2731, opened 2020-05-04
Mark inject-nmi stress test case invalid on s390x

smitterl:
On s390x 'inject-nmi' causes restart and kernel dump. Therefore,
'inject-nmi' can't be used to put stress on guest because it cannot
be run more than once immediately one after another.

Don't run test case on s390x.

### DIFF
--- a/libvirt/tests/cfg/timer_management.cfg
+++ b/libvirt/tests/cfg/timer_management.cfg
@@ -37,6 +37,7 @@
                 - vm_operations:
                     variants:
                         - inject_nmi:
+                            no s390-virtio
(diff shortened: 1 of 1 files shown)


smitterl:
https://www.ibm.com/support/knowledgecenter/linuxonibm/com.ibm.linux.z.ldva/ldva_r_virsh_inject-nmi.html

sathnaga:
@smitterl better to add this link in commit msg for future reference, https://www.ibm.com/support/knowledgecenter/linuxonibm/com.ibm.linux.z.ldva/ldva_r_virsh_inject-nmi.html...

PR #2745, opened 2020-05-07
Fix error of UnboundLocalError: top_index not assigned

chloerh:
When top_active, the value top_index is not assigned.


chloerh:
@Yingshun

PR #2750, opened 2020-05-09
Add exception detection when trying to unload a linux module

chloerh:
To avoid throwing out RuntimeError

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domfstrim.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domfstrim.py
@@ -220,4 +220,8 @@ def run(test, params, env):
         # Do domain recovery
         vm.shutdown()
         xml_backup.sync()
-        linux_modules.unload_module("scsi_debug")
+
+        def _unload():
+            linux_modules.unload_module("scsi_debug")
+            return True
+        utils_misc.wait_for(_unload, timeout=30, ignore_errors=True)


chloerh:
@Yingshun

libvirt/tests/src/virsh_cmd/domain/virsh_domfstrim.py
```diff
@@ -220,4 +220,7 @@ def _trim_completed():
         # Do domain recovery
         vm.shutdown()
         xml_backup.sync()
-        linux_modules.unload_module("scsi_debug")
+        try:
+            linux_modules.unload_module("scsi_debug")
+        except RuntimeError as e:
```
Yingshun, line 225: do you need to delete scsi disk and wait for a while before unloading module?

chloerh: I found something interesting, maybe I could wait after these are merged.
https://github.com/autotest/tp-libvirt/pull/2795
https://github.com/avocado-framework/avocado-vt/pull/2565

chloerh: @Yingshun would you take a look?

libvirt/tests/src/virsh_cmd/domain/virsh_domfstrim.py
```diff
@@ -220,4 +220,8 @@ def _trim_completed():
         # Do domain recovery
         vm.shutdown()
         xml_backup.sync()
-        linux_modules.unload_module("scsi_debug")
+
+        def _unload():
+            linux_modules.unload_module("scsi_debug")
+            return True
+        utils_misc.wait_for(_unload, timeout=30, ignore_errors=True)
```
kylazhang, line 227: What to do if timout and still raise exception? Then it will not raise error and continued, is that expected?

chloerh: I guess it is. The current issue is that after test passed, this piece of code in 'finally' block 'unload_module'  sometimes fails, but it doesn't seem matter whether it succeeds or not, at least to later tests. Since it won't do anything if it failed, I guess this modification is fine without changing it.

PR #2782, opened 2020-05-15
Fix the issue of fail to compare flags

chloerh:
The value of flag was set wrong.


chloerh:
@smitterl missed an '0'.

PR #2791, opened 2020-05-19
Fix virsh event case failure

chloerh:
On ppc machines, detach interface will cost more time. If doing
attach before detach fully completed, then 'add' will be in front
of 'remove' in the later virsh event output, which will cause failures
of tests. Add a wait_for to wait untill detach completed.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_event.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_event.py
@@ -269,12 +269,19 @@ def run(test, params, env):
                     virsh.detach_disk(dom.name, 'vdb', **virsh_dargs)
                     expected_events_list.append("'device-removed' for %s:"
                                                 " virtio-disk1")
-                    iface_xml_obj = create_iface_xml()
-                    iface_xml_obj.xmltreefile.write()
-                    virsh.detach_device(dom.name, iface_xml_obj.xml, **virsh_dargs)
+                    ifaces = vmxml.devices.by_device_tag('interface')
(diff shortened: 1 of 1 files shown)


chloerh:
PASS 1-type_specific.io-github-autotest-libvirt.virsh.event.positive_test.virsh_event.test_events.device-added-removed_event.no_timeout.loop

libvirt/tests/src/virsh_cmd/domain/virsh_event.py
```diff
@@ -269,12 +269,25 @@ def trigger_events(dom, events_list=[]):
...
                                                 " net0")
-                    time.sleep(2)
+
+                    def _device_detached(count):
+                        tmp_vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name)
+                        tmp_ifaces = tmp_vmxml.devices.by_device_tag('interface')
+                        return True if len(tmp_ifaces) == count - 1 else False
+
+                    utils_misc.wait_for(lambda: _device_detached(len(ifaces)), 60, step=1,
+                                        text='Waiting for device to be detached.')
```
Yingshun, line 290: virsh.detach_device() has a parameter wait_remove_event, you can set it to True.

chloerh: Great! I'll try, if it does not affect event getting.

chloerh: @Yingshun updated!

PR #2799, opened 2020-05-22
Wait for a few seconds to get iface status

chloerh:
To avoid getting outdated iface status.


chloerh:
@Yingshun

PR #2800, opened 2020-05-22
Set vm memory to 2G to avoid permission error

chloerh:
According to BZ1350772, the memory could not be to large for vms
created by unprivileged users.
Also set wait time to 60s to avoid timeout when vm starting too
slow.


chloerh:
@Yingshun

PR #2805, opened 2020-05-25
Fix ceph migrate command typo

chunfuwen:

### DIFF
--- a/libvirt/tests/src/migration/migrate_ceph.py
+++ b/libvirt/tests/src/migration/migrate_ceph.py
@@ -581,7 +581,7 @@ def run(test, params, env):
             logging.info(output)
             if status:
                 destroy_cmd = "virsh destroy %s" % vm_name
-                remote.run_remote_cmd(cmd, params, runner_on_target)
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -592,7 +592,7 @@ def run(test, params, env):
                                             cleanup=True)
         # Ensure VM can be cleaned up on remote host even migrating fail.
         destroy_vm_cmd = "virsh destroy %s" % vm_name
-        remote.run_remote_cmd(cmd, params, runner_on_target)
+        remote.run_remote_cmd(destroy_vm_cmd, params, runner_on_target)
```
Yingshun, line 595: same error in Line 584. BTW, you can use MigrationTest.cleanup_dest_vm to destroy vm in destination.

chunfuwen: updated for quick fix. Regarding code factor, it will be expected with incoming newly added cases

PR #2815, opened 2020-05-27
Allow pseries to configure the amount of numa cells separatedly

chloerh:
The default value of vcpus per numa cell is 2, since ppc hosts usually
have more cpus than x86 hosts, we would set the number to a larger
value to have less numa cells during test.


chloerh:
Test result:
1-type_specific.io-github-autotest-libvirt.virtual_devices.filesystem_device.positive_test.nop.fs_test.xattr_on.flock_on.lock_posix_off.cache_mode_none.one_fs.one_guest

dzhengfy:
When running cases in https://github.com/autotest/tp-libvirt/pull/2718 on PPC, we found the too many numa cells (like 10) will cause the vm fail to start (The bug will be opened). So we have to reduce the cell number and it is better to make this configurable.

PR #2821, opened 2020-05-28
Memory size will automatically align to 256m on pseries

chloerh:
According to Bug 1780506. This negative case is no longer aplicable
on pseries.


chloerh:
@dzhengfy

PR #2822, opened 2020-05-28
Remove unsupported case for pseries

chloerh:
Got error: internal error: qemu does not support SGA


chloerh:
@dzhengfy

PR #2824, opened 2020-05-28
Extend exit timeout for pseries

chloerh:
It usually cost more time than x86


chloerh:
@dzhengfy

PR #2825, opened 2020-05-28
Fail test if vm is not shutdown as expected

chloerh:


libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -91,7 +91,8 @@ def vm_recover_check(option, libvirtd, check_shutdown=False):
             # Shutdown and start the domain,
             # it should be in runing state and can be login.
             vm.shutdown()
-            vm.wait_for_shutdown()
+            if not vm.wait_for_shutdown(240):
```
sathnaga, line 94: let's use a `shutdown_timeout` variable, will be handy to tune for different arch incase needed later...

chloerh: OK

chloerh: @sathnaga updated

PR #2835, opened 2020-06-02
travis: Update python version of allow failures to 2.7

qiankehan:
Since python 2.7 is not maintained after 2020, it is time to update the
version of allow failures.


qiankehan:
> I agree with you. But there're a few details required a little tuning and I have submitted #2934
> How about put this on hold and take a look at #2934 ?

I think your PR is better

PR #2836, opened 2020-06-02
Fix the error test miss issue

yalzhang:
As the config file changes, the original pattern can not match any
test cases in the error test part. Update the pattern to fix it.

### DIFF
--- a/libvirt/tests/cfg/sriov/sriov.cfg
+++ b/libvirt/tests/cfg/sriov/sriov.cfg
@@ -39,7 +39,7 @@
                                     managed = "yes"
                                 - managed_no:
                                     managed = "no"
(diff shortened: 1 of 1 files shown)


yalzhang:
> LGTM, just missed two more broken filters, I believe,
> https://github.com/autotest/tp-libvirt/blob/734a112a2150eb718dbcf0f3551cad0d53331091/libvirt/tests/cfg/sriov/sriov.cfg#L77
> 
> 
> https://github.com/autotest/tp-libvirt/blob/734a112a2150eb718dbcf0f3551cad0d53331091/libvirt/tests/cfg/sriov/sriov.cfg#L42

You are right, Thank you for the catch!

PR #2843, opened 2020-06-05
Fix wrong genid is created when the hex length of vm.genid is less th…

xiaodwan:
…an 18

When vm.genid is too small, the hex length will be less than 18. it will
cause an incorrect expected genid.

The hex string should be left zero padded when it happened.


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.windows.input_mode.vmx.vmx_default.win10.output_mode.libvirt: PASS (1236.03 s)

xiaodwan:
@chunfuwen @chloerh @yantiao0 Please help to review this PR, thanks.

PR #2844, opened 2020-06-08
Add cephfs backend storage migration cases

chunfuwen:
Covering scenarios including mounting cephfs into local directly or by creating syslink

Refactor some code using newly migration utilities

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_ceph.cfg
+++ b/libvirt/tests/cfg/migration/migrate_ceph.cfg
@@ -3,13 +3,8 @@
     take_regular_screendumps = no
     ssh_timeout = 60
     migration_setup = "yes"
-    # please replace your configuration
-    server_ip = "${migrate_dest_host}"
-    server_user = "root"
-    server_pwd = "${migrate_dest_pwd}"
-    client_ip = "${migrate_source_host}"
-    client_user = "root"
-    client_pwd = "${migrate_source_pwd}"
+    # Please set your config for local_ip,local_passwd,remote_ip,remote_passwd,migrate_source_host
+    # migrate_source_pwd,migrate_dest_host and migrate_dest_pwd in base.cfg file under avocado-vt
     start_vm = "no"
     transport = "ssh"
     port = "22"
@@ -17,6 +12,9 @@
     start_vm = "no"
     ssh_port = "${port}"
     migration_timeout = 300
+    mon_host = "EXAMPLE_MON_HOST"
+    virsh_migrate_connect_uri = "qemu:///system"
+    migrate_vm_back = "yes"
     variants:
         - with_postcopy:
             postcopy_options = "--postcopy"
@@ -24,28 +22,18 @@
             postcopy_options = ""
     variants:
         - positive_testing:
-            status_error = "no"
             variants:
                 - migrating_ceph:
-                    mon_host = "EXAMPLE_MON_HOST"
                     replace_vm_disk = "yes"
                     disk_type = "network"
                     disk_target = "vda"
                     disk_target_bus = "virtio"
                     disk_format = "raw"
                     image_size = "10G"
-                    setup_nfs = "no"
-                    enable_virt_use_nfs = "no"
-                    create_target_image = "no"
-                    nfs_mount_dir =
-                    virsh_options = "--live --verbose "
-                    migration_timeout = 300
                     ceph_disk = "yes"
                     disk_source_protocol = "rbd"
                     vol_name = "avocado-vt-pool"
-                    disk_img = "rbd.img"
-                    migrate_vm_back = "yes"
-                    migration_source_uri = "qemu+ssh://${client_ip}/system"
+                    disk_img = "migrating_ceph_rbd.img"                                  
                     variants:
                         - auth:
                             client_name = "EXAMPLE_CLIENT_NAME"
@@ -56,3 +44,19 @@
                             secret_type = "ceph"
                             secret_usage = "EXAMPLE_SECRET_USAGE"
                             secret_uuid = "2782db67-ade6-4413-bbe8-636a1cce088a"
+                - migrating_cephfs_backend:
+                    ceph_cephfs_disk = "yes"
+                    migrate_vm_back = "no"
+                    disk_source_protocol = "fstype"
+                    auth_key = "EXAMPLE_AUTH_KEY"
+                    auth_user = "EXAMPLE_AUTH_USER"
+                    # Please replace this guest image path as needed
+                    guest_image_download_path = "EXAMPLE_GUEST_IMAGE_DOWNLOAD_PATH"
+                    default_guest_image_download_path = "http://download.libvirt.redhat.com/libvirt-CI-resources"
+                    variants:
+                        - mount_dir:              
+                            cephfs_mount_dir = "/mnt/cephfs"
+                        - mount_dir_softlink:
+                            cephfs_create_symlink = "yes"
+                            cephfs_mount_dir = "/mnt/cephfs"
+                            cephfs_symlink = "/var/lib/libvirt/images/cephfs"
--- a/libvirt/tests/src/migration/migrate_ceph.py
+++ b/libvirt/tests/src/migration/migrate_ceph.py
@@ -1,230 +1,31 @@
 import logging
 import os
 import re
+import socket
+
 
 from avocado.core import exceptions
 from avocado.core import data_dir
 from avocado.utils import process
 
 from virttest import ssh_key
-from virttest import data_dir
 from virttest import remote
 from virttest import utils_package
 from virttest import utils_selinux
-from virttest import utils_package
 from virttest import utils_conn
+from virttest import utils_misc
 from virttest import virsh
 from virttest import virt_vm
 from virttest import migration
+from virttest import ceph
+from virttest import libvirt_vm
+from virttest import utils_secret
 
 from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml import secret_xml
 from virttest.libvirt_xml.devices.disk import Disk
 
 from virttest.utils_test import libvirt
-
-from virttest import libvirt_version
-
-MIGRATE_RET = False
-
-
-def check_output(test, output_msg, params):
-    """
-    Check if known messages exist in the given output messages.
-
-    :param test: the test object
-    :param output_msg: the given output messages
-    :param params: the dictionary including necessary parameters
-
-    :raise TestSkipError: raised if the known error is found together
-                          with some conditions satisfied
-    """
-    err_msg = params.get("err_msg", None)
-    status_error = params.get("status_error", "no")
-    if status_error == "yes" and err_msg:
-        if err_msg in output_msg:
-            logging.debug("Expected error '%s' was found", err_msg)
-            return
-        else:
-            test.fail("The expected error '%s' was not found in output '%s'" % (err_msg, output_msg))
-
-    ERR_MSGDICT = {"Bug 1249587": "error: Operation not supported: " +
-                   "pre-creation of storage targets for incremental " +
-                   "storage migration is not supported",
-                   "ERROR 1": "error: internal error: unable to " +
-                   "execute QEMU command 'migrate': " +
-                   "this feature or command is not currently supported",
-                   "ERROR 2": "error: Cannot access storage file",
-                   "ERROR 3": "Unable to read TLS confirmation: " +
-                   "Input/output error",
-                   "ERROR 4": "error: Unsafe migration: Migration " +
-                   "without shared storage is unsafe"}
-
-    # Check for special case firstly
-    migrate_disks = "yes" == params.get("migrate_disks")
-    status_error = "yes" == params.get("status_error")
-    if migrate_disks and status_error:
-        logging.debug("To check for migrate-disks...")
-        disk = params.get("attach_A_disk_source")
-        last_msg = "(as uid:107, gid:107): No such file or directory"
-        if not libvirt_version.version_compare(4, 5, 0):
-            expect_msg = "%s '%s' %s" % (ERR_MSGDICT["ERROR 2"],
-                                         disk,
-                                         last_msg)
-        else:
-            expect_msg = ERR_MSGDICT["ERROR 4"]
-        if output_msg.find(expect_msg) >= 0:
-            logging.debug("The expected error '%s' was found", expect_msg)
-            return
-        else:
-            test.fail("The actual output:\n%s\n"
-                      "The expected error '%s' was not found"
-                      % (output_msg, expect_msg))
-
-    if params.get("target_vm_name"):
-        if output_msg.find(ERR_MSGDICT['ERROR 3']) >= 0:
-            logging.debug("The expected error is found: %s", ERR_MSGDICT['ERROR 3'])
-            return
-        else:
-            test.fail("The actual output:\n%s\n"
-                      "The expected error '%s' was not found"
-                      % (output_msg, ERR_MSGDICT['ERROR 3']))
-
-    for (key, value) in ERR_MSGDICT.items():
-        if output_msg.find(value) >= 0:
-            if key == "ERROR 1" and params.get("support_precreation") is True:
-                logging.debug("The error is not expected: '%s'.", value)
-            elif key == "ERROR 2":
-                break
-            else:
-                logging.debug("The known error was found: %s --- %s",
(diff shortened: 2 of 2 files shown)


chunfuwen:
PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_cephfs_backend.mount_dir.without_postcopy


Sending command: firewall-cmd --reload
Sending command: echo $?
Sending command: grep cpu /proc/cpuinfo | awk '{print $3}' | head -n 1
Sending command: echo $?
PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_cephfs_backend.mount_dir.with_postcopy


Sending command: grep cpu /proc/cpuinfo | awk '{print $3}' | head -n 1
Sending command: echo $?
PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_cephfs_backend.mount_dir_softlink.without_postcopy

PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_cephfs_backend.mount_dir_softlink.without_postcopy

chunfuwen:
#https://github.com/avocado-framework/avocado-vt/pull/2572

chunfuwen:
Kindly note that regarding cephfs, be sure not copy or convert image into cephfs mount folder directly since the image is damaged, instead it need wget one image directly from http server.

libvirt/tests/cfg/migration/migrate_ceph.cfg
```diff
@@ -3,49 +3,37 @@
...
-                    nfs_mount_dir =
-                    virsh_options = "--live --verbose "
-                    migration_timeout = 300
                     ceph_disk = "yes"
                     disk_source_protocol = "rbd"
                     vol_name = "avocado-vt-pool"
-                    disk_img = "rbd.img"
-                    migrate_vm_back = "yes"
-                    migration_source_uri = "qemu+ssh://${client_ip}/system"
+                    disk_img = "rbd.img"
```
Yingshun, line 36: Can we update the rbd.img to another specific name? many ceph related cases are used rbd.img as it's disk image,
it's easy to meet issue when creating an existing image which is used in other case.

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +236,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
+    else:
+        test.fail("Fail to install wget")
+
+
+def write_hostname_into_etc_host():
+    hostname = socket.gethostname()
+    if re.match(r'.*(nay|pek2).redhat.com$', hostname):
+        write_host_cmd = 'echo 10.66.86.24  download.libvirt.redhat.com >> /etc/hosts'
+    else:
+        write_host_cmd = 'echo 10.8.2.20  download.libvirt.redhat.com >> /etc/hosts'
```
Yingshun, line 268: How about moving it to CI? I don't think it's a good idea that hard-code ip address in tp-libvirt.

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +236,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
+    """
+    if utils_package.package_install("wget"):
+        logging.debug('begin download guest images ..............................')
+
+        def _download():
+            download_cmd = ("wget %s -O %s" % (guest_image_url, target_image_name))
+            if process.system(download_cmd, verbose=False, shell=True):
+                test.error("Failed to download image file")
+            return True
+        utils_misc.wait_for(_download, timeout=900)
```
Yingshun, line 256: It will be good if we can check if it's completed correctly.

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +236,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
-    driver = test_dict.get("test_driver", "qemu")
-    transport = test_dict.get("transport")
-    plus = test_dict.get("conn_plus", "+")
-    source_type = test_dict.get("vm_disk_source_type", "file")
-    virsh_options = test_dict.get("virsh_options", "--verbose --live")
-    vol_name = test_dict.get("vol_name")
+    driver = params.get("test_driver", "qemu")
+    transport = params.get("transport")
+    plus = params.get("conn_plus", "+")
+    options = params.get("virsh_migrate_options", "--live --verbose --timeout=1200")
```
Yingshun, line 354: Can you please check with fjin if we need 'p2p'? I had some changes about adding p2p option before.

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -544,23 +462,35 @@ def run(test, params, env):
...
+        migrate_test.ping_vm(vm, params)
+
+        vms = [vm]
+        migrate_test.do_migration(vms, None, dest_uri, 'orderly',
+                                  options, thread_timeout=1200,
+                                  ignore_status=True, virsh_opt=virsh_options,
+                                  func=func_name, extra_opts=extra,
+                                  func_params=params)
+        mig_result = migrate_test.ret
+        migrate_test.check_result(mig_result, params)
```
Yingshun, line 493: Can we add a step to check network connection after migration? We had it in L484 before migration, I think we'd better check it again after migration.

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +236,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
+        download_guest_image(test, replace_guest_image_url, "%s/%s" % (mount_point, first_disk_img))
+        libvirt.set_vm_disk(vm, cephfs_backend_disk)
+    else:
+        ceph.cephfs_umount(ceph_uri, mount_point, verbose=True, session=None)
+        remote_ssh_session = remote.remote_login("ssh", server_ip, "22", server_user,
+                                                 server_pwd, r"[\#\$]\s*$", timeout=480)
+        ceph.cephfs_umount(ceph_uri, mount_point, verbose=True, session=remote_ssh_session)
+        if cephfs_create_symlink:
+            utils_misc.rm_link(symlink_name)
+            utils_misc.rm_link(symlink_name, remote_ssh_session)
```
Yingshun, line 325: Should we unlink the symlink first and then do umount?

libvirt/tests/cfg/migration/migrate_ceph.cfg
chunfuwen: updated

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: it aims to support download image from internal site. And it only add host into etc, no harm to business logic.
If testing environment is outside of internal site, guest_image_download_path in cfg file can be provided this option.

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: the previous download logic can validate the download effect.

if process.system(download_cmd, verbose=False, shell=True):
                test.error("Failed to download image file")

libvirt/tests/cfg/migration/migrate_ceph.cfg
Yingshun: Sorry @chunfuwen , I can't find the update, is there anything missing?

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: updated

libvirt/tests/cfg/migration/migrate_ceph.cfg
chunfuwen: changed to "disk_img = "migrating_ceph_rbd.img" "

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: updated with 'p2p' within it

chunfuwen: updated with p2p option

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: It sound good, but in actual testing environment, finding that it is not stable whether it can get correct IP after migration due to unknown reasons. Considering the main purpose that the case mainly cover ceph storage migration business logic , so I made some trade off and not add check ip address here.  So, what's your suggestion?

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +202,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
     vmxml.sync()
 
 
+def download_guest_image(test, guest_image_url, target_image_name="/mnt/cephfs/RHEL-8.0-x86_64-latest.qcow2"):
```
kylazhang, line 205: I think it is better not give the default target image name as it is not valuable

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +202,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
+                test.error("Failed to download image file")
+            return True
+        utils_misc.wait_for(_download, timeout=900)
+        logging.debug('finish download guest images ..............................')
+        return target_image_name
+    else:
+        test.fail("Fail to install wget")
+
+
+def write_hostname_into_etc_host():
```
kylazhang, line 229: Missing docstring

libvirt/tests/src/migration/migrate_ceph.py
```diff
@@ -406,88 +202,170 @@ def build_disk_xml(vm_name, disk_format, host_ip, disk_src_protocol,
...
+        libvirt.set_vm_disk(vm, cephfs_backend_disk)
+    else:
+        if cephfs_create_symlink:
+            utils_misc.rm_link(symlink_name)
+            utils_misc.rm_link(symlink_name, remote_ssh_session)
+        ceph.cephfs_umount(ceph_uri, mount_point, verbose=True, session=None)
+        remote_ssh_session = remote.remote_login("ssh", server_ip, "22", server_user,
+                                                 server_pwd, r"[\#\$]\s*$", timeout=480)
+        ceph.cephfs_umount(ceph_uri, mount_point, verbose=True, session=remote_ssh_session)
+        remote_ssh_session.close()
```
kylazhang, line 292: Could it be a common library?

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: So could we have any other check to make sure the migration vm can be get in remote host? Else with only the cmd result check we can not make sure it is successful.

libvirt/tests/src/migration/migrate_ceph.py
Yingshun: I'm still confused about this part. 
utils_misc.wait_for() waits 900s until _download() returns True. But as you said, if process.system fails and raise an exception in 900s, will it still try to run again to meet the timout?
Would you mind explain more?

chunfuwen: regarding wait_for(func, timeout, first=0.0, step=1.0, text=None, ignore_errors=False):
if ignore_errors is False(default value),the exception will be raised again in wait_for()implementation.
def wait_for(func, timeout, first=0.0, step=1.0, text=None, ignore_errors=False):
...
        try:
            output = func()
        except:  # pylint: disable=W0702
            if not ignore_errors:
                raise
...

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: migrate_test.check_result(mig_result, params) will handle major part checking. So please share what kind of checking you are proposing if IP is not stable.

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: could you give more hints why it is better not give default value? I understand that even in python, this is common usage that default value is given  to one variable.

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: updated

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: Currently I can not forecast whether it can be used by others. Maybe when it comes, we can do code refactor.

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: From my point of view, the image name is different for each test, and the common value can not correctly reflect the real image name. Or you should change to a more common one. Like upsteam image name jeos?

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: check_result only check the migration virsh cmd return status, it can not show the rhel guest status on target machine. I think we can use ping_vm as it also support checking vm on remote.

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: ok, thanks

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: Ping_VM is the same suggest by Yingshun on Jul 15, and reply is inline.

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: When this function is called, the name is override by passing in parameter.
But anyway, the name is changed to /mnt/cephfs/jeos-common-x86_64-latest.qcow2

chunfuwen: changed to /mnt/cephfs/jeos-common-x86_64-latest.qcow2. But actually the name is overrided by passing in parameter

libvirt/tests/src/migration/migrate_ceph.py
Yingshun: Actually I've updated timeout in ping_vm from 120 to 240s which made it more stable I think. But to be honest I have no  idea what your problem is, could you paste the error message?

libvirt/tests/src/migration/migrate_ceph.py
Yingshun: Yes, I understand the default ignore_errors is False, actually that's my concern. If you want _download to raise an exception, I think you should remove wait_for. Please correct me if I'm wrong.

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: I also agree that do not add it in avocado. And I think the CI should already do it.

libvirt/tests/src/migration/migrate_ceph.py
kylazhang: I agree with Yingshun, no vm ping check after migration is not acceptable.

kylazhang (commented):
comments already inline. thanks

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: updated

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: change  ignore_errors=True

libvirt/tests/src/migration/migrate_ceph.py
chunfuwen: The testing is stuck in waiting for VM IP if execute ping_vm() and timeout value doesn't help it out

Yingshun: Okay, in this case, you'd better check if your network setting in dest and source hosts are correct. They should be in same segment.

chunfuwen:
the matching case is manual only now since the case lack reliablility ,close it first

PR #2846, opened 2020-06-09
Remove pv_eoi cases from pseries

chloerh:
The 'eoi' attribute of the 'apic' feature is not supported for
architecture 'ppc64le' or machine type 'pseries-xxxxx'


chloerh:
Requires @dzhengfy 's review

PR #2848, opened 2020-06-09
v2v debug code

xiaodwan:

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -86,11 +86,25 @@ def run(test, params, env):
             target_path = os.path.join("/home", v2v_user, pool_target)
             cmd = su_cmd + "'mkdir -p %s'" % target_path
             process.system(cmd, verbose=True)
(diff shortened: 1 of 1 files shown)


v2v/tests/src/v2v_options.py
```diff
@@ -74,6 +74,15 @@ def run(test, params, env):
     error_flag = 'strict'
     estimate_file = ''
 
+    def debug_func():
+        try:
+            pwd.getpwnam(v2v_user)
+            cmd = su_cmd + "'virsh pool-list --all'"
```
yantian0, line 80: Here only debug for 'virsh pool-list' command. How about make cmd as the parameter, so other command could invoke this debug function?

xiaodwan: This is only a pr for debugging problems, it will be closed or be changed to fix real problems.

yantian0: ok, got it.

chunfuwen:
Please resolve code conflict

PR #2851, opened 2020-06-10
Set vcpu plug number larger for pseries

chloerh:
pseries hosts usually have more cpus than x86 hosts, 8 is not large
enough to make this negative test case.

### DIFF
--- a/libvirt/tests/cfg/libvirt_vcpu_plug_unplug.cfg
+++ b/libvirt/tests/cfg/libvirt_vcpu_plug_unplug.cfg
@@ -107,6 +107,8 @@
             variants:
                 - greater_plug_number:
                     vcpu_plug_num = "8"
+                    pseries:
+                        vcpu_plug_num = "256"
                     vcpu_unplug = "no"
                     check_after_plug_fail = "yes"
                 - readonly_setvcpu:


chloerh:
Requires @dzhengfy 's review

chloerh:
(1/1) type_specific.io-github-autotest-libvirt.libvirt_vcpu_plug_unplug.negative_test.greater_plug_number: PASS (91.70 s)
JOB TIME   : 95.56 s

libvirt/tests/cfg/libvirt_vcpu_plug_unplug.cfg
```diff
@@ -107,6 +107,8 @@
             variants:
                 - greater_plug_number:
                     vcpu_plug_num = "8"
+                    pseries:
```
sathnaga, line 110: we already cover the positive case for maxvcpu, this is a negative test and it just need to be greater than maxvcpu number,  I guess this change is not relevant.

chloerh: What do you mean? We do not need to test this case any more?

sathnaga: test should fail if we give a plug_num > max_num, which is 8 > 4 without this change, will behave the same way as 256 > 4, there is no significance of this change with respect this particular case.

chloerh: Turns out we have done replacing during cases running, the max '4' has been replaced with 20. I'll find a better way to make it work.

PR #2856, opened 2020-06-11
Remove topology if test does not require

chloerh:
Since test could hit BZ1426220 if topology not removed.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
@@ -54,7 +54,7 @@ def run(test, params, env):
     remote_uri = params.get("remote_uri")
     tmpxml = os.path.join(data_dir.get_tmp_dir(), 'tmp.xml')
     topology_correction = "yes" == params.get("topology_correction", "yes")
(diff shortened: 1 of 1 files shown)


chloerh:
Requires @dzhengfy 's review

libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py
```diff
@@ -163,6 +163,13 @@ def set_expected(vm, options):
             vmxml['cpu'] = vmcpu_xml
             vmxml.sync()
 
+        # Remove topology if test does not requires
```
dzhengfy, line 166: I saw bz1426220 is only PPC bug. So I think we'd better not to make the degradation for x86_64 or other arches. Please limit this change too PPC only.

chloerh: I think it's a coincidence that this bug I found was reported on ppc. Then I found another bug related to this issue, bz1370066.

dzhengfy: OK. As this is also an issue on x86_64, I agree to remove it here.

PR #2859, opened 2020-06-15
v2v: Fix the problem that testing pool cannot be found

xiaodwan:
When create a pool in qemu:///session by normal user, sometimes
'virsh pool-create-as xxx' returns success but the pool cannot
be found. It might be libvirt bug.

This patch is a workaround  to this problem, it tries to create
it until sucess or timeout, so that v2v testing will not be blocked.

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -86,11 +86,25 @@ def run(test, params, env):
             target_path = os.path.join("/home", v2v_user, pool_target)
             cmd = su_cmd + "'mkdir -p %s'" % target_path
             process.system(cmd, verbose=True)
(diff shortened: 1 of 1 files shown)


xiaodwan:
https://libvirt-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/v2v/view/RHEL-8.3/job/v2v-RHEL-8.3-runtest-x86_64-function-v2v_options-rhel/8/testReport/rhel/v2v_options/positive_test_option_oc_unprivileged_input_mode_disk_image_qcow2_format_sparse_output_mode_libvirt/

xiaodwan:
@chunfuwen @chloerh @yantian0 Please help to review this PR, thanks.

PR #2855, opened 2020-06-11
acpi is not supported by pseries

chloerh:
Signed-off-by: haizhao <haizhao@redhat.com>

PR #2860, opened 2020-06-15
Revert "acpi is not supported by pseries"

chunfuwen:
Reverts autotest/tp-libvirt#2855


chunfuwen:
reverted previous one commit:#2855

PR #2861, opened 2020-06-15
Add cfg file definition conventions

chunfuwen:
In order to provide one consistent, human-readable, maintainable test case names, it is necessary to
apply some conventions rules to Cartesian Configuration format.

### DIFF
--- a/cfg_file_definition_convention.rst
+++ b/cfg_file_definition_convention.rst
@@ -1,0 +1,83 @@
+Cartesian configuration files(.cfg) will define test case names.
+"How do cfg files work? Please refer to https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html
+In order to provide consistent, human-readable, maintainable test case names, it is necessary to 
+apply some conventions rules to Cartesian Configuration format.
+
+1. Variant name is joined by underscore _ ,for example:
+   variants:
+       - hot_plug:
+           key1 = xx
+       - cold_plug:
+           key2 = xx
+
+2. Each variant defines a single dimension array with similar provision methods or operations, for example:
+   variants:
+       - block_test:
+           backend_storage_type = "iscsi"
+       - gluster_test:
+           backend_storage_type = "gluster"
+       - ceph_test:
+           backend_storage_type = "ceph"
+       - nfs_test:
+           backend_storage_type = "nfs"
+       - dir_pool_test:
+           backend_storage_type = "dir"
+   
+3. Variants may be nested within other variants, the most left (outside) is used to represents higher level
+   operations or options, and the most left has shorter name, for example:
+   variants:
+       - enable_auth:
+           enable_auth = "yes"
+           variants:
+               - use_auth_uuid:
+                   use_auth_uuid = "yes"
+               - use_auth_usage:
+                   use_auth_usage = "yes"
+
+
+4. Variant relationship will define case name order. Variant precedence is evaluated in top-down. In other words, 
+   the last parsed variants has precedence over earlier definitions,therefore the lowest variant name will be placed at the beginning
+   of whole test case name. The higher level one variant represents, the lower its position should be placed in. for example:
+   variants:
+       - device_disk:
+       - device_volume
+   variants:
+       - persistent
+       - config
+       - live
+       - current
+   variants:
+       - cold_plug:
+       - hot_plug:
+   variants:
+       -positive_test:
+           status_error = "no"
+       -negative_test:
+           status_error = "yes"
+
+5. Generally speaking,a good tip to define how many or what variants there should be is to check virsh command usage, where test cases may touch.
+   for example, by execute virsh attach-device --help command, and the output shows at least it should cover four types options: 
+     --persistent --config  --live --current.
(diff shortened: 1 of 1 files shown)


chunfuwen:
@santwana 
@smitterl 
@kylazhang 
@dzhengfy 
Welcome inputs and thoughts.

cfg_file_definition_convention.rst
```diff
@@ -0,0 +1,61 @@
...
+       - current
+   variants:
+       - cold_plug:
+       - hot_plug:
+   variants:
+       -positive_test:
+           status_error = "no"
+       -negative_test:
+           status_error = "yes"
+
```
chloerh, line 57: Currently we have fewer cases that need to be constructed from matrix, but more cases with different structures and sets of parameters. And they usually look like this in our test cfg files:

```
- positive_test:
    variants:
        - test_1:
        - test_2:
- negative_test:
    variants:
        - n_test_1:
        - n_test_2:
```
The advantage of this way is that we only need to place test cases to the correct part (positive or negative) without having to concern about the whole structure of the cfg file which influences all its test cases.  And no need to put a lot of cases' patterns under '-positive_test' or '-negative_test' like those matrix cfg files: virsh_attach_detach_disk_matrix.cfg  or virsh_attach_detach_interface_matrix.cfg  with one single line that has 2153 characters just because it's filled with only xxx, xxx, xxx ...

kylazhang: Agree

kylazhang:
@chunfuwen I think you can add more for some tips you ever mentioned for condition in cfg  and parmas for specific cases in cfg

cfg_file_definition_convention.rst
smitterl: +1

cfg_file_definition_convention.rst
```diff
@@ -0,0 +1,61 @@
+Cartesian configuration files(Cfg) will define test case names.
```
smitterl, line 1: "files(Cfg)" -> "files (.cfg)"

cfg_file_definition_convention.rst
```diff
@@ -0,0 +1,61 @@
...
+How cfg works? please refer to https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html
```
smitterl, line 2: "How cfg works? please" -> "How do cfg files work? Please"

cfg_file_definition_convention.rst
```diff
@@ -0,0 +1,61 @@
...
+In order to provide one consistent, human-readable, maintainable test case names, it is necessary to
```
smitterl, line 3: "one [..] names" -> "[..] names"

cfg_file_definition_convention.rst
```diff
@@ -0,0 +1,61 @@
...
+           variants:
+               - use_auth_uuid:
+                   use_auth_uuid = "yes"
+               - use_auth_usage:
+                   use_auth_usage = "yes"
+
+
+4. Variant relationship will define case name order. Variant precedence is evaluated in top-down. In other words, 
+   the last parsed variants has precedence over earlier definitions,therefore the lowest variant name will be placed at the beginning
+   of whole test case name. The higher level one variant represents, the more lower its position should be placed in. for example:
```
smitterl, line 40: "more lower" -> "lower"

smitterl (changes requested):
Thank you @chunfuwen Could you check some minor grammatical issues I think to have spotted?

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

cfg_file_definition_convention.rst
chunfuwen: updated

chloerh (approved):

chunfuwen:
@smitterl 
@kylazhang 
comments updated accordingly ,please take a look.

kylazhang (approved):

PR #2863, opened 2020-06-15
Set correct hugepage size for power8 and power9 hosts

chloerh:
Pagesizes on power8 and power9 hosts are different

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -10,6 +10,7 @@ import random
 
 from six.moves import xrange
 
+from avocado.utils import cpu as cpu_util
 from avocado.utils import process
 
 from virttest import virsh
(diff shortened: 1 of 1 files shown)


chloerh:
PASS 1-type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.hugepages.with_numatune.hot_plug

libvirt/tests/src/libvirt_mem.py
```diff
@@ -439,6 +440,12 @@ def modify_domain_xml():
     config = utils_config.LibvirtQemuConfig()
     setup_hugepages_flag = params.get("setup_hugepages")
     if (setup_hugepages_flag == "yes"):
+        cpu_arch = cpu_util.get_cpu_arch()
+        if cpu_arch == 'power8':
+            pg_size = '16384'
+        elif cpu_arch == 'power9':
+            pg_size = '2048'
+        [x.update({'size': pg_size}) for x in huge_pages]
```
dzhengfy, line 448: What is the reason to create a list here? I think we only need to do a update size.

chloerh: That't not my intent either. The var 'huge_pages' was set to be a list by the original code, therefore...

PR #2866, opened 2020-06-16
Drop code of dmesg checking for hpt tests

chloerh:
Content of dmesg output is not constant, dmesg checking could be
dropped.


chloerh:
Requires @dzhengfy 's review

Yingshun:
@chloerh I'm wondering if we can use regular expressions to search expected dmesg's output.

chloerh:
> @chloerh I'm wondering if we can use regular expressions to search expected dmesg's output.

The thing is, we were expecting 'Attempting to resize HPT to shift %d|HPT resize to shift %d complete' in dmesg output, those are 2 different kinds of content. But now, we cannot find some keywords inside the dmesg output, therefore regex cannot work here.

Yingshun:
Ok, see. Then LGTM.

chunfuwen:
is it possible to have test result posted here since there are code changes?

PR #2867, opened 2020-06-16
Update to check output of vol-list

Yingshun:
To avoid IndexError, update to check 'virsh vol-list' command's
output first.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/volume/virsh_vol_download_upload.py
+++ b/libvirt/tests/src/virsh_cmd/volume/virsh_vol_download_upload.py
@@ -227,6 +227,9 @@ def run(test, params, env):
         vol_list = virsh.vol_list(pool_name, debug=True).stdout.strip()
         # iscsi volume name is different from others
         if pool_type == "iscsi":
+            # Due to BZ 1843791, the volume cannot be obtained sometimes.
+            if len(vol_list.splitlines()) < 3:
+                test.fail("Failed to get iscsi type volume.")
             vol_name = vol_list.split('\n')[2].split()[0]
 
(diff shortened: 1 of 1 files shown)


Yingshun:
It reports "IndexError" with the previous code rarely. Actually, I cannot reproduce it in my environment but it fails in CI sometimes.

libvirt/tests/src/virsh_cmd/volume/virsh_vol_download_upload.py
```diff
@@ -227,6 +227,9 @@ def run(test, params, env):
         vol_list = virsh.vol_list(pool_name, debug=True).stdout.strip()
         # iscsi volume name is different from others
         if pool_type == "iscsi":
+            # Due to BZ 1843791, the volume cannot be obtained sometimes.
+            if len(vol_list.splitlines()) < 3:
```
chunfuwen, line 231: Is it possible to allow more time to try let volume popup by utils_misc. waitfor(xx)?

Yingshun: No, wait_for does not work this issue.

chunfuwen: I assume https://bugzilla.redhat.com/show_bug.cgi?id=1843791  mentions that  the same two records will be returned, rather than not record. Is this real root clause for  no volume return?

Yingshun: Yes, I have confirmed with feature owner. I will blacklist these cases in CI once this PR's merged.

PR #2870, opened 2020-06-17
Check vm's existance before removing it

chloerh:
Remove vm only if vm exists.


chloerh:
@yanqzhan

chunfuwen:
Is it possible to have test result posted here before merge?

PR #2876, opened 2020-06-19
Set different hugepage number other than 2000 on ppc hosts

chloerh:
PPC hosts have different hugepage size from x86, therefore the number
of hugepages on ppc should not be fixed to 2000 as x86. Set to a
smaller number to avoid memory size exceeds.

### DIFF
--- a/libvirt/tests/cfg/libvirt_mem.cfg
+++ b/libvirt/tests/cfg/libvirt_mem.cfg
@@ -87,6 +87,9 @@
                     test_qemu_cmd = "no"
                     test_mem_binding = "no"
                     setup_hugepages = "yes"
+                    huge_page_num = 2000
+                    pseries:
+                        huge_page_num = 200
                     max_mem_rt = 25600000
                     max_mem = "2609152"
                     current_mem = "2609152"
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -428,6 +428,7 @@ def run(test, params, env):
     tg_node = params.get("tg_node", 0)
     pg_size = params.get("page_size")
     pg_unit = params.get("page_unit", "KiB")
+    huge_page_num = int(params.get('huge_page_num', 2000))
     node_mask = params.get("node_mask", "0")
     mem_addr = ast.literal_eval(params.get("memory_addr", "{}"))
     huge_pages = [ast.literal_eval(x)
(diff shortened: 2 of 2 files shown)


chloerh:
PASS 1-type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.hugepages.with_numatune.hot_plug

chloerh:
Requires @dzhengfy 's approval to be merged

libvirt/tests/cfg/libvirt_mem.cfg
```diff
@@ -87,6 +87,9 @@
                     test_qemu_cmd = "no"
                     test_mem_binding = "no"
                     setup_hugepages = "yes"
+                    huge_page_num = 2000
+                    pseries:
+                        huge_page_num = 200
```
dzhengfy, line 92: Default hugepage size is 2M on x86_64 and Power9, while 16M on Power8. So does the problem happen on Power8?
The existing code to hard code the page number is not a good idea. That should be calculate from the vm memory, I think.

chloerh: Of course. But I don't know the strategy of the original case, and it'd be not cautious to modify test code that has not been affected. I agreed that memory should be calculated, but this fix is not about it.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -446,7 +447,7 @@ def modify_domain_xml():
         elif cpu_arch == 'power9':
             pg_size = '2048'
         [x.update({'size': pg_size}) for x in huge_pages]
-        setup_hugepages(int(pg_size))
+        setup_hugepages(int(pg_size), shp_num=huge_page_num)
```
sathnaga, line 450: my suggestion is to better to use https://github.com/avocado-framework/avocado-vt/blob/master/virttest/test_setup.py#L363, this has functions that are needed to set/ release hugepage required for test vm,

chloerh: I'm just trying to fix one problem, don't want to bring in other potential risks on other arches, hope you understand.

libvirt/tests/cfg/libvirt_mem.cfg
chloerh: FYI, the number '2000' is not set by me. The default value of param 'shp_num' of function 'setup_hugepages' is 2000.

PR #2882, opened 2020-06-22
Check mac address when attach interface with --live --config

yalzhang:
There is a bug that when attach-device with --live and --config but
no mac address for the interface, it will generate 2 different mac
address separately to live and inactive xml, which will cause the mac
changes after a vm destroy-start. Add the test scenario into the hotplug
tests.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_hotplug.py
+++ b/libvirt/tests/src/virtual_network/iface_hotplug.py
@@ -28,7 +28,7 @@ def run(test, params, env):
     if vm.is_alive():
         vm.wait_for_login()
 
-    def create_iface_xml(mac):
+    def create_iface_xml(mac=None):
         """
         Create interface xml file
         """
@@ -37,11 +37,25 @@ def run(test, params, env):
         iface.model = iface_model if iface_model else "virtio"
         if iface_target:
             iface.target = {'dev': iface_target}
-        iface.mac_address = mac
+        if mac:
+            iface.mac_address = mac
         if iface_rom:
             iface.rom = eval(iface_rom)
         logging.debug("Create new interface xml: %s", iface)
         return iface
+
+    def get_all_mac_in_vm():
+        """
+        get the mac address list of all the interfaces from a running vm os
+
+        return: a list of the mac addresses
+        """
+        mac_list = []
+        interface_list = vm.get_interfaces()
+        for iface_ in interface_list:
+            mac_ = vm.get_interface_mac(iface_)
+            mac_list.append(mac_)
+        return mac_list
 
     # Interface specific attributes.
     iface_num = params.get("iface_num", '1')
@@ -73,6 +87,7 @@ def run(test, params, env):
     set_mac = "yes" == params.get("set_mac", "no")
     status_error = "yes" == params.get("status_error", "no")
     pci_addr = params.get("pci")
+    check_mac = "yes" == params.get("check_mac", "no")
 
     # stree_test require detach operation
     stress_test_detach_device = False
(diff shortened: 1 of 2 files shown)


yalzhang:
new add 1 case, run locally pass, will run all the cases for this file later and add the test results.
virtual_network.iface_hotplug.at_device.iface_attach.model_virtio.options_test.check_mac.active

$ avocado run --vt-type libvirt virtual_network.iface_hotplug.at_device.iface_attach.model_virtio.options_test.check_mac.active
/root/code/avocado/avocado/plugins/run.py:299: FutureWarning: The following arguments will be changed to boolean soon: sysinfo, output-check, failfast and keep-tmp. 
  FutureWarning)
JOB ID     : a364f7d83f5468ae3d456e1f37dceae3b00b4e8c
JOB LOG    : /root/avocado/job-results/job-2020-06-22T01.03-a364f7d/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_hotplug.at_device.iface_attach.model_virtio.options_test.check_mac.active: PASS (96.20 s)
JOB TIME   : 97.85 s

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -37,12 +37,27 @@ def create_iface_xml(mac):
...
             iface.rom = eval(iface_rom)
         logging.debug("Create new interface xml: %s", iface)
         return iface
 
+    def get_all_mac_in_vm():
+        """
+        get the mac address list of all the interfaces from a running vm os
+
+        return: a list of the mac addresses
+
```
Yingshun, line 52: It'll be great if we remove this blank line.

libvirt/tests/src/virtual_network/iface_hotplug.py
```diff
@@ -170,6 +191,11 @@ def create_iface_xml(mac):
                     libvirt.check_exit_status(ret)
                 else:
                     if attach_device:
+                        if check_mac:
+                            mac_aft = get_all_mac_in_vm()
+                            add_mac = [i for i in mac_aft if i not in mac_bef]
+                            mac = add_mac[0]
```
Yingshun, line 197: I think it may raise IndexError if add_mac returns an empty list. Can you please update?
BTW, there is nothing wrong with your code in L196, but if you want to get new added items you can have a try with set(mac_aft).difference(set(mac_bef)) .

libvirt/tests/src/virtual_network/iface_hotplug.py
yalzhang: Have updated, Thank you!

libvirt/tests/src/virtual_network/iface_hotplug.py
yalzhang: Have add the "try...except" part, Please help to check if it is misused. Thank you!
                        if check_mac:
                            mac_aft = get_all_mac_in_vm()
                            add_mac = set(mac_aft).difference(set(mac_bef))
                            try:
                                mac = add_mac[0]
                                logging.debug("The mac address of the attached interface is %s" % mac)
                            except IndexError:
                                test.fail("Can not find the new added interface in the guest os!")

Yingshun: Aha, you can keep your original code, I just want to share a method.
Set type is an unordered collection of unique values, it does not support indexing,so you can't get elements by index. Here you can change add_mac to a list -- add_mac = list(set(mac_aft).difference(set(mac_bef))).

yalzhang: Thank you for the sharing! I think it's more elegant to use set, but forget to consider the difference between set and list. Thank you :-)

PR #2886, opened 2020-06-23
Replace operations using virsh instance with using params

chloerh:
Current test result shows that there's a little messy when running
through different unprivileged users using virsh instance. But I'm
not 100% sure changing to virsh params could fix this problem, since
the problem cannot reproduced by me. But even it will not fix this
specific issue, it's a better way to conduct these virsh operations.


chloerh:
@Yingshun

chloerh:
Test result from jenkins


iface_unprivileged.precreated.host_macvtap | 48 秒 | Fixed
-- | -- | --
iface_unprivileged.precreated.host_tap | 1 分 16 秒 | Passed

PR #2897, opened 2020-06-29
Fix randomly getting ip address failure problem

xiaodwan:
Sometimes the --mac argument cannot be set correctly. This is because
the bridge or network parameter is deleted from params dictionay
object before passing to v2v_cmd function. Then when composing
--mac option, if the deleted net type is chosed, the value will be
None, then following wrong parameter will be set.

wrong setting:
--mac 52:54:00:b9:9b:db:network:None

### DIFF
--- a/v2v/tests/src/specific_kvm.py
+++ b/v2v/tests/src/specific_kvm.py
@@ -877,12 +877,10 @@ def run(test, params, env):
             logging.info('Attach network')
             virsh.attach_interface(
                 vm_name, 'network default --current', debug=True)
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.network.only_br.kvm.output_mode.rhev.rhv_upload.it_default: PASS (486.68 s)

xiaodwan:
@chunfuwen @chloerh @yantian0 Please help to review this PR, thanks.

PR #2899, opened 2020-06-30
Fix problem of NPIV multipathd unable to start

jgao12345:
Signed-off-by: jgao <jgao@localhost.localdomain>

### DIFF
--- a/libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
+++ b/libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
@@ -110,6 +110,8 @@ def run(test, params, env):
                 {"nodedev_parent": first_online_hba,
                  "scsi_wwnn": scsi_wwnn,
                  "scsi_wwpn": scsi_wwpn})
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/npiv/npiv_hostdev_passthrough.py
```diff
@@ -110,6 +110,8 @@ def get_attached_disk():
                 {"nodedev_parent": first_online_hba,
                  "scsi_wwnn": scsi_wwnn,
                  "scsi_wwpn": scsi_wwpn})
+        # enable multipath service
+        process.run("mpathconf --enable", shell=True)
```
chunfuwen, line 114: maybe add ignore_status=True if mpathconf is already enabled

jgao12345: I don’t think it’s necessary, even if the service is started, the execution of this command can reset the environment to the cleanest state, so the execution is better

PR #2915, opened 2020-07-08
Update pv_eoi cases for pseries

chloerh:
There has been an error message update on this feature, which is
the test focuses on.
Unlike x86, eoi is not supported by pseries, the result of virsh
define command should reflect this info.

### DIFF
--- a/libvirt/tests/src/libvirt_qemu_cmdline.py
+++ b/libvirt/tests/src/libvirt_qemu_cmdline.py
@@ -4,6 +4,7 @@ BTW it not limited to hypervisors CPU/machine features.
 """
 import re
 import logging
+import platform
 
 from virttest import virsh
 from virttest.libvirt_xml import vm_xml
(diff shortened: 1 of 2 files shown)


chloerh:
PASS 1-type_specific.io-github-autotest-libvirt.libvirt_qemu_cmdline.hypervisor_features.pv_eoi.disable

chloerh:
@Yingshun @dzhengfy

libvirt/tests/src/libvirt_qemu_cmdline.py
```diff
@@ -48,16 +49,22 @@ def config_feature_pv_eoi(test, vmxml, **kwargs):
...
+        vmxml_feature.remove_feature('apic')
+    vmxml_feature.add_feature('apic', 'eoi', eoi_enable)
+    vmxml.features = vmxml_feature
+    logging.debug("Update VM XML:\n%s", vmxml)
+    expect_fail = False if 'expect_define_vm_fail' not in kwargs \
+        else kwargs['expect_define_vm_fail']
+    result = virsh.define(vmxml.xml, debug=True)
+    libvirt.check_exit_status(result, expect_fail)
+    if expect_fail:
+        libvirt.check_result(result, kwargs.get('expected_msg'))
```
dzhengfy, line 66: @chloerh , could we directly use libvirt.check_result() instead of L64?

chloerh: I don't think that's an regular way. libvirt.check_result() was designed to check output or a negative result, or just a result, as a supplement of check_exit_status(). I have reviewed the code and the logic of check_result() is not as simple and clear as it is in check_exit_status. Maybe it could do the job, but in my opinion, checking them separately seems better and clearer, and only check result when it needs to be checked.

dzhengfy: I still think check_exit_status() is something redundant because check_result() can cover its logic and check points.
But anyway, it is up to you, and not a must to change.

chloerh: I understand. Can I make a suggestion? What if we modify the docstring, and param settings of check_result() to make the logic clearer and more direct. Since there's not a parameter to determine the legality of the output's exit_status. If there is one, then I'd be happy to use it.

dzhengfy: Could you describe what the 'the legality of the output's exit_status'? Does the parameter 'any_error' in check_result() can meet your needs? I think it is similar with 'expect_fail' in your changes here.

chloerh: This is the docstring of check_result()
    """
    Check the result of a command and check command error message against
    expectation.

    :param result: Command result instance.
    :param expected_fails: a string or list of regex of expected stderr patterns.
                           The check will pass if any of these patterns matches.
    :param skip_if: a string or list of regex of expected stderr patterns. The
                    check will raise a TestSkipError if any of these patterns matches.
    :param any_error: Whether expect on any error message. Setting to True will
                      will override expected_fails
    :param expected_match: a string or list of regex of expected stdout patterns.
                           The check will pass if any of these patterns matches.
    """


This is check_exit_status()
    """
    Check the exit status of virsh commands.

    :param result: Virsh command result object
    :param expect_error: Boolean value, expect command success or fail
    """

As it shows, "expect_error" is a clear param, it shows the expectation of "command success or fail". But there is none in check_result(), at least not a direct one.  In that case, I would rather be safe to use check_exit_status() since it's clearly stated

chloerh: Sorry, I missed the 'any_error' part.

chloerh: What I'm saying is the docstring is not very clear about the exit_status check, that is all.

PR #2929, opened 2020-07-17
Update image type to raw for lun tests

chloerh:
Based on change on libvirt:
qemu: Forbid non-raw images for disk type='lun' with vitio-blk frontend
https://gitlab.com/libvirt/libvirt/-/commit/b40ec75296769884ba873d2a72cdda01c2c03c24


chloerh:
@qiankehan

PR #2931, opened 2020-07-20
Update libvirt_version imported module from provider to virttest

chunfuwen:
Those changes are part of efforts to maintain only one place libvirt_version across all python files


chunfuwen:
(p3-ve) [root@dell-r440-04 virttest]# libvirtd -V
libvirtd (libvirt) 6.0.0
(p3-ve) [root@dell-r440-04 virttest]# python
Python 3.6.8 (default, Dec  5 2019, 15:45:45) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from virttest import libvirt_version
>>> value = libvirt_version.version_compare(5,6,0)
>>> value
True
>>>

PR #2934, opened 2020-07-20
Update travis setting

chloerh:
- Update travis requirements, remove outdated ones, and make a few
version upgrades
- Set checking version from python 3.4 to 3.6
- Set ignore python version from 3.4 to None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,11 @@
 language: python
 python:
     - "2.7"
-    - "3.4"
+    - "3.6"
 
 branches:
     only:
         - master
-
-matrix:
-    allow_failures:
-        - python: "3.4"
 
 sudo: false
 
--- a/requirements-travis.txt
+++ b/requirements-travis.txt
@@ -1,9 +1,7 @@
-coverage==3.6
+coverage==5.1
 nose==1.3.0
 nosexcover==1.0.8
 tox==1.5.0
 virtualenv==1.9.1
 simplejson==3.8.1
-MySQL-python==1.2.5
-inspektor==0.4.5
-autotest==0.16.2
+inspektor==0.5.2


chloerh:
Inspired by https://github.com/autotest/tp-libvirt/pull/2835

.travis.yml
```diff
@@ -1,15 +1,15 @@
...
+    - "3.6"
 
 branches:
     only:
         - master
 
 matrix:
     allow_failures:
-        - python: "3.4"
+        - python: "2.7"
```
chunfuwen, line 12: I assume I still need maintain python 2.7 version, so better not all failures in python2.7

requirements-travis.txt
```diff
@@ -1,9 +1,7 @@
-coverage==3.6
+coverage==5.1
```
chunfuwen, line 1: is it possible to define version as per python version?such as 

coverage==3.6; python_version <= '2.7'
coverage==5.1; python_version >= '3.6'

.travis.yml
chloerh: I agree. And current result shows that this also passes on python 2.7. I'll wait for a few other opinions and then make the change.

requirements-travis.txt
chloerh: Of course, I'll update.

chloerh: But, it seems not necessary, since the test already passes on 2.7, right?

chunfuwen:
I remember there are some python files ,which will fail python 3.4 checking, that is why we allow 3.4 failures in travics ci config. I assume current travis CI results only shows current PR will pass python 2.7 and 3.6 checking.

chloerh:
> I remember there are some python files ,which will fail python 3.4 checking, that is why we allow 3.4 failures in travics ci config. I assume current travis CI results only shows current PR will pass python 2.7 and 3.6 checking.

Yes, and I have updated those files: https://github.com/autotest/tp-libvirt/pull/2933

chloerh:
@Yingshun @cliping  Would you mind take a look too?

PR #2938, opened 2020-07-21
Change the libvirt_version usage from tp-libvirt to avocado-vt

kylazhang:
Signed-off-by: kylazhang <weizhan@redhat.com>


chunfuwen:
Is any validated result available?

kylazhang:
I think no need to test each one as it just string replacement

PR #2942, opened 2020-07-24
migration: Add a case for a paused VM postcopy migration

Yingshun:
This PR adds a case - Do live postcopy migration for a paused VM.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_options_shared.cfg
+++ b/libvirt/tests/cfg/migration/migrate_options_shared.cfg
@@ -408,6 +408,14 @@
                                 actions_during_migration = "suspendvm"
                             with_postcopy:
                                 actions_during_migration = "suspendvm,setmigratepostcopy"
+                        - pause_vm_before_migration:
+                            only with_postcopy
+                            pause_vm_before_migration = "yes"
+                            asynch_migrate = "yes"
+                            stress_in_vm = "yes"
+                            stress_args = "--cpu 4 --io 4 --vm 2 --vm-bytes 256M --timeout 70"
+                            actions_during_migration = "setmigratepostcopy"
+                            actions_after_migration = "checkdomstate"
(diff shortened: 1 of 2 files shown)


Yingshun:
**Test result:** The failure was caused by an existing BZ 1464792.

```
avocado run --vt-type libvirt virsh.migrate_options_shared.positive_test.p2p_migration.pause_vm_before_migration.with_postcopy
JOB ID     : 22b7fdd8972a612d936000c2041f4f518a0469a6
JOB LOG    : /root/avocado/job-results/job-2020-07-24T01.36-22b7fdd/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.p2p_migration.pause_vm_before_migration.with_postcopy: FAIL: error: argument unsupported: post-copy migration is not supported with non-live or paused migration (173.62 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
```
Note: There was a similar case for precopy migration in migrate_vm, but it's not easy to do async migration in migrate_vm.py, so I added it here.

dzhengfy:
Before I merge this PR, I have a question.
bz1464792 is in New state for about 3 years, and no discussion for it. So I am very considering if this test is valuable, especially it will fail due to this bug. @fangge1212 , could you evaluate this case's priority?

fangge1212:
> Before I merge this PR, I have a question.
> bz1464792 is in New state for about 3 years, and no discussion for it. So I am very considering if this test is valuable, especially it will fail due to this bug. @fangge1212 , could you evaluate this case's priority? cc @Yingshun

I think it's no harm to add a case now, we can add it to blacklist in ci job before bug fixed. 
It's a valid scenario to postcopy migration a paused vm.

PR #2568, opened 2020-02-28
Adopt Avocado cpu library changes

sathnaga:
Cpu library is getting refactored through this
PR https://github.com/avocado-framework/avocado/pull/3350,
This PR modified respective test code to reflect the same.

Signed-off-by: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>

PR #2956, opened 2020-07-30
Fix wrong usage of getting cpu arch/family

chloerh:
The function get_cpu_arch is now deprecated on latest version of
avocado. We need to use get_family if we want to get the info like
'power8'.


chloerh:
Close this one since it should be included in https://github.com/autotest/tp-libvirt/pull/2568

PR #3009, opened 2020-08-19
libvirt_vm:set different page number to power8 and power9

chloerh:
- Power8 and power9 have different pagesize, page numbers need
to be different when power8 machines have small total memory
power9 machines don't have this kind of problems.
- Print numa info before starting vm

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -444,8 +444,10 @@ def run(test, params, env):
         cpu_arch = cpu_util.get_cpu_arch()
         if cpu_arch == 'power8':
             pg_size = '16384'
+            huge_page_num = 200
         elif cpu_arch == 'power9':
             pg_size = '2048'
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/libvirt_mem.py
```diff
@@ -474,6 +474,7 @@ def modify_domain_xml():
         if vm.is_alive():
             vm.destroy(gracefully=False)
         modify_domain_xml()
+        process.run('numactl --hard', shell=True, verbose=True)
```
dzhengfy, line 477: It is better to refer to src/numa/numa_memory.py to use utils_misc.NumaInfo() to get the useful information. It has package installation step.

chloerh:
Test result on power9

Test name | Duration | Status
-- | -- | --
positive_test.hugepages.with_numatune.hot_plug | 4 min 28 sec | Fixed

dzhengfy (approved):
LGTM

PR #3051, opened 2020-09-08
Add set threshold for disk mirroring node BZ1807741 cases

chunfuwen:
Test under mirroring of blockcommit and blockcopy, threshold setting
can work normally

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
@@ -111,6 +111,67 @@ def run(test, params, env):
         ret = virsh.vol_create(p_name, v_xml.xml, **virsh_dargs)
         libvirt.check_exit_status(ret)
 
+    def trigger_block_commit(vm_name, target, blockcommit_options, **virsh_dargs):
+        """
+        Trigger blockcommit.
+
+        :param vm_name: VM name
+        :param target: Disk dev in VM.
+        :param blockcommit_options: blockcommit option
+        :param virsh_dargs: additional parameters
+        """
+        result = virsh.blockcommit(vm_name, target,
+                                   blockcommit_options, ignore_status=False, **virsh_dargs)
+
+    def trigger_block_copy(vm_name, target, dest_path, blockcopy_options, **virsh_dargs):
+        """
+        Trigger blockcopy
+
+        :param vm_name: string, VM name
+        :param target: string, target disk
+        :param dest_path: string, the path of copied disk
+        :param blockcopy_options: string, some options applied
+        :param virsh_dargs: additional options
+        """
+        result = virsh.blockcopy(vm_name, target, dest_path, blockcopy_options, **virsh_dargs)
+        libvirt.check_exit_status(result)
+
+    def trigger_mirror_threshold_event(vm_domain, target):
+        """
+        Trigger mirror mode block threshold event.
+
+        :param vm_domain: VM name
+        :param target: Disk target in VM.
+        """
+        try:
+            session = vm_domain.wait_for_login()
+            # Sleep 10 seconds to let wait for events thread start first in main thread
+            time.sleep(10)
+            cmd = ("dd if=/dev/urandom of=file bs=1G count=3")
+            status, output = session.cmd_status_output(cmd)
+            if status:
+                test.error("Failed to fill data in VM target: %s with %s" % (target, output))
+        except (remote.LoginError, virt_vm.VMError, aexpect.ShellError) as e:
+            logging.error(str(e))
+            raise
+        except Exception as ex:
+            raise
+
+    def get_mirror_source_index(vm_name, dev_index=0):
+        """
+        Get mirror source index
+
+        :param vm_name: VM name
+        :param dev_index: Disk device index.
+        :return mirror source index in integer
+        """
+        disk_list = vm_xml.VMXML.get_disk_source(vm_name)
+        disk_mirror = disk_list[dev_index].find('mirror')
+        if disk_mirror is None:
+            test.fail("Failed to get disk mirror")
+        disk_mirror_source = disk_mirror.find('source')
+        return int(disk_mirror_source.get('index'))
+
     # Disk specific attributes.
     device = params.get("virt_disk_device", "disk")
     device_target = params.get("virt_disk_device_target", "vdd")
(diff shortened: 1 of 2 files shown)


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virsh.domblkthreshold.positive_test.coldplug.snapshot_test.mirror_mode_blockcommit


Command 'virsh list' finished with 0 after 0.46877121925354004s
[stdout]  1    gls    running
[stdout] 
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virsh.domblkthreshold.positive_test.coldplug.snapshot_test.mirror_mode_blockcopy


PASS 1-type_specific.io-github-autotest-libvirt.virsh.domblkthreshold.positive_test.coldplug.snapshot_test.default

chloerh:
@chunfuwen  Could you @ the feature owner to help review this pr? Thanks

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
```diff
@@ -111,6 +111,66 @@ def create_vol(p_name, vol_params):
...
+        :param target: Disk target in VM.
+        """
+        try:
+            session = vm_domain.wait_for_login()
+            time.sleep(10)
+            cmd = ("dd if=/dev/urandom of=file bs=1G count=3")
+            status, output = session.cmd_status_output(cmd)
+            if status:
+                test.error("Failed to fill data in VM target: %s with %s" % (target, output))
+        except (remote.LoginError, virt_vm.VMError, aexpect.ShellError) as e:
```
chloerh, line 155: What about other kinds of errors/exceptions? Seems we're not handling them.

chunfuwen: updated

chloerh (approved):

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
```diff
@@ -111,6 +111,68 @@ def create_vol(p_name, vol_params):
...
+
+        :param vm_name: VM name
+        :param target: Disk dev in VM.
+        :param blockcommit_options: blockcommit option
+        :param virsh_dargs: additional parameters
+        """
+        result = virsh.blockcommit(vm_name, target,
+                                   blockcommit_options, **virsh_dargs)
+        # Check status
+        libvirt.check_exit_status(result)
```
kylazhang, line 126: Although we agree this kind of usage but I have to say the above all from 114-126 part can be omitted and using virsh.blockcommit(vm_name, target,                                blockcommit_options, ignore_status=False, **virsh_dargs) when do blockcommit

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
```diff
@@ -111,6 +111,68 @@ def create_vol(p_name, vol_params):
...
+            time.sleep(10)
+            cmd = ("dd if=/dev/urandom of=file bs=1G count=3")
+            status, output = session.cmd_status_output(cmd)
+            if status:
+                test.error("Failed to fill data in VM target: %s with %s" % (target, output))
+        except (remote.LoginError, virt_vm.VMError, aexpect.ShellError) as e:
+            logging.error(str(e))
+            raise
+        except Exception as ex:
+            raise
```
kylazhang, line 159: Can I know the reason that some exception logging the error but some not? Or we can log all exception error?

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
```diff
@@ -111,6 +111,68 @@ def create_vol(p_name, vol_params):
...
+    def trigger_mirror_threshold_event(vm_domain, target):
+        """
+        Trigger mirror mode block threshold event.
+
+        :param vm_domain: VM name
+        :param target: Disk target in VM.
+        """
+        try:
+            session = vm_domain.wait_for_login()
+            time.sleep(10)
```
kylazhang, line 150: Please give the reason for the hardcode sleep 10 min instead of waiting some specific operation/value

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
chunfuwen: updated

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
chunfuwen: updated

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
chunfuwen: wait is here used to Thread synchronization. In this case, wait for ten seconds to let virsh wait event (block-method)start first in main thread .

kylazhang: I have to say it have risk to only wait for long time but not the real event, it may cause 2 types of failure, not start or already finished, I really suggest to use wait for

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
kylazhang: I mean as you only deal with the exception with logging.error(str(e)) so why not just remove the  except (remote.LoginError, virt_vm.VMError, aexpect.ShellError) as e: ? As we can not get any different from those 2 exceptions

kylazhang (commented):
Please check the comments inline, thanks

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
chunfuwen: one more logging.error line  somehow helps remind error message if exception is expected (caught).
Regarding unexpected except, raise directly is more common used

libvirt/tests/src/virsh_cmd/domain/virsh_domblkthreshold.py
chunfuwen: I think you misunderstand the situation. The wait in child thread is to let main thread to have chance to start wait-for-events ,which is block-method. sleep/wait is very common and effective way to sync threads operation. Previous similar usage experience prove that it is reliable and safe.

kylazhang (approved):

PR #3060, opened 2020-09-16
Fix xml error when setting numa_cell to vmcpuxml

chloerh:
The original way is to get an xml element object, which is not an
object of VMCPUXML. The new way is to create new cpu xml if cannot
find one in vmxml, then set numa_cells


chloerh:
Test result: pending

chloerh:
@dzhengfy Would you like to review? It's about ppc test. Thank u.

dzhengfy:
The change logic is simple, and few lines are changed for ppc only. So I mark it with minorfix.

dzhengfy (approved):
LGTM

PR #3068, opened 2020-09-21
Fix disk_bus_virtio.device_disk.default case failure

chunfuwen:
After libvirt 6.6.0, scsi=on/off flag is removed from qemu command line after libvirt 6.6.0,
so set device_option .* to make code compatible

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -1787,8 +1787,13 @@ def run(test, params, env):
                         device_option = "scsi=on"
                     else:
                         device_option = "scsi=off"
-                    cmd += (" | grep virtio-blk-pci,%s,bus=pci.%x,addr=0x%x"
-                            % (device_option, dev_bus, pci_slot))
+                    # scsi=on/off flag is removed from qemu command line after libvirt 6.6.0, so update cmd to make code compatible.
+                    if libvirt_version.version_compare(6, 6, 0):
+                        cmd += (" | grep virtio-blk-pci,bus=pci.%x,addr=0x%x"
+                                % (dev_bus, pci_slot))
+                    else:
+                        cmd += (" | grep virtio-blk-pci,%s,bus=pci.%x,addr=0x%x"
+                                % (device_option, dev_bus, pci_slot))
                 if device_bus[0] in ["ide", "sata", "scsi"]:
                     dev_unit = int(vm_xml.VMXML.get_disk_attr(vm_name, device_targets[0],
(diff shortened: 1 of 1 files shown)


chunfuwen:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multidisks.coldplug.single_disk_test.disk_bus_device_option.disk_bus_virtio.device_disk.default

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1787,6 +1787,9 @@ def test_device_update_boot_order(disk_type, disk_path, error_msg):
                         device_option = "scsi=on"
                     else:
                         device_option = "scsi=off"
+                    # scsi=on/off flag is removed from qemu command line after libvirt 6.6.0, so set device_option .* to make code compatible.
+                    if libvirt_version.version_compare(6, 6, 0):
+                        device_option = ".*"
```
sathnaga, line 1792: better to handle grep command option aswell, we can simply omit `device_option` if libvirt_version greater than 6.6 why to update it with `.*`, any specific reason, `.*` kind of regular expression might match unwanted strings aswell, better to drop it completely.

sathnaga (changes requested):

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
chunfuwen: the regular expression are composed by server string including device_option, see line 1793
                    cmd += (" | grep virtio-blk-pci,%s,bus=pci.%x,addr=0x%x"
                            % (device_option, dev_bus, pci_slot))
so as per my understanding, device_option as'.*' will  bring very limited effect due to constraints of other string

sathnaga: @chunfuwen 
my suggestion was to update the grep cmd itself, for 6.6.0 something like
```
cmd += (" | grep virtio-blk-pci,bus=pci.%x,addr=0x%x" % (dev_bus, pci_slot))
```
not sure if there is any other constraints by doing it though.

`.*` will eventually pass anything even with `scsi=on/off` too, so my suggestion was to drop to completely.

sathnaga (changes requested):
Comment given above

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
chunfuwen: updated accordingly

sathnaga (approved):
LGTM

PR #3074, opened 2020-09-21
Fix snapshot-revert case failures

chunfuwen:
Virsh snapshot-revert is not supported on combined internal and external snapshots
See more details from,https://bugzilla.redhat.com/show_bug.cgi?id=1733173
Moreover, add logic to fix no more pci slots issue encountered in certain cases

### DIFF
--- a/libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
+++ b/libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
@@ -96,6 +96,22 @@ def run(test, params, env):
     # Back VM XML
     vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
 
+    # Fix no more PCI slots issue in certain cases.
+    vm_dump_xml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    machine_type = params.get("machine_type", "pc")
+    if machine_type == 'q35':
+        vm_dump_xml.remove_all_device_by_type('controller')
+        machine_list = vm_dump_xml.os.machine.split("-")
+        vm_dump_xml.set_os_attrs(**{"machine": machine_list[0] + "-q35-" + machine_list[2]})
+        q35_pcie_dict0 = {'controller_model': 'pcie-root', 'controller_type': 'pci', 'controller_index': 0}
+        q35_pcie_dict1 = {'controller_model': 'pcie-root-port', 'controller_type': 'pci'}
+        vm_dump_xml.add_device(libvirt.create_controller_xml(q35_pcie_dict0))
+        # Add enough controllers to match multiple times disk attaching requirements
+        for i in list(range(1, 12)):
+            q35_pcie_dict1.update({'controller_index': "%d" % i})
+            vm_dump_xml.add_device(libvirt.create_controller_xml(q35_pcie_dict1))
+        vm_dump_xml.sync()
+
     virsh_dargs = {'debug': True, 'ignore_status': True}
     try:
         start_vm = "yes" == params.get("start_vm", "yes")
@@ -283,6 +299,8 @@ def run(test, params, env):
         elif domain_operation == "snapshot":
             # Run snapshot related commands: snapshot-create-as, snapshot-list
             # snapshot-info, snapshot-dumpxml, snapshot-create
+            # virsh snapshot-revert is not supported on combined internal and external snapshots
+            # see more details from,https://bugzilla.redhat.com/show_bug.cgi?id=1733173
             snapshot_name1 = "snap1"
             snapshot_name2 = "snap2"
             cmd_result = virsh.snapshot_create_as(vm_name, snapshot_name1,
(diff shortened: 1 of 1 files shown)


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.at_dt_iscsi_disk.positive_test.cold_plug.volume_disk.iscsi_direct_pool.snapshot_domain


cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.at_dt_iscsi_disk.positive_test.hot_plug.volume_disk.iscsi_pool.snapshot_domain

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
```diff
@@ -96,6 +96,22 @@ def run(test, params, env):
...
+    machine_type = params.get("machine_type", "pc")
+    if machine_type == 'q35':
+        vm_dump_xml.remove_all_device_by_type('controller')
+        machine_list = vm_dump_xml.os.machine.split("-")
+        vm_dump_xml.set_os_attrs(**{"machine": machine_list[0] + "-q35-" + machine_list[2]})
+        q35_pcie_dict0 = {'controller_model': 'pcie-root', 'controller_type': 'pci', 'controller_index': 0}
+        q35_pcie_dict1 = {'controller_model': 'pcie-root-port', 'controller_type': 'pci'}
+        vm_dump_xml.add_device(libvirt.create_controller_xml(q35_pcie_dict0))
+        # Add enough controllers to match multiple times disk attaching requirements
+        for i in list(range(1, 12)):
```
kylazhang, line 110: Could you please explain why use range 12? Or may be it should be a param in cfg

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
```diff
@@ -96,6 +96,22 @@ def run(test, params, env):
     # Back VM XML
     vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
 
+    # Fix no more PCI slots issue in certain cases.
+    vm_dump_xml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    machine_type = params.get("machine_type", "pc")
+    if machine_type == 'q35':
+        vm_dump_xml.remove_all_device_by_type('controller')
```
kylazhang, line 103: Is that ok to move all controller? scsi/sata also controllers which are usable for disk. I think only remove pci type controller is ok

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
```diff
@@ -309,11 +327,8 @@ def get_vol():
             cmd_result = virsh.snapshot_current(vm_name, **virsh_dargs)
             libvirt.check_exit_status(cmd_result)
 
-            snapshot_file = os.path.join(data_dir.get_tmp_dir(), snapshot_name2)
-            sn_create_op = ("%s --disk-only --diskspec %s,file=%s"
-                            % (snapshot_name2, disk_target, snapshot_file))
-            cmd_result = virsh.snapshot_create_as(vm_name, sn_create_op,
-                                                  **virsh_dargs)
+            virsh.snapshot_create_as(vm_name, snapshot_name2,
+                                     ignore_status=True, debug=True)
```
kylazhang, line 331: I think here ignore_status should be False? Or you missing cmd_result=?

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
chunfuwen: it is experiences number in actual trival testing with very additional buffer

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
chunfuwen: Removing all controller had been proven safe by a few previous fixes.

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
kylazhang: I think the number may be different for different test, and maybe it can be a parameter in cfg, and for special cases give the special number is reasonable

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
```diff
@@ -309,13 +327,9 @@ def get_vol():
             cmd_result = virsh.snapshot_current(vm_name, **virsh_dargs)
             libvirt.check_exit_status(cmd_result)
 
-            snapshot_file = os.path.join(data_dir.get_tmp_dir(), snapshot_name2)
-            sn_create_op = ("%s --disk-only --diskspec %s,file=%s"
-                            % (snapshot_name2, disk_target, snapshot_file))
-            cmd_result = virsh.snapshot_create_as(vm_name, sn_create_op,
-                                                  **virsh_dargs)
+            virsh.snapshot_create_as(vm_name, snapshot_name2,
```
kylazhang, line 330: Please give the reason of delete the snapshot-create-as parameters, and also add them in the commit message

kylazhang: Just check with chunfu, the test scenario changed and also discussed with feature owner so also need to change the snapshot options

libvirt/tests/src/virtual_disks/at_dt_iscsi_disk.py
kylazhang: Just discussed with chunfu, the test with this update may only related single attach operation so no need to set special number for cases

kylazhang (approved):

PR #3075, opened 2020-09-21
Add new cases for mixed detach disk

dzhengfy:
Since libvirt detach device APIs(detach-device, detach-device-alias)
works asynchronously, that means every call returns before the detach
operation returns.This patch is to test multiple continuous detach to
make sure a detach call will not be interrupted by other calls.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_detach_device.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_detach_device.cfg
@@ -12,6 +12,11 @@
             no cdrom_test,file_type
             only vm_running_config,vm_running_live,vm_running_current
             dt_device_alias = "yes"
+        - detach_mixed:
+            only normal_test..vm_running_live..disk_test..file_type..detach_mixed
+            dt_device_alias = "yes"
+            detach_times = 2
+            detach_mixed = "yes"
     variants:
         - cdrom_test:
             dt_device_pre_vm_state = "shut off"
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
@@ -3,6 +3,7 @@ import time
 import logging
 import shutil
 import uuid
+import threading
 
 import aexpect
 
@@ -139,6 +140,47 @@ def run(test, params, env):
             shutil.copyfile(iface.xml, device_xml_file)
         return device_xml_file
 
+    def handle_detach_mix():
+        """
+        Do detach_device and detach_device_alias parallel for multiple times
+        """
+        detach_times = int(params.get("detach_times"))
+        thread_list = []
+
+        def _detach_device(by_alias, detach_target):
+            """
+            Route the invoke for suitable detach functions
+
+            :param by_alias: True to use detach_device_alias
+            :param detach_target: Device xml or device alias
+            :return: None
+            """
+            func_name = virsh.detach_device_alias if by_alias else virsh.detach_device
+            func_name(vm_ref, detach_target,
+                      extra=dt_options,
+                      wait_remove_event=True,
+                      event_timeout=7,
+                      debug=True)
+
+        for count in range(0, detach_times):
+            dt_thread = threading.Thread(target=_detach_device,
+                                         args=(False, device_xml))
+
+            dt_alias_thread = threading.Thread(target=_detach_device,
+                                               args=(True, device_alias))
+
+            dt_thread.start()
+            thread_list.append(dt_thread)
+            logging.debug("Start thread using detach_device %s", dt_thread.name)
+
+            dt_alias_thread.start()
+            thread_list.append(dt_alias_thread)
+            logging.debug("Start thread using detach_device_alias %s", dt_alias_thread.name)
+
+        for one_thread in thread_list:
+            logging.debug("Ensure thread '{}' to exist".format(one_thread.name))
+            one_thread.join()
+
     vm_ref = params.get("dt_device_vm_ref", "name")
     dt_options = params.get("dt_device_options", "")
     pre_vm_state = params.get("dt_device_pre_vm_state", "running")
@@ -149,6 +191,7 @@ def run(test, params, env):
     readonly = "yes" == params.get("detach_readonly", "no")
     tmp_dir = data_dir.get_tmp_dir()
     test_cmd = "detach-device"
+    detach_mixed = "yes" == params.get("detach_mixed", "no")
     detach_alias = "yes" == params.get("dt_device_alias", "no")
     if detach_alias:
         test_cmd = "detach-device-alias"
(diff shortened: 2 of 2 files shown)


dzhengfy:
Run on beaker host:
```
# avocado run --vt-type libvirt --vt-machine-type q35 virsh.detach_device..detach_mixed
JOB ID     : d7cc7c3cc1199010608935f6564f40b2146b7ca0
JOB LOG    : /root/avocado/job-results/job-2020-09-21T05.41-d7cc7c3/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.detach_device.normal_test.vm_running_live.disk_test.file_type.detach_mixed: PASS (74.07 s)
JOB TIME   : 76.54 s

```

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
```diff
@@ -248,8 +249,23 @@ def create_device_xml(params, xml_path, device_source):
...
+            for count in range(0, detach_times):
+                virsh.detach_device_alias(vm_ref, device_alias,
+                                          dt_options, True, 7,
+                                          readonly=readonly,
+                                          debug=True)
+                virsh.detach_device(vm_ref, device_xml,
+                                    readonly=readonly,
+                                    flagstr=dt_options,
+                                    debug=True)
+                time.sleep(1)
```
chunfuwen, line 267: def detach_device(domainarg=None, filearg=None,
                  domain_opt=None, file_opt=None,
                  flagstr=None, wait_remove_event=False, event_timeout=7, **dargs):
detach_device extend one more parameter wait_remove_event, so it may help removal of time.sleep statement here

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
```diff
@@ -248,8 +249,23 @@ def create_device_xml(params, xml_path, device_source):
...
-
-        if detach_alias:
+        if detach_mixed:
+            detach_times = int(params.get("detach_times"))
+            status = virsh.detach_device(vm_ref, device_xml,
+                                         readonly=readonly,
+                                         flagstr=dt_options,
+                                         debug=True).exit_status
+            for count in range(0, detach_times):
+                virsh.detach_device_alias(vm_ref, device_alias,
```
chunfuwen, line 259: do we need create separate threads to trigger those detach operations?

sathnaga: yes, from the commit message it looks like parallel requests are needed, better to trigger via threading...
this seems not cover the case.

sathnaga (changes requested):

dzhengfy:
Run with libvirt-ci provisioned machine:
```
# avocado run --vt-type libvirt --vt-machine-type q35 virsh.detach_device..detach_mixed
JOB ID     : 879f053ea34e170ece74ec7d32bb2626d9b02580
JOB LOG    : /root/avocado/job-results/job-2020-10-12T20.37-879f053/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.detach_device.normal_test.vm_running_live.disk_test.file_type.detach_mixed: PASS (47.00 s)
JOB TIME   : 49.99 s
```

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
dzhengfy: @chunfuwen @sathnaga yes, updated.

chunfuwen (approved):
LGTM

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
```diff
@@ -139,6 +140,39 @@ def create_device_xml(params, xml_path, device_source):
...
+                      debug=True)
+
+        for count in range(0, detach_times):
+            dt_thread = threading.Thread(target=_detach_device,
+                                         args=(False, device_xml))
+
+            dt_alias_thread = threading.Thread(target=_detach_device,
+                                               args=(True, device_alias))
+
+            dt_thread.start()
```
sathnaga, line 171: wait for thread to finish?, i guess dt_thread.join() is missing...
same for other thread

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
```diff
@@ -248,8 +283,10 @@ def create_device_xml(params, xml_path, device_source):
             vm_ref = domuuid
         else:
             vm_ref = ""
-
-        if detach_alias:
+        if detach_mixed:
+            handle_detach_mix()
+            status = 0  # We do not care the virsh command return result
```
sathnaga, line 288: this does not look good, better make it return from the function and have the comment there, or controlled by ignore_status etc..

sathnaga (changes requested):

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
dzhengfy: @sathnaga, originally I also think to use join(). But finally I do not think it is necessary because virsh detach operations are always asynchronous that means the functions will return immediately and the thread will end up. So I think no need to join() again.
So what do you think about it?

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
dzhengfy: @santwana , the scenario is to execute multiple detach at same time. And the first one will succeed and the followings will fail which are we expected.  Here I set status =0 to avoid of the following checking codes failed by the failure of second detach. And it is enough to check the device within the guest and guest xml.

Currently second detach failure already does not raise any exception with ignore_status=True. I am not sure what you mean by 'controlled by ignore_status'. 
And you want the function to return a result, then which detach result should I return?

kylazhang: The comments in the code here is acceptable for me to understand, so approve for not update here.

libvirt/tests/src/virsh_cmd/domain/virsh_detach_device.py
dzhengfy: @kylazhang @sathnaga , updated.

libvirt/tests/cfg/virsh_cmd/domain/virsh_detach_device.cfg
```diff
@@ -12,6 +12,11 @@
             no cdrom_test,file_type
             only vm_running_config,vm_running_live,vm_running_current
             dt_device_alias = "yes"
+        - detach_mixed:
+            only normal_test.vm_running_live.disk_test.file_type.detach_mixed
```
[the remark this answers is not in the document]
dzhengfy: @kylazhang This customer bug does not cover other configurations, so I fixed the case pattern like this.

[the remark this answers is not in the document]
dzhengfy: @kylazhang updated.

kylazhang (approved):

PR #3089, opened 2020-09-24
Update test params for blockcopy

kylazhang:
change the block+lun test scenaro mapping from qcow2+virtio to raw+scsi as it do not supported with the old mapping

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockcopy.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_blockcopy.cfg
@@ -101,9 +101,6 @@
                 - block_disk:
                     replace_vm_disk = "yes"
                     disk_source_protocol = "iscsi"
-                    disk_target = "vda"
-                    disk_target_bus = "virtio"
-                    disk_format = "qcow2"
                     image_size = "10G"
                     emulated_image = "emulated-iscsi"
                     variants:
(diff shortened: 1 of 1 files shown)


kylazhang:
The test result failure message changed from 
"Failed to define avocado-vt-vm1 from xml"
to
"Libvirtd log error happened: Timed out during operation: cannot acquire state change lock" which also have bug not fixed

libvirt/tests/cfg/virsh_cmd/domain/virsh_blockcopy.cfg
```diff
@@ -120,8 +117,14 @@
                         - lun_t:
                             no volume_type, no_blockdev, shallow
                             disk_device = 'lun'
+                            disk_format = "raw"
+                            disk_target = "sda"
+                            disk_target_bus = "scsi"
```
chunfuwen, line 122: As my understanding, we need consider code compatibility, so the change need consider specific libvirt version.

kylazhang: I think it can also work with the old version, so may no need to do judgement in the code

kylazhang: I can run with old version to make sure

chunfuwen (dismissed):
LGTM

libvirt/tests/cfg/virsh_cmd/domain/virsh_blockcopy.cfg
kylazhang: After test on jenkins, all block_disk.lun_t cases passed for rhel8.2 job libvirt-RHEL-8.2-runtest-x86_64-function-block_copy_no_shallow

kylazhang:
@Yingshun can you help to review? Thanks

Yingshun (approved):

chunfuwen (approved):
LGTM

kylazhang:
meili also submit a same PR for it so from functional side, it have no problem

PR #3091, opened 2020-09-24
Update scsi controller on ppc vms to support scsi hotplug

chloerh:
Since hotplug scsi disk with ibmvscsi controller is not supported,
replacing with virtio-scsi model could fix it.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
@@ -7,6 +7,7 @@ import os.path
 import logging
 import aexpect
 import itertools
+import platform
 from string import ascii_lowercase
 
 from six import iteritems
@@ -867,7 +868,7 @@ def remove_chardevs(vm_name, vmxml):
     for chardev_type in chardev_types:
         vmxml.del_device(chardev_type, by_tag=True)
     vmxml.sync()
-    virsh.start(vm_name, debug=True)
+    virsh.start(vm_name, debug=True, ignore_status=False)
 
 
 def remove_non_disks(vm_name, vmxml):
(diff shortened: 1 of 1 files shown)


chloerh:
Test result

PASS 1-type_specific.io-github-autotest-libvirt.virsh.attach_device.block.single_scsi_file.without_alias.normal_test.hot_attach_hot_vm.persistent.name_ref.file_positional.domain_positional

libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
```diff
@@ -888,6 +889,22 @@ def remove_non_disks(vm_name, vmxml):
...
+    :param vm_name: domain name
+    :param vmxml: domain xml
+    :return:
+    """
+    machine = platform.machine().lower()
+    if 'ppc64le' in machine:
+        # Remove old scsi controller, replace with virtio-scsi controller
+        vmxml.del_controller('scsi')
+        vmxml.sync()
+        virsh.start(vm_name, debug=True, ignore_status=False)
```
chunfuwen, line 905: Do you need judge whether previous VM state is in running, rather that start it directly regardless of previous VM state?

chloerh: @chunfuwen Actually `vmxml.sync()` will destroy vm, therefore vm needs to be started after `sync()`

Yingshun (approved):

libvirt/tests/src/virsh_cmd/domain/virsh_attach_device.py
chunfuwen: I mean if previous VM is not running, then at last you try to start it due to this change. In this way, the test logic could be changed.

chloerh: @chunfuwen I guess you haven't read code coming after these lines

```
    if test_params.start_vm:
        # Make sure VM is working
        test_params.main_vm.verify_alive()
        test_params.main_vm.wait_for_login().close()
    else:  # VM not suppose to be started
        if test_params.main_vm.is_alive():
            test_params.main_vm.destroy(gracefully=True)
```

This piece of code makes sure that what you're worrying won't happen.

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

PR #3092, opened 2020-09-24
Fix wrong format of setting params for specific machine type

chloerh:


dzhengfy (approved):
LGTM

dzhengfy:
As this is an obvious fix, we need no test result.

PR #3102, opened 2020-09-29
virtual_disks_multidisks: rm the cases of usb bus together with devic…

qiankehan:
…e lun

According to the libvirt validation function
qemuValidateDomainDeviceDefDiskFrontend[1], the deivce lun is only
valid with virtio or scsi bus.

[1]: https://gitlab.com/libvirt/libvirt/-/raw/master/src/qemu/qemu_validate.c

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_multidisks.cfg
@@ -864,14 +864,6 @@
                             virt_disk_device_target = "sda"
                             virt_disk_device_bus = "usb"
                             variants:
-                                - device_lun:
-                                    status_error = "yes"
-                                    virt_disk_device_type = "block"
-                                    virt_disk_device_source = "test"
-                                    virt_disk_device = "lun"
-                                    virtio_scsi_controller = "yes"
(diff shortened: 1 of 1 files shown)


qiankehan:
no test result because of only removed cases

chunfuwen:
@qiankehan 
Do you want to totally remove it ,or it need judge based on libvirt version?

qiankehan:
> @qiankehan
> Do you want to totally remove it ,or it need judge based on libvirt version?

Yes. See the following code in the reference linl:
```c
    if (disk->device == VIR_DOMAIN_DISK_DEVICE_LUN) {
        /* make sure that both the bus supports type='lun' (SG_IO). */
        if (disk->bus != VIR_DOMAIN_DISK_BUS_VIRTIO &&
            disk->bus != VIR_DOMAIN_DISK_BUS_SCSI) {
            virReportError(VIR_ERR_CONFIG_UNSUPPORTED,
                           _("disk device='lun' is not supported for bus='%s'"),
                           virDomainDiskBusTypeToString(disk->bus));
            return -1;
        }
```
The comment there means **device lun** only could be used for the buses support **SG_IO**.
Usb bus is not the target bus for lun

chunfuwen (approved):

PR #3104, opened 2020-09-30
Add vfio-ap test cases

smitterl:
Depends on https://github.com/avocado-framework/avocado-vt/pull/2780

Add test cases for single device passthrough hot- and coldplug.
This feature is only available on s390x.

### DIFF
--- a/libvirt/tests/src/svirt/libvirt_ap_passthrough.py
+++ b/libvirt/tests/src/svirt/libvirt_ap_passthrough.py
@@ -1,0 +1,99 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2020
+# Author: Sebastian Mitterle <smitterl@redhat.com>
+
+import logging
+
+from virttest import virsh
+from virttest.utils_zcrypt import CryptoDeviceInfoBuilder, \
+    APMaskHelper, MatrixDevice, load_vfio_ap
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest.libvirt_xml.devices import hostdev
+from virttest.utils_misc import wait_for
+
+# minimal supported hwtype
+MIN_HWTYPE = 10
+
+
+def run(test, params, env):
+    """
+    Tests vfio-ap passthrough on s390x
+
+    1. Control guest lifecycle for cold- vs. hotplug
+    2. Set up passthrough attaching new device
+    3. Confirm device availability in guest
+    """
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+    vmxml_backup = VMXML.new_from_inactive_dumpxml(vm_name)
+
+    plug = params.get("plug")
+    mask_helper = None
+    matrix_dev = None
+
+    try:
+        if plug == "cold" and vm.is_alive():
+            vm.destroy()
+        if plug == "hot" and vm.is_dead():
+            vm.start()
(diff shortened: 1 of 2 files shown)


smitterl:
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qe
mu:///system libvirt_ap_passthrough
JOB ID     : 05b40f23606c045294f85d914138099b489e194f
JOB LOG    : /root/avocado/job-results/job-2020-09-30T05.06-05b40f2/job.log                                                                
 (1/2) type_specific.io-github-autotest-libvirt.libvirt_ap_passthrough.hotplug: PASS (17.23 s)                                             
 (2/2) type_specific.io-github-autotest-libvirt.libvirt_ap_passthrough.coldplug: PASS (31.56 s)                                            
JOB TIME   : 50.28 s
```

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
```diff
@@ -0,0 +1,99 @@
...
+
+    plug = params.get("plug")
+    mask_helper = None
+    matrix_dev = None
+
+    try:
+        if plug == "cold" and vm.is_alive():
+            vm.destroy()
+        if plug == "hot" and vm.is_dead():
+            vm.start()
```
dzhengfy, line 48: Could you help add try...except clause to detect the unexpected starting failure?

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
```diff
@@ -0,0 +1,99 @@
...
+            vm.wait_for_login()
```
dzhengfy, line 49: If we do not need a login session, we'd better use 'vm.wait_for_login().close()' to close the session.

dzhengfy (commented):
Others looks fine to me

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
smitterl: @dzhengfy This is the hotplug branch, this line makes sure the machine has finished booting so that the actual hotplug precondition is met. IMO this is necessary because the test might attach the device while the guest is still booting but that would then not be a valid state for any hotplug scenario.

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
```diff
@@ -0,0 +1,99 @@
...
+        uuid = matrix_dev.uuid
+        hostdev_xml.source = hostdev_xml.new_source(**{"uuid": uuid})
+        hostdev_xml.xmltreefile.write()
+
+        logging.debug("Attaching %s", hostdev_xml.xmltreefile)
+        virsh.attach_device(vm_name, hostdev_xml.xml, flagstr="--current",
+                            ignore_status=False)
+
+        if plug == "cold":
+            vm.start()
```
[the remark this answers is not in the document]
smitterl: @dzhengfy I'm not sure I understand correctly: Do you mean to add a `except` clause to handle the "VMStartError" that might be raised e.g. in https://github.com/avocado-framework/avocado-vt/blob/613b909f8f7e7c359cd4ca4b5f4800bab74b7d13/virttest/libvirt_vm.py#L2477 when issuing `vm.start`? If so, I don't understand what the value of adding that logic would be because if the start fails the test execution will jump to the finally clause and bubble up the `VMStartError` afterwards which will result in ERROR with the corresponding message from "VMStartError". Maybe I'm missing something.

IIUC, handling any exception thrown in the try block is only useful in the test script if we react on them with some behavior besides error logging (e.g. some clean up task or retrying some action etc.)

Please, can you help clarify?

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
smitterl: s. above

libvirt/tests/src/svirt/libvirt_ap_passthrough.py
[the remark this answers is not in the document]
dzhengfy: Yeah, your description makes sense to me.

dzhengfy (approved):
LGTM

chunfuwen (approved):
LGTM

PR #3111, opened 2020-10-16
[batch 1]Fixing split daemon cases for remote connection of virsh cmds

iccaszhulili:
In modular daemon environment, remote mode is set to 'direct' by
default, which is not supported for SSH transport.

Signed-off-by: iccaszhulili <lizhu@redhat.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_dominfo.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_dominfo.py
@@ -1,6 +1,7 @@
 from virttest import ssh_key
 from virttest import virsh
 from virttest import utils_libvirtd
+from virttest.utils_libvirt import libvirt_config
 
 
 def run(test, params, env):
@@ -31,6 +32,7 @@ def run(test, params, env):
     remote_pwd = params.get("remote_pwd", "")
     remote_user = params.get("remote_user", "root")
     remote_uri = params.get("remote_uri")
+    client_libvirt_file = None
 
     # run test case
     if vm_ref == "id":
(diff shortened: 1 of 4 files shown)


chunfuwen:
Let's hold on this review.
As per my understanding, setting mode=direct in libvirt.conf is not necessary in enabling libvirtd working under modularity mode.
Need work with yafu to work out it in alternative way.
I pasted original email with Daniel here.
**************************************************************************************************************
----- Original Message -----
> From: "Daniel P. Berrangé" <berrange@redhat.com>
> To: "Chunfu Wen" <chwen@redhat.com>
> Cc: "Yan Fu" <yafu@redhat.com>, "Jaroslav Suchanek" <jsuchane@redhat.com>, "Xuesong Zhang" <xuzhang@redhat.com>,
> "Lijuan Men" <lmen@redhat.com>, "dyuan" <dyuan@redhat.com>, "Chenli Hu" <chhu@redhat.com>, "Dan Zheng"
> <dzheng@redhat.com>
> Sent: Monday, 24 February, 2020 8:28:24 PM
> Subject: Re: Test matrix review for split daemons
>
> On Mon, Feb 24, 2020 at 07:18:00AM -0500, Chunfu Wen wrote:
> > Hi Daniel,
> >
> > Thank you for your reply!
> > I add one more question here.
> >
> > **************************************************
> > For any virsh command(need start virtqemud first), it needs explicitly
> > specify uri now, is this expected ?
> >
> > [root@dell-per430-02 libvirt]# virsh -c 'qemu:///system' list
> > Id Name State
> > --------------------
> >
> > [root@dell-per430-02 libvirt]# virsh list
> > error: failed to connect to the hypervisor
> > error: Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such
> > file or directory
>
> This is related to our compile time default, which still prefers the
> legacy libvirtd. So with no URI, virsh still tries to connect to
> libvirtd's socket. To deal with this you must have virtproxyd sockets
> enabled, which will let virproxdy auto-start.
>
> Alternatively use /root/.config/libvirt/libvirt.conf to set
>
>    remote_mode=direct

Does that only work for non-root session mode? I tested starting virtqemud from systemd or foreground, it still needed to use 'virsh -c qemu:///system list'.

>
> which should cause it to detect virtqemud and connect to its
> socket without setting a URI.
>
>
> Regards,
> Daniel

dzhengfy:
Hold on this PR as we are in the progress of removing 'remote_mode=direct' from /etc/libvirt/libvirt.conf in CI. @iccaszhulili @chunfuwen

chunfuwen:
close this PR since it is not necessary after remove remote_mode=direct' from /etc/libvirt/libvirt.conf

PR #3115, opened 2020-10-17
migrate_vm: Remove meaningless "libvirtd crash" check

fangge1212:
Libvirtd will be restarted immediately by systemd when it crashed,
so it is meaningless to do "libvirtd crash" check.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -3,7 +3,6 @@ import os
 import re
 import signal
 import time
-import aexpect
 import platform
 
 from subprocess import PIPE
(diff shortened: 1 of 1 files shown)


fangge1212:
**Before fix:**
Test in modular daemon mode:
FAIL 1-type_specific.io-github-autotest-libvirt.virsh.migrate_vm.positive_testing.p2p_migration.basic.with_tls_reverse.with_postcopy -> TestFail: Local libvirtd service is crashed unexpectedly.

**After fix:**
Test in modular daemon mode:
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.positive_testing.p2p_migration.basic.with_tls_reverse.with_postcopy: PASS (216.93 s)

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2560,25 +2559,8 @@ def get_target_hugepage_num(params):
             migrate_vm(test, test_dict)
 
         if target_vm_name:
-            # Check the libvirtd service is running on both hosts.
-            logging.debug("Local libvirtd service should still be running. "
-                          "Check it...")
-            libvirtd = utils_libvirtd.Libvirtd()
-            if not libvirtd.is_running():
-                test.fail("Local libvirtd service is crashed unexpectedly.")
-            session = None
```
Yingshun, line 2569: Does version checking require here? BTW, may I ask if you've updated the related manual case?

fangge1212: 1. Systemd had been supporting this years ago(I don't know the exact version), so I think there is no need to do the version check because we don't need to test the "very old" version
2. I don't think "libvirtd crash" check should be a test step in specific test case, I will delete the step in manual case. I think if we need to check "libvirtd crash", we need to check it in all the test cases, but it is impossible to add this step in all the manual cases.

3. But your comment reminds me another point about case design:
   1) We could check whether libvirtd is crashed during test by: check libvirtd pid before test -> do test -> check libvirtd pid after test, but is it meaningful to do so? Because if libvirtd crashed during test, the operation that make libvirtd crash will fail with following error(unless you are doing negative test and you expect the operation fail, you may not notice that libvirtd crashed if you don't check the error message)
> error: Disconnected from qemu:///system due to end of file
> error: End of file while reading data: Input/output error

2) If we need to check libvirtd crash during test, the better way is to put this check in the process of avocado-vt and has a configuration in base.cfg to turn it on/off.
I will talk about this within our team

fangge1212: No matter whether it is meaningful to check "libvirtd crash", the "check" in this py is really useless and should be removed

Yingshun: Okay fine, if you insist...

Yingshun (approved):

kylazhang (approved):

PR #3116, opened 2020-10-19
migrate_network: Refactor this mudule based on migration_template

fangge1212:
Refactor this mudule to make the code more readable and maintanable,
and reduce the effort of adding new features in the future.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_network.cfg
+++ b/libvirt/tests/cfg/migration/migrate_network.cfg
@@ -3,30 +3,17 @@
     # Console output can only be monitored via virsh console output
     only_pty = True
     take_regular_screendumps = no
-    # Options to pass to virsh migrate command before <domain> <desturi>
-    virsh_migrate_options = ""
-    # Extra options to pass after <domain> <desturi>
-    virsh_migrate_extra = ""
     # SSH connection time out
     ssh_timeout = 60
-    migration_setup = "yes"
-    storage_type = 'nfs'
-    setup_local_nfs = 'yes'
-    virsh_migrate_dest_state = running
-    virsh_migrate_src_state = running
-    virsh_migrate_options = "--live --p2p --verbose"
-    # Local URI
-    virsh_migrate_connect_uri = "qemu:///system"
 
     variants:
-        - with_postcopy:
-            postcopy_options = "--postcopy"
-        - without_postcopy:
-            postcopy_options = ""
+        - precopy:
+        - postcopy:
+            postcopy = True
     variants:
         - ovs_br:
-            only without_postcopy
-            migr_vm_back = "yes"
+            only precopy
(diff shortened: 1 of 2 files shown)


fangge1212:
This pr depends on https://github.com/avocado-framework/avocado-vt/pull/2709

fangge1212:
I can't provide the test result because the original script without this pr also can't run PASS

dzhengfy:
@fangge1212, as your  avocado-framework/avocado-vt#2709 is still in review, I suggest you need not to modify existing migration python files so urgent. Actually we'd better to discuss if there is a need to modify those existing python files aligning with the migration template. Could you share your thoughts?

fangge1212:
> @fangge1212, as your [avocado-framework/avocado-vt#2709](https://github.com/avocado-framework/avocado-vt/pull/2709) is still in review, I suggest you need not to modify existing migration python files so urgent. Actually we'd better to discuss if there is a need to modify those existing python files aligning with the migration template. Could you share your thoughts?

First, I want to see whether the migration template need update by using it 
Second, from the test requirement's viewpoint, the virsh migrate options in existing migrate_*.py is hard coded, I need to cover bigger test matrix, this can be achieved by migration template easily
Third, feature owner can add new case about migrate with specific feature more easily by using migration template

fangge1212:
> > @fangge1212, as your [avocado-framework/avocado-vt#2709](https://github.com/avocado-framework/avocado-vt/pull/2709) is still in review, I suggest you need not to modify existing migration python files so urgent. Actually we'd better to discuss if there is a need to modify those existing python files aligning with the migration template. Could you share your thoughts?
> 
> First, I want to see whether the migration template need update by using it
> Second, from the test requirement's viewpoint, the virsh migrate options in existing migrate_*.py is hard coded, I need to cover bigger test matrix, this can be achieved by migration template easily
Besides of the virsh migrate options, we need to cover "migrate vm back" , while some existing migrate_*.py doesn't do this step
> Third, feature owner can add new case about migrate with specific feature more easily by using migration template

fangge1212:
One another benefit would be that when somethings not related to specific feature(network ,cpu, etc) changed, we can update one single place (migration_template.py) instead of updating every single *.py.
I spent a lot of time to debug the failures in modular daemon mode recently, which can be avoided by using migration_template in the future

dzhengfy:
@fangge1212 even though the original script run Fail, you can copy the fail message and the test result with your changes. We can compare and make sure no regression.

fangge1212:
> @fangge1212 even though the original script run Fail, you can copy the fail message and the test result with your changes. We can compare and make sure no regression.

There is no test result, the host network is down during testing

fangge1212:
And I'm confused now, should I close this pr as nobody agree to update existing py file based on the new migration_template?

PR #3117, opened 2020-10-19
migrate_vm: Remove steps of setting selinux mode to permissive

fangge1212:
We should always do testing with selinux mode set to enforcing, or
we may miss bugs related to libvirt sVirt or selinux-policy.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1598,18 +1598,6 @@ def run(test, params, env):
             test_dict['brick_path'] = os.path.join(test.virtdir, pool_name)
 
         if gluster_disk:
-            logging.info("Put local SELinux in permissive mode")
-            utils_selinux.set_status("permissive")
-            LOCAL_SELINUX_ENFORCING = False
-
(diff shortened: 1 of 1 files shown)


Yingshun:
Could you please upload a test result?

fangge1212:
> Could you please upload a test result?

I think there is no need to provide a test result for this pr, if the case will fail in selinux enforcing mode, there would be a libvirt or selinux-policy bug, which has nothing to do with this pr

chunfuwen:
@fangge1212 , this change the code ,as per established reviewed rules and guideline, please upload test result

dzhengfy:
As this change 

> > Could you please upload a test result?
> 
> I think there is no need to provide a test result for this pr, if the case will fail in selinux enforcing mode, there would be a libvirt or selinux-policy bug, which has nothing to do with this pr

+1

dzhengfy:
Considering this is only to remove the selinux setting on both hosts, it does not change the case logic, so I agree we can merge it for now. If the case fails, that should be the feature issue because enforcing is required for migration setting.

dzhengfy (approved):
LGTM

PR #3123, opened 2020-10-22
virsh_migrate_setmaxdowntime: Fix conf issue in modular daemon mode

fangge1212:
Translation of conf_type from monolithic daemon mode to modular daemon
mode is complex and depends on the parameter to be configured heavily,
while the reverse translation is much more simple by tranlating all the 
conf_type of modular daemon mode to "libvirtd". This commit sets the 
conf_type to the value of modular daemon mode in <test>.cfg and the 
<test>.py or avocado-vt utils will translate to "libvirtd" if test env 
is monolithic daemon mode.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py
@@ -7,8 +7,6 @@ from avocado.utils import process
 
 from virttest import virsh
 from virttest import ssh_key
-from virttest import utils_libvirtd
-from virttest import utils_config
 from virttest import utils_misc
 from virttest import libvirt_version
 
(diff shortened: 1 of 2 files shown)


fangge1212:
**Before fix:**
In modular daemon mode:
 (02/11) type_specific.io-github-autotest-libvirt.virsh.migrate_setmaxdowntime.error_test.invalid_uuid.without_postcopy:  ERROR: [Errno 2] No such file or directory: '/var/log/libvirt/libvirtd.log' (255.43 s)

**After fix:**
Tested in both modular daemon mode and non-modular daemon mode, both PASSED:
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_setmaxdowntime.error_test.invalid_uuid.without_postcopy: PASS (254.16 s)

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py
```diff
@@ -78,18 +77,77 @@ def cleanup_dest(vm, src_uri, dest_uri):
...
+    :return: conf type
+    """
+
+    if utils_split_daemons.is_modular_daemon():
+        conf_type = conf_type_dict["modular_daemon_mode"]
+    else:
+        conf_type = conf_type_dict["normal_mode"]
+
+    return conf_type
+
```
Yingshun, line 106: If 'virtqemud' is a replacement of 'libvirtd' in your case, you can simply recall libvirt.customize_libvirt_config().

Yingshun (changes requested):

libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py
fangge1212: I don't want to hard code it in py file, in case we need to add new test case with different conf_type.

Yingshun: No, you don't need to hardcode, the default value of config_type is 'libivrtd' and it will become 'virtqemud' when modular daemon is enabled.

fangge1212: I don't want to hard code the conf_type of modular daemon mode to "virtqemud"

Yingshun: Could you please go through the code below? I think you didn't understand what I said.
https://github.com/avocado-framework/avocado-vt/blob/e41354352909a2a8bca9ee36b412ed3af7dfb436/virttest/utils_test/libvirt.py#L3577-L3606

fangge1212: I know  what you mean, in avocado-vt, conf type for daemon is hardcoded to virtqemud when libvirt is modular daemon, but this is not what I expect. Because conf type for daemon may not be virtqemud, it can also be virtproxyd or others. I think it will be more straitforward and clear for it to be configured in *.cfg. 
We can meet if there is still gap in our understanding.

fangge1212:
Retest case virsh.migrate_setmaxdowntime.error_test.invalid_uuid.without_postcopy, PASSED

fangge1212:
@Yingshun @dzhengfy 
I have updated this pr, please review again

Yingshun (approved):

kylazhang (approved):

PR #3126, opened 2020-10-22
migrate_vm: Set selinux state before migration

fangge1212:
Set selinux state before migration, default value is "enforcing".
There are two reasons to do this:
1) We should always do testing in "enforcing" mode, or we may miss
bugs related to libvirt sVirt or selinux-policy
2) If selinux is not set to enforcing before migration, some tests
will fail, e.g. *.disable_virt_use_nfs_remotely.*

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -214,6 +214,41 @@ def check_output(test, output_msg, params):
                             % (key, value, output_msg))
 
 
+def setup_selinux_local(selinux_state):
+    """
+    Set up selinux on local host
+
+    :param selinux_state: Enforcing, or Permissive
+    """
+
+    logging.debug("Set selinux to %s on local host before migration",
+                  selinux_state)
+    utils_selinux.set_status(selinux_state)
+
+
+def setup_selinux_remote(test, selinux_state, server_ip, server_user,
+                         server_pwd):
+    """
+    Set up selinux on remote host
+
+    :param test: the test object
+    :param selinux_state: Enforcing, or Permissive
+    :param server_ip: remote host ip
+    :param server_user: remote host user to login with
+    :param server_pwd: remote host pwd
+    """
+
+    logging.debug("Set selinux to %s on remote host before migration",
+                  selinux_state)
+
+    cmd = "setenforce %s" % selinux_state
+    status, output = run_remote_cmd(cmd, server_ip, server_user,
+                                    server_pwd)
+    if status:
+        test.error("Failed to set SELinux to %s mode on remote host" %
+                   selinux_state)
+
+
 def migrate_vm(test, params):
     """
     Connect libvirt daemon
@@ -239,6 +274,18 @@ def migrate_vm(test, params):
     for option in options.split():
         if option.startswith("--"):
             check_virsh_command_and_option(test, "migrate", option)
+
+    # Setup seliux before migration
+    logging.debug("Set selinux mode before migration")
+    selinux_state = params.get("selinux_state", "enforcing")
+    # On local:
+    setup_selinux_local(selinux_state)
+    # On remote:
+    server_ip = params.get("server_ip")
+    server_user = params.get("server_user")
+    server_pwd = params.get("server_pwd")
+    setup_selinux_remote(test, selinux_state,
+                         server_ip, server_user, server_pwd)
 
     logging.info("Prepare migrate %s", vm_name)
     global MIGRATE_RET


fangge1212:
**Before fix:**
 (03/61)   type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.disable_virt_use_nfs_remotely.with_postcopy:   FAIL: Unexpected return result:\nMigration: [ 14 %]Migration: [ 82  %]Migration: [100 %]Migration: [100 %]\n (143.05 s)

**After fix:**
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.disable_virt_use_nfs_remotely.with_postcopy: PASS (143.46 s)

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -240,6 +275,18 @@ def migrate_vm(test, params):
...
+    # Setup seliux before migration
+    logging.debug("Set selinux mode before migration")
+    selinux_state = params.get("selinux_state", "enforcing")
+    # On local:
+    setup_selinux_local(selinux_state)
+    # On remote:
+    server_ip = params.get("server_ip")
+    server_user = params.get("server_user")
+    server_pwd = params.get("server_pwd")
+    setup_selinux_remote(test, selinux_state,
```
Yingshun, line 287: I think we should restore the environment if we change the sestatus during our test. How do you think?

fangge1212: Ok, I will update later

Yingshun: Okay!

qiankehan:
This update has the assumption that the cases will not be run on the system without selinux (such as debian, ubuntu). I am not sure if the assumption is right. I tend to consider moving the selinux state setting at the ENV provision stage.

fangge1212:
> This update has the assumption that the cases will not be run on the system without selinux (such as debian, ubuntu). I am not sure if the assumption is right. I tend to consider moving the selinux state setting at the ENV provision stage.

1. For your concern, if you are setting selinux on local host, utils in aovcado-vt already handled it by:
`  95 def set_status(status, selinux_force=False):                                                                                         
  96     """                                                                                                                              
  97     Set status of selinux.                                                                                                           
......                                                                                                                 
 107     if ubuntu and not selinux_force:                                                                                                 
 108         logging.warning("Ubuntu doesn't support selinux by default")                                                                 
 109         return                                                                                                                       
 110                                                                                                                                      
`
2. I will think about how to handler this situation on remote host

3. The reason I choose to set selinux to enforcing before migration is because I found some test cases or utils in avocado-vt will set selinux to permissive secretly when they are implementing other functions and doesn't restore it afterward, which may result to migration be performed in permissive mode which is not expected

qiankehan:
> > This update has the assumption that the cases will not be run on the system without selinux (such as debian, ubuntu). I am not sure if the assumption is right. I tend to consider moving the selinux state setting at the ENV provision stage.
> 
>     1. For your concern, if you are setting selinux on local host, utils in aovcado-vt already handled it by:
>        ` 95 def set_status(status, selinux_force=False):                                                                                            96     """                                                                                                                                 97     Set status of selinux.                                                                                                            ......                                                                                                                   107     if ubuntu and not selinux_force:                                                                                                   108         logging.warning("Ubuntu doesn't support selinux by default")                                                                   109         return                                                                                                                         110                                                                                                                                      `
>

OK. No problem here.
 
>     2. I will think about how to handler this situation on remote host
> 
>     3. The reason I choose to set selinux to enforcing before migration is because I found some test cases or utils in avocado-vt will set selinux to permissive secretly when they are implementing other functions and doesn't restore it afterward, which may result to migration be performed in
permissive mode which is not expected

Since the default configuration on centos/rhel/fedora is enforcing selinux, that's the fault of these cases here because they don't restore the env. Make sure the setting are restored after every case is another possible way to resolve this issue. So here are 2 methods:
1. Initiate the settings in the cases requiring the settings
2. Restore the settings after every case

The 1st needs the case contributor clearly knows what settings may affect the test case. The 2nd needs the reviewers help check if the env is restored correctly after a case.
I am not sure if the 2nd way is better.

fangge1212:
> > > This update has the assumption that the cases will not be run on the system without selinux (such as debian, ubuntu). I am not sure if the assumption is right. I tend to consider moving the selinux state setting at the ENV provision stage.
> > 
> > 
> > ```
> > 1. For your concern, if you are setting selinux on local host, utils in aovcado-vt already handled it by:
> >    ` 95 def set_status(status, selinux_force=False):                                                                                            96     """                                                                                                                                 97     Set status of selinux.                                                                                                            ......                                                                                                                   107     if ubuntu and not selinux_force:                                                                                                   108         logging.warning("Ubuntu doesn't support selinux by default")                                                                   109         return                                                                                                                         110                                                                                                                                      `
> > ```
> 
> OK. No problem here.
> 
> > ```
> > 2. I will think about how to handler this situation on remote host
> > 
> > 3. The reason I choose to set selinux to enforcing before migration is because I found some test cases or utils in avocado-vt will set selinux to permissive secretly when they are implementing other functions and doesn't restore it afterward, which may result to migration be performed in
> > ```
> 
> permissive mode which is not expected
> 
> Since the default configuration on centos/rhel/fedora is enforcing selinux, that's the fault of these cases here because they don't restore the env. Make sure the setting are restored after every case is another possible way to resolve this issue. So here are 2 methods:
> 
>     1. Initiate the settings in the cases requiring the settings
> 
>     2. Restore the settings after every case
> 
> 
> The 1st needs the case contributor clearly knows what settings may affect the test case. The 2nd needs the reviewers help check if the env is restored correctly after a case.
> I am not sure if the 2nd way is better.

Method 2 is too risky, and it seems very hard to correct the fault in previous code..

dzhengfy:
@fangge1212 , will you update the code again?

fangge1212:
I can't think up a good solution that will work on all platforms, so I closed this pr.

PR #3150, opened 2020-11-10
Update ceph pool or image name as per validation feature

chunfuwen:
This is the step to move ceph pool name separation


chunfuwen:
those changes are just string replacement, doesn't touch any python code logic change, so no test report are needed

chunfuwen:
The change will give pool_name for individual cfg file to give upstream viewers a reminder, but also those pool or image name can be override in CI environment.

kylazhang (approved):

PR #3155, opened 2020-11-14
Skip test driver cases under split daemon mode

iccaszhulili:
test driver connection will fail under split daemon
mode because of bug 1828121

### DIFF
--- a/libvirt/tests/src/remote_access/remote_access.py
+++ b/libvirt/tests/src/remote_access/remote_access.py
@@ -8,6 +8,7 @@ from avocado.utils import process
 from virttest import remote
 from virttest import utils_iptables
 from virttest import utils_libvirtd
+from virttest import utils_split_daemons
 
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/remote_access/remote_access.py
```diff
@@ -195,6 +196,11 @@ def run(test, params, env):
     auth_unix_rw = test_dict.get("auth_unix_rw")
     kinit_pwd = test_dict.get("kinit_pwd")
 
+    if driver == "test":
+        if utils_split_daemons.is_modular_daemon():
+            test.skip("test driver connection will fail under split daemon "
+                      "mode because of bug 1828121")
```
Yingshun, line 202: @iccaszhulili  How about keeping them failed and blacklist these cases in CI?

Yingshun: s/test.skip/test.cancel

PR #3161, opened 2020-11-17
auto_dump: replace process.kill with process.terminate

chloerh:
func 'kill' of multiprocessing.Process was brought in from python3.7
therefore not availabe on python3.6. Replacing with 'terminate' could
allow test running on python3.6 without throwing out error like:
`AttributeError: 'Process' object has no attribute 'kill'`


dzhengfy (approved):
LGTM

PR #3168, opened 2020-11-23
Fix error_policy_report case failure

chunfuwen:
As per https://bugzilla.redhat.com/show_bug.cgi?id=1609701,
use virsh.domblkerror to check whether disk error report policy  works or not

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
@@ -2,6 +2,8 @@ import os
 import logging
 import aexpect
 
+from avocado.utils import process
+
 from virttest import utils_selinux
 from virttest import data_dir
 from virttest import virt_vm
@@ -265,7 +267,14 @@ def run(test, params, env):
                             s, o = session0.cmd_status_output(cmd)
                             logging.debug("session in vm0 exit %s; output: %s", s, o)
                             if error_policy == "report":
-                                if s:
+                                process.run("rm -rf %s" % disk_source, ignore_status=False, shell=True)
+                                vms_list[0]['vm'].destroy(gracefully=False)
+
+                                def _check_error():
+                                    cmd_result = virsh.domblkerror(vms_list[0]['name'])
+                                    return 'Segmentation fault' in cmd_result.stdout_text.strip()
+                                status = utils_misc.wait_for(lambda: _check_error, timeout=90)
+                                if not status:
                                     test.fail("Test error_policy %s: cann't report"
                                               " error" % error_policy)
                             elif error_policy == "ignore":


chunfuwen:
[stdout] 
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multivms.coldplug.vms_error_policy_test.error_policy_report

chunfuwen:
@lento-sun , please help take a look.

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
```diff
@@ -265,7 +267,14 @@ def get_vm_disk_xml(dev_type, dev_name, **options):
                             s, o = session0.cmd_status_output(cmd)
                             logging.debug("session in vm0 exit %s; output: %s", s, o)
                             if error_policy == "report":
-                                if s:
+                                process.run("rm -rf %s" % disk_source, ignore_status=True, shell=True)
```
dzhengfy, line 270: Do you need to set 'ignore_status' to False in order to raise exception?

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
```diff
@@ -265,7 +267,14 @@ def get_vm_disk_xml(dev_type, dev_name, **options):
...
+                                vms_list[0]['vm'].destroy(gracefully=False)
+
```
lento-sun, line 272: Here we destroy the vm first and then use a 'virsh domblkerror' to check the vm? I am not quite familar with this fuction but guess the 'domblkerror' is working on a running vm? Add @qiankehan as feature owner.

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
chunfuwen: I think not since it is one pre-caution step only.
In case disk_source is absent, ignore_status=False will cause false warning.

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
chunfuwen: report policy is a general mechanism where error can be notified in whatever situation disk error happen.
The scenario validate report can be sent out once any kind of error happen.

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_multivms.py
chunfuwen: previously hanhan ever shared one bugzilla, as per the bug 's description, the report function can not work at all. 
Here just use alternative way to verify it work.

kylazhang (approved):

PR #3173, opened 2020-11-25
Update incorrect param

dzhengfy:
The pool name was set incorrectly. This is to fix it by setting correct
default libvirt pool name.


chunfuwen (approved):
LGTM

chunfuwen:
I assume this don't need test result posted

PR #3175, opened 2020-11-26
Fix test.cancel was not thrown

dzhengfy:
Original test.cancel was not thrown by the avocado framework which is
because another invalid file removing in final cleanup phase fails. So
this is to avoid of that invalid removing step in order to throw the
original exception.


dzhengfy:
HPT case should be skipped on Power 9. It only is supported on Power 8.

Original error:
ERROR 1-type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.hpt_resize.required.with_postcopy -> TestError: Path is needed for deleting local disk

chloerh (approved):

PR #3177, opened 2020-11-26
Add test case for extended SCCB support on s390x

smitterl:
By having bigger SCCB size, diag318 data can be recorded for more
than 240 vcpus.

### DIFF
--- a/libvirt/tests/src/cpu/diagnose_data.py
+++ b/libvirt/tests/src/cpu/diagnose_data.py
@@ -1,0 +1,77 @@
+import logging
+import re
+
+from virttest.libvirt_xml.vm_xml import VMXML, VMFeaturesXML
+from virttest import virsh
+
+
+def run(test, params, env):
+    """
+    Test limits on the number of vcpus and recorded diagnose data
+
+    :param test: test object
+    :param params: Dict with the test parameters
+    :param env: Dict with the test environment
+    :return:
+    """
+    final_number_of_vcpus = int(params.get("final_number_of_vcpus"))
+    els = params.get("els")
+    diag318 = params.get("diag318")
+    check_stat = params.get("check_stat") == "yes"
+    plug = params.get("plug")
+    vm_name = params.get("main_vm")
+
+    vm = env.get_vm(vm_name)
+    vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml_backup = vmxml.copy()
+
+    try:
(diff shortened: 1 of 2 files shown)


smitterl:
```
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system diagnose_data
JOB ID     : 26ad533d06b4bf813c144b5ca6283d01e54013f5
JOB LOG    : /root/avocado/job-results/job-2020-11-26T11.42-26ad533/job.log
 (1/2) type_specific.io-github-autotest-libvirt.diagnose_data.hotplug.with_diag318: PASS (30.29 s)
 (2/2) type_specific.io-github-autotest-libvirt.diagnose_data.coldplug.with_diag318: PASS (31.87 s)
JOB TIME   : 63.83 s

```

dzhengfy (approved):
LGTM

libvirt/tests/src/cpu/diagnose_data.py
```diff
@@ -0,0 +1,77 @@
...
+    Raises an assertion error if diag 318 debug counter has only zeros
+
+    :param session: guest session
+    :raises AssertionError: if check fails
+    """
+    out = session.cmd_output('cat /sys/kernel/debug/diag_stat|grep "diag 318"')
+    logging.debug("diag_stat contains: %s", out)
+    no_zeros = out.replace("0", "").replace("diag 318", "")
+    if not re.search(r"\d", no_zeros):
+        raise AssertionError("Expected counters to be none zero, got: %s"
```
chunfuwen, line 76: Does  'raise AssertionError' usage bring more benefit than Test.Fail()? otherwise we may keep consistent method to assert results.

smitterl: @chunfuwen I don't need to pass the `test` while at the same time I can use a built-in error to have a meaning full error type. Alternatively, I could consider importing the avocado `exceptions.TestFail` and raise it here manually instead of using the helper method on `test`. It would just require an additional import. Please, let me know.

chunfuwen: In my opinion, this is related in what scope you use Assert or test.fail. If in avocado or avocado-vt scope, they are more about utilities module, I think using exception and assert make sense. But  in the tp-libvirt, it is strongly related to test case logic where we need a final judgement that case fail or pass in the end.

chunfuwen (approved):

PR #3185, opened 2020-12-03
migration: set start_vm to yes in some <test>.cfg files

fangge1212:
Some test modules require that vm is started during env_process before
test, it is better to set start_vm to yes in each <test>.cfg rather
than replying on base.cfg.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_bandwidth.cfg
+++ b/libvirt/tests/cfg/migration/migrate_bandwidth.cfg
@@ -17,6 +17,8 @@
     virsh_migrate_options = "--live --p2p --verbose"
     # Local URI
     virsh_migrate_connect_uri = "qemu:///system"
+    # Start vm in env_process before test
+    start_vm = "yes"
 
     log_outputs = "/var/log/libvirt/libvirt_daemons.log"
     diff_rate = '0.5'
--- a/libvirt/tests/cfg/migration/migrate_gluster.cfg
+++ b/libvirt/tests/cfg/migration/migrate_gluster.cfg
@@ -12,6 +12,8 @@
     image_size = "10G"
     vol_name = "vol_migrate_vm"
     pool_name = "glusterfs"
+    # Start vm in env_process before test
+    start_vm = "yes"
 
     # enable virt_use_glusterd SELinux boolean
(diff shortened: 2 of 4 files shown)


fangge1212:
Actually I don't quite understand why they need start_vm=yes, but this pr should fix the script failure

Yingshun:
I think it's caused by ci-shared-datas/-/commit/f2513d40c88c231dc13592b85c21ec2c508fa6b1.
To be honest, I think it's incorrect to set start_vm to no for whole migration jobs because all of the existing cases are based on 'yes' and I don't see any reason to make it to 'no'.
Can you please check it again in ci?

fangge1212:
> I think it's caused by ci-shared-datas/-/commit/f2513d40c88c231dc13592b85c21ec2c508fa6b1.
> To be honest, I think it's incorrect to set start_vm to no for whole migration jobs because all of the existing cases are based on 'yes' and I don't see any reason to make it to 'no'.
> Can you please check it again in ci?

In most migration jobs, there is no need to start vm before test, because we need to update vm disk to nfs storage, and in libvirt.set_vm_disk(vm, params), vm will be destroyed and started again.

fangge1212:
> I think it's caused by ci-shared-datas/-/commit/f2513d40c88c231dc13592b85c21ec2c508fa6b1.
> To be honest, I think it's incorrect to set start_vm to no for whole migration jobs because all of the existing cases are based on 'yes' and I don't see any reason to make it to 'no'.
> Can you please check it again in ci?

But it seems start_vm has other usage in env_process, it looks complicated, could you explain to me?

fangge1212:
"start_vm = yes" is consumed in preprocess() to start vm and configure vm kernel parameters, and in postprocess() to verify_guest_dmesg and remove vm kernel paramters.

I checked migration test cfg files, in some cfg files, start_vm is set to no(migrate_vm, virsh_migrate_copy_storage, migrate_ceph, etc).
If it is needed to use the above functions in preprocess() and postprocess(), then we need to remove all the "start_vm=no" in all the cfg files in tp-libvirt to remove "start_vm = no" to make the strategy consistent for all the auto tests of all the features.

PR #3189, opened 2020-12-04
migration: Cleanup vm on dest host before migration

fangge1212:
If vm with same uuid exists on dest host, migration will fail. Previous
code only considered the situation when vm with same name exists on dest
host.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_option_mix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_option_mix.py
@@ -280,8 +280,7 @@ def run(test, params, env):
         logging.debug("Clean up vm on dest host before migration")
         if dname:
             cleanup_vm(vm, dname, dest_uri)
-        else:
-            cleanup_vm(vm, vm.name, dest_uri)
+        cleanup_vm(vm, vm.name, dest_uri)
 
         # Prepare host env: set selinux state before migration
         logging.debug("Set selinux to enforcing before migration")


libvirt/tests/src/virsh_cmd/domain/virsh_migrate_option_mix.py
```diff
@@ -280,8 +280,7 @@ def cleanup_vm(vm, vm_name='', uri=''):
         logging.debug("Clean up vm on dest host before migration")
         if dname:
             cleanup_vm(vm, dname, dest_uri)
-        else:
-            cleanup_vm(vm, vm.name, dest_uri)
+        cleanup_vm(vm, vm.name, dest_uri)
```
chunfuwen, line 283: Does mean that line 281 and 282 are not necessary?

fangge1212: No, Line 281 and 282 are necessary, it is for test with --dname.
Line 283 is for both test without --dname and with --dname

fangge1212: No, Line 281 and 282 are necessary for test with --dname.
Line 283 is for all tests

chunfuwen: As per your logic, if test with -dname, cleanup_vm will be called twice. Is it that you intend to do?

fangge1212: Yes, line 282 is to clean up dest vm which has same name as "dname"; line 283 is to clean up dest vm which has same name as original name.

fangge1212: If line 283 is not executed when dname is specified, there is a chance that migration fails with error like "domain xx with same uuid exists on dest host"

kylazhang (approved):

kamilvarga:
LGTM.

Yingshun (approved):

PR #3214, opened 2020-12-17
[migrate_vm] enable kvm hpage via configuration

smitterl:
Some hosts might load kvm without hugepage support, e.g. on s390x
because hugepage support is currently incompatible with nested support.

Add logic to test cases `with_hugepages` so that kvm is reloaded with
hugepage support on source and/or target host, if configured.

Default is "no", that is the code will work as before.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_vm.cfg
+++ b/libvirt/tests/cfg/migration/migrate_vm.cfg
@@ -104,6 +104,8 @@
                             grep_str_from_local_libvirt_log = "virDomainMigrateSetCompressionCache.*cacheSize=${set_migrate_compcache_size}"
                             virsh_options = "--live --verbose --compressed"
                         - with_hugepages:
+                            enable_kvm_hugepages = "no"
+                            enable_remote_kvm_hugepages = "no"
                             setup_hugepages = "yes"
                             mb_enable = "yes"
                             config_remote_hugepages = "yes"
@@ -609,6 +611,8 @@
                             restart_vm = "yes"
                             virsh_options = "--live --verbose"
                         - with_hugepages:
+                            enable_kvm_hugepages = "no"
+                            enable_remote_kvm_hugepages = "no"
                             variants:
                                 - no_config_target:
                                     setup_hugepages = "yes"
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1233,6 +1233,8 @@ def run(test, params, env):
     mb_enable = "yes" == test_dict.get("mb_enable", "no")
     config_remote_hugepages = "yes" == test_dict.get("config_remote_hugepages",
                                                      "no")
+    enable_kvm_hugepages = "yes" == test_dict.get("enable_kvm_hugepages", "no")
+    enable_remote_kvm_hugepages = "yes" == test_dict.get("enable_remote_kvm_hugepages", "no")
     remote_tgt_hugepages = get_target_hugepage_num(test_dict)
     remote_hugetlbfs_path = test_dict.get("remote_hugetlbfs_path")
     delay = int(params.get("delay_time", 10))
@@ -1382,6 +1384,13 @@ def run(test, params, env):
 
     # Back up xml file.
     vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+
+    # Enable hugepages if necessary.
+    if enable_kvm_hugepages:
+        cmds = ["modprobe -r kvm",
+                "modprobe kvm hpage=1"]
+        for cmd in cmds:
+            process.run(cmd)
 
     # Get current VM's memory
     current_mem = vmxml_backup.current_mem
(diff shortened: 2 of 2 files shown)


smitterl:
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio virsh.migrate_vm..with_hugepages --vt-connect-uri qemu:///system
JOB ID     : f0f288f1e669d2442edd832986fb869d92bafb52
JOB LOG    : /root/avocado/job-results/job-2020-12-17T13.05-f0f288f/job.log
 (1/4) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.positive_testing.live_migration.with_hugepages.without_postcopy: PASS (175.80 s)
 (2/4) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.with_hugepages.no_config_target.without_postcopy: PASS (119.12 s)
 (3/4) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.with_hugepages.target_hugemem_less_than_vm_mem.without_postcopy: PASS (154.67 s)
 (4/4) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.with_hugepages.no_restart_target_libvirtd.without_postcopy: PASS (153.28 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 604.50 s
```

smitterl:
@dzhengfy @chloerh @chunfuwen Can you help review? Alternatively, I can make sure the hugepages are activated on remote host through CI, too. Do you prefer not to change the code and close this PR?

libvirt/tests/cfg/migration/migrate_vm.cfg
```diff
@@ -104,6 +104,8 @@
                             grep_str_from_local_libvirt_log = "virDomainMigrateSetCompressionCache.*cacheSize=${set_migrate_compcache_size}"
                             virsh_options = "--live --verbose --compressed"
                         - with_hugepages:
+                            enable_kvm_hugepages = "no"
```
chunfuwen, line 107: I assume one parameter is enough since local and remote should be same enable or disable.

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -2961,3 +2973,14 @@ def get_target_hugepage_num(params):
             port = disk_port if disk_port else uri_port[1:]
             migrate_setup.migrate_pre_setup("//%s/" % server_ip, test_dict,
                                             cleanup=True, ports=port)
+
+        cmds = ["modprobe -r kvm",
+                "modprobe kvm"]
+        if enable_kvm_hugepages:
+            for cmd in cmds:
+                process.run(cmd)
```
chunfuwen, line 2981: Better add shell=True when call process.run()

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1800,6 +1809,9 @@ def get_target_hugepage_num(params):
             cmds = ["mkdir -p %s" % remote_hugetlbfs_path,
                     "mount -t hugetlbfs none %s" % remote_hugetlbfs_path,
                     "sysctl vm.nr_hugepages=%s" % remote_tgt_hugepages]
+            if enable_remote_kvm_hugepages:
+                cmds.append("modprobe -r kvm")
+                cmds.append("modprobe kvm hpage=1")
```
dzhengfy, line 1814: sebas, is this hpage=1 required on s390x? I do not notice this is required on x86_64.

smitterl: @dzhengfy Yes, hugepages are not enabled downstream by default as they are not compatible with nesting - which isn't enabled by default.

libvirt/tests/cfg/migration/migrate_vm.cfg
smitterl: @chunfuwen Having two different parameters allows for scripting of error cases. Maybe we can keep it this way?

libvirt/tests/src/migration/migrate_vm.py
smitterl: @chunfuwen What is the reason? IIUC, shell=True is necessary to run bash/sh scripts[1], however, here we call a system command modprobe and furthermore using a shell can be exploited for shell injection[2].

[1] "If shell is True, the specified command will be executed through the shell. This can be useful if you are using Python primarily for the enhanced control flow it offers over most system shells and still want convenient access to other shell features such as shell pipes, filename wildcards, environment variable expansion, and expansion of ~ to a user’s home directory. However, note that Python itself offers implementations of many shell-like features (in particular, glob, fnmatch, os.walk(), os.path.expandvars(), os.path.expanduser(), and shutil)." - https://docs.python.org/3/library/subprocess.html
[2] https://docs.python.org/3/library/subprocess.html#security-considerations

chunfuwen: @smitterl , I just share one of my experience that  IF cmd has specical characters, it may not return expected result in practice

dzhengfy (approved):
LGTM

chunfuwen (approved):
LGTM

PR #3216, opened 2020-12-18
[virsh_migrate_copy_storage] Fix multi-disk lvm_image test cases

smitterl:
This fix depends on https://github.com/avocado-framework/avocado-vt/pull/2866

1. Fix broken tear down

If `utlv.attach_disks` fails, `cp_mig` is not defined. Therefore, teardown breaks.
As a result, the second vm (suffix "_smtest") will still exist and break the next test
if it also tries to define a new vm with the same name.

2. Improve logging

The current log can be confusing as it shows
"Bug 1249587: error: Operation not supported: pre-creation of storage targets for incremental storage migration is not supported"
although it's only checking for that value in the command output.

Prepend the log message to explicitly state it's being checked.

3. Use lvm disk type and clean up

Using 'block' worked before but will now lead to an exception in `create_local_disk`.
Use 'lvm' instead.

'block' should still worked at avocado-vt commit 12d1962
However, in commit 18fd732 the switch for 'lvm' was made explicit.

4. Fix lvm creation

Migration failed with error
'error: operation failed: migration of disk vdb failed: Input/output error'.

The destination lvm was created too small, used command:
'lvcreate -V 0.1G SMTEST --name vdb_avocado-vt-vm1_smtest --size 1M'

Use the non-sparse version of the command:
'lvcreate -L 0.1G SMTEST --name vdb_avocado-vt-vm1_smtest'

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_copy_storage.py
@@ -88,7 +88,7 @@ def check_output(test, output_msg, params):
                    "this feature or command is not currently supported"}
 
     for (key, value) in list(ERR_MSGDICT.items()):
-        logging.debug("%s: %s", key, value)
+        logging.debug("Checking expected errors - %s: %s", key, value)
         if output_msg.find(value) >= 0:
             if (key == "ERROR 1" and
                     params.get("support_precreation") is True):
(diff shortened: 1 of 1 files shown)


smitterl:
Test runs
```bash
13:47:26 (1/6) virsh.migrate_copy_storage.normal_test.single_disk.copy_storage_inc.file_image Result: PASS 123.34 s
13:49:30 (2/6) virsh.migrate_copy_storage.normal_test.single_disk.copy_storage_all.file_image Result: PASS 113.77 s
13:51:26 (3/6) virsh.migrate_copy_storage.normal_test.multi_disks.copy_storage_inc.file_image Result: PASS 135.80 s
13:53:43 (4/6) virsh.migrate_copy_storage.normal_test.multi_disks.copy_storage_inc.lvm_image Result: PASS 170.31 s
13:56:35 (5/6) virsh.migrate_copy_storage.normal_test.multi_disks.copy_storage_all.file_image Result: PASS 136.88 s
13:58:53 (6/6) virsh.migrate_copy_storage.normal_test.multi_disks.copy_storage_all.lvm_image Result: PASS 169.52 s
```

chunfuwen (approved):
LGTM

PR #3218, opened 2020-12-21
[migrate_options_shared] reboot guest on dest

smitterl:
Networking is not guaranteed to work after migration of NAT setup without reboot.
Reboot the vm before pinging it.

### DIFF
--- a/libvirt/tests/src/migration/migrate_options_shared.py
+++ b/libvirt/tests/src/migration/migrate_options_shared.py
@@ -1461,6 +1461,9 @@ def run(test, params, env):
             remote_virsh_session.close_session()
 
         if int(mig_result.exit_status) == 0:
+            if not remote_virsh_session:
+                remote_virsh_session = virsh.VirshPersistent(**remote_virsh_dargs)
+                remote_virsh_session.reboot(vm_name)
             migration_test.ping_vm(vm, params, dest_uri)
 
             if cmd_in_vm_after_migration:


smitterl:
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio virsh.migrate_options_s
hared.positive_test.p2p_migration.check_domstats.with_postcopy --vt-connect-uri qemu:///system
JOB ID     : 5de099d75fbdfa260c97f82e84afa8ba32b1ca76
JOB LOG    : /root/avocado/job-results/job-2020-12-21T04.00-5de099d/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.positive_test.p2p_migration.check_domstats.with_postcopy: PASS
(172.86 s)
JOB TIME   : 175.01 s
```

Yingshun:
Checking network accessibility is a mandatory step after migration in this py file for now. We need to check if the migrated VM is working well or not and, here, we check network accessibility via migration.ping_vm. And after that, we may need to check other things, such as logs, some command lines in VM or other something else. If it's a problem of getting IP from arpping cmd, I recommend  getting VM's IP from src before migration or extending timeout of getting the migrated VM's IP from destination.If none of them works, we may need to replace the network check with another basic functional check.

smitterl:
How about we check if the machine is running on destination? Do you know
why we're checking for IP first of all?

Yingshun:
> How about we check if the machine is running on destination? 
I think it's good.

dzhengfy:
@smitterl @Yingshun Checking network of migrated vm on target machine is the basic requirement to make sure the vm is accessible. We also have other checkpoints covered in other cases.  If dev has confirmed that Networking is not guaranteed to work after migration of NAT setup without reboot, it is acceptable to do reboot. Here I propose below：
1. use ip got from source vm to ping target vm at first, if fail then
2. reboot target vm and ping again. It seems we need not ti reboot at any time.
What is your opinion？ 
@fangge1212 as feature owner, how di you think about？

Yingshun:
@dzhengfy I think checking network connection is not the onvirsh.is_alive(vm_name, uri=dest_urily way here. Actually I agree with @smitterl , I think checking the status of the migrated vm is fine. How about just checking virsh.is_alive(vm_name, uri=dest_uri)?

fangge1212:
If NAT network is not guaranteed to work after migration, I think we can remove this checkpoint.
Anyway, we have other specific test cases for migration with various network devices, so this checkpoint can be covered in the specific test cases instead of all the migration test cases

smitterl:
@fangge1212 @Yingshun @dzhengfy I think it's risky to remove the checkpoint from the automation script (or substitute it by is_alive(...)) in case any other checks in some test cases do require network connectivity. I am not familiar with all test cases. Therefore, personally I would want to keep the reboot logic here in the script to stabilize the test. If NATs are expected to work after reboot, that's then covered, too. But I'm fine with whatever you prefer, if we first run all the tests to make sure removing the checkpoint doesn't break any tests. What do you think?

fangge1212:
Hi @smitterl 
I'm also not familiar with all the test cases. If you need to keep the reboot logic to stabilize the test, that's OK. But I think we'd better move the reboot and network check logic to the end of all the checkpoints. Because other checkpoints may should be executed before vm is reboot.

smitterl:
%MCEPASTEBIN%

> Hi @smitterl
> I'm also not familiar with all the test cases. If you need to keep the reboot logic to stabilize the test, that's OK. But I think we'd better move the reboot and network check logic to the end of all the checkpoints. Because other checkpoints may should be executed before vm is reboot.

@fangge1212 Right, I hadn't thought of other checkpoints that become invalid if rebooting first. OK; I think we shall go with first trying to remove checkpoint as you mention before then. Thanks.

libvirt/tests/src/migration/migrate_options_shared.py
```diff
@@ -1461,6 +1461,9 @@ def cancel_bg_migration():
             remote_virsh_session.close_session()
 
         if int(mig_result.exit_status) == 0:
+            if not remote_virsh_session:
+                remote_virsh_session = virsh.VirshPersistent(**remote_virsh_dargs)
+                remote_virsh_session.reboot(vm_name)
```
dzhengfy, line 1466: @smitterl After discussion with @Yingshun and @fangge1212 , we decided to remove ping_vm operation and use console login instead of ssh login in the following lines. Network connection checking will be covered in specific migration cases for network feature. @Yingshun will submit a pr for this change. After that, could you agree to close this PR?

Yingshun: Done, please check https://github.com/autotest/tp-libvirt/pull/3218. Thanks!

PR #3223, opened 2020-12-23
Update bridge type interface with network type xml

yalzhang:
When update an <interface>, and the interface type changed from type='network'
where the network was an unmanaged bridge (so actualType == bridge) to type='bridge'
(i.e. actualType also == bridge), the update without the recently fix would fail
due to the perceived change in type. Add the test case about updating bridge type
interface with network type xml.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_bridge.py
+++ b/libvirt/tests/src/virtual_network/iface_bridge.py
@@ -13,6 +13,7 @@ from virttest.staging import service
 from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml.devices import interface
 from virttest import utils_package
+from virttest.libvirt_xml import network_xml
 
 NETWORK_SCRIPT = "/etc/sysconfig/network-scripts/ifcfg-"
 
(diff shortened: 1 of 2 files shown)


yalzhang:
$ avocado run --vt-type libvirt virtual_network.iface_bridge.update_with_diff_type.shared_physical_network
JOB ID     : 79a33b37f161fafb2b689bedae638396cc47e513
JOB LOG    : /root/avocado/job-results/job-2020-12-23T08.47-79a33b3/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_bridge.update_with_diff_type.shared_physical_network: PASS (67.01 s)
JOB TIME   : 69.06 s

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -52,6 +53,15 @@ def create_bridge(br_name, iface_name):
               ' pkill dhclient; sleep 6; dhclient {0}; ifconfig {1} 0"'.format(br_name, iface_name)
         process.run(cmd, shell=True, verbose=True)
 
+    def create_bridge_network(br_name, net_name):
```
chunfuwen, line 56: Does the create_bridge_xml in libvirt.py satisfy your request?

libvirt/tests/src/virtual_network/iface_bridge.py
```diff
@@ -173,10 +185,12 @@ def check_net_functions(guest_ip, ping_count, ping_timeout, guest_session, host_
                                              ignore_status=True)
             else:
                 logging.info("Try to hot-plug device")
+                if create_network:
+                    create_bridge_network(bridge_name, iface_source["network"])
```
chunfuwen, line 189: Better check whether bridge_name exist or not before call create_xx().

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: No, actually I don't know why there is "ip = utils_net.get_ip_address_by_interface(iface)" in the create_bridge_xml. And it will return ip. Bridge type network do not need an ip.

libvirt/tests/src/virtual_network/iface_bridge.py
yalzhang: The "bridge_name exist or not" has been checked when creating the bridge in the function "create_bridge". But this remind me that before create the network, I need to check if the network with the same name already exists. I have updated it. Thank you!

chunfuwen (approved):
LGTM

PR #3227, opened 2020-12-28
migration: fixup for migration with rng devices

fangge1212:
This commit includes 2 fixes:
1. Remove native_tls from test matrix
2. Add step to remove rng device from vm xml before start vm in case
vm already has a rng device.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_with_virtual_devices.cfg
+++ b/libvirt/tests/cfg/migration/migrate_with_virtual_devices.cfg
@@ -6,10 +6,6 @@
         - postcopy:
             postcopy = True
     variants:
-        - plaintext:
-        - native_tls:
-            native_tls = True
-    variants:
         - rng:
             hotplug = "yes"
             device_type = "rng"
--- a/libvirt/tests/src/migration/migrate_with_virtual_devices.py
+++ b/libvirt/tests/src/migration/migrate_with_virtual_devices.py
@@ -197,6 +197,9 @@ class MigrationWithRng(MigrationVirtualDevicesBase):
             self._create_tcp_rng_source(self.migrate_source_host,
                                         mode='server')
 
+        # Remove rng devices from vm xml defination
+        self._remove_rng_from_vmxml(self.main_vm)
+
     def _post_start_vm(self):
         # Hotplug device
         if self.hotplug == "yes":
@@ -343,6 +346,18 @@ class MigrationWithRng(MigrationVirtualDevicesBase):
     def func_kill_rng_source():
         cmd = 'pkill -9 nc'
         process.run(cmd, ignore_status=True)
+
+    def _remove_rng_from_vmxml(self, vm):
+        """
+        Remove rng device from domain xml defination
+
+        :param vm: vm object
+        """
+        logging.debug("Remove rng device from domain xml defination")
+
+        vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+        vmxml.remove_all_device_by_type("rng")
+        vmxml.sync()
 
     def wait_for_check_rng_in_dumpxml(self, vm, xml_set, rng_present,
                                       timeout=120):


fangge1212:
Before fix:
avocado run --vt-type libvirt --vt-machine-type q35 migrate.virtual_devices.rng.backend_builtin.plaintext.precopy --vt-connect-uri qemu:///system
JOB ID     : e0e039dd5865818e17f176d8076e8723bbd3829b
JOB LOG    : /root/avocado/job-results/job-2020-12-28T01.04-e0e039d/job.log
 (1/1) type_specific.io-github-autotest-libvirt.migrate.virtual_devices.rng.backend_builtin.plaintext.precopy: ERROR: Error in phase POST_MIGRATE_BACK:\nDevice xml is not removed from domain xml (423.60 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 425.04 s

After fix:
 avocado run --vt-type libvirt --vt-machine-type q35 migrate.virtual_devices.rng.backend_builtin.plaintext.precopy --vt-connect-uri qemu:///system
JOB ID     : 7c7a45796a705a77b591c45fe5240bc0a5f80ea3
JOB LOG    : /root/avocado/job-results/job-2020-12-28T02.26-7c7a457/job.log
 (1/1) type_specific.io-github-autotest-libvirt.migrate.virtual_devices.rng.backend_builtin.plaintext.precopy: PASS (303.60 s)
JOB TIME   : 305.03 s

libvirt/tests/src/migration/migrate_with_virtual_devices.py
```diff
@@ -344,6 +347,20 @@ def func_kill_rng_source():
         cmd = 'pkill -9 nc'
         process.run(cmd, ignore_status=True)
 
+    def remove_rng_from_vmxml(self, vm):
```
chunfuwen, line 350: remove_rng_from_vmxml to _remove_rng_from_vmxml?

libvirt/tests/src/migration/migrate_with_virtual_devices.py
```diff
@@ -344,6 +347,20 @@ def func_kill_rng_source():
...
+
+        :param vm: vm object
+        """
+        logging.debug("Remove rng device from domain xml defination")
+
+        virsh_instance = virsh.Virsh(uri=vm.connect_uri)
+        vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name,
+                                              virsh_instance=virsh_instance)
+        vmxml.remove_all_device_by_type("rng")
+        vmxml.sync()
```
Yingshun, line 362: I think vmxml.sync() may fail if the vm is on remote host, I mean vm.connect_uri is point to remote uri. 
A dumped xml file will be located in local but the vm is on remote, it may be an issue I think.
could you please check it again?

libvirt/tests/cfg/migration/migrate_with_virtual_devices.cfg
```diff
@@ -5,10 +5,6 @@
         - precopy:
         - postcopy:
             postcopy = True
-    variants:
-        - plaintext:
-        - native_tls:
-            native_tls = True
```
Yingshun, line 11: Would you mind explaining why you delete these lines?

libvirt/tests/src/migration/migrate_with_virtual_devices.py
fangge1212: A lot of other functions will also fail if vm is on remote host. I can do nothing for this.

libvirt/tests/cfg/migration/migrate_with_virtual_devices.cfg
fangge1212: After reconsidering, I think native_tls don't need to be in common test matrix.

libvirt/tests/src/migration/migrate_with_virtual_devices.py
fangge1212: ok

fangge1212: updated

libvirt/tests/src/migration/migrate_with_virtual_devices.py
Yingshun: Yes, we have some modules that only work with localhost, but there are also many modules that support remote run.
I think the easiest way is to copy  the dumped xml to remote and sync xml  with virsh.  What do you think?

chunfuwen: > Yes, we have some modules that only work with localhost, but there are also many modules that support remote run.
> I think the easiest way is to copy the dumped xml to remote and sync xml with virsh. What do you think?

@fangge1212 , please reply yingshun's comment, we may move this PR forward since it is pending for long time

fangge1212: Hi @Yingshun 
Could you give me an example? So I can save some time to write the code from scratch

Yingshun: You can follow below steps:
Firstly, using remote.scp_to_remote to copy the xml to a remote host. Next, calling virsh with uri to sync up the xml.

dzhengfy: @fangge1212 @Yingshun , 
If _remove_rng_from_vmxml() is only used before starting vm on source host, then we currently do not need to consider remote, right?

Yingshun: If we don't need to consider remote operations, how about deleting L358 to avoid unnecessary risks?

fangge1212: Maybe change it to "virsh_instance = virsh.Virsh()"? I forgot the usage totally.

Yingshun: We don't need to pass it I think, it has a default value.

fangge1212: updated

Yingshun (approved):

chunfuwen (approved):
lgtm

dzhengfy (approved):
LGTM

PR #3233, opened 2021-01-05
To support cross migration from 8.2 to 8.3, the qemu cmdline has some change

jinqi7:
Since the qemu command changed from 8.2 to 8.3, the different comparation strings
is needed for 8.2 as the source of the migration.

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_mem.cfg
+++ b/libvirt/tests/cfg/migration/migrate_mem.cfg
@@ -48,7 +48,8 @@
             memdev_tg_size = "512"
             memdev_tg_sizeunit = "MiB"
             memdev_tg_node = "0"
-            qemu_checks = "id=ram-node1,size=536870912`-numa node,nodeid=1,cpus=2-3,memdev=ram-node1`-object memory-backend-ram,id=memdimm0,size=536870912"
+            # migrate from 8.2 to 8.3.1, the qemu cmdline is "-numa node,nodeid=1,cpus=2-3,mem=1024 -object memory-backend-ram,id=memdimm0,size=536870912"
+            qemu_checks = "(id=ram-node1,size=536870912|)`-numa node,nodeid=1,cpus=2-3,mem(=1024|dev=ram-node1)`-object memory-backend-ram,id=memdimm0,size=536870912"
         - mem_balloon:
             check = "mem_balloon"
             ballooned_mem = "716800"


chunfuwen:
please update commit message ,and better have test result pasted

jinqi7:
avocado run --vt-type libvirt virsh.migrate_mem.mem_device
JOB ID : 33a4d036ea4af360f6185745e6ba766e441afa9b
JOB LOG : /root/avocado/job-results/job-2020-12-30T05.17-33a4d03/job.log
(1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_mem.mem_device: PASS (128.76 s)
RESULTS : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

jinqi7:
> please update commit message ,and better have test result pasted

Updated. @chunfuwen , please help to review it again ? Thanks

libvirt/tests/cfg/migration/migrate_mem.cfg
```diff
@@ -29,7 +29,7 @@
     variants:
         - mem_device:
             check = "mem_device"
-            cpuxml_cpu_mode = "host-model"
+            cpuxml_cpu_mode = "custom"
```
Yingshun, line 32: Can you explain why this change is needed? I don't understand the purpose of this change...

fangge1212: I don't suggest modifying vm cpu mode in single migration test case, as cpu mode is set at job level(for migration job).
For now, we set cpu  mode/model to host-model/qemu64 to avoid cpu compatibility issue between src and dest host(because it is hard to borrow two hosts with compatible cpu in ci).
In RHEL9, qemu64 will be deprecated, we need to figure out a new way to set cpu mode. If we set cpu mode in every single test case, we will have to maintain them separately.

jinqi7: Removed the change.

fangge1212: As host-model is usedin this case, and ci can't promise to borrow two same hosts, the case always failed in auto job. Please fix this in new pr

fangge1212: And I don't think it is reasonable to set cpu in memory test case

jinqi7: @fangge1212  Which one is  correct from your two comments ?
> For now, we set cpu mode/model to host-model/qemu64 to avoid cpu compatibility issue between src and dest host(because it is hard to borrow two hosts with compatible cpu in ci)
> As host-model is used in this case, and ci can't promise to borrow two same hosts, the case always failed in auto job.

In this case, the cpu mode is also host-model and qemu64 as fallback.

fangge1212: I mean we set cpu mode/model to host-model/qemu64 to avoid cpu compatibility issue between src and dest host at job level instead of in each test case. 
In this test case, host-model takes effect actually, qemu64 is not used. So this case always failed in ci job. 
I suggest removing the cpu config part in this tesr case.

jinqi7: Can you give the domain xml abou the "host-model/qemu64" that you used?

jinqi7: From other cases, I saw the xml is like "<cpu check="none" match="exact" mode="custom">    <model fallback="forbid">qemu64</model>". right?

jinqi7: The format of xml can't be print well,  and the content is just like:
cpu check="none" match="exact" mode="custom"   
model fallback="forbid" qemu64  /model

Yingshun: May I ask why you modify cpu mode in this script? There's no step to set cpu mode in polarion: RHEL-175191, so if it's unnecessary, I agree with @fangge1212  .

jinqi7: Since I don't know what cpu mode is set in other migration test cases and where it's set by default. I am asking it now.

jinqi7: @fangge1212 , please answer the question above, thx

jinqi7: As I talked with fangge offline, the cpu mode in use is "custom/qemu64" , not "host-model/qemu64".   
And since  qemu64  will deprecated in rhel9,  the cpu mode change will be implemented later. 
The PR only fix the issue of migrating from 8.2 to 8.3.  Please help to review @Yingshun

Yingshun (approved):

fangge1212 (approved):

libvirt/tests/cfg/migration/migrate_mem.cfg
```diff
@@ -48,7 +48,8 @@
             memdev_tg_size = "512"
             memdev_tg_sizeunit = "MiB"
             memdev_tg_node = "0"
-            qemu_checks = "id=ram-node1,size=536870912`-numa node,nodeid=1,cpus=2-3,memdev=ram-node1`-object memory-backend-ram,id=memdimm0,size=536870912"
+            # migrate from 8.2 to 8.3.1, the qemu cmdline is "-numa node,nodeid=1,cpus=2-3,mem=1024 -object memory-backend-ram,id=memdimm0,size=536870912"
+            qemu_checks = "(id=ram-node1,size=536870912|)`-numa node,nodeid=1,cpus=2-3,mem(=1024|dev=ram-node1)`-object memory-backend-ram,id=memdimm0,size=536870912"
```
dzhengfy, line 52: @fangge1212 @jinqi7 @Yingshun , as we discussed about qemu command line checking before, is this checking still required any more?

jinqi7: @dzhengfy, it's the only checkpoint for memory device migration.

dzhengfy (approved):
LGTM

PR #3237, opened 2021-01-06
Check whether pkg is successfully installed

chloerh:
ppc tests requires pkg ndctl. The installation of the pkg might
fail sometimes. Stop testing if pkg cannot be installed.


chunfuwen (approved):
LGTM

PR #3247, opened 2021-01-13
nvdimm:check minimum target size when defining vm

chloerh:
The old behavior is throwing out error when starting vm. Currently
defining vm would fail. The case needs update to move the check
ahead to defining step.

### DIFF
--- a/libvirt/tests/src/memory/nvdimm.py
+++ b/libvirt/tests/src/memory/nvdimm.py
@@ -149,7 +149,10 @@ def run(test, params, env):
                          for k, v in params.items() if k.startswith('nvdimmxml_')}
         nvdimm_xml = create_nvdimm_xml(**nvdimm_params)
         vmxml.add_device(nvdimm_xml)
-        if check in ['ppc_no_label', 'discard']:
+        check_define_list = ['ppc_no_label', 'discard']
+        if libvirt_version.version_compare(7, 0, 0):
+            check_define_list.append('less_than_256')
+        if check in check_define_list:
             result = virsh.define(vmxml.xml, debug=True)
             libvirt.check_result(result, expected_fails=[error_msg])
             return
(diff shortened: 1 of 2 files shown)


dzhengfy (dismissed):
Code changes LGTM, but please help test results for sure. Thanks.

libvirt/tests/src/memory/nvdimm.py
```diff
@@ -168,12 +168,6 @@ def check_nvdimm_file(file_name):
             target_size = vm_nvdimm_xml.target.size
             logging.debug('Target size: %s', target_size)
 
-            if check == 'less_than_256':
```
Yingshun, line 171: I’m just wondering if it works for all version. do we need to think about backward compatibility?

chloerh: Do you mean for older version of libvirt, we should keep the original check? That's a good point. @dzhengfy What do you think?

dzhengfy: Yes, that's a good catch. Please keep the compatibilty.

chloerh:
Before fix:

```
 (1/1) type_specific.io-github-autotest-libvirt.nvdimm.back_file.label.invalid_target_size.less_than_256: ERROR: Failed to define avocado-vt-vm1 for reason:\nerror: Failed to define domain from /tmp/xml_utils_temp_gq9uj486.xml\nerror: XML error: minimum target size for the NVDIMM must be 256MB plus the label size\n (5.92 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 6.58 s
```

After fix:
```
 (1/1) type_specific.io-github-autotest-libvirt.nvdimm.back_file.label.invalid_target_size.less_than_256: PASS (5.56 s)
JOB TIME   : 6.25 s
```

libvirt/tests/src/memory/nvdimm.py
chloerh: @Yingshun @dzhengfy updated

Yingshun (approved):

dzhengfy (approved):

PR #3296, opened 2021-01-31
[code review guide] extend review guide

smitterl:
1. remove a tag from domain definition and define from scratch instead of updating existing
2. make sure comments are still valid
3. wait don't sleep

### DIFF
--- a/tp-libvirt_review_comment_summary.rst
+++ b/tp-libvirt_review_comment_summary.rst
@@ -8,14 +8,15 @@ Overview
 
 This doc summarize the most common review comments in tp-libvirt open source project.
 
-For beginners or practice executor, it may provide some self-check before submitting to normal review in github.
+On one hand it may serve as a check list for reviewers but also for coders to make sure their PR meets expected quality level before submission.
 
 ======================================================================
 Goal
 ======================================================================
 
 1. Provide one place to accumulate knowledge.
-2. Provide some help to beginners or executors in coding or style format for tp-libvirt.
+2. Provide help to coders in expected coding or style format for tp-libvirt.
+3. Provide a reference for reviewers.
 
 ======================================================================
 Common issues
@@ -72,6 +73,11 @@ Coding
 
  - libvirt_vm.check_exit_status
 
+- comments
+
+ - if a test is added or modified and there's a comment - usually for the 'run' function - then we should make sure
+   the comment still is valid
+ 
 - resource cleanup
 
  - vm_connection_session close(exception gracefully close)
@@ -108,6 +114,8 @@ Enhancement (best practice):
 - use :code:`with` to open files
 - either use global constants for timeouts or test :code:`params` to set values for timeouts (e.g. :code:`wait_for_loging(..., timeout=LOGIN_TIMEOUT)`);
   this way it is easier for others to tweak timeouts on slower systems
+- wait don't sleep: try to avoid using time.sleep(...); instead try waiting for a condition to hold (utils_misc.wait_for) - this speeds up testing and enables us to use larger timeout
+  values without necessarily increasing test duration
 - make sure this is run before sending patch::
 
     inspekt checkall --disable-style E501,E265,W601,E402,E722,E741 --no-license-check <test-script-name>.py


smitterl:
@fangge1212

tp-libvirt_review_comment_summary.rst
```diff
@@ -72,6 +72,17 @@ Coding
 
  - libvirt_vm.check_exit_status
 
+- domain manipulation
+ 
+ - if an xml element is to be updated by the test, consider first removing it fully and defining it from scratch (this
```
chunfuwen, line 77: I am a little concerning about this comment since from the customer scenarios view, customer generally update what he or she intend to do, rather than remove it firstly.

tp-libvirt_review_comment_summary.rst
```diff
@@ -72,6 +72,17 @@ Coding
...
+   helps avoid test failures due to slightly different guest xmls) - avocado-vt offers for example the method
+   vmxml.xmltreefile.remove_by_xpath
+
+- comments
+
+ - if a test is added or modified and there's a comment - usually for the 'run' function - then we should make sure
```
chunfuwen, line 83: I agree this. Maybe we can further  point out what are "we" .e.g reviewer, author, PR merger ?

tp-libvirt_review_comment_summary.rst
```diff
@@ -108,6 +119,8 @@ Enhancement (best practice):
 - use :code:`with` to open files
 - either use global constants for timeouts or test :code:`params` to set values for timeouts (e.g. :code:`wait_for_loging(..., timeout=LOGIN_TIMEOUT)`);
   this way it is easier for others to tweak timeouts on slower systems
+- wait don't sleep: avoid using time.sleep(...); instead try waiting for a condition to hold (utils_misc.wait_for) - this speeds up testing and enables us to use larger timeout
```
chunfuwen, line 122: I totally agree

tp-libvirt_review_comment_summary.rst
smitterl: > I am a little concerning about this comment since from the customer scenarios view, customer generally update what he or she intend to do, rather than remove it firstly.

@chunfuwen Can we predict how customers might have their xmls? Please note that this recommendation is not about "testing updating a domain xml" but about stabilizing functional test flow from breaking because of slightly different test cases. The cases I target here is to prevent breakages that are fixed by removing elements, e.g.
https://github.com/autotest/tp-libvirt/pull/2531
https://github.com/autotest/tp-libvirt/pull/2273

Maybe I should choose another wording?

tp-libvirt_review_comment_summary.rst
smitterl: @chunfuwen I believe what 'we' is intended to mean is already part of the Overview and Goal section:
"For beginners or practice executor, it may provide some self-check before submitting to normal review in github."
"Provide some help to beginners or executors in coding or style format for tp-libvirt."

Do you think those sections should be updated, something like "On one hand it may serve as a check list for reviewers but also for coders to make sure their PR meets expected quality level before submission."

chunfuwen:
@kylazhang 
@dzhengfy 
loop them into this review

tp-libvirt_review_comment_summary.rst
chunfuwen: > @chunfuwen I believe what 'we' is intended to mean is already part of the Overview and Goal section:
> "For beginners or practice executor, it may provide some self-check before submitting to normal review in github."
> "Provide some help to beginners or executors in coding or style format for tp-libvirt."
> 
> Do you think those sections should be updated, something like "On one hand it may serve as a check list for reviewers but also for coders to make sure their PR meets expected quality level before submission."?

I prefer  "Do you think those sections should be updated, something like "On one hand it may serve as a check list for reviewers but also for coders to make sure their PR meets expected quality level before submission"

tp-libvirt_review_comment_summary.rst
kylazhang: I also have concern for it, as some update maybe only an attribute and do not related to much to its current settings, just add an extra items in its current xml section, why need to remove them all? As for running successfully you may also need to get the information of current xml then there will be no difference with just updating it. Especially for different arch it also different setting for the xml, and totally generate a new xml section may means give specific setting in conf or just hard code it, which should be more complex

tp-libvirt_review_comment_summary.rst
kylazhang: May be we can not avoid it but as long as wait_for can replace, we will use wait_for instead

kylazhang (commented):
Please check comments, thanks

tp-libvirt_review_comment_summary.rst
chunfuwen: @smitterl , as per past experiences, there are some missed bugs related to xml which was not covered by qemuxml2args . The density of qemuxml2args is in the hand of developer, and  how often they are executed is out of control by QE too. So from QE's pointview, it provides one more tier xml validation.

tp-libvirt_review_comment_summary.rst
dzhengfy: Agree with @kylazhang

smitterl: @kylazhang @dzhengfy Do you want me to change this part?

tp-libvirt_review_comment_summary.rst
smitterl: @kylazhang @chunfuwen Are you requesting to remove this paragraph?

chunfuwen: To avoid confusion it may bring up,  Better remove it firstly

tp-libvirt_review_comment_summary.rst
kylazhang: Yes, use wait_for as long as it can replace sleep. But we can not avoid using sleep.

tp-libvirt_review_comment_summary.rst
kylazhang: Yes, as the real situation is complex and it is hard to say how to update is more stable

kylazhang (commented):
please check comments inline, thanks

dzhengfy (approved):
LGTM

chunfuwen (approved):
LGTM

kylazhang (approved):

PR #3302, opened 2021-02-02
v2v: clean vms variant for v2v_options job

xiaodwan:
The vms requires guest 'avocado-vt-vm1' which is not used by v2v.
It's annoying when running it on local host manually.


xiaodwan:
A minor change won't have any side effect to other components.

PR #3304, opened 2021-02-03
Remove tests of input devices with ps2 bus from pseries tests

chloerh:
ps2 is not supported by pseries.


dzhengfy (approved):
LGTM

PR #3323, opened 2021-02-10
Set multiple_controller_ppc tests to only run on pseries

chloerh:


dzhengfy (approved):
I am surprised that we have not done this before.

PR #3330, opened 2021-02-22
max_vcpus: Skip virsh_maxvcpus on aarch64

liuyd96:
libvirt gets the maximum from the kernel, so the output of maxvcpu is
not consistent among multi-archs.

Skip maxvcpus tests on aarch64 since the default report_num is
incorrect for aarch64.


liuyd96:
@chloerh , thanks.

chloerh (approved):

PR #3340, opened 2021-02-24
Fix the unplug failure

yalzhang:
As there is a patch recently to add the alias into account when detach
an interface device. The original detach device will fail as it has
alias set as 'None', which can pass in the old libvirt which do not
care about the alias, but fail in current libvirt version. Add a customer
alias into the cfg file to fix the issue. It should pass both in old and
latest libvirt.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_bridge.cfg
@@ -39,6 +39,7 @@
                 - hotplug_device:
                     hotplug = "yes"
                     iface_type = "bridge"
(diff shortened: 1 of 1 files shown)


yalzhang:
run locally pass
$avocado run --vt-type libvirt virtual_network.iface_bridge.multiqueues.hotplug_device.shared_physical_network
JOB ID     : 35657fadfab19b05c331c084214ebf8439d12d46
JOB LOG    : /root/avocado/job-results/job-2021-02-24T02.39-35657fa/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_bridge.multiqueues.hotplug_device.shared_physical_network: PASS (75.11 s)
JOB TIME   : 76.84 s

chloerh (approved):

PR #3349, opened 2021-03-01
migration: Use a parameter in base.cfg to enable VM migration back

Yingshun:
One of the migration jobs needs to enable VM migration back for all
cases, so update the parameter to migrate_vm_back that already
exists in base.cfg.


Yingshun:
@fangge1212  Would you please help review this PR? Thanks!

Yingshun:
There is no code logic change, so no need to provide a test result. Could you please help review it?

kamilvarga (approved):
I don't fully understand why this change was necessary, but it seems to be applied in all places, so LGTM.

fangge1212 (approved):

dzhengfy (approved):
LGTM

PR #3352, opened 2021-03-02
Remove incorrect skip aarch64 cases

liuyd96:
Some virtual devices are not supported by RHEL qemu-kvm but upstream
qemu-system-aarch64 (I checked on qemu-4.2.0-7.fc32) supports them.

So skip these cases on tp-libvirt upstream is incorrect, remove these
incorrect skip cases.


chunfuwen (approved):
LGTM

chunfuwen:
no result is needed due to filter result

PR #3356, opened 2021-03-04
virsh_blockpull: Fix an error while converting raw

Yingshun:
It reports "File exists" when converting data sometimes. Update to
use a random image name and remove it before conversion.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockpull.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockpull.py
@@ -8,6 +8,7 @@ from avocado.utils import process
 from virttest import virsh
 from virttest import data_dir
 from virttest import utils_libvirtd
+from virttest import utils_misc
 from virttest import utils_package
 from virttest import libvirt_storage
 from virttest import ceph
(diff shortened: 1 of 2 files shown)


Yingshun:
I had a similar change in blockcommit(b4fb091edc9a) that worked well.

```
JOB ID     : aa049e001eb202f35e6a03a3f3090d69c810bafc
JOB LOG    : /root/avocado/job-results/job-2021-03-03T21.29-aa049e0/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.blockpull.normal_test.network_disk.ceph.notimeout.nobase.async: PASS (352.45 s)
JOB TIME   : 353.21 s
```

libvirt/tests/src/virsh_cmd/domain/virsh_blockpull.py
```diff
@@ -299,6 +300,17 @@ def check_chain_backing_files(disk_src_file, expect_backing_file=False):
...
                 if src_host.count("EXAMPLE") or mon_host.count("EXAMPLE"):
                     test.cancel("Please provide ceph host first.")
+
+                params.update(
+                   {"disk_source_name": os.path.join(
+                      pool_name,
+                      'rbd_blockpull_' + utils_misc.generate_random_string(4) +
+                      '.img')})
+                if utils_package.package_install(["ceph-common"]):
+                    ceph.rbd_image_rm(
```
chunfuwen, line 310: 1）if it is random image name , I assume ceph.rbd_image_rm is not necessary since High probability  image name does not exist considering the fact that image name is created firstly by random string
2)If it is fixed image name, call ceph.rbd_image_rm before convert can remove the risk of converting failure

Yingshun: If multiple jobs use the same name, it usually occurs a conflict.  So I'd like to use the random name and early deletion together to avoid this kind of issue. Early deletion seems like unnecessary, but I've met it one time, so I think we'd better to add this part.

PR #3364, opened 2021-03-08
[virtio-transitional][s390x] add no filter on s390x

smitterl:
On s390x all disks (and more generally most devices) are
attached not as PCI but as CCW. Virtio (non-)transitional is PCI-only.

Therefore, filter out those test cases for s390x.


chunfuwen (approved):
LGTM

chunfuwen:
no results is needed due to filter cases

PR #3368, opened 2021-03-09
controller_functional: remove multifunction test case for s390x

smitterl:
Multifunction is PCI related. On s390x SCSI works over channel system
that doesn't have this concept. Filter those test cases out.


dzhengfy (approved):
LGTM

dzhengfy:
Exclude the case, so no test result is needed.

PR #3392, opened 2021-03-25
Delete the host interface management related tests

yalzhang:
virsh_iface_bridge.py, virsh_iface_trans.py and virsh_iface_edit.py can
manage the host interface. As 1)it is not used in uplayer, 2)it may not function
well for a long time since the depercation of network-scripts. 3)The commands
will not be supported since rhel 9, 4)Most of the cases are added into skiplist
as they may not safe to update the host network during auto testrun, with all the
above reasons, delete these script in tp-libvirt.


yalzhang:
@kylazhang  Please evaluate it, Thank you!

chunfuwen:
for depreciated feature, we may keep test scripts there.
But we can remove them in jobs.yaml

yalzhang:
> for depreciated feature, we may keep test scripts there.
> But we can remove them in jobs.yaml

OK, I will close this PR.

PR #3397, opened 2021-03-25
Update the check point about libvirtd crash

yalzhang:
Currently when libvirtd crash, it will restart. So if we want to
check if there was a libvirtd crash, check the libvirtd status if
not enough, we should also check the pid of libvirtd.

### DIFF
--- a/libvirt/tests/src/daemon/crash_regression.py
+++ b/libvirt/tests/src/daemon/crash_regression.py
@@ -54,9 +54,15 @@ def run_restart_save_restore(params, libvirtd, vm):
     Save and restore a domain after restart daemon.
     """
     libvirtd.restart()
+    ori_libvirtd_pid1 = process.run("pidof libvirtd").stdout_text
     save_path = os.path.join(data_dir.get_tmp_dir(), 'tmp.save')
     virsh.save(vm.name, save_path)
     virsh.restore(save_path)
+    if libvirtd.was_running:
+        aft_libvirtd_pid1 = process.run("pidof libvirtd").stdout_text
+        if ori_libvirtd_pid1 != aft_libvirtd_pid1:
+            logging.debug("libvirtd crash after vm save and restore!")
+            return 1
 
 
 def post_restart_save_restore(params, libvirtd, vm):
@@ -236,6 +242,7 @@ def run(test, params, env):
     """
     Run various regression tests and check whether libvirt daemon crashes.
     """
+    global re
     func_name = 'run_' + params.get("func_name", "default")
     post_func_name = 'post_' + params.get("func_name", "default")
     repeat = int(params.get("repeat", "1"))
@@ -255,10 +262,13 @@ def run(test, params, env):
                 shell=True, ignore_status=True)
     try:
         libvirtd.start()
+        ori_pid_libvirtd = process.run("pidof libvirtd").stdout_text
 
         run_func = globals()[func_name]
         for i in xrange(repeat):
-            run_func(params, libvirtd, vm)
+            re = run_func(params, libvirtd, vm)
+            if re:
+                test.fail("libvirtd crashed when execute the function %s" % func_name)
 
         stopped = libvirtd.wait_for_stop(timeout=5)
         if stopped:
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/daemon/crash_regression.py
```diff
@@ -54,9 +54,15 @@ def run_restart_save_restore(params, libvirtd, vm):
     Save and restore a domain after restart daemon.
     """
     libvirtd.restart()
+    ori_libvirtd_pid1 = process.getoutput("pidof libvirtd")
     save_path = os.path.join(data_dir.get_tmp_dir(), 'tmp.save')
     virsh.save(vm.name, save_path)
     virsh.restore(save_path)
+    if libvirtd.was_running:
+        aft_libvirtd_pid1 = process.getoutput("pidof libvirtd")
```
chunfuwen, line 62: We suggest use process.run() to get command ouput. 
why Process.getouput exists is to just  to keep code compatability.

yalzhang: Update, Thank you!

chunfuwen: it sounds like you miss the change for process.getoutput

libvirt/tests/src/daemon/crash_regression.py
```diff
@@ -54,9 +54,15 @@ def run_restart_save_restore(params, libvirtd, vm):
     Save and restore a domain after restart daemon.
     """
     libvirtd.restart()
+    ori_libvirtd_pid1 = process.run("pidof libvirtd").stdout_text
```
Yingshun, line 57: We'd better consider the situation when the split daemon is enabled, so I recommend that you use libvirtd.service_exec here...

libvirt/tests/src/daemon/crash_regression.py
```diff
@@ -54,9 +54,15 @@ def run_restart_save_restore(params, libvirtd, vm):
...
     save_path = os.path.join(data_dir.get_tmp_dir(), 'tmp.save')
     virsh.save(vm.name, save_path)
     virsh.restore(save_path)
+    if libvirtd.was_running:
+        aft_libvirtd_pid1 = process.run("pidof libvirtd").stdout_text
```
Yingshun, line 62: Same as above..

libvirt/tests/src/daemon/crash_regression.py
```diff
@@ -236,6 +242,7 @@ def run(test, params, env):
     """
     Run various regression tests and check whether libvirt daemon crashes.
     """
+    global re
```
Yingshun, line 245: May I ask why this line is added?

libvirt/tests/src/daemon/crash_regression.py
```diff
@@ -255,10 +262,13 @@ def run(test, params, env):
                 shell=True, ignore_status=True)
     try:
         libvirtd.start()
+        ori_pid_libvirtd = process.run("pidof libvirtd").stdout_text
 
         run_func = globals()[func_name]
         for i in xrange(repeat):
-            run_func(params, libvirtd, vm)
+            re = run_func(params, libvirtd, vm)
```
Yingshun, line 269: How about changing 're' to more meaningful name? like 'result', 'res'. I was thinking it as 're' module when I first saw it😅

qiankehan: > How about changing 're' to more meaningful name? like 'result', 'res'. I was thinking it as 're' module when I first saw itsweat_smile

Agree. ret make more sense

qiankehan:
NACK. How do you know the libvirtd pid update is caused by libvirtd crash? It could be caused by libvirtd resrtart.
I think implementing **coredumpctl** with time range (_--since_ or _--until_) is better

qiankehan:
What's more, libvirtd could crash in any cases. It's hard to prove that it is more likely to crash in these cases rather than the others.
I suggest to add this checking in avocado-vt and then apply to all the tp-libvirt cases.

yalzhang:
As we have talked to use abrt to catch the libvirtd crash, close this PR

PR #3401, opened 2021-03-29
Add the bug fixed version into the error message

yalzhang:
Add the bug fixed version into the error message, it can reduce the
debugging time.


chunfuwen:
There is no logic change, doesn't need require test result

chunfuwen (approved):
LGTM

PR #3423, opened 2021-04-08
controller_functional: fix host-model missing

dzhengfy:
'host-model' is recommended to use in <cpu> configuration in vm xml.
Otherwise, vm might fail to start under some circumstances. So this
pr is to add it to the guest xml.

### DIFF
--- a/libvirt/tests/src/controller/controller_functional.py
+++ b/libvirt/tests/src/controller/controller_functional.py
@@ -855,8 +855,15 @@ def run(test, params, env):
             address_params = {'bus': "%0#4x" % int(pci_bus_number), 'slot': "%0#4x" % int(pci_bus_number)}
             libvirt.set_disk_attr(vm_xml, 'vda', 'address', address_params)
         if cpu_numa_cells:
-            vmxml_cpu = VMCPUXML()
-            vmxml_cpu.xml = "<cpu><numa/></cpu>"
+            if not vm_xml.cpu:
+                vmxml_cpu = VMCPUXML()
+                vmxml_cpu.xml = "<cpu mode='host-model'><numa/></cpu>"
+            else:
+                vmxml_cpu = vm_xml.cpu
+                logging.debug("Existing cpu configuration in guest xml:\n%s", vmxml_cpu)
+                vmxml_cpu.mode = 'host-model'
+                vmxml_cpu.remove_elem_by_xpath('/model')
+                vmxml_cpu.remove_elem_by_xpath('/numa')
             vmxml_cpu.numa_cell = VMCPUXML.dicts_to_cells(eval(cpu_numa_cells))
             vm_xml.cpu = vmxml_cpu
             vm_xml.vcpu = int(params.get('vcpu_count', 4))


dzhengfy:
Before the fix:
```
# avocado run --vt-type libvirt --vt-machine-type q35 controller.functional.positive_tests.pcie_expander_bus_numa --vt-connect-uri qemu:///system
2021-02-25 03:20:31,180 stacktrace       L0045 ERROR|     raise VMIPAddressMissingError(mac, ip_version)
2021-02-25 03:20:31,180 stacktrace       L0045 ERROR| virttest.virt_vm.VMIPAddressMissingError: No ipv4 DHCP lease for MAC 52:54:00:8e:51:61
```

dzhengfy:
After the fix when guest xml has cpu model 'custom' or 'host-model':
```
# avocado run --vt-type libvirt --vt-machine-type q35 controller.functional.positive_tests.pcie_expander_bus_numa
JOB ID     : 5e45b81066319250a3e904c463d57f255327b85a
JOB LOG    : /root/avocado/job-results/job-2021-04-08T02.05-5e45b81/job.log
 (1/1) type_specific.io-github-autotest-libvirt.controller.functional.positive_tests.pcie_expander_bus_numa: PASS (43.27 s)
JOB TIME   : 43.96 s

```

dzhengfy:
depend on https://github.com/avocado-framework/avocado-vt/pull/2993

libvirt/tests/src/controller/controller_functional.py
```diff
@@ -855,8 +855,15 @@ def check_multifunction():
...
-            vmxml_cpu.xml = "<cpu><numa/></cpu>"
+            if not vm_xml.cpu:
+                vmxml_cpu = VMCPUXML()
+                vmxml_cpu.xml = "<cpu mode='host-model'><numa/></cpu>"
+            else:
+                vmxml_cpu = vm_xml.cpu
+                logging.debug("Existing cpu configuration in guest xml:\n%s", vmxml_cpu)
+                vmxml_cpu.mode = 'host-model'
+                vmxml_cpu.remove_elem_by_xpath('/model')
+                vmxml_cpu.remove_elem_by_xpath('/numa')
```
kamilvarga, line 866: Why the cpu model and numa elements are removed? Isn't safer to just replace existing cpu configuration in the guest by an empty one as it was done previously with just a mode='host-model' defined?

chunfuwen:
Quote "'host-model' is recommended to use in configuration in vm xml.
Otherwise, vm might fail to start under some circumstances. So this
pr is to add it to the guest xml."
1)Do we already confirm with feature team  that this is their recommendation ,rather than resolve issue that non-host model VM fail launch ?
2)If host model is recommended, it sound like CI is a good place-in to provision VM with host-model at the beginning, otherwise we may find other python files may face similar changes

libvirt/tests/src/controller/controller_functional.py
dzhengfy: Removing numa is to avoid conflicts in case there is already a numa setting in the guest xml before the test.
Removing model is to avoid conflicts in case there is already a <model >setting in the guest xml before the test.
For  example, if I have below guest xml
  <cpu mode='custom' match='exact' check='none'>
    <model fallback='forbid'>qemu64</model>
  </cpu>
After vmxml_cpu.mode = 'host-model', then libvirt will save it as below, that's not my needs. So I would like to remove it.
  <cpu mode='host-model' check='none'>
    <model fallback='forbid'>qemu64</model>
  </cpu>

But it is optional because when starting the vm, libvirt will remove <model> and replace with below.
<cpu mode='custom' match='exact' check='full'>
    <model fallback='forbid'>Skylake-Server-IBRS</model>
    <vendor>Intel</vendor>
    <feature policy='require' name='ss'/>
    <feature policy='require' name='vmx'/>

dzhengfy:
> Quote "'host-model' is recommended to use in configuration in vm xml.
> Otherwise, vm might fail to start under some circumstances. So this
> pr is to add it to the guest xml."
> 1)Do we already confirm with feature team that this is their recommendation ,rather than resolve issue that non-host model VM fail launch ?
> 2)If host model is recommended, it sound like CI is a good place-in to provision VM with host-model at the beginning, otherwise we may find other python files may face similar changes

Answers:
1> Yes, I have confirmed with feature owner that we recommend to test with host-model in RHEL8 and later.
2> Our CI already supports to use 'host-model' in predefined guest xml. But in some special test cases, we need to configure special items about cpu, like numa setting. To make sure 'host-model' used, we have to explicitly set this value. And this PR also intends to adapt  to different configurations about <cpu>, not only limited to our CI environment.

dzhengfy:
@kamilvarga could you help double review?

libvirt/tests/src/controller/controller_functional.py
kamilvarga: Got It, it LGTM then.

kamilvarga (approved):

chunfuwen (approved):
lgtm

chunfuwen:
let's merge it

PR #3424, opened 2021-04-08
svirt: skip some tests for aarch64

liuyd96:
aarch64 guest use firmware efi which requires nvram file.
E.g
```
<os>
  <type arch='aarch64' machine='virt-rhel8.4.0'>hvm</type>
  <loader readonly='yes' type='pflash'>/usr/share/edk2/aarch64/QEMU_EFI-silent-pflash.raw</loader>
  <nvram>/var/lib/libvirt/qemu/nvram/avocado-vt-vm1_VARS.fd</nvram>
  <boot dev='hd'/>
</os>
```
Some tests set user/group as qemu and disable dynamic_ownership in qemu.conf.

There is a permission conflict when we disable dynamic_ownership and set qemu
user/group as qemu:
Run qemu process as user qemu but nvram belongs to root.

So skip these tests for aarch64

Signed-off-by: Liu <liuyd.fnst@fujitsu.com>

### DIFF
--- a/libvirt/tests/cfg/svirt/dac_nfs_save_restore.cfg
+++ b/libvirt/tests/cfg/svirt/dac_nfs_save_restore.cfg
@@ -32,6 +32,7 @@
             dynamic_ownership = "no"
     variants:
         - qemu_user:
+            no aarch64
         - root_user:
             qemu_user = "root"
             qemu_group = "root"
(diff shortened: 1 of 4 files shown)


liuyd96:
@chloerh thanks

chloerh (approved):
Though I think it's a little bit 'not pretty', but I don't know another way of doing so. So I'll approve

PR #3433, opened 2021-04-13
Add wait_for_event for change-media operation

kylazhang:

Depends on https://github.com/avocado-framework/avocado-vt/pull/2985

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
@@ -294,7 +294,12 @@ def run(test, params, env):
                 source = device_source
             all_options = action_twice + options_twice + " " + source
             time.sleep(5)
+            if options_twice == "--config" or pre_vm_state == "shutoff":
+                wait_for_event = False
+            else:
+                wait_for_event = True
             ret = virsh.change_media(vm_ref, target_device, all_options,
+                                     wait_for_event=wait_for_event,
                                      ignore_status=True, debug=True)
             status_error = False
             if pre_vm_state == "shutoff":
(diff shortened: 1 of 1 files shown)


kylazhang:
During my test seems also depens on https://github.com/avocado-framework/avocado-vt/pull/3000 to make the running pass

libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
```diff
@@ -294,7 +294,7 @@ def check_result(vm_name, disk_source, disk_type, disk_target,
                 source = device_source
             all_options = action_twice + options_twice + " " + source
             time.sleep(5)
-            ret = virsh.change_media(vm_ref, target_device, all_options,
+            ret = virsh.change_media(vm_ref, target_device, all_options, wait_for_event=True,
```
chunfuwen, line 297: I assume we need pass event_type in if enable wait_for_event

kylazhang: I say your previous patch event_type = 'tray-change' which is the default one in change_media command param. So I think no need to set here

kylazhang:
I have test on 506 cases for all change_media_matrix and change_media cases, and all passed

kylazhang:
@chunfuwen I do some update with conditional add event wait as config change-media and shutdown guest change-media will not have event. you can have a check to see if logic is correct. Thanks

chloerh (approved):

libvirt/tests/src/virsh_cmd/domain/virsh_change_media_matrix.py
```diff
@@ -294,7 +294,12 @@ def check_result(vm_name, disk_source, disk_type, disk_target,
                 source = device_source
             all_options = action_twice + options_twice + " " + source
             time.sleep(5)
+            if options_twice == "--config" or pre_vm_state == "shutoff":
```
chunfuwen, line 297: @kylazhang 
Previously I figure out min set of impacted cases(cover failure cases), from safety point view, suggest just copy that filter
```
            wait_change_event = False
            if options in ["--force", "--current", "--config"] and pre_vm_state == "paused" and action_twice in ["--update "]:
                wait_change_event = True
```

chunfuwen:
#https://github.com/avocado-framework/avocado-vt/pull/3007
create_disk is moved to libvirt_utils.py as new API

chunfuwen (approved):
lgtm

PR #3458, opened 2021-04-23
v2v: add a new case for weak denpendency of libguestfs-xfs

xiaodwan:
Add a new case for weak dependency of libguestfs-xfs.

### DIFF
--- a/v2v/tests/src/v2v_options.py
+++ b/v2v/tests/src/v2v_options.py
@@ -15,6 +15,7 @@ from avocado.utils.astring import to_text
 
 from virttest import data_dir
 from virttest import utils_misc
+from virttest import utils_package
 from virttest import utils_v2v
 from virttest import utils_sasl
 from virttest import virsh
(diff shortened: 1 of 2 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.v2v_options.positive_test.cmd_free.weak_dependency.input_mode.none.output_mode.none: PASS (6.46 s)

v2v/tests/src/v2v_options.py
```diff
@@ -886,6 +886,8 @@ def check_alloc():
     finally:
         if hypervisor == "esx":
             process.run("rm -rf %s" % vpx_passwd_file)
+        if checkpoint == "weak_dendency":
+            process.run("yum install libguestfs-xfs -y", ignore_status=True)
```
chunfuwen, line 890: we usually add shell=True when call process.run to avoid special characters issue

xiaodwan: @chunfuwen I changed it to utils_package.package_install function.

xiaodwan:
Merged this PR since the change request were modified.

PR #3462, opened 2021-04-25
Add p2p migration case for with_migrate_disks

Yingshun:
RHEL-201640: [Migration][--migrate-disks] Specify the disks to be
    migrated for storage migration - p2p

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_vm.cfg
+++ b/libvirt/tests/cfg/migration/migrate_vm.cfg
@@ -546,6 +546,19 @@
                                 - part_non_shared_disks:
                                     status_error = "yes"
                                     virsh_options = "--live --verbose --copy-storage-all --migrate-disks vda"
+                        - with_migrate_disks_p2p:
+                            setup_nfs = "yes"
+                            enable_virt_use_nfs = "yes"
+                            migrate_disks = "yes"
+                            attach_A_disk_source = "${nfs_mount_src}/test1.qcow2"
+                            attach_B_disk_source = "${nfs_mount_dir}/test2.qcow2"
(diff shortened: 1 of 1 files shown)


Yingshun:
no code changes, only copy a live migration case, so no need to provide a test result.

libvirt/tests/cfg/migration/migrate_vm.cfg
```diff
@@ -546,6 +546,19 @@
                                 - part_non_shared_disks:
                                     status_error = "yes"
                                     virsh_options = "--live --verbose --copy-storage-all --migrate-disks vda"
+                        - with_migrate_disks_p2p:
```
dzhengfy, line 549: Is it better to reuse existing '- with_migrate_disks:' to avoid many duplicated lines?
For example,
 - with_migrate_disks:
                  variants:
                                - all_non_shared_disks:
                                variants:
                                      - @default:
                                         virsh_options = "--live --verbose --copy-storage-all --migrate-disks vda,vdb"
                                      - p2p:
                                        virsh_options = "--live --p2p --verbose --copy-storage-all --migrate-disks vda,vdb"

Yingshun: Yes, the change will be small If I do so, but then, I have to update libvirt-ci, caselink and Jira manually because the patterns of the previous cases are changed. Actually, I have several cases to add p2p cases, just like this pr does, so I'd like to keep it. How do think? Is it acceptable?

dzhengfy: Understand.

dzhengfy (approved):
LGTM

PR #3483, opened 2021-05-11
Fix the vm can not login issue

yalzhang:
In cases 'cold_plug.nfv..', we need to set the numa node for the vm.
So there is a function add_numa(), which will delete the original cpu
setting and initiate the cpu part to '<cpu><numa/></cpu>' which ignored
the cpu mode setting, it will default to 'qemu64', which has bug
on some releases. Workaround it to be 'host-model' which is the original
vm setting.


chloerh (approved):

chloerh:
@dzhengfy It seems to be the 'host-model' issue. Would you like to take a look?

chunfuwen (approved):
There is similar fix seen previously on other cases

dzhengfy (approved):
LGTM

PR #3484, opened 2021-05-12
Add 1 network case of mtu

chloerh:
- RHEL-198271: [pre-created tap] Set mtu for pre-created tap device
and start vm with another mtu setting and managed='no'

### DIFF
--- a/libvirt/tests/src/virtual_network/mtu.py
+++ b/libvirt/tests/src/virtual_network/mtu.py
@@ -6,6 +6,7 @@ import shutil
 from avocado.utils import process
 
 from virttest import data_dir
+from virttest import libvirt_version
 from virttest import utils_misc
 from virttest import utils_net
 from virttest import utils_package
@@ -39,6 +40,12 @@ def run(test, params, env):
     model = params.get('model', 'virtio')
     timeout = int(params.get('timeout', 240))
     wait_for_up = int(params.get('wait_for_up', 0))
+    create_tap = 'yes' == params.get('create_tap', 'no')
+    tap_mtu_size = int(params.get('tap_mtu_size', 2000))
+
+    if create_tap and 'managed_no' in params['name']:
+        if not libvirt_version.version_compare(7, 0, 0):
+            test.cancel('This test is not supported until libvirt-7.0.0')
 
     def set_network(size, net='default'):
         """
@@ -51,22 +58,27 @@ def run(test, params, env):
         logging.debug(virsh.net_dumpxml(net))
 
     def set_interface(mtu_size='', source_network='default',
-                      iface_type='network', iface_model='virtio'):
+                      iface_type='network', iface_model='virtio',
+                      iface_target=None):
         """
         Set mtu size to a certain interface
         """
         interface_type = 'bridge' if iface_type in ('bridge', 'openvswitch') else iface_type
         iface_dict = {
             'type': interface_type,
-            'source': "{'%s': '%s'}" % (interface_type, source_network),
             'model': iface_model
         }
+        if source_network:
+            iface_dict.update({'source': "{'%s': '%s'}" % (interface_type, source_network)})
 
         if iface_type == 'openvswitch':
             iface_dict.update({'virtualport_type': 'openvswitch'})
 
         if mtu_size:
             iface_dict.update({'mtu': "{'size': %s}" % mtu_size})
+
+        if iface_target:
+            iface_dict.update({'target': iface_target})
 
         libvirt.modify_vm_iface(vm_name, 'update_iface', iface_dict)
         logging.debug(virsh.dumpxml(vm_name).stdout)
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -51,23 +54,28 @@ def set_network(size, net='default'):
...
         """
         Set mtu size to a certain interface
         """
         interface_type = 'bridge' if iface_type in ('bridge', 'openvswitch') else iface_type
         iface_dict = {
             'type': interface_type,
-            'source': "{'%s': '%s'}" % (interface_type, source_network),
             'model': iface_model
         }
+        if source_network is not None:
```
kamilvarga, line 67: This if condition can be simplified: "if source_network:" instead of "if source_network is not None:"

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -51,23 +54,28 @@ def set_network(size, net='default'):
...
+            iface_dict.update({'source': "{'%s': '%s'}" % (interface_type, source_network)})
 
         if iface_type == 'openvswitch':
             iface_dict.update({'virtualport_type': 'openvswitch'})
 
         if mtu_size:
             iface_dict.update({'mtu': "{'size': %s}" % mtu_size})
 
+        if iface_target is not None:
```
kamilvarga, line 76: Can be simplified, please see my comment above.

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -183,7 +191,53 @@ def check_mtu_in_vm(fn_login, mtu_size, timeout):
             process.run('ovs-vsctl add-br %s' % br, shell=True, verbose=True)
             process.run('ovs-vsctl show', shell=True, verbose=True)
 
-        if not check or check in ['save', 'managedsave', 'hotplug_save']:
+        if create_tap:
+            tap_name = 'tap_' + utils_misc.generate_random_string(3)
+            tap_cmd = params.get('tap_cmd')
+            if tap_cmd is None:
+                test.error('No tap creating command provided.')
+            tap_cmd = tap_cmd.format(tap_name=tap_name)
```
kamilvarga, line 199: Shouldn't be a "mtu_size" provided to tap_cmd as well?

libvirt/tests/src/virtual_network/mtu.py
chloerh: You're right! Updated.

libvirt/tests/src/virtual_network/mtu.py
chloerh: Updated

libvirt/tests/src/virtual_network/mtu.py
chloerh: @kamilvarga Yes it should, but I did it by putting them in mtu.cfg, line:45~46
```
                    tap_mtu_size = 2000
                    tap_cmd = 'ip tuntap add mode tap name {tap_name};ip link set dev {tap_name} mtu ${tap_mtu_size}'
```
Then tap_mtu_size gets into tap_cmd. 

Here's one piece of log of the test case
```
    tap_cmd = ip tuntap add mode tap name {tap_name};ip link set dev {tap_name} mtu 2000
    tap_mtu_size = 2000
```
So this way works.

chloerh:
Currently host mtu size check failed due to bug. But after removing that checkpoint, cases passed.
```
t, please install python-imaging or the equivalent for your distro.
JOB ID     : 4882462ff6d9a423cb3ff7bac3b563bc977856f1
JOB LOG    : /root/avocado/job-results/job-2021-05-26T23.52-4882462/job.log
 (1/2) type_specific.io-github-autotest-libvirt.virtual_network.mtu.positive_test.tap.managed_no.smaller: PASS (43.14 s)
 (2/2) type_specific.io-github-autotest-libvirt.virtual_network.mtu.positive_test.tap.managed_no.larger: PASS (42.30 s)
JOB TIME   : 85.87 s
```

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -183,7 +191,52 @@ def check_mtu_in_vm(fn_login, mtu_size, timeout):
...
+                    test.cancel('This test is not supported until libvirt-7.0.0')
+                iface_target = params.get('iface_target') % tap_name
+                set_interface(mtu_size, source_network=None,
+                              iface_type='ethernet', iface_target=iface_target)
+                vm.start()
+                logging.debug(virsh.dumpxml(vm_name).stdout)
+
+                # Check mtu of tap on host
+                host_iface_info = process.run(ip_link_cmd % tap_name, verbose=True).stdout_text
+                if 'mtu %s' % tap_mtu_size in host_iface_info:
```
Yingshun, line 218: I think it'll be great if you define a function for these lines. There is a similar check in L231~L237. What do you think

libvirt/tests/src/virtual_network/mtu.py
kamilvarga: Got it, thanks.

kamilvarga (approved):
LGTM.

Yingshun (approved):

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -183,7 +191,52 @@ def check_mtu_in_vm(fn_login, mtu_size, timeout):
...
+            tap_cmd = tap_cmd.format(tap_name=tap_name)
+            logging.debug('Tap creating command: \n %s', tap_cmd)
+            # Create tap device
+            process.run(tap_cmd, verbose=True, shell=True)
+            # Check tap device's detail
+            ip_link_cmd = 'ip link show %s'
+            process.run(ip_link_cmd % tap_name, verbose=True)
+
+            if 'managed_no' in params['name']:
+                if not libvirt_version.version_compare(7, 0, 0):
```
dzhengfy, line 208: It is better to check libvirt version at the beginning of the test. What do you think?

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -183,7 +191,52 @@ def check_mtu_in_vm(fn_login, mtu_size, timeout):
             process.run('ovs-vsctl add-br %s' % br, shell=True, verbose=True)
             process.run('ovs-vsctl show', shell=True, verbose=True)
 
-        if not check or check in ['save', 'managedsave', 'hotplug_save']:
+        if create_tap:
```
dzhengfy, line 194: Considering the run() function is already very long, the readability is coming worse. So I suggest we put this tap device testing case into several functions with meaningful names which will make run() shorter and easy to read. For example, setup_tap_device(),  check_mtu_info() ... What do you think?

libvirt/tests/src/virtual_network/mtu.py
chloerh: Updated.

libvirt/tests/src/virtual_network/mtu.py
chloerh: Thanks. But I'd like to keep the way it is if you don't mind. Creating multiple function seems not quite necessary right now. I'll keep it in mind when trying to re-format this test.

dzhengfy (approved):

PR #3493, opened 2021-05-18
[vfio_ccw] Add ccw passthrough tests

smitterl:
Depends on https://github.com/avocado-framework/avocado-vt/pull/3064

1. Test device is visible inside guest.
2. Test device is removed gracefully when all channel paths
   are removed.

### DIFF
--- a/libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
+++ b/libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
@@ -1,0 +1,7 @@
+- libvirt_ccw_passthrough:
+    type = libvirt_ccw_passthrough
+    only s390-virtio
+    variants:
+        - happy_path:
+        - device_removal:
+            device_removal_case = yes
--- a/libvirt/tests/src/libvirt_ccw_passthrough.py
+++ b/libvirt/tests/src/libvirt_ccw_passthrough.py
@@ -1,0 +1,186 @@
+import logging
+
+from uuid import uuid4
+
+from avocado.core.exceptions import TestError
+
+from virttest import utils_package
+from virttest import virsh
+from virttest.utils_zchannels import SubchannelPaths, ChannelPaths
+from virttest.utils_misc import cmd_status_output
+from virttest.libvirt_xml.devices.hostdev import Hostdev
+from virttest.libvirt_xml.vm_xml import VMXML
+
+
+def device_is_listed(session, chpids):
+    """
+    Checks if the css device is listed by comparing the channel
+    path ids.
+
+    :param session: guest console session
+    :param chipds: chpids where the disk is connected, e.g. "11122122"
+    """
+
+    paths = SubchannelPaths(session)
+    paths.get_info()
+    devices_inside_guest = [x for x in paths.devices
+                            if x[paths.HEADER["CHPIDs"]] == chpids]
+    return len(devices_inside_guest) > 0
+
+
+def set_override(schid):
+    """
+    Sets the driver override for the device' subchannel.
+
+    :param schid: Subchannel path id for device.
+    """
+
+    cmd = "driverctl -b css set-override %s vfio_ccw" % schid
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise TestError("Can't set driver override. %s" % out)
+
+
+def unset_override(schid):
+    """
+    Unsets the driver override for the device' subchannel.
+
+    :param schid: Subchannel path id for device.
+    """
+
+    cmd = "driverctl -b css unset-override %s" % schid
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise TestError("Can't set driver override. %s" % out)
+
+
+def start_device(uuid, schid):
+    """
+    Starts the mdev with mdevctl
+
+    :param uuid: device uuid
+    :param schid: subchannel id for the device
+    """
+
+    cmd = "mdevctl start -u %s -p %s -t vfio_ccw-io" % (uuid, schid)
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise TestError("Can't start mdev. %s" % out)
+
+
+def stop_device(uuid):
+    """
+    Stops the mdev with mdevctl
+
+    :param uuid: device uuid
+    """
+
+    cmd = "mdevctl stop -u %s" % (uuid)
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        logging.warning("Couldn't stop device. %s", out)
+
+
+def attach_hostdev(vm_name, uuid):
+    """
+    Attaches the mdev to the machine.
+
+    :param vm_name: VM name
+    :param uuid: mdev uuid
+    """
+
+    hostdev_xml = Hostdev()
+    hostdev_xml.mode = "subsystem"
+    hostdev_xml.model = "vfio-ccw"
+    hostdev_xml.type = "mdev"
+    hostdev_xml.source = hostdev_xml.new_source(**{"uuid": uuid})
+    hostdev_xml.xmltreefile.write()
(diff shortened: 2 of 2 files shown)


smitterl:
Test run:
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system libvirt_ccw_passthrough
JOB ID     : 908d02b4daf300e15855b8c055b72d2ff3e4b347
JOB LOG    : /root/avocado/job-results/job-2021-05-18T11.32-908d02b/job.log
 (1/2) type_specific.io-github-autotest-libvirt.libvirt_ccw_passthrough.happy_path: PASS (39.83 s)
 (2/2) type_specific.io-github-autotest-libvirt.libvirt_ccw_passthrough.device_removal: PASS (46.60 s)
JOB TIME   : 86.89 s
```

kamilvarga (approved):
LGTM.

libvirt/tests/src/libvirt_ccw_passthrough.py
```diff
@@ -0,0 +1,186 @@
...
+                        ignore_status=False)
+
+
+def assure_preconditions():
+    """
+    Makes sure that preconditions are established.
+    """
+
+    utils_package.package_install("mdevctl")
+    utils_package.package_install("driverctl")
```
dzhengfy, line 108: List parameter is supported, so  better to use `utils_package.package_install(['mdevctl', 'driverctl'])`

dzhengfy (commented):
Other codes LGTM.
Another suggestion:
To make the tp-libvirt/src (cfg) clear structure, it is better to put your cases into a new directory like 'ccw' or 'passthrough'. I know we have already many cases under src/cfg directly, but I hope we can improve this step by step. What's your opinion?

libvirt/tests/src/libvirt_ccw_passthrough.py
smitterl: Thanks for pointing that out @dzhengfy ! I changed it.

smitterl:
> Other codes LGTM.
> Another suggestion:
> To make the tp-libvirt/src (cfg) clear structure, it is better to put your cases into a new directory like 'ccw' or 'passthrough'. I know we have already many cases under src/cfg directly, but I hope we can improve this step by step. What's your opinion?

@dzhengfy I agree this makes sense but I prefer not to do this as part of this PR but in a new one because this will affect more than just my changes.

I propose I'll open another PR after this one has been merged with the following structure for passthrough tests
```bash
- passthrough
+---/ccw/passthrough_ccw*.(py|cfg)
+---/pci/passthrough_pci*.(py|cfg)
+---/pci/vfio.py
+---/usb/usb_passthrough.(py|cfg)
```
Alternatively, I can add that commit in this PR but would like to not squash commits into one in this case as is required by this repository's code rules. What do you think?

libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
```diff
@@ -0,0 +1,7 @@
+- libvirt_ccw_passthrough:
```
chunfuwen, line 1: Can we put those files under subfolder:virtual_devices, rather than under cfg directly?

libvirt/tests/src/libvirt_ccw_passthrough.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Sets the driver override for the device' subchannel.
+
+    :param schid: Subchannel path id for device.
+    """
+
+    cmd = "driverctl -b css set-override %s vfio_ccw" % schid
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise TestError("Can't set driver override. %s" % out)
```
chunfuwen, line 41: raise TestError  change to test.error(xx) ?

smitterl: @chunfuwen I want to avoid having to pass `test` parameter. Especially because later these functions will have to be extracted from this file, see future change waiting for this to be merged: https://github.com/smitterl/tp-libvirt/pull/2

libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
smitterl: @chunfuwen Please, check discussion with @dzhengfy on comment above - https://github.com/autotest/tp-libvirt/pull/3493#issuecomment-861342924

dzhengfy:
> > Other codes LGTM.
> > Another suggestion:
> > To make the tp-libvirt/src (cfg) clear structure, it is better to put your cases into a new directory like 'ccw' or 'passthrough'. I know we have already many cases under src/cfg directly, but I hope we can improve this step by step. What's your opinion?
> 
> @dzhengfy I agree this makes sense but I prefer not to do this as part of this PR but in a new one because this will affect more than just my changes.
> 
> I propose I'll open another PR after this one has been merged with the following structure for passthrough tests
> 
> ```shell
> - passthrough
> +---/ccw/passthrough_ccw*.(py|cfg)
> +---/pci/passthrough_pci*.(py|cfg)
> +---/pci/vfio.py
> +---/usb/usb_passthrough.(py|cfg)
> ```
> 
> Alternatively, I can add that commit in this PR but would like to not squash commits into one in this case as is required by this repository's code rules. What do you think?

@smitterl  That new structure seems good to me.

dzhengfy (approved):

libvirt/tests/src/libvirt_ccw_passthrough.py
chunfuwen: @smitterl 
This is common agreement that test.xx() is used in tp-libvirt project, and raise XX is used in avocado-vt since early stage of those projects

libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
chunfuwen: @smitterl 
I notice the above mentioned comment.
But isn't it(/libvirt/tests/src/virtual_device)a  proper folder since ccw is category of virtual devices

chunfuwen (commented):
see inline comments

chunfuwen (approved):
just leave minor issue , just LGTM

libvirt/tests/cfg/libvirt_ccw_passthrough.cfg
smitterl: I believe vfio are not virtual devices. IIUC, vsock, specific watchdogs etc. are virtual devices because they only exist in the qemu/kvm stack and not on real hardware. However, vfio are drivers for real hardware to be used by the guests as if directly attached.

smitterl:
> > > Other codes LGTM.
> > > Another suggestion:
> > > To make the tp-libvirt/src (cfg) clear structure, it is better to put your cases into a new directory like 'ccw' or 'passthrough'. I know we have already many cases under src/cfg directly, but I hope we can improve this step by step. What's your opinion?
> > 
> > 
> > @dzhengfy I agree this makes sense but I prefer not to do this as part of this PR but in a new one because this will affect more than just my changes.
> > I propose I'll open another PR after this one has been merged with the following structure for passthrough tests
> > ```shell
> > - passthrough
> > +---/ccw/passthrough_ccw*.(py|cfg)
> > +---/pci/passthrough_pci*.(py|cfg)
> > +---/pci/vfio.py
> > +---/usb/usb_passthrough.(py|cfg)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Alternatively, I can add that commit in this PR but would like to not squash commits into one in this case as is required by this repository's code rules. What do you think?
> 
> @smitterl That new structure seems good to me.

Cool. Implemented in https://github.com/autotest/tp-libvirt/pull/3566

PR #3494, opened 2021-05-19
v2v: set an initial setting for luks password

xiaodwan:
A better solution is replacing the marco with real value in libvirt-ci,
but '\n' in the string will cause a malform testcase cfg file in
tp-libvirt.

e.g.
variant:
  - case1:
    env_settings = "echo -e '${{{{luks_password}}}}
${{{{luks_password}}}}
${{{{luks_password}}}}
${{{{luks_password}}}}
' |"
  - case2:
    env_settings = "wrold"

There is not a good solution for keeping the '\n', meanwhile, don't
append the '\n' at the end of the string.

And this is not a typical senario, let's just set the real value to
avoid the risk of breaking libvit-ci.


xiaodwan:
@chloerh @chunfuwen @yantian Please help to review this PR, thanks.

PR #3501, opened 2021-05-20
Close aexpect session to avoid leaving processes

chloerh:
Not-closed session could leave aexpect-helper processes which
could affect future test process depending on the command that
was running in the session. Close it to fix the problem.


chloerh:
Test result

pending

Yingshun (approved):
LGTM

PR #3511, opened 2021-05-24
v2v: add a new case about unknown_vmware_controller_type

xiaodwan:
see bz#1510801


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.ova.normal.linux.output_mode.rhev.rhv_upload: PASS (807.25 s)

xiaodwan:
@chloerh @yantian0 Please help to review this PR, thanks.

chunfuwen (commented):
@xiaodwan 
In tp-libvirt(libvirt part), now we usually add polarion case id(without specific link) in git commit log.
Not sure whether there is matched polarion case id, if there is ,better add it.

xiaodwan:
@chunfuwen I think polarion ID is not permanent, if the case management tool is changed in future, it's difficult to find the source.
but anyway, I'll add both polarion id and bugzilla id.

PR #3522, opened 2021-05-28
Fix port checking ignored

dzhengfy:
Due to the function name is wrongly checked, the two ports comparison
does not really happen. So this is to fix the wrong name.


dzhengfy:
@Yingshun ,could you help have a look?

chunfuwen (approved):
lgtm

chunfuwen:
mark it as minor_fix since only one real line change(others are logging.debug)

dzhengfy:
(1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_storage.disks_port.copy_storage_all: PASS (187.67 s)

PR #3546, opened 2021-06-15
v2v: delete all windows configuration from xen

xiaodwan:
Because the windows guests is too old on xen server, they were
removed in 9a600c4df4b194d34ec8d800f12dacc9f50c3f21, but the
windows pre-configuration were left by mistake, this patch fixes
the problem.


xiaodwan:
Just delete the useless configuration, no test should be done.

PR #3555, opened 2021-06-15
Fix Unknown protocol iscsi issue when using qemu-img

chunfuwen:
qemu-img create -f luks xx iscsi://127.0.0.1:3260/iqn.2021-06.com.virttest:emulated-iscsi.target/0 500M
is not supported in latest version due to feature depreciation
and need use direct file path instead

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_luks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_luks.py
@@ -291,8 +291,8 @@ def run(test, params, env):
                     iscsi_target, lun_num = libvirt.setup_or_cleanup_iscsi(
                         is_setup=True, is_login=False, image_size=storage_size,
                         portal_ip=iscsi_host)
-                device_source = "iscsi://%s:%s/%s/%s" % (iscsi_host, iscsi_port,
-                                                         iscsi_target, lun_num)
+                # hardcode
+                device_source = '/var/lib/avocado/data/avocado-vt/emulated-iscsi'
                 disk_src_dict = {"attrs": {"protocol": "iscsi",
                                            "name": "%s/%s" % (iscsi_target, lun_num)},
                                  "hosts": [{"name": iscsi_host, "port": iscsi_port}]}


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.luks.positive_test.coldplug.device_disk.iscsi_test.enable_auth.auth_in_source.use_auth_uuid.encryption_in_source

lento-sun:
Seems the fix is under 'disk type=network' and 'protocl=iscsi' branch. Due to Bug 1967133, the libiscsi will be deprecated in rhel9 anyway. So when that bug fixed, this whole branch will be failed again. 
I don't think we need to fix the 'qemu-img' error here in this case, but maybe just skip the whole case in rhel9 in ci's config. what do you think?

kamilvarga (approved):
LGTM.

chunfuwen:
> Seems the fix is under 'disk type=network' and 'protocl=iscsi' branch. Due to Bug 1967133, the libiscsi will be deprecated in rhel9 anyway. So when that bug fixed, this whole branch will be failed again.
> I don't think we need to fix the 'qemu-img' error here in this case, but maybe just skip the whole case in rhel9 in ci's config. what do you think?

@lento-sun 
We still need it on rhel 8.x(rhel8.5) afterwards, so maybe we still can fix it somehow.
Regarding on rhel9, as you mentioned, we can filter cases related to them in CI.

lento-sun:
> > Seems the fix is under 'disk type=network' and 'protocl=iscsi' branch. Due to Bug 1967133, the libiscsi will be deprecated in rhel9 anyway. So when that bug fixed, this whole branch will be failed again.
> > I don't think we need to fix the 'qemu-img' error here in this case, but maybe just skip the whole case in rhel9 in ci's config. what do you think?
> 
> @lento-sun
> Not sure whether it also impact rhel8.5 afterwards, so maybe we still can fix it somehow.
> Regarding on rhel9, as you mentioned, we can filter cases related to them in CI.

IMHO, I don't get why we need a tmp fix in both rhel9 or rhel8, let's just talk about the rhel8 situation due to you agree to skip the case in rhel9. 

1. If qemu in rhel8.5 still supports the libiscsi, the original 'device_source = "iscsi://%s:%s/%s/%s"' won't be failed when use 'qemu-img create iscsi://xxx', so the change is not necessary.
2. If qemu in rhel8.5 decides not to support libiscsi, the case also could be skipped by skiplist.yaml in that time. and the fix here is also just a tmp workaround which is not necessary.

chunfuwen:
close it as not necessary fix

PR #3562, opened 2021-06-17
Increase current memory size

jinqi7:

Increase current memory size to avoid vm crash in memory_discard


chunfuwen:

chunfuwen (approved):
lgtm

PR #3564, opened 2021-06-18
Update the openvswitch package name

yalzhang:
As the openvswitch package name varies on differenct os version,
update it.

### DIFF
--- a/libvirt/tests/src/virtual_network/mtu.py
+++ b/libvirt/tests/src/virtual_network/mtu.py
@@ -175,8 +175,16 @@ def run(test, params, env):
         if add_pkg:
             add_pkg = add_pkg.split()
             new_pkg = add_pkg.copy()
-            if 'openvswitch' in add_pkg and shutil.which('ovs-vsctl'):
-                new_pkg.remove('openvswitch')
+            if 'openvswitch' in add_pkg:
+                # Install openvswitch
+                if process.system("yum info openvswitch", ignore_status=True) == 0:
+                    utils_package.package_install("openvswitch")
+                if process.system("yum info openvswitch2.11", ignore_status=True) == 0:
+                    utils_package.package_install("openvswitch2.11")
+                if process.system("yum info openvswitch2.15", ignore_status=True) == 0:
+                    utils_package.package_install("openvswitch2.15")
+                if shutil.which('ovs-vsctl'):
+                    new_pkg.remove('openvswitch')
             utils_package.package_install(new_pkg)
         if 'openvswitch' in add_pkg:
             br = 'ovsbr0' + utils_misc.generate_random_string(3)


yalzhang:
before:
Error Details
FileNotFoundError: [Errno 2] No such file or directory (ovs-vsctl del-br ovsbr06UA): 'ovs-vsctl'&#10;
after:
$ sudo avocado --vt-type libvirt run virtual_network.mtu.positive_test.normal.ovswitch.iface
JOB ID     : 4488d3941b169c7a35cd0125810346259e742326
JOB LOG    : /root/avocado/job-results/job-2021-06-17T20.55-4488d39/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.mtu.positive_test.normal.ovswitch.iface: PASS (73.73 s)
JOB TIME   : 74.70 s

yalzhang:
@chloerh  Please help to review, Thank you! utils.net has a function setup_ovs_vhostuser(), but it will set vhostuser ports, which is not suitable for this case. So I copy the lines in that method into this file.

chloerh (commented):
IMO, we could save some code by extracting openvswitch installation from function `setup_ovs_vhostuser` , then call the extracted function here. What do you think? After all, copying code from here to there is something that we should try to avoid.

e.g.
```
def install_openvswith():
    # Install openvswitch
    if process.system("yum info openvswitch", ignore_status=True) == 0:
        utils_package.package_install("openvswitch")
    if process.system("yum info openvswitch2.11", ignore_status=True) == 0:
        utils_package.package_install("openvswitch2.11")
    if process.system("yum info openvswitch2.15", ignore_status=True) == 0:
        utils_package.package_install("openvswitch2.15")
```

Something like this, but with a better name :joy:

libvirt/tests/src/virtual_network/mtu.py
```diff
@@ -175,8 +175,16 @@ def check_mtu_in_vm(fn_login, mtu_size, timeout):
         if add_pkg:
             add_pkg = add_pkg.split()
             new_pkg = add_pkg.copy()
-            if 'openvswitch' in add_pkg and shutil.which('ovs-vsctl'):
-                new_pkg.remove('openvswitch')
+            if 'openvswitch' in add_pkg:
+                # Install openvswitch
+                if process.system("yum info openvswitch", ignore_status=True) == 0:
+                    utils_package.package_install("openvswitch")
+                if process.system("yum info openvswitch2.11", ignore_status=True) == 0:
```
chunfuwen, line 182: change if to elif? if multiple openvswitch exists, it leads to  install each version of it. Is that your intention?

Yingshun (commented):
This problem can be fixed in libvirt-ci I think - adding openvswitch2.15 or changing ovs2 from openvswitch2.11 to  openvswitch2.*(not sure if it works, may need to check with jil. 🙂 )
```
      install-pkgs: !extend
          - lsof
          - openvswitch
          - openvswitch2.11
```
The version may continue to change, I think we'd better find a low-cost way to modify it. What do you think?

yalzhang:
> This problem can be fixed in libvirt-ci I think - adding openvswitch2.15 or changing ovs2 from openvswitch2.11 to openvswitch2.*(not sure if it works, may need to check with jil.  )
> 
> ```
>       install-pkgs: !extend
>           - lsof
>           - openvswitch
>           - openvswitch2.11
> ```
> 
> The version may continue to change, I think we'd better find a low-cost way to modify it. What do you think?

I have submitted a MR to add it in the CI config. I think no matter we will add something in avocado-vt, adding it in the ci config file is no harm. I will close this PR.

PR #3567, opened 2021-06-21
[cpu/diagnose] Add test for spinlock yield forwarding 

smitterl:
Spinlock yield forwarding can potentially let the host solve
spinlocks in the guest by forwarding diag9c. The test confirms
the forwarding. Testing the rate value is out of scope (non-functional
test). However, we choose a high value to make sure yields are forwarded
because during testing lower rate settings didn't seem to be functional
as they didn't forward anything.

More info: https://www.kernel.org/doc/html/latest/virt/kvm/s390-diag.html

### DIFF
--- a/libvirt/tests/cfg/cpu/diagnose_spinlock_yield_forward.cfg
+++ b/libvirt/tests/cfg/cpu/diagnose_spinlock_yield_forward.cfg
@@ -1,0 +1,9 @@
+- diagnose_9c_forward:
+    type = diagnose_spinlock_yield_forward
+    only s390-virtio
+    variants:
+        - forwarding_on:
+              diag9c_forward_hz = 10000
+        - forwarding_off:
+              diag9c_forward_hz = 0
+
--- a/libvirt/tests/src/cpu/diagnose_spinlock_yield_forward.py
+++ b/libvirt/tests/src/cpu/diagnose_spinlock_yield_forward.py
@@ -1,0 +1,135 @@
+import logging
+import time
+
+from avocado.core.exceptions import TestError, TestFail
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest.utils_misc import cmd_status_output, wait_for
+
+
+DEACTIVATE_FORWARDING = 0
+FORWARD_COUNTER_TIMEOUT = 360
+CHECK_INTERVAL = 60
+
+
+def update_kvm_parameter(hz):
+    """
+    Updates kvm paramter for forwarding
+
+    :param hz: diag9c_forwarding_hz
+    :raises TestError: if parameter couldn't be updated
+    """
+
+    cmd = "echo %s > /sys/module/kvm/parameters/diag9c_forwarding_hz" % hz
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise TestError("Could not update parameter. %s" % out)
+
+
+def start_locktorture(session):
+    """
+    Starts spin lock torture test to cause spinlocks in guest
+
+    :param session: guest session
+    :raises TestError: if the locktorture can't be started
+    """
+    logging.debug("session is %s" % session)
+    cmd = "modprobe locktorture torture_type=spin_lock"
+    err, out = cmd_status_output(cmd, shell=True, session=session)
+    if err:
+        raise TestError("Could not start locktorture. %s" % out)
+
+
+def confirm_forwarding(hz):
+    """
+    Confirm if forwarding is happening or not depending on set
+    hz parameter
+
+    :param hz: if 0 expect no forwarding, if positive expect forward
+    """
+
+    if int(hz) > 0:
+        wait_for_forward_count_to_be_non_zero(FORWARD_COUNTER_TIMEOUT)
+    else:
+        confirm_forward_count_is_still_zero_after(CHECK_INTERVAL)
+
+
+def confirm_forward_count_is_still_zero_after(interval):
+    """
+    Wait for a certain time interval and then confirm
+    that the counter didn't go up
+
+    :param interval: the time in seconds we expect the counter not to change
+    :rases TestFail: if the counter increase despite test expectation
+    """
+    time.sleep(interval)
+    err, out = get_forward_value()
+    if err or int(out) != 0:
+        raise TestFail("Counter wasn't zero although forward"
+                       " was inactive. Value is %s" % out)
+
+
+def get_forward_value():
+    """
+    Reads sysfs for forward value
+    """
+    cmd = "cat /sys/kernel/debug/kvm/diag_9c_forward"
+    return cmd_status_output(cmd, shell=True)
+
+
+def wait_for_forward_count_to_be_non_zero(timeout):
+    """
+    Wait for the forward counter to become > 0. Fail if it doesn't
+    after timeout.
+
+    :param timeout: maximum time to wait for counter to go up
+    :rases TestFail: if the counter doesn't increase despite test expectation
+    """
+
+    def count_is_non_zero():
+        """ Confirm if the count is non-zero """
+        err, out = get_forward_value()
+        if err:
+            return False
+        elif int(out) <= 0:
+            return False
+        elif int(out) > 0:
+            return True
+
+    is_non_zero = wait_for(count_is_non_zero, step=10, timeout=timeout)
+    if not is_non_zero:
+        raise TestFail("Counter didn't go up though forwarding active.")
+
+
+def run(test, params, env):
+    """
+    Test if Spinlock Yield Forwarding works
+
+    We use the locktorture kernel module to provoke spinlocks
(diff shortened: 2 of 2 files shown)


smitterl:
@fbq815 Could you have a look? Thanks!

fbq815:
JOB ID     : c979a954b17bdf874145b29ee6fd46b68af3bf6e
JOB LOG    : /root/avocado/job-results/job-2021-06-22T12.59-c979a95/job.log
 (1/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_on: PASS (20.95 s)
 (2/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_off: PASS (71.57 s)

As the test result on RHEL9, this PR seems good to me

smitterl:
> JOB ID : c979a954b17bdf874145b29ee6fd46b68af3bf6e
> JOB LOG : /root/avocado/job-results/job-2021-06-22T12.59-c979a95/job.log
> (1/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_on: PASS (20.95 s)
> (2/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_off: PASS (71.57 s)
> RESULTS : PASS 2 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
> 
> As the test result on RHEL9, this PR seems good to me

Thank you @fbq815 !

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,129 @@
...
+        if err:
+            return False
+        elif int(out) <= 0:
+            return False
+        elif int(out) > 0:
+            return True
+
+    is_non_zero = wait_for(count_is_non_zero, step=10, timeout=timeout)
+    if not is_non_zero:
+        raise TestFail("Counter didn't go up though forwarding active.")
```
dzhengfy, line 94: As a common rule, we do use test.fail() instead of raising TestFail exception directly in this repo.

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,129 @@
...
+
+def confirm_forward_count_is_still_zero_after(interval):
+    """
+    Wait for a certain time interval and then confirm
+    that the counter didn't go up
+    """
+    time.sleep(interval)
+    err, out = get_forward_value()
+    if err or int(out) != 0:
+        raise TestFail("Counter wasn't zero although forward"
```
dzhengfy, line 62: As a common rule, we do use test.fail() instead of raising TestFail exception directly in this repo.

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Starts spin lock torture test to cause spinlocks in guest
+
+    :param session: guest session
+    """
+    logging.debug("session is %s" % session)
+    cmd = "modprobe locktorture torture_type=spin_lock"
+    err, out = cmd_status_output(cmd, shell=True, session=session)
+    if err:
+        raise OSError("Could not start locktorture. %s" % out)
```
dzhengfy, line 37: As a common rule, we do use test.error() instead of raising an exception directly in this repo. But If you like to raise an exception here, please help update the docstring for :return: xxx

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Updates kvm paramter for forwarding
+
+    :param hz: diag9c_forwarding_hz
+    """
+
+    cmd = "echo %s > /sys/module/kvm/parameters/diag9c_forwarding_hz" % hz
+    err, out = cmd_status_output(cmd, shell=True)
+    if err:
+        raise OSError("Could not update parameter. %s" % out)
```
dzhengfy, line 24: As a common rule, we do use test.error() instead of raising an exception directly in this repo. But If you like to raise an exception here, please help update the docstring for :return: xxx

libvirt/tests/src/cpu/diagnose_9c_forward.py
smitterl: @dzhengfy I prefer to raise the exception here directly because then I don't have to pass `test` to the top level function and the top level function can be more easily moved, e.g. to avocado-vt if found useful for other tests. What do you think?

libvirt/tests/src/cpu/diagnose_9c_forward.py
smitterl: @dzhengfy I prefer to raise the exception here directly because then I don't have to pass test to the top level function and the top level function can be more easily moved, e.g. to avocado-vt if found useful for other tests. What do you think?

libvirt/tests/src/cpu/diagnose_9c_forward.py
smitterl: @dzhengfy I changed the error type to the more specific avocado `TestError` which is the error raised by `test.error`. Does this work for you?

libvirt/tests/src/cpu/diagnose_9c_forward.py
smitterl: @dzhengfy I changed the error type to the more specific avocado `TestError` which is the error raised by `test.error`. Does this work for you?

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,135 @@
+import logging
```
chunfuwen, line 1: If I remember correctly, python file name is not recommended as digital number in it .
see, https://www.reddit.com/r/learnpython/comments/5oxkfv/python_file_naming_standard/

libvirt/tests/src/cpu/diagnose_9c_forward.py
```diff
@@ -0,0 +1,135 @@
...
+
+def confirm_forward_count_is_still_zero_after(interval):
+    """
+    Wait for a certain time interval and then confirm
+    that the counter didn't go up
+
+    :param interval: the time in seconds we expect the counter not to change
+    :rases TestFail: if the counter increase despite test expectation
+    """
+    time.sleep(interval)
```
chunfuwen, line 64: Can we have a way to avoid hardcode sleep here ?

fbq815: Hi Sebas, to avoid hardcode sleep, maybe you could try "taskset -pc % <PID>" , this should also work on libvirt side, this would improve the increasing rate of diagnose_9c_forward I think

smitterl: > Hi Sebas, to avoid hardcode sleep, maybe you could try "taskset -pc % " , this should also work on libvirt side, this would improve the increasing rate of diagnose_9c_forward I think

I don't think this will work. The cpupinning via taskset could increase the likelihood of counter to go up but only in the nested setup
host(LPAR) -> cpupin -> guest1(kvm) -> guest2(kvm)
because it makes sure that cpus in guest1 will overcommit. However, this test case doesn't use the nested setup.

smitterl: > Can we have a way to avoid hardcode sleep here ?

I don't think we can avoid a sleep here totally because the test condition is "make sure that the counter does NOT go up" and we observe the value during the first `interval` seconds. This is a rather fragile test case as it depends on the overcommitment on the host., IMO, so personally I would prefer to remove it fully.

libvirt/tests/src/cpu/diagnose_9c_forward.py
smitterl: OK, I renamed them.

smitterl: I also re-ran the tests to make sure I didn't forget anything
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system diagnose_9c_forward
JOB ID     : 5f5876e8cfe26d9ff7600c1ac68396581ecdad4a
JOB LOG    : /root/avocado/job-results/job-2021-07-02T11.28-5f5876e/job.log
 (1/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_on: PASS (15.13 s)
 (2/2) type_specific.io-github-autotest-libvirt.diagnose_9c_forward.forwarding_off: PASS (72.78 s)
JOB TIME   : 88.19 s
```

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

fbq815 (approved):

PR #3575, opened 2021-06-23
v2v: fix chipset type checkpoint in rhv4.4.6.8

xiaodwan:
See bz1961945.
Because of above bug, the chipset setting was broken in rhv4.4.6.8.
And it blocks all our testing. A temporary workaround is done to
avoid all scripts failure. Once it's fixed, a better solution will
replace this one.


xiaodwan:
@yantian0 please help to review this PR, thanks.

yantian0 (approved):

PR #3576, opened 2021-06-24
sriov: Add 1 net failover case

Yingshun:
RHEL-187001 - [save/restore][net failover] Save & restore for vm
with hostdev type interface and teaming setting

Test result: Failed because of BZ1815426.
` (1/1) type_specific.io-github-autotest-libvirt.sriov_net_failover.save_restore_hostdev_iface_with_teaming: ERROR: Command '/usr/bin/virsh save avocado-vt-vm1 /tmp/avocado_t4n0wh7t/save_file ' failed.\nstdout: b''\nstderr: b''\nadditional_info: None (188.27 s)`

### DIFF
--- a/libvirt/tests/src/sriov/sriov_net_failover.py
+++ b/libvirt/tests/src/sriov/sriov_net_failover.py
@@ -1,5 +1,7 @@
+import aexpect
 import logging
 import os
+import re
 import time
 
 from avocado.utils import process
@@ -150,7 +152,8 @@ def run(test, params, env):
         virsh.restore(save_file, debug=True, ignore_status=False)
         if not libvirt.check_vm_state(vm_name, "running"):
             test.fail("The guest should be running after executing 'virsh restore'.")
-
+        vm.cleanup_serial_console()
+        vm.create_serial_console()
         vm_session = vm.wait_for_serial_login()
         check_vm_network_accessed(vm_session)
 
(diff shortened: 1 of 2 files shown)


Yingshun:
@yalzhang @chloerh Would you please help review? Thanks!

chloerh (approved):

libvirt/tests/src/sriov/sriov_net_failover.py
```diff
@@ -186,24 +230,42 @@ def check_vm_iface_num(session, exp_num=3):
...
+            tcpdump_session.sendline(cmd)
+
         if not utils_misc.wait_for(lambda: not utils_test.ping(
            ping_dest, count=3, timeout=5, output_func=logging.debug,
            session=vm_session)[0], first=5, timeout=timeout):
             test.fail("Failed to ping %s." % ping_dest)
+        if tcpdump_iface:
+            output = tcpdump_session.get_stripped_output()
+            pat_str = "ICMP echo request"
+            if re.search(pat_str, output) == tcpdump_status_error:
```
yalzhang, line 264: LGTM. Do you plan to add this check to the other scenarios? I think when there are hostdev+bridge and we test connection, then it is better to add this check with tcpdump_status_error=True to ensure the traffic go through hostdev not bridge, like after hot plug hostdev interface/device, or after migration on the target host.  If there is only a bridge interface, like after unplug hostdev, or during migration, no need to check with tcpdump_status_error=False because there is only one bridge interface, ensure it can access successfully is enough.

Yingshun: Ok, will update other test scenarios in other PR.

yalzhang (approved):

PR #3592, opened 2021-06-29
v2v: don't test to rhv when VIRITO_WIN or virtio-win is not set

xiaodwan:
see bz#1913577


xiaodwan:
very minor fix, so merge it to move forward the testing.

PR #3632, opened 2021-07-09
Fix the pattern issue

yalzhang:
The original pattern can not get the desired connction name, update it.


yalzhang:
when the connection name is " enp1s0"

>>> out = process.run('nmcli dev show enp1s0').stdout_text
before:
>>> x = re.findall(r'GENERAL.CONNECTION:(.+?)\n', out)
>>> x
['                     enp1s0']

after:
>>> x = re.findall(r'GENERAL.CONNECTION:\s+(\S+.*)$', out, re.MULTILINE)
>>> x
['enp1s0']

when the connection name is "wired connection 2"

PR #3638, opened 2021-07-13
sriov: Add 2 'managed' related cases

Yingshun:
RHEL7-18519 - [network] Start vm with VF from VF Pool
    with "managed=no" or default setting
RHEL7-43422 - [attach-device] Hotplug/unplug VF with managed='no'

**depends on:**
https://github.com/avocado-framework/avocado-vt/pull/3133
https://github.com/avocado-framework/avocado-vt/pull/3136
https://github.com/avocado-framework/avocado-vt/pull/3137
https://github.com/avocado-framework/avocado-vt/pull/3138
**Test results:**
 ```
(1/3) type_specific.io-github-autotest-libvirt.sriov.managed.networks.managed_no: PASS (109.36 s)
 (2/3) type_specific.io-github-autotest-libvirt.sriov.managed.networks.default: PASS (170.95 s)
 (3/3) type_specific.io-github-autotest-libvirt.sriov.managed.device_hotplug.managed_no: PASS (40.63 s)
JOB TIME   : 322.76 s

```

### DIFF
--- a/libvirt/tests/src/sriov/sriov_managed.py
+++ b/libvirt/tests/src/sriov/sriov_managed.py
@@ -1,0 +1,152 @@
+import logging
+
+from provider.sriov import sriov_base
+
+from virttest import utils_net
+from virttest import utils_sriov
+from virttest import virsh
+
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices import interface
+from virttest.utils_libvirt import libvirt_network
+from virttest.utils_libvirt import libvirt_vfio
+from virttest.utils_libvirt import libvirt_vmxml
+from virttest.utils_test import libvirt
+
+
+def run(test, params, env):
+    """
+    SR-IOV: managed related test.
+    """
+    def start_vm(vm, test_login=False, destroy_vm=False):
+        """
+        Start up VM
+
+        :param vm: The vm object
+        :param test_login: Whether to login VM
+        :param destroy_vm: Whether to destroy VM
+        """
+        if vm.is_alive():
+            vm.destroy()
+        vm.start()
+        if test_login:
+            vm.wait_for_serial_login(timeout=180).close()
+        if destroy_vm:
+            vm.destroy()
+
+    def create_vf_pool():
+        """
+        Create VF pool
+        """
+        net_hostdev_dict = {"net_name": params.get("net_name"),
+                            "net_forward": params.get("net_forward"),
+                            "vf_list_attrs": "[%s]" % utils_sriov.pci_to_addr(vf_pci)}
+        libvirt_network.create_or_del_network(net_hostdev_dict)
+
+    def check_vm_iface_managed(vm_name, iface_dict):
+        """
+        Check 'managed' in VM's iface
+
+        :param vm_name: Name of VM
+        :param iface_dict: The parameters dict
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Yingshun** [comment]
@yalzhang @chloerh would you please help review this PR? Thanks!

File: libvirt/tests/src/sriov/sriov_managed.py [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+        libvirt_vfio.check_vfio_pci(vf_pci)
+        virsh.nodedev_reattach(dev_name, debug=True, ignore_status=False)
+        libvirt_vfio.check_vfio_pci(vf_pci, status_error=True)
+
+    test_case = params.get("test_case", "")
+    run_test = eval("test_%s" % test_case)
+
+    vm_name = params.get("main_vm", "avocado-vt-vm1")
+    vm = env.get_vm(vm_name)
+    pf_pci = utils_sriov.get_pf_pci()
```
yalzhang, line 131: Suggest to add something like:
if not pf_pci:
    test.cancel("no available PF found")

The status check in utils_sriov.get_pf_pci() is a little too strict to check "operstate" file. But after I think twice, I think it is reasonable, especially when we need to test the network connection by 'ping'('operstate' shows "up" means Interface is operational up and can be used.).  Libvirt also check the PF's status, if the PF status is "down", the vm with its vf passthrough can not start. But this "down" is not depend on "operstate", it is much looser, just check if the interface is configured up/down.

provider/sriov/sriov_base.py
```diff
@@ -0,0 +1,40 @@
...
+    default_vf = process.run(cmd, shell=True, verbose=True).stdout_text
+    if not utils_sriov.set_vf(pf_pci_path, vf_no):
+        raise exceptions.TestError("Failed to set vf.")
+    return default_vf
+
+
+def recover_vf(pf_pci, default_vf, params):
+    """
+    Recover vf setting
+
```
yalzhang, line 32: If this function is to clear the vf setting and restore the vf_no to 0, the default_vf should be "0"

libvirt/tests/src/sriov/sriov_managed.py
Yingshun: pf_pci value checking added.

provider/sriov/sriov_base.py
Yingshun: I'd like to make it configurable so I set a default value to '0', is it ok?

yalzhang: It's OK, though I can not imagine the scenario that restore it to non-zero value. Because when the value is 4, next time when we want to change it to 2, it is impossible and will report error. The correct operation it to restore to 0, then change it to 2.

yalzhang: I have ran it on my local system and checked the logs, it's very good.

yalzhang (approved):

provider/sriov/sriov_base.py
Yingshun: The default vf value was from the system before testing. So if the VFs was not disabled after testing, it should be "incorrectly" set before you ran the case.

chloerh (approved):

PR #3650, opened 2021-07-15
Set one memory hotplug case to x86 only

chloerh:
According to the feature itself

### DIFF
--- a/libvirt/tests/cfg/memory/memory_hotplug.cfg
+++ b/libvirt/tests/cfg/memory/memory_hotplug.cfg
@@ -3,6 +3,7 @@
     start_vm = no
     variants case:
         - report_failure:
+            only x86_64
             setvm_max_mem_rt = 15243264
             setvm_max_mem_rt_unit = 'KiB'
             setvm_max_mem_rt_slots = 16


jinqi7:
Seems the expect result in rhel9 is changed. So, can you  set it only tested in rhel8 ?

jinqi7 (commented):
It doesn't fit for rhel9 also.

chloerh:
> It's need to be add to skiplist for rhel9.

@jinqi7 I'd prefer to add it to skiplist, what do you think?

libvirt/tests/cfg/memory/memory_hotplug.cfg
```diff
@@ -3,6 +3,7 @@
     start_vm = no
     variants case:
         - report_failure:
+            only x86_64
```
dzhengfy, line 6: What's the reason for this memory hotplug case will be only supported on x86_64?

jinqi7:
> > It's need to be add to skiplist for rhel9.
> 
> @jinqi7 I'd prefer to add it to skiplist, what do you think?

That's I am suggesting.

chloerh:
Then it's decided, the case will be added to skiplist.
@jinqi7 @dzhengfy Please help to merge this one, thanks.

libvirt/tests/cfg/memory/memory_hotplug.cfg
chloerh: @dzhengfy it's the design of the case, confirmed with Luyao.

jinqi7 (approved):

dzhengfy (approved):
LGTM

chloerh:
I was going to added the case to skiplist and then found that @jinqi7 has already done it for me. Thanks @jinqi7 :+1:

PR #3676, opened 2021-08-01
Wait a while to confirm the vf initialize and vm boot

yalzhang:
Fix the occasional failure caused by the vm not boot successfully,
or vf not initialized successfully.

### DIFF
--- a/libvirt/tests/src/sriov/sriov_ro_sysfs.py
+++ b/libvirt/tests/src/sriov/sriov_ro_sysfs.py
@@ -44,7 +44,10 @@ def run(test, params, env):
         if 'hostdev' not in vm_iface_types:
             test.fail('Unable to get hostdev interface!')
         if cmd_in_vm:
-            vm_session.cmd(cmd_in_vm)
+            if not utils_misc.wait_for(lambda:
+                                       not vm_session.cmd_status(cmd_in_vm),
+                                       30, 10):
+                test.fail("Can not get the Virtual Function info on vm!")
         vm_session.close()
 
     libvirt_version.is_libvirt_feature_supported(params)


yalzhang:
before:
3 cases fail when auto job run on x710 cards:
rhel.sriov.ro_sysfs.vf_hotplug
rhel.sriov.managed.networks.managed_no
rhel.sriov_net_failover.save_restore_hostdev_iface_with_teaming
after run with this PR, the 2 failure disappear, check on rhel 9 sriov run test build 36

yalzhang:
@Yingshun  Please help to check, Thank you!

libvirt/tests/src/sriov/sriov_ro_sysfs.py
```diff
@@ -44,6 +45,8 @@ def test_vf_hotplug():
         if 'hostdev' not in vm_iface_types:
             test.fail('Unable to get hostdev interface!')
         if cmd_in_vm:
+            # to confirm the vm has boot successfully
+            time.sleep(10)
```
Yingshun, line 49: How about waiting for a while via utils_misc.wait_for() instead of time.sleep()?
for example, it can be `utils_misc.wait_for(lambda: vm_session.cmd(cmd_in_vm), 10, 10)`, or something like that.

libvirt/tests/src/sriov/sriov_managed.py
```diff
@@ -133,6 +134,8 @@ def test_device_hotplug():
         test.cancel("NO available pf found.")
     default_vf = sriov_base.setup_vf(pf_pci, params)
 
+    # after set vm, wait a while for vf to initialize
+    time.sleep(5)
```
Yingshun, line 138: Does it make a problem when getting vf's pci id? If yes, I recommend updating sriov_base.setup_vf().

libvirt/tests/src/sriov/sriov_ro_sysfs.py
yalzhang: Updated, Thank you!

libvirt/tests/src/sriov/sriov_managed.py
yalzhang: It did happen sometimes, not each time. And I found the case will pass with vf already enabled, and probably fail when there is no existing vfs. The error says " Unable to get VF". The "ls /sys/bus/pci/devices/{}/virtfn{}/net".format(pf_pci, vf_index) may fail with "ls: cannot access '/sys/bus/pci/devices/0000:3b:00.0/virtfn0/net': No such file or directory". I think we need to wait for the directory creation even the echo command returned successfully. It is not 100% happen. I have ran pass locally for 3 times now(echo 0 before run the case). I think we can  add a sleep(5) in utils_sriov.set_vf() if there is further failures. Just keep it as it is, and wait and see.

chloerh (approved):

libvirt/tests/src/sriov/sriov_ro_sysfs.py
```diff
@@ -44,7 +44,8 @@ def test_vf_hotplug():
         if 'hostdev' not in vm_iface_types:
             test.fail('Unable to get hostdev interface!')
         if cmd_in_vm:
-            vm_session.cmd(cmd_in_vm)
+            # to confirm the vm has boot successfully
```
Yingshun, line 47: I think this comment is incorrect, right?

libvirt/tests/src/sriov/sriov_ro_sysfs.py
```diff
@@ -44,7 +44,8 @@ def test_vf_hotplug():
...
+            utils_misc.wait_for(lambda: vm_session.cmd_status(cmd_in_vm), 10, 5)
```
Yingshun, line 48: I think we'd better add a check here, and  `lambda: vm_session.cmd_status(cmd_in_vm)` should be ` lambda:  not vm_session.cmd_status(cmd_in_vm)`, 

Test passed with below change in my env, you can update accordingly. 
```
-            vm_session.cmd(cmd_in_vm)
+            if not utils_misc.wait_for(lambda: not vm_session.cmd_status(cmd_in_vm), 30, 10):
+                test.fail('xxx')
```

libvirt/tests/src/sriov/sriov_managed.py
Yingshun: Ok, thanks for you explanation. I'll keep eyes on this problems.

libvirt/tests/src/sriov/sriov_ro_sysfs.py
yalzhang: Updated, Thank you!

libvirt/tests/src/sriov/sriov_ro_sysfs.py
yalzhang: Yes, it is confusing, I have deleted it. It is not because vm boot but the hotplug operation trigger some msg in vm, which cause the 'lspci | grep' command fail.

Yingshun (approved):
LGTM

libvirt/tests/src/sriov/sriov_ro_sysfs.py
yalzhang: updated

PR #3683, opened 2021-08-03
live_migration:fix err pattern

dzhengfy:

### DIFF
--- a/libvirt/tests/cfg/migration/live_migration.cfg
+++ b/libvirt/tests/cfg/migration/live_migration.cfg
@@ -161,7 +161,7 @@
             variants:
                 - migrate_x509_no:
                     default_x509_no, default_x509_one:
-                        err_msg = "Cannot write to TLS channel | Cannot read from TLS channel"
+                        err_msg = "Cannot write to TLS channel|Cannot read from TLS channel"
                         migrate_again = 'yes'
                         status_error = 'yes'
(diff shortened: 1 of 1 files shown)


dzhengfy:
Before fix:
```
virsh.live_migration.native_tls.migrate_x509_one.default_x509_zero.by_shared.p2p_live.without_postcopy
TestFail: Can not find the expected patterns 'Cannot write to TLS channel | Cannot read from TLS channel' in output 'error: operation failed: migration out job: Cannot write to TLS channel: Input/output error'
```

dzhengfy:
This is a very simple fix, so I do not think we need to rerun the case to test the fix. This is because the extra blank before and after '|' which make the re can not match the expected string.
A simple test using python session is enough as below:

```
>>> pat = 'Cannot write to TLS channel | Cannot read from TLS channel'
>>> output='error: operation failed: migration out job: Cannot write to TLS channel: Input/output error'
>>> print(re.search(pat, output))
None
>>> pat = 'Cannot write to TLS channel|Cannot read from TLS channel'
>>> print(re.search(pat, output))
<_sre.SRE_Match object; span=(44, 71), match='Cannot write to TLS channel'>
>>> 
```

cliping (approved):
LGTM

Yingshun (approved):

PR #3716, opened 2021-08-19
migrate_storage: Fixup some script issues

Yingshun:
1. Setup ssh connection for copy_storage_all cases to avoid
connection failures
2. Update bandwidth and thread_timeout to avoid timeout issue


Yingshun:
@cliping @dzhengfy would you please help review it? Thanks!

Yingshun:
@smitterl would you please help review it? Thanks!

cliping (approved):
LGTM

libvirt/tests/cfg/migration/migrate_storage.cfg
```diff
@@ -2,6 +2,11 @@
     type = migrate_storage
     migration_setup = "yes"
     log_outputs = "/var/log/libvirt/libvirt_daemons.log"
+    client_ip = "${local_ip}"
```
chunfuwen, line 5: If I understand correctly ,newly introduced method: update_params(xx) in migration_test aims to resolve similar issue. Can it apply here as well?

Yingshun: No, we don't have a function to set these values. Or maybe you can show me directly. And they will be used in environment preparation period, so I think they should be in cfg directly.

chunfuwen: what migration_test.update_virsh_migrate_extra_args aims for?

Yingshun: It's for updating extra_args of do_migration.

chunfuwen: @Yingshun 
@dzhengfy 
Do you think it is proper to wrap client_ip related parameters handling logic in updating extra_args of do_migration as well in case client_ip is missed in cfg file ? By it , it sounds like one unified solution can be achieved.

Yingshun: I don't think so. The value of extra_args will be used in do_migration only. But in the case of mine, they'll be used by others, so I intend to keep the current way.
Actually, we had a discussion about  the settings like client_ip, and decided to set them in cfg at that time.

dzhengfy: It looks ok to me in current solution. It is also the usual way to set these parameters in cfg for better awareness to the user.  update_virsh_migrate_extra_args() has special usage, so it is not good to mess them up.

dzhengfy (approved):
LGTM

PR #3724, opened 2021-08-23
Exclude s390x for vhostuser related test

yalzhang:
As s390x do not support openvswitch, skip the test which has
openvswitch dependency.


chunfuwen (approved):
LGTM

chunfuwen:
There is no result required since it aims to filter cases.
Moreover, it only has two line changes, mark it as minor_fix

PR #3740, opened 2021-08-25
usb: fix unsupported usb model issue

brucepatricklee:
usb: fix unsupported usb model issue - remove cpu model vt82c686b-usb-uhci


cliping (approved):
LGTM

PR #3742, opened 2021-08-25
Fix value compare error for rx

kylazhang:

For compare_iface_stat when host rx_errs and rx_drop are not 0(as in vm it may always 0), the delta/max_value will larger than expect value. Already checked with feature owner that we do not need to compare these 2 value. So remove the compare for them

before:
```
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_stat.compare.direct_type: FAIL: Iface stat of host and vm should be close. (64.33 s)
```
after:
```
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_stat.compare.direct_type: PASS (63.30 s)
```


cliping (approved):
LGTM

PR #3743, opened 2021-08-26
Fix virtual disks vhostuser use cases by increasing num-queues

chunfuwen:
As per bug[1], larger number of queues are needed when start qemu-storage-daemon

[1]https://bugzilla.redhat.com/show_bug.cgi?id=1992390

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_vhostuser.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_vhostuser.py
@@ -37,6 +37,7 @@ def start_vhost_sock_service(image_path, sock_path):
         ' --blockdev \'{"driver":"file","filename":"%s","node-name":"libvirt-1-storage","auto-read-only":true,"discard":"unmap"}\''
         ' --blockdev \'{"node-name":"libvirt-1-format","read-only":false,"driver":"raw","file":"libvirt-1-storage"}\''
         ' --export vhost-user-blk,id=vhost-user-blk0,node-name=libvirt-1-format,addr.type=unix,addr.path=%s,writable=on'
+        ',num-queues=3'
         ' --chardev stdio,mux=on,id=char0; sleep 3'
         % (image_path, sock_path))
     cmd_output = process.run(start_sock_service_cmd, ignore_status=False, shell=True).stdout_text.strip()


libvirt/tests/src/virtual_disks/virtual_disks_vhostuser.py
```diff
@@ -37,6 +37,7 @@ def start_vhost_sock_service(image_path, sock_path):
         ' --blockdev \'{"driver":"file","filename":"%s","node-name":"libvirt-1-storage","auto-read-only":true,"discard":"unmap"}\''
         ' --blockdev \'{"node-name":"libvirt-1-format","read-only":false,"driver":"raw","file":"libvirt-1-storage"}\''
         ' --export vhost-user-blk,id=vhost-user-blk0,node-name=libvirt-1-format,addr.type=unix,addr.path=%s,writable=on'
+        ',num-queues=3'
```
kamilvarga, line 40: Just out of curiosity, is there any system variable we can look/check to setup this value per the system? For example number of the virtual cpus?

kamilvarga (commented):
Others LGTM.

libvirt/tests/src/virtual_disks/virtual_disks_vhostuser.py
chunfuwen: As per bug 1992390, it sounds like just because it pick default value , which is usually the number of virtual CPUs, then qemu guest failed boot

kylazhang: I think you need to check how to calculate the number here, with smp +1?(not sure)  please check with feature owner to find the result instead of hard code here. Thanks

kylazhang: I think you need to check how to calculate the number here, with smp +1?(not sure)  please check with feature owner to find the result instead of hard code here. Thanks

chunfuwen: @kylazhang , we don't specify smp parameter here when launch qemu-storage-daemon
@qiankehan , do you have idea how  we  can dynamically specify num_queues value

qiankehan: > @kylazhang , we don't specify smp parameter here when launch qemu-storage-daemon
> @qiankehan , do you have idea how we can dynamically specify num_queues value?
You can do it by three ways. The default values are
1.  **num_queues of qemu-storage-daemon == 1**
2. **vcpus==queues  of virtio device**

And one limitation:  **queues  of virtio device <= num_queues of qemu-storage-daemon**
I have 3 ideas to resolve it:
1. As what you did in this PR, set a large enough value for num_queues of qemu-storage-daemon
2. As what I did in PR3749, set queues  of virtio device as 1
3. Get the vcpu number(xpath /domain/vcpu/text()). Then set num_queues of qemu-storage-daemon as vcpus.

chunfuwen: @qiankehan 
@kylazhang 
Let me clarify something to avoid making simple thing become complex.
If we go through related Bug  description, the root clause is sometimes queues of virtio device > num_queues of qemu-storage-daemon.
1)By PR PR3749, queues of virtio device is set as 1
2)Here in this PR,  I set num_queues of qemu-storage-daemon as 3(the max allowed is 4)
3) So queues of virtio device =1 < num_queues of qemu-storage-daemon =3, which provide enough safe threthod value

kylazhang: I prefer to @qiankehan 's  suggestion that set the num_queues = vcpus. So I think you can get the vcpus and give its value to num_queues.

kylazhang: And if 3 is a large enough value for any condition as num_queues value then I can also accept current change. @qiankehan @chunfuwen Do I understand correctly?

chunfuwen: @kylazhang 
Go back the fundamental purpose case itself.
What we should do is to  ensure libvirt has solid vhost-socket provided by qemu-storage-daemon.
Libvirt don't focus on testing qemu-storage-daemon  with  variants num_queues config, which should be qemu team focus

chunfuwen: @kylazhang 
In PR PR3749, queues of virtio device is set as 1.
By this PR, we set num_queues of qemu-storage-daemon as 3
Therefore, 3 is a large enough value for any condition

kamilvarga: > @kylazhang
> In PR PR3749, queues of virtio device is set as 1.
> By this PR, we set num_queues of qemu-storage-daemon as 3
> Therefore, 3 is a large enough value for any condition

What if someone decides to change the queues of virtio device to 3 or more? The test will fail again and the debugging and review will be required again as well. What about to make the test more robust for such changes? I believe the 3rd bulet from @qiankehan answer above should not be too dificult to implement, right?

> Get the vcpu number(xpath /domain/vcpu/text()). Then set num_queues of qemu-storage-daemon as vcpus.

chunfuwen: @kamilvarga 
Please walk through the whole history about this PR and related case.
It is not how difficult to implement it , more about how to approach the solution in the clear and simple way.
The case itself is test vhost-user disk VM can be operated normally. It is not test variants we provision vhost socket(that is what qemu need adddress).
What we need ensure that  num_queues of qemu-storage-daemon >=  num_qeuues of virtio devices, which is guranted by PR PR3749(virtio device is set as 1).

By the way, theotically, this PR is not necessary now since PR3749 is merged.
After observing test results for a while ,I expect to close it.

chunfuwen:
close it since it may not need any more

libvirt/tests/src/virtual_disks/virtual_disks_vhostuser.py
kamilvarga: Sorry for late response, I missed this somehow. Anyway, you have better history background of this update, so I trust you and fix is OK from my point of view.

kamilvarga (commented):
LGTM.

PR #3749, opened 2021-09-01
virtual_disks_vhostuser: Set queues of vhost blk to 1

qiankehan:
Since QEMU 5.2 (commit a4eef0711b2)[1] the default queues of
vhost-user-blk-pci will be set to the numbers of smp. In case it is over
the default queues value 1 in qemu-storage-daemon, set it to 1.

Fix the 

[1]: https://gitlab.com/qemu/qemu/-/commit/a4eef0711b2cf7a7476c3e2c202a414b68a1baa0

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_vhostuser.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_vhostuser.cfg
@@ -11,6 +11,7 @@
     target_dev = "vdb"
     target_bus = "virtio"
     type_name = "vhostuser"
+    queues = 1
     disk_snapshot_attr = "no"
     variants:
         - start_vhostuser_vm:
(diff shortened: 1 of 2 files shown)


chunfuwen:
@qiankehan 
previously there is similar fix by increasing num-queues, https://github.com/autotest/tp-libvirt/pull/3743.
Do we need both of them, or they fix the same issue?

qiankehan:
> @qiankehan
> previously there is similar fix by increasing num-queues, #3743.
> Do we need both of them, or they fix the same issue?

yes. The same issue.  But `num-queues=3` is not completed. As the commit msg says, if the default queues(vcpus) is over the num-queues in qemu-storage-daemon, it will be failed, too.

chloerh (approved):

kamilvarga (approved):
LGTM.

chunfuwen (approved):
lgtm

PR #3752, opened 2021-09-02
Add memory case

chloerh:
- RHEL7-93110 - Multi-operation for dimm devices

### DIFF
--- a/libvirt/tests/cfg/memory/memory_misc.cfg
+++ b/libvirt/tests/cfg/memory/memory_misc.cfg
@@ -50,3 +50,12 @@
                     idmap_attrs = {'uid': {'start': '0', 'target': '1000', 'count': '10'}, 'gid': {'start': '0', 'target': '1000', 'count': '10'}}
                     cpu_attrs = {'numa_cell': [{'id': '0', 'cpus': '0,2', 'memory': '513024', 'unit': 'KiB'}, {'id': '1', 'cpus': '1,3', 'memory': '513024', 'unit': 'KiB'}]}
                     memxml_attrs = {'mem_model': 'dimm', 'target': {'size': 512000, 'node': 1}}
+        - dimm:
+            variants case:
+                - multiop:
+                    vm_attrs = {'max_mem_rt': 4096, 'max_mem_rt_slots': 16, 'max_mem_rt_unit': 'M', 'vcpu': 4, 'cpu': {'numa_cell': [{'id': '0', 'cpus': '0-1', 'memory': '512', 'unit': 'M'}, {'id': '1', 'cpus': '2-3', 'memory': '512', 'unit': 'M'}]}}
+                    dimm_device_0 = {'mem_model': 'dimm', 'target': {'size': 524288, 'size_unit': 'KiB', 'node': 0}, 'alias': {'name': 'ua-c80aba6e-b6d8-448b-ab6e-8c7b5c29f354'}, 'address': {'attrs': {'type': 'dimm', 'slot': '0', 'base': '0x100000000'}}}
+                    dimm_device_1 = {'mem_model': 'dimm', 'target': {'size': 524288, 'size_unit': 'KiB', 'node': 0}, 'alias': {'name': 'dimm2'}, 'address': {'attrs': {'type': 'dimm', 'slot': '2', 'base': '0x120000000'}}}
+                    dimm_device_2 = {'mem_model': 'dimm', 'target': {'size': 524288, 'size_unit': 'KiB', 'node': 0}, 'alias': {'name': 'dimm4'}, 'address': {'attrs': {'type': 'dimm', 'slot': '4', 'base': '0x140000000'}}}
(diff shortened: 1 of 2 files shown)


chloerh:
(1/1) type_specific.local.memory_misc.dimm.multiop: PASS (4.71 s)

libvirt/tests/cfg/memory/memory_misc.cfg
```diff
@@ -50,3 +50,13 @@
...
+                    dimm_device_1 = {'mem_model': 'dimm', 'target': {'size': 524288, 'size_unit': 'KiB', 'node': 0}, 'alias': {'name': 'dimm2'}, 'address': {'attrs': {'type': 'dimm', 'slot': '2', 'base': '0x120000000'}}}
+                    dimm_device_2 = {'mem_model': 'dimm', 'target': {'size': 524288, 'size_unit': 'KiB', 'node': 0}, 'alias': {'name': 'dimm4'}, 'address': {'attrs': {'type': 'dimm', 'slot': '4', 'base': '0x140000000'}}}
+                    qemu_cmdline = '-object memory-backend-ram,id=memua-c80aba6e-b6d8-448b-ab6e-8c7b5c29f354,size=536870912 -device pc-dimm,node=0,memdev=memua-c80aba6e-b6d8-448b-ab6e-8c7b5c29f354,id=ua-c80aba6e-b6d8-448b-ab6e-8c7b5c29f354,slot=0,addr=4294967296.*-object memory-backend-ram,id=memdimm2,size=536870912 -device pc-dimm,node=0,memdev=memdimm2,id=dimm2,slot=2,addr=4831838208.*-object memory-backend-ram,id=memdimm4,size=536870912 -device pc-dimm,node=0,memdev=memdimm4,id=dimm4,slot=4,addr=5368709120'
```
kamilvarga, line 60: Just out of curiosity, are those hex hashes same each qemu run?

kamilvarga (commented):
Others LGTM.

libvirt/tests/cfg/memory/memory_misc.cfg
chloerh: @kamilvarga I suppose you mean `id=memua-c80aba6e-b6d8-448b-ab6e-8c7b5c29f354` ? To my understanding, it came from alias setting of dimm_device_0. So I think it's supposed to be the same. I'll run this test later on another machine to make sure, and then get back to you.

kamilvarga: OK, thank you.

chloerh: @kamilvarga 
After consulting with feature owners, we decided to update the check steps a little. The qemu command line checking now updated. Could you help review again? Thanks

chloerh: Oh, I have run the test case on both 8 and 9 and passed.

kamilvarga (approved):
LGTM.

libvirt/tests/src/memory/memory_misc.py
```diff
@@ -233,6 +235,52 @@ def run_test_xml_check(case):
...
                     test.fail('Xml comparision of %s failed.', attr)
 
+    def run_test_dimm(case):
+        """
+        Multi-operation for dimm devices
+
+        :param case: test case
+        """
+        vm_attrs = eval(params.get('vm_attrs', '{}'))
+        vmxml.setup_attrs(**vm_attrs)
```
chunfuwen, line 245: vm_attrs is long hardcoded string from cfg.
If we can stand with maintain one long string, I wonder what advantages set_attrs() has compared with set_xml(xx) directly?

chloerh: @chunfuwen what's your suggestion about the code? Could you provide an example of the code that you suggest? I don't think the current way is not easy to maintain.

dzhengfy: @chloerh , I would like to say for this PR,  set_attrs() is acceptable to be used which aligns with our discussion within the team.
@chunfuwen, regarding set_xml(xx) usage, it is another way to operate the device xml. We can discuss it more within the team. But I do not think it is a blocker to this PR, right?

dzhengfy (approved):
LGTM

libvirt/tests/src/memory/memory_misc.py
chunfuwen: @dzhengfy 
@chloerh 
set_attr() is added to recommended list for Libvirt automation tips in upstream.
I just want to confirm the scope "recommend list"
I just wonder whether we have to  use set_attrs() as default option to construct xml  in the following practice, or set_attr() is one of option for contributors to choose?

chloerh: @chunfuwen I don't think the 'recommend list' is something we need to discuss on this pr...
Do you still have concerns about **this pr**?

chunfuwen (approved):
Approve this PR firstly , let's move recommend list to other thread for discussion although it impact this decision making.

PR #3755, opened 2021-09-03
sriov: Add a case to test max vfs

Yingshun:
RHEL7-18503 - Hotplug MAX VFs to guest

**Test result:**
` (1/1) type_specific.io-github-autotest-libvirt.sriov.scalability.max_vfs: PASS (334.02 s)`

### DIFF
--- a/libvirt/tests/src/sriov/sriov_scalability.py
+++ b/libvirt/tests/src/sriov/sriov_scalability.py
@@ -1,0 +1,199 @@
+import logging
+
+from provider.sriov import sriov_base
+
+from virttest import utils_misc
+from virttest import utils_sriov
+from virttest import virsh
+
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_libvirt import libvirt_network
+from virttest.utils_libvirt import libvirt_vmxml
+from virttest.utils_test import libvirt
+
+
+def get_vm_iface_num(vm_name):
+    """
+    Get VM virtual interfaces' number
+
+    :param vm_name: The name of VM
+    :return: VM interfaces' number
+    """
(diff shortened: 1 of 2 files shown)


Yingshun:
@yalzhang @chloerh would you plz help review this pr? Thanks!

chloerh (approved):

libvirt/tests/src/sriov/sriov_scalability.py
```diff
@@ -0,0 +1,199 @@
...
+        Setup for max_vfs case
+
+        1. Check test environment
+        2. Enable VFs
+        3. Create networks
+        """
+        if not utils_misc.compare_qemu_version(4, 0, 0, False):
+            test.cancel("This test is supported from qemu-kvm 4.0.0.")
+        if len(pf_info) < 2:
+            test.cancel("This test requires at least 2 PFs.")
```
yalzhang, line 78: I think it might be good to maybe check the value of /sys/class/net/$PF/device/sriov_totalvfs for PFs regardless of the PF's 'operate' value, and calculate the total value of 'sriov_totalvfs', then cancel the case only if the total value is less than 65.

Yingshun: @yalzhang That's a good idea! But we only have machines that have PFs with maximum '63' vfs. If we need to support other cards, we have to update the deployments, cases and codes. And if the max vfs value of a PF is not '63',  the current code will raise an exception when setting '63' to sriov_numvfs, we won't miss the error. So I'd like to keep the current way now. What's your opinion?

yalzhang: LGTM

yalzhang (approved):

PR #3758, opened 2021-09-06
Add memballoon case

chloerh:
- RHEL7-16778: [Memory management] Present memballoon statistic
period in live xml

### DIFF
--- a/libvirt/tests/src/memory/memballoon.py
+++ b/libvirt/tests/src/memory/memballoon.py
@@ -5,52 +5,152 @@ from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
 
 
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
+
+
 def run(test, params, env):
     """
     Test memballoon device
     """
-    vm_name = params.get('main_vm')
-    vm = env.get_vm(vm_name)
 
-    case = params.get('case', '')
-    status_error = "yes" == params.get('status_error', 'no')
+    def run_test_alias(case):
+        """
+        Test memballoon alias
 
-    bkxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+        :param case: test case
+        """
+        model = params.get('model')
+        alias_name = params.get('alias_name')
+        has_alias = 'yes' == params.get('has_alias', 'no')
 
-    try:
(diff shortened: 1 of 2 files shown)


chloerh:
PASS 1-type_specific.local.memballoon.period.memstat

Yingshun (approved):
LGTM

libvirt/tests/src/memory/memballoon.py
```diff
@@ -5,52 +5,163 @@
...
+        logging.info('No specific setup step for %s', case)
+
+    def cleanup_test_default(case):
+        """
+        Default cleanup for test cases
+
+        :param case: test case
+        """
+        logging.info('No specific cleanup step for %s', case)
+
```
kamilvarga, line 31: I am wondering why the setup and cleanup functions are listed there if they do nothing at a moment.

kamilvarga (approved):
LGTM.

libvirt/tests/src/memory/memballoon.py
chloerh: @kamilvarga  from line 150 to 156, I was actually using these 2 function as placeholder. In fact, I also think there's something could be done to avoid putting 2 empty functions here.  

How about:

1. Define another empty function to replace these 2 function, and it would be 'empty function -1' 
2. Move variables assigning code to `setup` function, move cleanup code in `finally` to cleanup function, then these 2 functions would not be empty.

Or any other solutions.

I'd really appreciate your opinion. I'd also ask other colleagues' opinions too. Thank you all for helping me polishing my code.  :grin:

kamilvarga: @chloerh I like the option 2, so from the code is obvious, what is happening during the setup, what part is related to test execution and what is happening at the end. However, the reason why I commented there is that I just don't like placeholders in the code. That is probably just my mental block from previous industry :), where such code is not allowed.

chloerh: @kamilvarga Thanks for the feedback. I like option 2 too! I'll make the change within this week.
About the placeholder, I found your comment very enlightening, we all have a lot to learn from you and your experience. So thanks again :smiley:

kamilvarga (approved):
LGTM.

PR #3770, opened 2021-09-09
nodedev-create-destroy: add comment about precondition

smitterl:
We found that due to performance issues with older mdevctl versions
the test might fail because it takes libvirt longer to identify the
ap_matrix device.

In order to avoid code to become more complex add a note about
pre-conditions for anybody who might run into this.


fbq815:
thanks for adding the comment, it make sense to me

chunfuwen (approved):
LGTM

fbq815 (approved):

PR #3773, opened 2021-09-10
iface_network: add version switch

smitterl:
This was fixed only with 7.0.0 and confirmed with devs
there are no plans to backport at this point because there's
a workaround.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_network.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_network.cfg
@@ -167,7 +167,9 @@
                             test_ipv4_address = "no"
                             variants:
                                 - default:
+                                    func_supported_since_libvirt_ver = (6, 6, 0)
                                 - bridge_type:
+                                    func_supported_since_libvirt_ver = (7, 0, 0)
                                     iface_model = "virtio"
                                     iface_type = "bridge"
                                     change_iface_option = "yes"
--- a/libvirt/tests/src/virtual_network/iface_network.py
+++ b/libvirt/tests/src/virtual_network/iface_network.py
@@ -653,16 +653,7 @@ TIMEOUT 3"""
     define_macvtap = "yes" == params.get("define_macvtap", "no")
     net_dns_forwarders = params.get("net_dns_forwarders", "").split()
 
-    # Destroy VM first
-    if vm.is_alive() and not update_device:
-        vm.destroy(gracefully=False)
-
-    # Back up xml file.
-    netxml_backup = NetworkXML.new_from_net_dumpxml("default")
-    iface_mac = vm_xml.VMXML.get_first_mac_by_name(vm_name)
-    params["guest_mac"] = iface_mac
-    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
-    vms_list = []
+    # Cancel if not yet supported in libvirt version under test
     if "floor" in ast.literal_eval(iface_bandwidth_inbound):
         if not libvirt_version.version_compare(1, 0, 1):
             test.cancel("Not supported Qos options 'floor'")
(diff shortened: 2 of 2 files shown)


smitterl:
```
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system virtual_network.iface_network.net_forward.net_nat.reconnect_tap.bridge_type                               
WARNING:root:No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.                                                        
WARNING:root:No python imaging library installed. PPM image conversion to JPEG disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
JOB ID     : de82c0282df5f6573b2fce29b6923db38ea49874
JOB LOG    : /root/avocado/job-results/job-2021-09-10T06.39-de82c02/job.log                                                  
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_forward.net_nat.reconnect_tap.bridge_type: CANCEL: This libvirt version doesn't support this function. (3.72 s)                                                          
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 1                                            
JOB TIME   : 4.03 s
```

smitterl:
@yalzhang Please, could you help reviewing this? 谢谢

libvirt/tests/cfg/virtual_network/iface_network.cfg
```diff
@@ -162,6 +162,7 @@
                             net2_dhcp_start_ipv4 = '192.168.150.2'
                             net2_dhcp_end_ipv4 = '192.168.150.254'
                         - reconnect_tap:
+                            func_supported_since_libvirt_ver = (7, 0, 0)
```
kylazhang, line 165: I think you should move this to - bridge_type: as default test seems are OK?

libvirt/tests/src/virtual_network/iface_network.py
```diff
@@ -675,6 +675,8 @@ def dig_test(session):
                         ' fixed with libvirt commit'
                         ' 876211ef4a192df1603b45715044ec14567d7e9f')
 
+    libvirt_version.is_libvirt_feature_supported(params)
```
kylazhang, line 678: I think you should move this line to the beginning of the main function.

libvirt/tests/cfg/virtual_network/iface_network.cfg
chunfuwen: it sound like it is specific to reconnect_tap cases, so it may be proper to put here

smitterl: @yalzhang Can you please help clarify?

libvirt/tests/src/virtual_network/iface_network.py
smitterl: Why? It's right after the another version check, is there a reason for this version check to be on top? Not sure I understand correctly.

libvirt/tests/cfg/virtual_network/iface_network.cfg
smitterl: Yes, this is specific to the reconnect_tap; bz1730084. @yalzhang Could confirm.

libvirt/tests/src/virtual_network/iface_network.py
yalzhang: I think it should be moved to L1159 after the "if reconnect_tap", otherwise it may impact other cases. Sorry to miss this version check when I push the original PR.

libvirt/tests/cfg/virtual_network/iface_network.cfg
yalzhang: Yes, it is only for reconnect_tap.

libvirt/tests/src/virtual_network/iface_network.py
kylazhang: @smitterl In fact I think we should move all version compare in front then no need to do the previous vm operation and xml backup operation. What do you think?

@yalzhang It will not affect others as if func_supported_since_libvirt_ver is None there will return true directly.

libvirt/tests/cfg/virtual_network/iface_network.cfg
kylazhang: I mean only set for reconnect_tap.bridge_type. Seems reconnect_tap.default should set another version check.

smitterl: OK; done :)

libvirt/tests/src/virtual_network/iface_network.py
smitterl: OK, I've moved all up right below the variable definitions because one check used a variable.

yalzhang (approved):

kylazhang (approved):

chunfuwen (approved):
LGTM

PR #3782, opened 2021-09-15
Update the test.fail() to test.cancel()

yalzhang:
It is better to use test.cancel() than test.fail() when the libvirt
version without the fix.


yalzhang:
@chloerh  Please help to review. It's a minor fix. Thank you!

chloerh (approved):

PR #3784, opened 2021-09-15
Add version check to interface check in failover related testing

yalzhang:
As there is a qemu bug which caused the issue that only 2 interfaces
are found on the vm, add a qemu version check in the function.

### DIFF
--- a/libvirt/tests/src/migration/sriov_migrate.py
+++ b/libvirt/tests/src/migration/sriov_migrate.py
@@ -69,8 +69,17 @@ def run(test, params, env):
         :raise: test.fail when interfaces' number is not equal to exp_num
         """
         if len(iface_list) != exp_num:
-            test.fail("%d interfaces should be found on the vm, "
-                      "but find %s." % (exp_num, iface_list))
+            # There is issue before qemu-kvm-5.2.0-10 that only 2 interfaces
+            # exists cancel test once qemu-kvm less than 5.2.0, refer to
+            # bug 1817003#c13
+            if not utils_misc.compare_qemu_version(5, 2, 0, is_rhev=False):
+                test.cancel("Cancel the test as there are only %s interfaces "
+                            "on vm, while expect is %s. It's caused by an qemu "
+                            "issue which was fixed since qemu-kvm-5.2.0" %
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
LGTM

chunfuwen:
If there is test result available, please upload that since it  is helpful to check some basic syntax by accident

libvirt/tests/src/migration/sriov_migrate.py
```diff
@@ -68,6 +68,13 @@ def check_vm_iface_num(iface_list, exp_num=3):
         :param exp_num: The expected number
         :raise: test.fail when interfaces' number is not equal to exp_num
         """
+        # There is issue before qemu-kvm-5.2.0-10 that only 2 interfaces exists
+        # cancel test once qemu-kvm less than 5.2.0, refer to bug 1817003#c13
+        if not utils_misc.compare_qemu_version(5, 2, 0, is_rhev=False):
+            test.cancel("Cancel the test as there are only %s interfaces on "
+                        "vm, while expect is %s. It's caused by an qemu issue "
+                        "which was fixed since qemu-kvm-5.2.0",
+                        len(iface_list), exp_num)
```
Yingshun, line 77: How about moving it to L79? I think it's better to compare the ifaces' number and cancel it on some  qemu versions. What do you think?

Yingshun, line 77:
```suggestion
                        "which was fixed since qemu-kvm-5.2.0"
                        % (len(iface_list), exp_num))
```

yalzhang: Thank you!  I mixed them again!

yalzhang: updated

Yingshun (approved):

kylazhang:
@yalzhang please give the test result for the cases related, thanks. Just lack the results, if pass, seem it can merge directly.

yalzhang:
It seems qemu has regression issue on 6.1 again. Let me close this PR, it is more reasonable to add it to the skip list.

PR #3788, opened 2021-09-17
Fix the log file can not found issue

yalzhang:
Fix the log file can not found issue in split daemon.


yalzhang:
before the fix, the case will fail with split daemon as the log configure set in libvirtd.conf, which is no use for split daemon:
virtual_network.iface_options.iface_type.type_vhostuser.multi_guests.interface_check.without_share_mem
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/avocado_jfg69cxl/libvirtd.log'&#10;

after the fix, the case can pass with split daemon, also with libvirtd
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_options.iface_type.type_vhostuser.multi_guests.interface_check.without_share_mem: PASS (138.50 s)
JOB TIME   : 141.53 s

yalzhang:
@Yingshun  Please help to review. And I found the virtqemud_conf.restore() can not restore the virtqemud.conf, do you know why? (In fact, others will not restores, either, but they can be restored in other ways just as we have talked.)

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -713,15 +725,18 @@ def check_tardev_xml(vm_name, set_tar_devs, set_sources, find=True):
...
-        if log_level:
-            libvirtd_conf["log_level"] = log_level
-            libvirtd_conf["log_filters"] = '"1:json 1:libvirt 1:qemu 1:monitor 3:remote 4:event"'
-        libvirtd_conf["log_outputs"] = '"1:file:%s"' % libvirtd_log_path
+        if utils_split_daemons.is_modular_daemon():
+            for drv in ['virtqemud', 'virtnetworkd', 'virtnodedevd',
+                        'virtnwfilterd', 'virtsecretd',
+                        'virtstoraged', 'virtinterfaced']:
+                drv_conf = utils_config.get_conf_obj(drv)
+                set_daemon_conf(drv_conf, log_level, libvirtd_log_path)
```
Yingshun, line 736: There is a recommended way(libvirt.customize_libvirt_config) to update and restore the configuration, would you like to try?

Yingshun:
> @Yingshun Please help to review. And I found the virtqemud_conf.restore() can not restore the virtqemud.conf, do you know why? (In fact, others will not restores, either, but they can be restored in other ways just as we have talked.)

To be honest, I'm not quite sure. I think may be we could try libvirt.customize_libvirt_config() first. It works well in migration test for now, so maybe it could solve your problem, I guess☺️

libvirt/tests/src/virtual_network/iface_options.py
yalzhang: Thank you for the suggestion, after update to libvirt.customize_libvirt_config, all case passed and the conf file recovered.

libvirt/tests/src/virtual_network/iface_options.py
```diff
@@ -713,15 +712,24 @@ def check_tardev_xml(vm_name, set_tar_devs, set_sources, find=True):
...
         logging.debug("libvirtd_log_path is %s", libvirtd_log_path)
-        libvirtd_conf = utils_config.LibvirtdConfig()
-        if log_level:
-            libvirtd_conf["log_level"] = log_level
-            libvirtd_conf["log_filters"] = '"1:json 1:libvirt 1:qemu 1:monitor 3:remote 4:event"'
-        libvirtd_conf["log_outputs"] = '"1:file:%s"' % libvirtd_log_path
+        conf_dict = {"log_level": log_level,
+                     "log_filters": "\"1:json 1:libvirt 1:qemu 1:monitor 3:remote 4:event\"",
+                     "log_outputs": "\"1:file:%s\"" % libvirtd_log_path}
+        if utils_split_daemons.is_modular_daemon():
```
chunfuwen, line 722: check  libvirt.customize_libvirt_config( ), it seem that it can detect whether it is in modular or monolith mode, so here we may not call if utils_split_daemons.is_modular_daemon()again. By passing damemon name, it will differ them itself.
```
virttest/utils_test/libvirt.py:3695:def customize_libvirt_config
```

chunfuwen, line 722: check  libvirt.customize_libvirt_config( ), it seem that it can detect whether it is in modular or monolith mode, so here we may not call if utils_split_daemons.is_modular_daemon()again. By passing damemon name, it will differ them itself.
```
virttest/utils_test/libvirt.py:3695:def customize_libvirt_config
```

PR #3791, opened 2021-09-18
Add interface cases with acpi

chloerh:
- RHEL-279587 - [acpi] Set ACPI index for interfaces
- RHEL-281192 - [acpi] Set ACPI index for interfaces:boundary/negative
values

### DIFF
--- a/libvirt/tests/src/virtual_network/network_misc.py
+++ b/libvirt/tests/src/virtual_network/network_misc.py
@@ -7,12 +7,16 @@ from avocado.utils import service
 
 from virttest import libvirt_version
 from virttest import utils_misc
+from virttest import utils_net
 from virttest import virsh
 
 from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices import interface
 from virttest.libvirt_xml.network_xml import NetworkXML
+from virttest.utils_test import libvirt
 
 DEFAULT_NET = 'default'
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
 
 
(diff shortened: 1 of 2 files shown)


chloerh:
Depends on https://github.com/avocado-framework/avocado-vt/pull/3230

libvirt/tests/src/virtual_network/network_misc.py
```diff
@@ -121,8 +127,91 @@ def run_test_zone_info():
         test_zone = eval('run_test_zone_%s' % case)
         test_zone()
 
+    def run_test_iface_acpi(case):
+        """
+        Test acpi setting of interface
+
+        :param case: test case
+        """
+        if not utils_misc.compare_qemu_version(6, 1, 0) and \
```
Yingshun, line 136: Do you need to add 'is_rhev=False' to utils_misc.compare_qemu_version()?
qemu-kvm-6.1.0-2.el9.x86_64 was installed on my local env, and the results are as follow:
```
>>> utils_misc.compare_qemu_version(6, 1, 0)
False
>>> utils_misc.compare_qemu_version(6, 1, 0, is_rhev=False)
True
>>> 
```

yalzhang (approved):
Have ran the cases and checked the logs, pass.

libvirt/tests/src/virtual_network/network_misc.py
chloerh: Thank you! Updated.

Yingshun (approved):

PR #3672, opened 2021-07-29
migrate_vm: Update hugepage number

cliping:
The value of 'remote_target_hugepages' need less than vm mermory.
If this value is too large, migration will successful.

Signed-off-by: lcheng <lcheng@redhat.com>

PR #3803, opened 2021-09-23
migrate_vm: simplify target hugepage size formula

smitterl:
Commit e6ff5b9c6d277696c884a3869ba38b5f22a5123b updated the formula
that calculates the needed target hugepages for successful and unsuccessful
migration with hugepages on the destination host.
This way it fixed failing negative test cases (unsuccessful migration).
However, now the positive test case fails on a system with
Guest VM memory MB: 1024
Hugepagesize: 1024 KB

The formula seems to assume a specific hugepage size (1024*2) and looks rather
complicated.

It is difficult to determine this number as it depends on the number
of already allocated hugepages locally and on the different hugepage sizes.
For the source host this is handled through the 'setup_hugepages = yes' parameter
in the test configuration that is used during 'env_process.py' to set up
a correct hugepage number so that vms can run successfully with memory backed
by hugepages.

If we assume a specific setup (s. comment in code) we can use a simpler
formula and still have both test scenarios (negative, postive) pass.

Additionally, log some global parameters that have an impact on how
'env_process.py' sets up the hugepages on the source host.

### DIFF
--- a/libvirt/tests/src/migration/migrate_vm.py
+++ b/libvirt/tests/src/migration/migrate_vm.py
@@ -1153,19 +1153,36 @@ def run(test, params, env):
     Test remote access with TCP, TLS connection
     """
 
-    def get_target_hugepage_num(params, mem):
+    def get_target_hugepage_num(params):
         """
-        Get the number of hugepage on target host
+        Get the number of hugepages to be set on target host
+
+        It is difficult to determine this number as it depends on the number
+        of already allocated hugepages locally and on the different hugepage sizes.
+
+        Therefore, let's assume
+        1. that the current hugepages_num was set up correctly for a single
+           vm to start correctly with hugepage memory backing.
+           (This is usually done by the setup_hugepages = yes setting and through
+           the env_process.py.)
+        2. Only a single vm is configured to run on source and on destination.
+        3. A negative value of 'remote_target_hugepages' indicates we want the
+           migration to fail.
+
+        Under these circumstances we can just choose the same hugepage_num that works
+        on the source host for the target host in case of positive testing (migration
+        succeeds) and a ridiculously low value of 1 hugepage on the target should be enough
+        to cause the migration to fail because the vm would not be backed by hugepages
+        on the target.
 
         :param params: The parameters used
-        :param mem: Current VM's memory
         :return: the number of hugepage to be allocated on target host
         """
         hugepage_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
         with open(hugepage_file, 'r') as fp:
             hugepage_num = int(fp.readline().strip())
         more_less_hp = int(params.get("remote_target_hugepages", "0"))
-        target_hugepage = min(int(mem/(1024*2)), hugepage_num) + more_less_hp
+        target_hugepage = hugepage_num if more_less_hp >= 0 else 1
         logging.debug("Number of huge pages on target host to be allocated:%d", target_hugepage)
         return target_hugepage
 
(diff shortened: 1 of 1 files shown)


smitterl:
```
# avocado run --vt-type libvirt --vt-machine-type s390-virtio virsh.migrate_vm.positive_testing.live_migration.with_hugepages.without_postcopy,virsh.migrate_vm.negative_testing.live_migration.with_hugepages.target_hugemem_less_than_vm_mem.without_postcopy --vt-connect-uri qemu:///system
JOB ID     : 8307bb9a881f5284237d08534ae623d0fed13848
JOB LOG    : /root/avocado/job-results/job-2021-09-23T11.44-8307bb9/job.log
 (1/2) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.positive_testing.live_migration.with_hugepages.without_postcopy: PASS (121.75 s)
 (2/2) type_specific.io-github-autotest-libvirt.virsh.migrate_vm.negative_testing.live_migration.with_hugepages.target_hugemem_less_than_vm_mem.without_postcopy: PASS (123.24 s)
JOB TIME   : 245.32 s
```

smitterl:
@cliping @chunfuwen @Yingshun Please have a look 谢谢

libvirt/tests/src/migration/migrate_vm.py
```diff
@@ -1153,19 +1153,36 @@ def run(test, params, env):
...
         :param params: The parameters used
-        :param mem: Current VM's memory
         :return: the number of hugepage to be allocated on target host
         """
         hugepage_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
         with open(hugepage_file, 'r') as fp:
             hugepage_num = int(fp.readline().strip())
         more_less_hp = int(params.get("remote_target_hugepages", "0"))
-        target_hugepage = min(int(mem/(1024*2)), hugepage_num) + more_less_hp
+        target_hugepage = hugepage_num if more_less_hp >= 0 else 1
```
chunfuwen, line 1185: As per https://github.com/autotest/tp-libvirt/pull/3672, it sounds like sometimes, remote_target_hugepages' need less than vm mermory, at least not directly set as host allocated hugepages(/proc/sys/vm/nr_hugepages).
Maybe can be:
```
target_hugepage = min(int(mem/(1024*n)), hugepage_num) + more_less_hp  if more_less_hp >= 0 else 1
```
Here n could be 2 ,or others depend on Hugepagesize Unit

fangge1212: My understanding:
1. If setup_hugepages is "yes" in cfg file(we already did this in migrate_vm.cfg), env_process will setup hugepages on host correctly so vm can start successfully. So we don't need to calculate hugepages number here in tp-libvirt unless vm memory size is changed in tp-libvirt.

2.In my experience, qemu overhead doesn't need hugepages, but this doesn't matter in this case

3.So I think this patch would work well.

fangge1212 (approved):

libvirt/tests/src/migration/migrate_vm.py
smitterl: > As per #3672, it sounds like sometimes, remote_target_hugepages' need less than vm mermory, at least not directly set as host allocated hugepages(cat /proc/sys/vm/nr_hugepages).
> Maybe can be:
> 
> ```
> factor = Hugepagesize // 1024 (grep Hugepagesize /proc/meminfo)
> target_hugepage = min(int(mem/(1024*factor)), hugepage_num) + more_less_hp  if more_less_hp >= 0 else 1
> ```
> 
> Here factor could be 2 ,or others(e.g 1) depending on Hugepagesize

@chunfuwen I thought of a similar solution. But then I thought this really complicates the code even more which also makes debugging harder. Therefore, I decided to supplier a simpler solution that works. Additionally, the comment and the added logging hopefully will give anybody who'd have to debug a failure some hints what to look into (the setup_hugepages logic with its parameters, and assumptions about the test environment).

chunfuwen (approved):
LGTM
Hopefully liping cheng can double check whether this change will impact her commit:e6ff5b9

chunfuwen:
@cliping , please help double check the change

PR #3813, opened 2021-09-29
Updated some checking for hooks in split daemon mode

jinqi7:

The daemon process for daemon mode is different from libvirtd  and updated it in the script. And remove the checking with the log  removed.


libvirt/tests/src/libvirt_hooks.py
```diff
@@ -267,8 +268,12 @@ def daemon_hook():
             # kill the daemon with SIGHUP
             if os.path.exists(hook_log):
                 os.remove(hook_log)
-            utils_misc.signal_program('libvirtd', 1,
-                                      '/var/run')
+            if utils_split_daemons.is_modular_daemon():
```
chunfuwen, line 271: Since this is more of common use case where parameter for one method is different due to is_modular_daemon().
Could you try to wrap one common method to allow it for common use?
e.g something like this
```
def adopt_func_on_ibvirtd(func):
      if utils_split_daemons.is_modular_daemon():
          func('virtqemu' , *args, **kargs)
     else:
          func('libvirtd' , *args, **kargs)
```

jinqi7:
@Yingshun Can you help to review this PR? Because it related to split daemon.

PR #3822, opened 2021-10-12
migrate_storage.py: update qemu-img create info with -F option

meinaLi:
In RHEL 8.6, qemu-img should specify the backing format when creating a new file based on backing file. So add -F option in this command, which don't affect the previous test result.


chunfuwen:
@meinaLi 
please upload test result when it is available

PR #3823, opened 2021-10-12
Fix virtual_disks usb failures due to detaching events detection

chunfuwen:
virtual_disks.usb.coldplug related cases fail due to unnessary event checking


chunfuwen:
This is only 3 lines change, and fixed solution is known solution, mark it as minor_fix

PR #3828, opened 2021-10-14
Fix Not found event tray-change issue

chunfuwen:
In extreme case, current 14 seconds may not be enough to receive event


chunfuwen:

PR #3832, opened 2021-10-20
Remove hypervisor_cpu_baseline cases from pseries tests

chloerh:
computing baseline hypervisor CPU is not supported for arch ppc64le


dzhengfy (approved):
Due to the xml files are from x86_64 which are not applied to ppc64le, we need to exclude those cases on ppc, too.
LGTM

PR #3839, opened 2021-10-26
libvirt_rng: relax test condition inside guest

smitterl:
The test case currently expects virtio to be rng hw_device
that's selected by the kernel but this depends on the
kernel code. Let's just check that some hw_device is set
and the virtio device is available. The way the kernel selects
the device is unlikely to change unintentionally and
the current_rng can always be set manually.


smitterl:
Closing in favor of https://github.com/autotest/tp-libvirt/pull/3946

PR #3860, opened 2021-11-09
v2v: fix the failed cases with error 'Not find cirrus device'

vwu-vera:
Fix the failed TCs with error 'Not find cirrus device'.

PR #3862, opened 2021-11-10
Revert "v2v: fix the failed cases with error 'Not find cirrus device'"

xiaodwan:
Reverts autotest/tp-libvirt#3860


xiaodwan:
The changes in PR#3860 will cause other cases failure, so revert it.

PR #3918, opened 2021-12-08
virsh_hypervisor_cpu_baseline: unsupport on ppc

dzhengfy:


dzhengfy:
Too simple, so no need test result. Those cases will not run on ppc.

Yingshun (approved):

PR #3956, opened 2021-12-23
Update mtu 4096

yalzhang:
The mtu '6000' or '4096' both can set successfully on x86 systems.
But on big endian systems, '4096' is judged as invalid as one kernel
bug 2026234. Update this test parameter to '4096' to cover this bug.


smitterl (approved):
LGTM

Yingshun (approved):

PR #3972, opened 2022-01-04
blockcommit_blockpull: Rename rbd image name

Yingshun:
If the case does not finish normally for some reasons(bug or
cancelled by user), the rbd image might unexpectedly remain in
the ceph server. To avoid using too many quotas, update to name
the image with distribution details.

Case failed due to a known issue, but the image name is updated as expected.
```
[root@libvirtceph ~]# rbd ls blockcommit-pool    
rbd_rhel_8_<release>_x86_64.img                   <-----the img exists before a case is executed, <release> is release version.

[root@libvirtceph ~]# rbd ls blockcommit-pool      <---- the image is deleted and recreated during the case execution
[root@libvirtceph ~]# rbd ls blockcommit-pool
rbd_rhel_8_<release>_x86_64.img
```

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_blockcommit.py
@@ -5,6 +5,7 @@ import collections
 
 import aexpect
 
+from avocado.utils import distro
 from avocado.utils import process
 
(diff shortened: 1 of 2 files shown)


chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

PR #3974, opened 2022-01-05
[minor fix] virsh_undefine: skip no_option test for aarch64

liuyd96:
aarch64 requires '--nvram' option to undefine guest and that conflicts with
no_option. So skip this test for aarch64


chunfuwen (approved):
LGTM

chunfuwen:
filter this case on aarch64 ,and mark it as minor_fix

PR #4012, opened 2022-01-26
Fix event not found in libvirtd log

chunfuwen:
Having libvirtd log only makes sense under monolithic libvirtd mode

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -17,6 +17,7 @@ from virttest import nfs
 from virttest import utils_libvirtd
 from virttest import utils_misc
 from virttest import utils_disk
+from virttest import utils_split_daemons
 from virttest import data_dir
 from virttest import utils_selinux
 from virttest import utils_package
@@ -1997,7 +1998,7 @@ def run(test, params, env):
                 if not disk_detach_error:
                     utils_misc.wait_for(_check_disk_detach, timeout=20)
                 # Give time to log file to collect more events
-                if virtio_disk_hot_unplug_event_watch:
+                if virtio_disk_hot_unplug_event_watch and not utils_split_daemons.is_modular_daemon():
                     result = utils_misc.wait_for(lambda: check_info_in_libvird_log_file('"event": "DEVICE_DELETED"'), timeout=20)
                     if not result:
                         test.fail("Failed to get expected messages from log file: %s."


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multidisks.hotplug.single_disk_test.virtio_disk_hot_unplug_event_watch

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1997,7 +1998,7 @@ def _check_disk_detach():
                 if not disk_detach_error:
                     utils_misc.wait_for(_check_disk_detach, timeout=20)
                 # Give time to log file to collect more events
-                if virtio_disk_hot_unplug_event_watch:
+                if virtio_disk_hot_unplug_event_watch and not utils_split_daemons.is_modular_daemon():
```
dzhengfy, line 2001: I have concern with the fix strategy for this one. Checking libvirt daemon log for the event should be working. The key is we need to use the correct log file in modular daemon environment

chunfuwen: @dzhengfy 
For log checking, I don't think we need enable all log checking cases into modular daemon mode one by one since logging is generic framework. Actually I assume there  are specialized  cases developed by yafu to cover logging when we enable modular daemon.
@yafu-1 
What's your opinion ?

yafu-1: I agree with @dzhengfy 's option. And it needs to check specialized log, which needs to deal cases by cases.

dzhengfy: @qiankehan As this is related virtual-disk feature, could you help confirm if this log checking is needed on modular daemon environment for this scenario? Thanks

qiankehan: No need here. For single libvirtd daemon, find the DEVICE_DELETED qmp event in the log of libvirtd; For split daemons, find the event in log of virtqemud. That is the only difference.



> @qiankehan As this is related virtual-disk feature, could you help confirm if this log checking is needed on modular daemon environment for this scenario? Thanks

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1997,7 +1998,7 @@ def _check_disk_detach():
...
                     result = utils_misc.wait_for(lambda: check_info_in_libvird_log_file('"event": "DEVICE_DELETED"'), timeout=20)
```
qiankehan, line 2002: After the PR, the code would be 
```python
if virtio_disk_hot_unplug_event_watch and not utils_split_daemons.is_modular_daemon():
    result = utils_misc.wait_for(lambda: check_info_in_libvird_log_file('"event": "DEVICE_DELETED"'), timeout=20)
```
That means the checking of "DEVICE_DELETED" would be **skipped** for split daemons.  However, it is necessary for split daemons, too.
The pseudocode should be like:
```python
if virtio_disk_hot_unplug_event_watch:
    if Libvirtd:
        Update libvirtd.conf
        Restart libvirtd
    if SplitDaemons:
        Update virtqemud.conf
        Restart virtqemud
```

qiankehan (changes requested):

chunfuwen:
close it since it may not be used anymore

PR #4021, opened 2022-01-27
v2v: update the dependency checking command

xiaodwan:
'yum deplist xxx' sometimes cannot query out correct result because
the db is not latest. I'm not sure how it happens. e.g. if you install
the pkg by a url link instead of repos, it always query from the old
package. so I think 'rpm -qR xxx' is better than that in this case.


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.v2v_options.positive_test.cmd_free.deplist.input_mode.none.output_mode.none: PASS (4.51 s)

chunfuwen (approved):
LGTM

PR #4026, opened 2022-02-07
backingchain:add case for blockcommit 

nanli1:
Add new case for RHEL7-16568:test blockcommit from top
 to base after creating external disk-only snapshot with "block" disk type

### DIFF
--- a/libvirt/tests/src/backingchain/blockcommit_basic_function.py
+++ b/libvirt/tests/src/backingchain/blockcommit_basic_function.py
@@ -1,0 +1,81 @@
+import logging
+
+from avocado.utils import process
+
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_test import libvirt
+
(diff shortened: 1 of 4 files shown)


libvirt/tests/src/backingchain/blockcommit_basic_function.py
```diff
@@ -0,0 +1,81 @@
...
+
+from provider.backingchain import blockcommand_base
+from provider.backingchain import check_functions
+
+LOG = logging.getLogger('avocado.backingchain.checkfunction')
+
+
+def run(test, params, env):
+    """
+    Case for RHEL-288322
```
Yingshun, line 17: test

nanli1: eeeee

nanli1: 111

PR #4028, opened 2022-02-07
Add polarion decorator to output polarion case id when execute specif…

chunfuwen:
…ic case

This may help map automation case with specific polarion case id

### DIFF
--- a/provider/decorator.py
+++ b/provider/decorator.py
@@ -1,0 +1,26 @@
+"""
+decorators to wrap up function behaviors
+"""
+
+import logging
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def polariondecorator(polarion_id):
+    """
+    one decorator to output polarion case id when execute specific test case
+    method
+    usage:
+        @polariondecorator("RHEL7-110402")
+        def check_multifunction_is_on(vm_name, test):
+
+    :param polarion_id: polarion case id
+    """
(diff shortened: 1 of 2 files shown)


chunfuwen:
Sometimes we may face some problems when mapping automation method with specific polarion case id, this decorator will help clarify them by applying polarion case id on specific automation method, and the decorator will output polarion id in the log.
```
@polariondecorator("RHEL7-110402")
def check_multifunction_is_on(vm_name, test):
    """
    Check multifunction is on

    :param vm_name: VM name
    :param test: test assert object
    """
    result = virsh.dumpxml(vm_name).stdout_text.strip()
    LOG.debug("test dumpxml:%s", result)
    pattern_str = r"multifunction=.*on.*"
    if not re.search(pattern_str, result):
        test.fail("multifunction should be set to 'on' if two disks share pci slot")
```
output follow as below
```
Found/Verified IP 192.168.122.35 for VM avocado-vt-vm1 NIC 0
Execute*********************polarion case id:RHEL7-110402

```

chunfuwen:
@dzhengfy 
@kylazhang 
This just provides one way to help mapping between automation method with polarion case id.
If someone feel it useful, just take easy to use it

chunfuwen:
close this PR since it depends on polarion id, and it still not clear whether this id is proper to be exposed.

PR #4035, opened 2022-02-09
Fix disks_duplicate_target_different_bus_error_test case failure

chunfuwen:
Make sure that VM is completely booted before detach operation

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -1973,6 +1973,20 @@ def run(test, params, env):
                 dt_options = ""
                 if devices[i] == "cdrom":
                     dt_options = "--config"
+                # Make sure that VM is completely booted before detach operation
+                if vm.is_alive():
+                    vm.wait_for_login().close()
+
+                def _check_disk(target):
+                    """
+                    Check disk with specific target
+
+                    :param target: target device
+                    """
+                    return target in vm.get_blk_devices()
+
+                utils_misc.wait_for(lambda: _check_disk(device_targets[i]), 10, 3)
+
                 ret = virsh.detach_disk(vm_name, device_targets[i],
                                         dt_options, wait_remove_event=True,
                                         **virsh_dargs)


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multidisks.hotplug.multi_disks_test.disks_duplicate_target_different_bus_error_test

chunfuwen:
two lines changes, mark it as minor_fix

cliping (approved):
LGTM

dzhengfy:
@chunfuwen ， If the existing failure information for this case could be provided, that will be helpful for reviewers to know the problem the case happens to.

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1973,6 +1973,20 @@ def test_device_update_boot_order(disk_type, disk_path, error_msg):
...
+
+                def _check_disk(target):
+                    """
+                    Check disk with specific target
+
+                    :param target: target device
+                    """
+                    return target in vm.get_blk_devices()
+
+                utils_misc.wait_for(lambda: _check_disk(device_targets[i]), 10, 3)
```
dzhengfy, line 1988: Maybe here you can test the return value of wait_for() to decide if  a test.fail(error) should be raised, which will ensure the target exists before detaching.

chunfuwen: @dzhengfy 
This is common practice we discussed multiple times previously

chunfuwen:
> @chunfuwen ， If the existing failure information for this case could be provided, that will be helpful for reviewers to know the problem the case happens to.

@dzhengfy 
This is common practice we discussed multiple times previously

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
chunfuwen: @dzhengfy 
Checking whether it failed or not will be made decision by just following line virsh.detach_disk 
We don't want to change current test case work flow

dzhengfy (approved):
LGTM

PR #4040, opened 2022-02-10
Add sleep between event track and trigger

kylazhang:

```
 (1/1) type_specific.io-github-autotest-libvirt.virsh.net_event.positive_test.no_timeout.loop.lifecycle_event.default: PASS (7.58 s)
```

### DIFF
--- a/libvirt/tests/src/virsh_cmd/network/virsh_net_event.py
+++ b/libvirt/tests/src/virsh_cmd/network/virsh_net_event.py
@@ -131,6 +131,8 @@ def run(test, params, env):
                                      options=net_event_option, **virsh_dargs)
             utlv.check_exit_status(result, status_error)
 
+        time.sleep(1)
+
         if not status_error:
             # Verify 'lifecycle' events
             if not net_event_list and net_event_name == 'lifecycle':


libvirt/tests/src/virsh_cmd/network/virsh_net_event.py
```diff
@@ -131,6 +131,8 @@ def check_output(output, expected_event_list):
                                      options=net_event_option, **virsh_dargs)
             utlv.check_exit_status(result, status_error)
 
+        time.sleep(1)
```
chunfuwen, line 134: is there any way to avoid hard coded wait time, for example check network status ?

kylazhang: As there only need to wait for a small time just to delay the event check a little later, so I think it is acceptable to just sleep 1s here.

dzhengfy (commented):
Could you describe what the problem was before your fix? I am curious about which operation requires additional wait, virsh.net_event() ? or virsh_session.sendline()?

kylazhang:
> Could you describe what the problem was before your fix? I am curious about which operation requires additional wait, virsh.net_event() ? or virsh_session.sendline()?

The issue is that sometimes the event can not be get correctly, just as the trigger event is too fast that the net event can not get the first event.

chunfuwen (approved):
LGTM

dzhengfy (approved):
Understand and LGTM

PR #4063, opened 2022-02-21
migrate_options_shared: Update error message

cliping:
As the error message changes in new version, update the error message.

Signed-off-by: lcheng <lcheng@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/migration/migrate_options_shared.cfg
+++ b/libvirt/tests/cfg/migration/migrate_options_shared.cfg
@@ -546,7 +546,8 @@
                     update_tpm_secret = yes
                     src_secret_value = "sec value test"
                     dst_secret_value = "sec value diff"
-                    err_msg = "qemu-kvm: tpm-emulator: Setting the stateblob \(type 1\) failed with a TPM error 0x21 decryption error"
+                    vm_err_msg = "qemu-kvm: tpm-emulator: Setting the stateblob \(type 1\) failed with a TPM error 0x21 decryption error"
+                    err_msg = "operation failed: domain is no longer running"
                 - p2p_migration:
                     virsh_migrate_options = "--live --p2p --verbose"
                     variants:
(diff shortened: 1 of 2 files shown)


cliping:
Before fix:
TestFail: Can not find the expected patterns 'qemu-kvm: tpm-emulator: Setting the stateblob \(type 1\) failed with a TPM error 0x21 decryption error' in output 'Migration: [  1 %]Migration: [  5 %]Migration: [  8 %]Migration: [ 10 %]Migration: [ 13 %]Migration: [ 16 %]Migration: [ 20 %]Migration: [ 23 %]Migration: [ 75 %]Migration: [ 89 %]error: operation failed: domain is no longer running'

After fix:
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_options_shared.negative_test.vtpm_emulator.without_postcopy: PASS (125.85 s)
JOB TIME   : 126.53 s

kylazhang:
Hi @cliping ,
what's the cause for the different error msg? I think it is not better to judge the msg with distro version as it can also be used as fedora or other release. Could we judge the some package version?

libvirt/tests/cfg/migration/migrate_options_shared.cfg
```diff
@@ -546,7 +546,8 @@
                     update_tpm_secret = yes
                     src_secret_value = "sec value test"
                     dst_secret_value = "sec value diff"
-                    err_msg = "qemu-kvm: tpm-emulator: Setting the stateblob \(type 1\) failed with a TPM error 0x21 decryption error"
+                    vm_err_msg = "qemu-kvm: tpm-emulator: Setting the stateblob \(type 1\) failed with a TPM error 0x21 decryption error"
```
chunfuwen, line 549: This is similar situation we had met before that error message is subject to changes.
I assume we have some common understanding  that the way to move this forward is to use wild matching pattern, which focus on assert key words in the error message

cliping:
Discuss with feature owner, file a bug about this error message. Thanks.

https://bugzilla.redhat.com/show_bug.cgi?id=2061275

PR #4066, opened 2022-02-23
After modular daemons is enabled, daemon config file and assicated lo…

chunfuwen:
…g is changed

The code is to adapt the caused changes

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
@@ -15,6 +15,7 @@ from virttest import virsh
 from virttest import remote
 from virttest import nfs
 from virttest import utils_libvirtd
+from virttest import utils_split_daemons
 from virttest import utils_misc
 from virttest import utils_disk
 from virttest import data_dir
@@ -24,6 +25,7 @@ from virttest import utils_package
 from virttest.utils_test import libvirt
 from virttest.utils_config import LibvirtQemuConfig
 from virttest.utils_config import LibvirtdConfig
+from virttest.utils_config import VirtQemudConfig
 
 from virttest.libvirt_xml import vm_xml, xcepts
 from virttest.libvirt_xml.devices.disk import Disk
@@ -862,7 +864,7 @@ def run(test, params, env):
         log_outputs = "1:file:%s" % log_config_path
         libvirtd_config.log_outputs = log_outputs
         libvirtd_config.log_filters = "1:json 1:libvirt 1:qemu 1:monitor 3:remote 4:event"
-        utils_libvirtd.libvirtd_restart()
+        utils_libvirtd.Libvirtd('virtqemud').restart()
 
     def check_info_in_libvird_log_file(matchedMsg=None):
         """
@@ -1058,7 +1060,7 @@ def run(test, params, env):
     # Configure libvirtd log level and path.
     log_file = params.get("log_file", "libvirtd.log")
     log_config_path = os.path.join(data_dir.get_data_dir(), log_file)
-    libvirtd_config = LibvirtdConfig()
+    libvirtd_config = VirtQemudConfig() if utils_split_daemons.is_modular_daemon() else LibvirtdConfig()
 
     if virtio_disk_hot_unplug_event_watch:
         config_libvirtd_log()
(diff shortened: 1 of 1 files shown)


chunfuwen:
Monolithic result follow as:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multidisks.hotplug.single_disk_test.virtio_disk_hot_unplug_event_watch

Modolarity daemons result:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.multidisks.hotplug.single_disk_test.virtio_disk_hot_unplug_event_watch

libvirt/tests/src/virtual_disks/virtual_disks_multidisks.py
```diff
@@ -1058,7 +1060,7 @@ def check_info_in_libvird_log_file(matchedMsg=None):
     # Configure libvirtd log level and path.
     log_file = params.get("log_file", "libvirtd.log")
     log_config_path = os.path.join(data_dir.get_data_dir(), log_file)
-    libvirtd_config = LibvirtdConfig()
+    libvirtd_config = utils_config.get_conf_obj('virtqemud') if utils_split_daemons.is_modular_daemon() else LibvirtdConfig()
```
Yingshun, line 1063: utils_config.get_conf_obj() returns LibvirtdConfig() if config_type is 'libvirtd', so how about changing it like
`libvirtd_config = utils_config.get_conf_obj(utils_libvirtd.Libvirtd('virtqemud').service_name)`?

chunfuwen: @Yingshun 
It sounds like current code is more of readable and understandable using  utils_split_daemons.is_modular_daemon() to differentiate mononthic and modular situation

Yingshun: Yes, you can get 'virtqemud' or 'libvirtd' by calling utils_split_daemons.is_modular_daemon(), but i think it's better to directly pass the config type to utils_config.get_conf_obj(). Because it's a wrapper function to get different configuration object by the given config_type. In other words, utils_config.get_conf_obj() returns LibvirtdConfig() if you pass 'libvirtd' and returns VirtQemudConfig() if pass 'virtqemud'.

chunfuwen: @Yingshun 
Quote above comment "but i think it's better to directly pass the config type to utils_config.get_conf_obj(). Because it's a wrapper function to get different configuration object by the given config_type"

Could you share one link with commonly recognized practice that support this comment ?

Let's go back your suggestion:
libvirtd_config = utils_config.get_conf_obj(utils_libvirtd.Libvirtd('virtqemud').service_name)

When reader take a glance at this code,  reader may understand that utils_libvirtd.Libvirtd('virtqemud').service_name may help return one string "virtqemud". But now it give reader one surprise why I get legacy libvirtd_config by string "virtqemud" instead of "libvirtd".

Yingshun: @chunfuwen I said that you can get 'virtqemud' and 'libvirtd' according to the environment setting...
But I don't quite understand what your point is? Could you explain it?

Yingshun: @dzhengfy  what do you think?

dzhengfy: @chunfuwen , 
`utils_libvirtd.Libvirtd(x).service_name` is recommended usage to get which daemon is used in current environment, 'libvirtd' or 'virtqemud' or others. This is already documented in our` Libvirt automation tips in upstream` google document (see No.21), and shared in whole team.

We all hope the case codes could be easy as possible without knowing the test environment's details, like mononthic or modular mode, even though there might be some exceptions. So we designed some avocado-vt utilities to help to shield the details. `utils_libvirtd.Libvirtd(x).` does just serve for that purpose. So utils_libvirtd.Libvirtd(x) is recommended to use for daemon related operations. Btw, utils_config.get_conf_obj(xx) can return LibvirtdConfig if 'libvirtd' is passed, 
`libvirtd_config = utils_config.get_conf_obj('virtqemud') if utils_split_daemons.is_modular_daemon() else LibvirtdConfig()`
so here your codes seems a little duplicated.

chunfuwen: @dzhengfy 
@Yingshun 
As discussed on the meeting, VirtQemudConfig() is used directly instead of utils_config.get_conf_obj('virtqemud')

dzhengfy (approved):
LGTM

Yingshun (approved):

PR #4080, opened 2022-03-07
domcapabilities_output: add case to remove ovmf path

dzhengfy:
Case ID: RHEL-202492
Test scenario:
Remove ovmf directory and check domcapabilities output does not
include that entry and does include when it comes back.

### DIFF
--- a/libvirt/tests/cfg/host_hypervisor/domcapabilities_output.cfg
+++ b/libvirt/tests/cfg/host_hypervisor/domcapabilities_output.cfg
@@ -1,0 +1,5 @@
+- domcapabilities.output:
+    type = domcapabilities_output
+    start_vm = no
+    variants test_case:
+        - rm_ovmf_path:
--- a/libvirt/tests/src/host_hypervisor/domcapabilities_output.py
+++ b/libvirt/tests/src/host_hypervisor/domcapabilities_output.py
@@ -1,0 +1,73 @@
+import os
+
+from virttest.libvirt_xml import domcapability_xml
+
+
+def get_ovmf_path():
+    """
+    Retrieve the ovmf path in virsh domcapabilities
+
+    :returns: str, ovmf path, otherwise None
+    """
+    domcapa_xml = domcapability_xml.DomCapabilityXML()
+    ovmf_nodes = domcapa_xml.xmltreefile.findall('/os/loader/value')
+    if not ovmf_nodes:
+        return None
+    return ovmf_nodes[0].text
+
+
+def test_rm_ovmf_path(test):
+    """
+    Remove ovmf path and check domcapabilities can be updated accordingly
+
+    Steps:
+    1. Get ovmf path in virsh domcapabilities
+    2. Rename the directory of ovmf path
+    3. Check virsh domcapabilities output does not include ovmf path entry
+    4. Recover the directory of ovmf path
+    5. Check virsh domcapabilities output includes ovmf path entry again
+
+    :param test: test object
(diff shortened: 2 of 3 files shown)


dzhengfy:
(1/1) type_specific.io-github-autotest-libvirt.domcapabilities.misc.rm_ovmf_path: PASS (7.69 s)

libvirt-8.0.0-6.el9.x86_64

libvirt/tests/src/host_hypervisor/domcapabilities_misc.py
```diff
@@ -0,0 +1,75 @@
+import logging
+import os
+
+from virttest.libvirt_xml import domcapability_xml
+
+
+logger = logging.getLogger('avocado.' + __name__)
```
Yingshun, line 7: I think we'd better to name it with uppercase because it's a value to be used as constant, what do you think?

Yingshun (commented):
Others LGTM.

libvirt/tests/src/host_hypervisor/domcapabilities_misc.py
dzhengfy: Do you mean 'logger' is a constant? I think it is just a log object, so it has debug() function.
Perhaps you think it is better using the name 'logging' as other existing python codes are using?

Yingshun: I don't think the best option is 'logging'. But we had too many tp-libvirt scripts at that time, I had to choose a way to solve the problem with the least effort and added comments as below:
# Using as lower capital is not the best way to do, but this is just a
# workaround to avoid changing the entire file.

We used 'LOG' in avocadot-vt, so Ithink we'd better to name it with uppercase for new created script. BTW, it's a good idea to use 'test.log'. I think.

dzhengfy: I still do not  understand why LOG is used in avocado-vt. But test.log seems a good choice. I updated with test.log.

Yingshun: Yes, I agree.

Yingshun (approved):
LGTM

libvirt/tests/cfg/host_hypervisor/domcapabilities_misc.cfg
```diff
@@ -0,0 +1,5 @@
+- domcapabilities.misc:
```
chunfuwen, line 1: we have libvirt/tests/cfg/virsh_cmd/host/virsh_domcapabilities.cfg to reside domcapabilities  related features cases, Could this scenario be fit in this file ?

chunfuwen: That 's fine to separate another cfg file. Just a little wonder about domcapabilities_misc.cfg, could it be something specific domcapabilities_xx.cfg since  'misc' means that no matter what  various types or various different sources, they are be put it in.

chunfuwen (commented):
see inline comment

libvirt/tests/cfg/host_hypervisor/domcapabilities_misc.cfg
dzhengfy: Change the file name and add docstring to say this pytyon file is used for the test scenarios which need to check outputs of virsh domcapabilities under certain configuration. Thanks.  @chunfuwen

chunfuwen (approved):
LGTM

PR #4083, opened 2022-03-08
sriov: Fix an UnboundLocalError issue - iface_name

Yingshun:
If there is no iface_name got by utils_net.get_linux_ifname,
it'll raise TestError, which breaks the iface_name initialization.
The 'Exception' is too common, update to a proper one in try-except
block and correct logging.warning format.

**Test result:**
_Before the fix:_
`UnboundLocalError: local variable 'iface_name' referenced before assignment&#10;`

_After the fix:_
` (1/1) type_specific.io-github-autotest-libvirt.sriov.normal_test.guest_with_vf.hotplug.default.mac_check.vf.managed_no: PASS (66.24 s)`

### DIFF
--- a/libvirt/tests/src/sriov/sriov.py
+++ b/libvirt/tests/src/sriov/sriov.py
@@ -148,21 +148,24 @@ def run(test, params, env):
             vm.cleanup_serial_console()
         vm.create_serial_console()
         session = vm.wait_for_serial_login(timeout=240)
+        iface_name = utils_misc.wait_for(
+            lambda: utils_net.get_linux_ifname(session, mac_addr), 30,
+            first=5, ignore_errors=True)
+        if not iface_name:
+            test.fail("no interface with MAC address %s found" % mac_addr)
 
         def get_ip():
             return utils_net.get_guest_ip_addr(session, mac_addr)
 
         try:
             ip_addr = ""
-            iface_name = utils_net.get_linux_ifname(session, mac_addr)
-            if iface_name is None:
-                test.fail("no interface with MAC address %s found" % mac_addr)
             session.cmd("pkill -9 dhclient", ignore_all_errors=True)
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
LGTM

libvirt/tests/src/sriov/sriov.py
```diff
@@ -148,21 +148,24 @@ def get_ip_by_mac(mac_addr, timeout=120):
...
-            if iface_name is None:
-                test.fail("no interface with MAC address %s found" % mac_addr)
             session.cmd("pkill -9 dhclient", ignore_all_errors=True)
             session.cmd("dhclient %s " % iface_name, ignore_all_errors=True)
             ip_addr = utils_misc.wait_for(get_ip, 20)
             logging.debug("The ip addr is %s", ip_addr)
-        except Exception:
-            logging.warning("Find %s with MAC address %s but no ip for it" % (iface_name, mac_addr))
+        except utils_net.IPAddrGetError:
+            logging.warning("Find %s with MAC address %s but no ip for it",
```
dzhengfy, line 167: Is it a checkpoint for getting ip? Do we need to raise it?

dzhengfy (commented):
Others LGTM

libvirt/tests/src/sriov/sriov.py
Yingshun: Thank you for your comment, confirmed with feature owner, it's not a checkpoint of this case(we have other cases to check connectable), so it's ok to log a messge with warn level.
I'm touching these 2 lines because 'Exception' is too general, we should point to a specific one( Personally, I think we should avoid such coding). And the other thing is that the previous logging format is incorrect. It's not for logic or something.

Yingshun: @dzhengfy could you please re-review this PR?

dzhengfy (approved):
LGTM

PR #4091, opened 2022-03-10
numa_node_memory_bind: expose test.cancel

dzhengfy:
Original codes make test.cancel exposed as test.error finally which is unclear.
So this fix is to expose test.cancel again.

### DIFF
--- a/libvirt/tests/src/numa/numa_node_memory_bind.py
+++ b/libvirt/tests/src/numa/numa_node_memory_bind.py
@@ -11,11 +11,13 @@ from virttest.utils_test import libvirt
 logging = log.getLogger('avocado.' + __name__)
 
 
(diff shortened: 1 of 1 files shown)


Yingshun (approved):

dzhengfy:
After the fix:
 (1/2) type_specific.io-github-autotest-libvirt.numa_node_memory_bind.vcpu_bind: CANCEL: The host only has 1 numa node, but 2 is required at least (47.72 s)
 (2/2) type_specific.io-github-autotest-libvirt.numa_node_memory_bind.placement_bind: CANCEL: The host only has 1 numa node, but 2 is required at least (48.23 s)

Before the fix:

cliping (approved):
LGTM

PR #4093, opened 2022-03-10
Fix the attach xml can not find error for virtual network

kylazhang:

```
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_target.flush_after_detach.flush_with_occupation.default: PASS (57.17 s)
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_ovs.net_ovs.ovs_qos.hotplug_iface: PASS (8.14 s)
```

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_ovs.py
+++ b/libvirt/tests/src/virtual_network/iface_ovs.py
@@ -1,11 +1,13 @@
 import logging as log
 import ast
+import os
 import re
 import shutil
 
 from avocado.utils import distro
 from avocado.utils import process
 
+from virttest import data_dir
 from virttest import virt_vm
 from virttest import virsh
 from virttest import utils_net
@@ -142,7 +144,8 @@ def run(test, params, env):
             if not hotplug:
                 libvirt.modify_vm_iface(vm_name, 'update_iface', iface_dict)
             else:
-                iface_attach_xml = libvirt.modify_vm_iface(vm_name, 'get_xml', iface_dict)
+                iface_attach_xml = os.path.join(data_dir.get_data_dir(), "iface_attach.xml")
+                shutil.copyfile(libvirt.modify_vm_iface(vm_name, 'get_xml', iface_dict), iface_attach_xml)
                 libvirt_vmxml.remove_vm_devices_by_type(vm, 'interface')
 
         try:
@@ -191,4 +194,6 @@ def run(test, params, env):
         # Try to recovery ovs bridge
         if utils_net.ovs_br_exists(bridge_name):
             utils_net.del_ovs_bridge(bridge_name)
+        if "iface_attach_xml" in locals():
+            os.remove(iface_attach_xml)
         vmxml_backup.sync()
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_network/iface_target.py
```diff
@@ -150,7 +153,8 @@ def check_target_name(index, direct=False):
         if_dict = prepare_iface_dict(test_macvtap)
         mac_addr = utils_net.generate_mac_address_simple()
         if_dict['mac'] = mac_addr
-        iface_add_xml = libvirt.modify_vm_iface(vm_name, 'get_xml', if_dict)
+        iface_add_xml = os.path.join(data_dir.get_data_dir(), "iface_add.xml")
```
dzhengfy, line 156: When is the file deleted?

libvirt/tests/src/virtual_network/iface_ovs.py
```diff
@@ -142,7 +144,8 @@ def check_ovs_port(ifname, brname):
             if not hotplug:
                 libvirt.modify_vm_iface(vm_name, 'update_iface', iface_dict)
             else:
-                iface_attach_xml = libvirt.modify_vm_iface(vm_name, 'get_xml', iface_dict)
+                iface_attach_xml = os.path.join(data_dir.get_data_dir(), "iface_attach.xml")
```
dzhengfy, line 147: When is the file deleted?

chunfuwen, line 147: It sounds like it only happens when hotplug.
I just guess whether it is related to the situation that VM is not completely booted.
Maybe we can add some logic here to ensure VM completely booted

kylazhang: I think it do not related to the vm start? The error is xml can not be found. It happened when hotplug because hotplug need the xml file.

kylazhang: Ok. I will think about delete them at finally.

kylazhang: Updated

libvirt/tests/src/virtual_network/iface_target.py
kylazhang: Updated

libvirt/tests/src/virtual_network/iface_ovs.py
chunfuwen: Yes,  we know it happen on line 157 when calling virsh.attach_device, 
But before that we just  call vm.start(), and subsequently call virsh.attach_device, so in anyway, we may need ensure VM is completely booted before call  virsh.attach_device

kylazhang: Hi Chunfuwen, maybe that's another issue for hotplug. But I think if that happened, there will be another failure occurred different from "attach/detach xml can not be found" in my discussion. So I think or we can waiting for that issues happened then deal with it? As we can not find such issues now for this test. Thanks

chunfuwen (approved):
LGTM

PR #4097, opened 2022-03-11
numa_memory_spread: use first numa node with mem

dzhengfy:
Original codes use numa node 0 to test, but on some numa hosts,
node 0 has no memory, so the vm will fail to start with it.
This is a fix to use the first available numa node with memory for
testing.

### DIFF
--- a/libvirt/tests/cfg/numa/numa_memory_spread.cfg
+++ b/libvirt/tests/cfg/numa/numa_memory_spread.cfg
@@ -2,7 +2,6 @@
     type = numa_memory_spread
     start_vm = "no"
     memory_mode = "restrictive"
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
it makes sense, but better upload one test result
LGTM

dzhengfy:
> it makes sense, but better upload one test result LGTM

@chunfuwen Thanks. Because the hardware needs to have multiple nodes and node 0 has zero memory, it is hard to find this special numa host. Could we merge it and run the job to see if it works. If there still has a problem, I would like to fix it accordingly.

chunfuwen:
Let's merge and monitor what could happen.

PR #4098, opened 2022-03-12
v2v: add model='virtio-transitional' check for XML in rhel9

vwu-vera:
Please check bz #2043333

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -29,7 +29,8 @@ FEATURE_SUPPORT = {
     'libosinfo': 'virt-v2v-1.40.2-2.el7',
     'virtio_rng': '2.6.26',
     'cache_none': 'virt-v2v-1.42.0-4',
-    'q35': 'virt-v2v-1.43.3-2'}
+    'q35': 'virt-v2v-1.43.3-2',
+    'virtio_model': 'virt-v2v-1.45.97-4'}
 # bz#1961107
 V2V_ADAPTE_SPICE_REMOVAL_VER = "[virt-v2v-1.45.92,)"
 
@@ -620,6 +621,14 @@ class VMChecker(object):
             for disk in root.findall("./devices/disk/driver[@cache]"):
                 if disk.get('cache') == 'none':
                     self.log_err(err_msg)
+
+        LOG.info("Checking model='virtio-transitional' not existing in VM XML")
+        if self.os_type == 'windows' and self.target == 'libvirt' and compare_version(
+                FEATURE_SUPPORT['virtio_model']):
+            root = ET.fromstring(self.vmxml)
+            err_msg = "Checking model='virtio-transitional' not existing failed"
+            if root.findall(".//*[@model='virtio-transitional']") or root.findall(".//*[@type='virtio-transitional']"):
+                self.log_err(err_msg)
 
     def check_linux_vm(self):
         """


vwu-vera:
(1/1) type_specific.io-github-autotest-libvirt.convert_vm_to_libvirt.esx.vm.7_0.windows.win2022.arch_x86_64.it_vddk: PASS (499.66 s)

vwu-vera:
@xiaodwan Please help to review. Thanks.

provider/v2v_vmcheck_helper.py
```diff
@@ -621,6 +622,13 @@ def check_vm_xml(self):
                 if disk.get('cache') == 'none':
                     self.log_err(err_msg)
 
+        LOG.info("Checking model='virtio-transitional' not existing in VM XML")
+        if self.os_type == 'windows' and self.target == 'libvirt' and compare_version(
+                FEATURE_SUPPORT['virtio_model']):
+            err_msg = "Checking model='virtio-transitional' not existing failed"
+            if self.vmxml.find('virtio-transitional') != -1:
```
xiaodwan, line 629: The find usage is not like this. It will always return '-1' in this case.
You should check the official document and learn it first.
https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.find
BTW, the XPath syntax is another choice.

provider/v2v_vmcheck_helper.py
```diff
@@ -932,14 +940,21 @@ def check_local_output(params):
...
+    else:
+        if compare_version(FEATURE_SUPPORT['cache_none']):
+            # Check 'cache_none' in xml file
+            LOG.info("Checking cache='none' not exist in %s" % xml_file)
+            root = ET.parse(xml_file).getroot()
+            for disk in root.findall("./devices/disk/driver[@cache]"):
+                if disk.get('cache') == 'none':
+                    result = False
+                    break
+        if compare_version(FEATURE_SUPPORT['virtio_model']):
```
xiaodwan, line 952: This part is not necessary. I'm not sure why the scenario is limited to local only. I guess 
'to rhv' is also applied.
'-o local' is not a typical case. Only add the check in check_xml function is enough.

vwu-vera: Thanks for suggestion. :) Delete this part in the latest code.

provider/v2v_vmcheck_helper.py
vwu-vera: self.vmxml is not an instance of class`xml.etree.ElementTree.Element` but string. And the keywords needs to be checked under the several tags. This is the efficient method.
Please refer to https://docs.python.org/3/library/stdtypes.html#str.find.

Also double checked in the TC runs for type(self.vmxml):
avocado.v2v.provider.v2v_vmcheck_helper INFO | vera debug:<class 'str'>

xiaodwan: ok, I see.  I think this way is not strong enough. Ex. if the guest name is 'virtio-transitional', the code will go wrong.
It's better to find it by the method in xml.etree.ElementTree.
Would you like to do it? or let me do it after merging this PR?

xiaodwan:
And please stop adding polarion case ID in git commit. You can find chunfu's email about the reason.
Thanks.

provider/v2v_vmcheck_helper.py
vwu-vera: Thanks for reminding about case ID. I missed the mail. 
And could you help to find the better way in xml.etree.ElementTree? Cause I found the keyword is under at least 5 different tags.

xiaodwan: You may find the 'Supported XPath syntax' part in https://docs.python.org/3/library/xml.etree.elementtree.html.
This is quite a good chance for you to learn and practice the xpath usage. You could learn it and solve this issue by
yourself first. 
You could follow the instruction in the official doc or any other sources, then do some experiment  in python interpreter.

vwu-vera:
@xiaodwan I tried to rewrite in xpath way. Could you help to check it? Thanks.

provider/v2v_vmcheck_helper.py
```diff
@@ -621,6 +622,24 @@ def check_vm_xml(self):
...
+                    "./devices/disk[@model]",
+                    "./devices/memballoon[@model]",
+                    "./devices/rng[@model]"
+                    ]
+            for p in paths:
+                for model in root.findall(p):
+                    if model.get('model') == 'virtio-transitional':
+                        self.log_err(err_msg)
+            for model in root.findall("./devices/interface/model[@type]"):
+                if model.get('type') == 'virtio-transitional':
```
xiaodwan, line 640: root.findall(".//*[@model='virtio-transitional']")

This will do.

vwu-vera: Great. Thanks! This method is more efficient. Please help to review.

vwu-vera: Thanks for the advice.

provider/v2v_vmcheck_helper.py
```diff
@@ -621,6 +622,14 @@ def check_vm_xml(self):
                 if disk.get('cache') == 'none':
                     self.log_err(err_msg)
 
+        LOG.info("Checking model='virtio-transitional' not existing in VM XML")
+        if self.os_type == 'windows' and self.target == 'libvirt' and compare_version(
+                FEATURE_SUPPORT['virtio_model']):
+            root = ET.fromstring(self.vmxml)
+            err_msg = "Checking model='virtio-transitional' not existing failed"
+            if root.findall(".//*[@model='virtio-transitional']") and root.findall(".//*[@type='virtio-transitional']"):
```
xiaodwan, line 630: No, it's '+' or 'or', not 'and'.  The 'and' logic is wrong here.

vwu-vera: Thanks. Has corrected it.

xiaodwan (approved):

PR #4101, opened 2022-03-15
update file after move create_secret function to libvirt_secret.py

nanli1:
update related file when move create_secret to libvirt_secret.py
    

chunfuwen (approved):
LGTM

PR #4116, opened 2022-03-21
numa_memory: fix bitmap error

dzhengfy:
This is to fix the wrong value assignment. A string value is expected,
but not a list value.

### DIFF
--- a/libvirt/tests/src/numa/numa_memory.py
+++ b/libvirt/tests/src/numa/numa_memory.py
@@ -221,8 +221,9 @@ def run(test, params, env):
                         if nodes_size > len(node_list):
                             test.cancel("nodeset %s out of range" % numa_memory['nodeset'])
                         else:
-                            numa_memory['nodeset'] = node_list[:nodes_size]
-                            logging.debug("Update the numa memory nodeset to '%s'", node_list[:nodes_size])
+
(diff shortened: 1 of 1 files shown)


dzhengfy:
Before the fix:
LibvirtXMLError: Failed to define avocado-vt-vm1 for reason:error: Failed to define domain from /tmp/xml_utils_temp_2tdh_ysc.xmlerror: invalid argument: Failed to parse bitmap '[1, 2]'&#10;

libvirt/tests/src/numa/numa_memory.py
```diff
@@ -198,7 +198,7 @@ def numa_mode_check(mode_nodeset):
                         if nodes_size > len(node_list):
                             test.cancel("nodeset %s out of range" % numa_memory['nodeset'])
                         else:
-                            numa_memory['nodeset'] = node_list[:nodes_size]
+                            numa_memory['nodeset'] = ','.join(map(str, node_list))
```
Yingshun, line 201: I wonder if we still need to consider nodes_size here?

dzhengfy: yes, thanks and updated

chunfuwen (approved):
LGTM

Yingshun (approved):

PR #4133, opened 2022-03-29
v2v: allow v2v execution by non-root user

xiaodwan:
This only works when converting guest from esx to ovirt by rhv-upload.
Other scenarios are not supported in automation.


chunfuwen (approved):
LGTM

PR #4146, opened 2022-04-07
v2v: seperate the dir permission from usr_partition case

xiaodwan:
Because the dir_permission case only works in rhel9. We should make
sure it won't fail in rhel8.


chunfuwen (approved):
LGTM

PR #4147, opened 2022-04-08
numa_numanode_cpu_info: fix incorrect skip

dzhengfy:
The required numa node number is 2, but the code is requiring a number bigger than 2.
So this is not consistent with the original purpose and over required. This PR is to
make 2 numa nodes host also be able to run the case instead of skipping.

### DIFF
--- a/libvirt/tests/src/numa/numa_numanode_cpu_info.py
+++ b/libvirt/tests/src/numa/numa_numanode_cpu_info.py
@@ -33,16 +33,18 @@ def update_xml(vm_name, online_nodes, params):
     vmxml.sync()
 
 
-def setup_host(online_nodes, pages_list, ori_page_set):
+def setup_host(required_node_num, online_nodes, pages_list, ori_page_set):
     """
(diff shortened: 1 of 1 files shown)


dzhengfy:
Class | Duration | Fail | (diff) | Skip | (diff) | Pass | (diff) | Total | (diff)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
numa_numanode_cpu_info | 11 sec | 0 |   | 0 | -1 | 1 | +1 | 1 |

dzhengfy:
Before the fix: 
Class | Duration | Fail | (diff) | Skip | (diff) | Pass | (diff) | Total | (diff)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
numa_numanode_cpu_info | 10 sec | 0 |   | 1 | +1 | 0 | -1 | 1 |

Yingshun (approved):
LGTM

chunfuwen (approved):
LGTM

PR #4151, opened 2022-04-13
Fix Failed to start domain 'avocado-vt-vm1' error due to the fact tha…

chunfuwen:
…t VM destroy may not be completed

Confirmed with feature owner, it is actually not related to selinux
Some eror message can be seen with Requested operation is not valid: Setting different SELinux label
on /var/lib/libvirt/qemu/domain-1-avocado-vt-vm1/master-key.aes which is already in use

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
@@ -404,6 +404,7 @@ def run(test, params, env):
 
         # Wait for vm in stable state
         if params.get("start_vm") == "yes":
+            utils_misc.wait_for(lambda: vm.state() == "shut off", 10)
             if vm.state() == "shut off":
                 vm.start()
                 vm.wait_for_login()


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virsh.managedsave.functional_test.libvirt_guests.single_guest

libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
```diff
@@ -404,6 +404,7 @@ def build_vm_xml(vm_name, **dargs):
 
         # Wait for vm in stable state
         if params.get("start_vm") == "yes":
+            utils_misc.wait_for(lambda: vm.state() == "shut off", 10)
```
dzhengfy, line 407: I am a little confused. Before the fix, I guest the error message was raised from line 409 vm.start(), and  vm.state() == "shut off" at that time.  So even the vm.state() == shut off that does not mean the vm is really shutdown which is the root cause for the error, am I right?  If that, your Line 407 can still not ensure the vm is totally shutdown because vm.state() might return True at once which has no difference with line 408.  Maybe I was missing something, but please share your thoughts. Thanks.

chunfuwen: @dzhengfy 
1)The root clause is vm.start() in line 409 in some extreme case that VM may not be destroyed completely although vm.state() return 'shuf_off'
2)yes, utils_misc.wait_for(lambda: vm.state() == "shut off", 10)  may not work as expected to fix this issue
3) We are not suggested to hard code sleep time, so currently I can not figure out what other proper solution I can have.
If you have, please share that here.

dzhengfy: @chunfuwen I do not have good solution for your problem. But I have another workaround. I inspected the python code, and found there is no need to set 'start_vm = yes' because all the configurations need to stop vm at first. So I suggest you set 'start_vm=no' in cfg to have a try. If it works, that will help avoid vm starting failure problem.

dzhengfy: @chunfuwen I do not have good solution for your problem. But I have another workaround. I inspected the python code, and found there is no need to set 'start_vm = yes' because all the configurations need to stop vm at first. So I suggest you set 'start_vm=no' in cfg to have a try. If it works, that will help avoid vm starting failure problem.

dzhengfy:
@chunfuwen I did not see any comment from you since my last comment. Could you give me any opinion?

libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
dzhengfy: @chunfuwen I do not have good solution for your problem. But I have another workaround. I inspected the python code, and found there is no need to set 'start_vm = yes' because all the configurations need to stop vm at first. So I suggest you set 'start_vm=no' in cfg to have a try. If it works, that will help avoid vm starting failure problem.

dzhengfy (commented):
See above

chunfuwen:
let's close it until we can find more proper to handle this

PR #4164, opened 2022-04-20
provider/vfio/ccw: improve logging

smitterl:
I suspect fdasd might succeed sometimes despite a non-zero
status code though I couldn't reproduce it. This made the test
case fail. (I think this could happen because of a warning
when the disk is detected as used by several hosts.)

Add the status report to the log in case this reproduces.


chunfuwen (approved):
LGTM

chunfuwen:
one line debug code is changed, mark it as minor_fix

Yingshun (approved):

PR #4165, opened 2022-04-21
Add more error message for mapping new result

kylazhang:
Before failure is 
```
TestFail: Expect should fail with one of ['error: unsupported configuration: unknown boot device'], but failed with:error: Failed to define domain from /tmp/xml_utils_temp_k6o4pe3h.xmlerror: XML error: Invalid value for attribute 'dev' in element 'boot': 'noexist'
```
After
``` 
(1/1) type_specific.io-github-autotest-libvirt.virsh.boot.by_ovmf.negative_test.invalid_boot_dev: PASS (12.00 s)
```


chunfuwen (approved):
LGTM

chunfuwen:
one line change in cfg file, mark it as minor_fix

PR #4172, opened 2022-04-24
migrate_mem: add case maxmemory

dzhengfy:
Test scenario: migration with maxmemory and check no specified error in libvirt daemon log
VIRT-38151

### DIFF
--- a/libvirt/tests/src/migration/migrate_mem.py
+++ b/libvirt/tests/src/migration/migrate_mem.py
@@ -160,7 +160,9 @@ def verify_test_default(vm, params, test):
     """
     check_str_local_log = params.get("check_str_local_log")
     if check_str_local_log:
-        libvirt.check_logfile(check_str_local_log, params.get("libvirtd_debug_file"))
+        libvirt.check_logfile(check_str_local_log,
+                              params.get("libvirtd_debug_file"),
+                              eval(params.get('str_in_log', 'True')))
 
 
 def verify_test_mem_balloon(vm, params, test):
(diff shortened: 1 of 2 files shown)


dzhengfy:
(1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_mem.mem_device: PASS (144.94 s)
libvirt-8.0.0-8.el9_0.x86_64

dzhengfy:
@cliping Could you help review?

cliping (approved):
LGTM

libvirt/tests/src/migration/migrate_mem.py
```diff
@@ -406,7 +406,9 @@ def run_migration(vm, params, test):
         migration_test.ping_vm(vm, params, uri=dest_uri)
 
     if check_str_local_log:
-        libvirt.check_logfile(check_str_local_log, params.get("libvirtd_debug_file"))
+        libvirt.check_logfile(check_str_local_log,
+                              params.get("libvirtd_debug_file"),
+                              eval(params.get('str_in_log')))
```
Yingshun, line 411: We need a default str_in_log value I think, otherwise eval() will raise an exception if there's no 'str_in_log' in case cfg.

Yingshun:
@dzhengfy conflicted, please rebase.

libvirt/tests/src/migration/migrate_mem.py
dzhengfy: Even the situation will not happen because it only executes when check_str_local_log is  not NULL, but I still think it is a good coding habit. I updated it in verify_test_default(). Thanks

Yingshun (approved):

PR #4185, opened 2022-05-05
migration: Update to recover the source host's hosts file

Yingshun:
The connection to the target host may be broken if the IP addr
set in the previous case is inaccessible. So update to recover
/etc/hosts for the source host.

Before the fix:
hosts contains `<ipv6 addr> <target host cn>` after executing 'live_migration.migration_host.set_src_hosts_ipv6.p2p_live.without_postcopy'

After the fix:
There's no `<ipv6 addr> <target host cn>` in /etc/hosts.


Yingshun:
@cliping  Could you please review this PR? Thank you!

cliping (approved):
LGTM

Yingshun:
Thank you @cliping  for reviewing, reran the 8.x job with this pr, all passed.

PR #4204, opened 2022-05-11
v2v: fix block_dev case issue

xiaodwan:
The output file name is different between '-o json' and '-o local'.
so the code should be changed to reflect that. If not, the test
will be invalid.


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.negative_test.block_dev.small.esx_70.local.it_vddk.vpx_uri: PASS (147.96 s)

chunfuwen (approved):
LGTM

PR #4208, opened 2022-05-12
v2v:add the required version for TC 'VM with sata disk'

vwu-vera:
Please refer bug #1677608


vwu-vera:
@xiaodwan  Please help to review the modification. Thanks.

v2v/tests/src/specific_kvm.py
```diff
@@ -678,6 +679,10 @@ def check_result(result, status_error):
     try:
         v2v_sasl = None
 
+        if version_required and not utils_v2v.multiple_versions_compare(
```
xiaodwan, line 682: This check could be put at line 100, so the check will be done as early as possible.

xiaodwan, line 682: This check could be put at line 100, so the check will be done as early as possible.

xiaodwan (approved):

PR #4214, opened 2022-05-18
v2v:add 8.7/9.1 guests for signoff/acceptance testing

vwu-vera:
Check with the matrix.


vwu-vera:
@xiaodwan please help to review. Thanks.

v2v/tests/cfg/convert_vm_to_libvirt.cfg
```diff
@@ -37,6 +37,14 @@
...
             vm_pwd = ${password}
             variants:
+                - latest9_dev:
+                    only source_esx.esx_70
+                    os_short_id = "RHEL9_DEV_SHORT_ID"
+                    os_version = "LATEST9_DEV"
+                - latest8_dev:
+                    only source_esx.esx_70
+                    os_short_id = "RHEL9_DEV_SHORT_ID"
+                    os_version = "LATEST8_DEV"
```
xiaodwan, line 47: 'LATEST8_DEV' will not be replaced with correct value sometimes by CI. 
Because there is also a 'LATEST8' replacement in cfg file. Sometimes it will replace 'LATEST8_DEV' with 'rhel8.7_DEV' first, this will cause errors.
So when defining those replacements, we should take care they won't be replaced by other existing replacements.
That's mean a replacement should not be a sub-string of another replacement.
In this case, it could be "LATEST_DEV_8".

v2v/tests/cfg/convert_vm_to_libvirt.cfg
```diff
@@ -37,6 +37,14 @@
             vm_user = ${username}
             vm_pwd = ${password}
             variants:
+                - latest9_dev:
+                    only source_esx.esx_70
+                    os_short_id = "RHEL9_DEV_SHORT_ID"
+                    os_version = "LATEST9_DEV"
```
xiaodwan, line 43: see the comment below.

v2v/tests/cfg/convert_vm_to_libvirt.cfg
vwu-vera: Thanks for pointing out.
Has modified accordingly.

xiaodwan (approved):

PR #4228, opened 2022-05-25
v2v: add vsock check for linux VMs

xiaodwan:
For linux VMs, if kernel supports vsock, v2v will add vsock to the
guest after conversion. This patch adds the check for it.

But some old releases may not support the feature by kernel. We just
simply skip those guests. This is not a good way to do it. But they
could be improved in future.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -24,6 +24,7 @@ LOG = logging.getLogger('avocado.v2v.' + __name__)
 
 RETRY_TIMES = 10
 # Temporary workaround <Fix in future with a better solution>
+# Deprecated struct, Don't use it.
 FEATURE_SUPPORT = {
     'genid': 'virt-v2v-1.40.1-1.el7',
     'libosinfo': 'virt-v2v-1.40.2-2.el7',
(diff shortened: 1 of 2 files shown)


provider/v2v_vmcheck_helper.py
```diff
@@ -627,9 +630,29 @@ def check_vm_xml(self):
...
             err_msg = "Checking model='virtio-transitional' not existing failed"
-            if root.findall(".//*[@model='virtio-transitional']") or root.findall(".//*[@type='virtio-transitional']"):
+            if root.findall(
+                    ".//*[@model='virtio-transitional']") or root.findall(".//*[@type='virtio-transitional']"):
                 self.log_err(err_msg)
 
+        if self.is_vsock_supported(self.os_version):
+            self.check_xml('./devices/vsock')
+
+    def check_xml(self, xpath, reverse=False):
```
chunfuwen, line 640: the parameter reverse is a little misleading,  how about changing to direct meaning one  .e.g. 'not_found'

xiaodwan: Thanks. I changed it to 'existence'.

chunfuwen (approved):
LGTM

PR #4255, opened 2022-06-08
fix to use --nvram in virsh undefine

dzhengfy:
--nvram is needed for ovmf guest from RHEL 9.x. And it also does not have side effect
on seabios guest. This option was supported since RHEL7.x which is enough old. The
assumption is that it is no value to test RHEL6 host with libvirt.

Depend on https://github.com/avocado-framework/avocado-vt/pull/3423


chunfuwen (approved):
LGTM

PR #4262, opened 2022-06-13
libvirt_vcpu_plug_unplug: disable test on s390x

smitterl:
The test case requires a vcpu hot-unplug which is
not supported on s390x.


chunfuwen (approved):
LGTM

PR #4264, opened 2022-06-13
libvirt_cputune: disable tests on s390x

smitterl:
On s390x,
- cpu flag 'mba' is not available
- numad is not supported by libvirt

Disable tests accordingly.


chunfuwen (approved):
LGTM

PR #4265, opened 2022-06-13
hmat: disable tests on s390x

smitterl:
The test suite requires NUMA which is not available
on s390x. Disable test cases.


chunfuwen (approved):
LGTM

PR #4266, opened 2022-06-13
memory_discard: disable tests on s390x

smitterl:
Test cases require memory hotplug which is not
supported on s390x at this point.


chunfuwen (approved):
LGTM

PR #4267, opened 2022-06-13
memory_misc: disable test cases on s390x

smitterl:
On s390x, no DIMM or NUMA support. Therefore, disable
test cases.


chunfuwen (approved):
LGTM

PR #4270, opened 2022-06-13
virtual_disks.ccw_addr: disable tests on s390x

smitterl:
Despite their name, these tests are meant to be run
on a system that uses PCI per default and check their behaviour.
Almost all of these tests are negative. They do not apply
to s390x where CCW is the default. Therefore, disable them.


chunfuwen (approved):
LGTM

PR #4327, opened 2022-07-01
Fixup:chown of VARS.fd file to avoid permission issue

chloerh:
Ownership will not be changed dynamically when 'dynamic_ownership'
is disabled. Therefore the ownership of certain files should be
maintained manually to avoid permission issue.

### DIFF
--- a/libvirt/tests/src/svirt/dac_nfs_save_restore.py
+++ b/libvirt/tests/src/svirt/dac_nfs_save_restore.py
@@ -1,4 +1,5 @@
 import os
+import pwd
 import logging as log
 
 from avocado.core import exceptions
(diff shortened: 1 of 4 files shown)


chloerh:
Test result: 

running

yafu-1 (approved):
LGTM

libvirt/tests/src/svirt/dac_nfs_save_restore.py
```diff
@@ -112,6 +113,14 @@ def run(test, params, env):
             elif qemu_user == "qemu":
                 os.chown(disk_path, 107, 107)
 
+        # Change ownership of VARS.fd file
+        if dynamic_ownership is False:
+            if vmxml.os.xmltreefile.find('nvram') is not None:
```
dzhengfy, line 118: As the latest virt-install has new behavior that there will be no nvram setting in <os> if the vm is not started. So it seems you need to update the fix accordingly.

libvirt/tests/src/svirt/dac_start_destroy.py
```diff
@@ -232,6 +232,18 @@ def _create_user():
             vmxml.sync()
             logging.debug("updated domain xml is: %s" % vmxml.xmltreefile)
 
+        # Change ownership of VARS.fd file
+        if dynamic_ownership is False:
+            if vmxml.os.xmltreefile.find('nvram') is not None:
```
dzhengfy, line 237: same as above.

dzhengfy (commented):
Others LGTM

libvirt/tests/src/svirt/dac_nfs_save_restore.py
chloerh: Updated and rerunning tests.

libvirt/tests/src/svirt/dac_start_destroy.py
chloerh: Updated and rerunning tests.

dzhengfy (approved):
LGTM

PR #4344, opened 2022-07-10
Fix the issues about flock/posix

iccaszhulili:
1) As shown in https://bugzilla.redhat.com/show_bug.cgi?id=2057882#c9,
posix/flock is not supported, some import scenarios in only tested when
posix/flock=on before, and these scenarios have be added into skiplist.
We should change posix/flock to off in these scenarios.

2) If I run the cases on my local machine, there would be a process of
avocado contains "virtiofsd". If one just grep the first process of
virtiofsd, it is not the virtiofsd process itself.

Signed-off-by: Lili Zhu <lizhu@redhat.com>

### DIFF
--- a/libvirt/tests/cfg/virtual_device/filesystem_device.cfg
+++ b/libvirt/tests/cfg/virtual_device/filesystem_device.cfg
@@ -4,8 +4,8 @@
     vms = "avocado-vt-vm1"
     start_vm = no
     cache_mode = "none"
-    lock_posix = "on"
-    flock = "on"
+    lock_posix = "off"
+    flock = "off"
     virtiofsd_path = "/usr/libexec/virtiofsd"
     queue_size = "512"
     driver_type = "virtiofs"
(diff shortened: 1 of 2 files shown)


iccaszhulili:
@kylazhang Please help to review. Thanks

libvirt/tests/cfg/virtual_device/filesystem_device.cfg
```diff
@@ -32,13 +32,13 @@
                 - two_guests:
                     vms = "avocado-vt-vm1 avocado-vt-vm2"
                     guest_num = 2
-                    only nop.fs_test.xattr_on.flock_on.lock_posix_on.cache_mode_none.one_fs, hotplug_unplug..fs_test.xattr_on.flock_on.lock_posix_on.cache_mode_none.one_fs
+                    only nop.fs_test.xattr_on.flock_off.lock_posix_off.cache_mode_none.one_fs, hotplug_unplug..fs_test.xattr_on.flock_off.lock_posix_off.cache_mode_none.one_fs
```
chunfuwen, line 35: Per my understanding, the fix depends on specific libvirt version.
If we change test case name here, it bring one uncompatibility issue in lower version.
One negative effect, the case name may not be found in lower libvirt version.
One way to go forward may be replace flock_off.lock_posix_off with .. ,and in our CI ,we filter explicit or known failed cases

iccaszhulili: @chunfuwen flock/posix is never supported before. The only difference is that in lower version, it will not report error when setting flock/posix to on, but it will report error now. And if you set flock/posix to off, it will not report any error when installed with lower version libvirt. Anyway, I will update the caselink linkage after the fix. 

For the solution you mentioned: "replace flock_off.lock_posix_off with ..", I think it will add more cases, these cases will not be tested at all. What do you think?

chunfuwen (approved):
LGTM

cliping (approved):
LGTM

dzhengfy:
> @kylazhang Please help to review. Thanks

@iccaszhulili , kylazhang does not review PR any more. FYI.

dzhengfy (commented):
Please provide  test results and failures before your fix.

iccaszhulili:
Test result: 
RHEL9.1: libvirt-RHEL-9.1-runtest-x86_64-function-misc_virtual_devices/33(1 failure, because of a bug, which has been fixed for libvirt-8.5.0-4)
RHEL8.7: libvirt-RHEL-8.7-runtest-x86_64-function-misc_virtual_devices/19/

dzhengfy (approved):
LGTM

PR #4365, opened 2022-07-20
virsh_blockcommit.py: fix case which is hard to get expected result

meinaLi:
When we remove --pivot option in virsh blockommit, the blockjob will always have an active job there. Then we can get the expected error info in this negative scenario. But if we use --pivot option, it may execute so quickly that it's hard to get error info, we can only get: ......expect fail, but run successfully. 


nanli1:
LGTM,
Just confirm , Did you also add "--pivot" on polarion case

nanli1 (approved):

meinaLi:
@nanli1 Not sure which one need to be updated. Because the polarion case matched in caselink is incorrect.

chunfuwen (approved):
LGTM

PR #4396, opened 2022-08-01
Fix iface security model dac not available issue

chunfuwen:
If define Security driver model 'dac' in vm.xml,  seclabel model="dac" under unprivileged user is not supported

Therefore, one error: unsupported configuration: Security driver model 'dac' is not available will be thrown

Fixing the issue by removing seclabel model="dac" since unprivileged user doesn't support this feature

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_unprivileged.py
+++ b/libvirt/tests/src/virtual_network/iface_unprivileged.py
@@ -197,6 +197,8 @@ def run(test, params, env):
             upu_vmxml.devices = all_devices
             logging.debug(upu_vmxml)
 
+            # Remove seclabel model="dac" since unprivileged user doesn't support this feature
+            upu_vmxml.del_seclabel([('model', 'dac'), ('relabel', 'yes')])
             # Define updated xml
             shutil.copyfile(upu_vmxml.xml, new_xml_path)
             upu_vmxml.xml = new_xml_path


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_network.iface_unprivileged.precreated.host_macvtap

chunfuwen:
This error happen s390x running jobs

chunfuwen:
@yalzhang 
@smitterl 
Please help take a look

yalzhang (approved):
Thank you for fixing this.

libvirt/tests/src/virtual_network/iface_unprivileged.py
```diff
@@ -199,6 +199,14 @@ def check_ping(dest_ip, ping_count, timeout, src_ip=None, session=None,
 
             # Define updated xml
             shutil.copyfile(upu_vmxml.xml, new_xml_path)
+            # Remove seclabel model="dac" since unprivileged user doesn't support this feature
```
dzhengfy, line 202: If you only want to remove seclabel with dac, how about using like below:
```
        del_dict = [('model', 'dac'), ('relabel', 'yes')]
        upu_vmxml.del_seclabel(del_dict)
```
I can see a detailed sample in virttest/unittests/libvirt_xml/test_vm_xml.py

chunfuwen: @dzhengfy 
Thank you for suggestion.
But the trick is here that we define VM from new_xml_path in the follow coding line 211, so the change only in new_xml_path can impact VM creating:
```
virsh.define(new_xml_path, **upu_args)
```

dzhengfy: @chunfuwen , I read through the previous codes and found most codes are for updating the `upu_vmxml` and then define a vm. So I am thinking if we need to ` shutil.copyfile() `on line 201.  I would like to remove line 201 and use above method to edit the `upu_vmxml` object and use `virsh.define(upu_vmxml.xml, **upu_args)`. 

Another way that if you still think ` shutil.copyfile() `on line 201 should not be removed, then I would like suggest below:
```
 del_dict = [('model', 'dac'), ('relabel', 'yes')]
 upu_vmxml.del_seclabel(del_dict)
 shutil.copyfile(upu_vmxml.xml, new_xml_path)
 upu_vmxml.xml = new_xml_path
 virsh.define(new_xml_path, **upu_args)
```
Anyway, from my side, the `with open(new_xml_path, "r+") as f:` seems a significant different style with existing codes while we do have same style's solution. So it is better to align with existing style, right?

dzhengfy (commented):
Replied in https://github.com/autotest/tp-libvirt/pull/4396/files#r945458712

libvirt/tests/src/virtual_network/iface_unprivileged.py
chunfuwen: @dzhengfy 
1)The case aims to define VM under unprevileged user, so under root user, it first virsh dumpxml  vm into upu_vmxml
2)Then copy upu_vmxml to /home/xx folder, and then use vm.define() to define new VM.
3)From logic view, it is more natural to modify copied Vm xml under /home/xx, rather than modify original upu_vm.xml although it look fewer code by using upu_vmxml.del_seclabel(del_dict)

dzhengfy: @chunfuwen , as we discussed offline, I am looking forward for your updates. Thanks a lot.

dzhengfy: @chunfuwen , as we discussed offline, I am looking forward for your updates. Thanks a lot.

dzhengfy: @chunfuwen , as we discussed offline, I am looking forward for your updates. Thanks a lot.

dzhengfy (commented):
@chunfuwen , as we discussed offline, I am looking forward for your updates. Thanks a lot.

chunfuwen:
@dzhengfy  updated as per offline discussion

dzhengfy (approved):
LGTM

PR #4398, opened 2022-08-01
Fix static check failure in travics ci

chunfuwen:
Fix static check failure in travics ci
inspekt checkall --disable-style E501,E265,W601,W605,E402,E722,E741 --no-license-check will fail
it need roll back pycodestyle==2.8.0 to fix it


chunfuwen:
close it since it could fix them in coding

PR #4441, opened 2022-08-22
v2v: remove deprecated floppy device codes

xiaodwan:
floppy device is deprecated. The test code should be removed.


chunfuwen (approved):
LGTM

PR #4444, opened 2022-08-23
Automate cases: start VM with various startuppolicy for USB Device

chunfuwen:
Automate cases: start VM with various startuppolicy for USB Device

startuppolicy:optional, mandatory and requisite can be specified as source attribute in USB disk
while with USB device plugged in, and whether usb device exists or not need to be checked after VM start

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
+++ b/libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
@@ -1,0 +1,47 @@
+- virtual_disks.usb_startuppolicy_config:
+    type = virtual_disks_usb_startuppolicy
+    take_regular_screendumps = "no"
+    start_vm = "no"
+    target_format = "raw"
+    type_name = "block"
+    driver_type = 'raw'
+    device_type = "disk"
+    target_dev = "sdh"
+    target_bus = "usb"
+    status_error = "no"
+    define_error = "no"
+    pkgs_host = "usbutils"
+    no s390-virtio
+    # something like verdor:product
+    usb_device_label = 'ENTER.YOUR.DEV.NAME'
+    variants:
+        - start_guest:
+    variants:
+        - set_mandatory_startuppolicy:
+            startup_policy_value = "mandatory"
+            variants:
+                - without_usb_device_plugged:
+                    virt_disk_device_source = "/dev/sdx"
+                    usb_start_error_message = "No such file or directory"
+                    only negative
+                - with_usb_device_plugged:
+                    virt_disk_device_source = "/dev/sdb"
+                    only positive
+        - set_requisite_startuppolicy:
+            startup_policy_value = "requisite"
+            virt_disk_device_source = "/dev/sdb"
+            define_error = "yes"
+            only negative
+        - set_optional_startuppolicy:
+            startup_policy_value = "optional"
+            virt_disk_device_source = "/dev/sdb"
+            only positive
+    variants:
+        - coldplug:
+            virt_device_hotplug = "no"
+    variants:
+        - positive:
+            status_error = "no"
+        - negative:
+            status_error = "yes"
+
--- a/libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
@@ -1,0 +1,185 @@
+import logging
+import platform
+import shutil
+import time
+
+import aexpect
+
+from virttest import utils_disk
+from virttest import utils_misc
+from virttest import virt_vm, remote
+
+from virttest.utils_libvirt import libvirt_usb
+from virttest.utils_libvirt import libvirt_disk
+
+from virttest.libvirt_xml import vm_xml, xcepts
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def run(test, params, env):
+    """
+    Test usb virtual disk startup policy.
+
+    1.Prepare a vm with usb type disk with related startup policy
+    2.Attach the virtual disk to the vm
+    3.Start vm
+    4.Check the disk in vm
+    """
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+
+    def get_new_disks(old_partitions):
+        """
+        Get new virtual disks in VM after disk plug.
+
+        :param old_partitions: already existing partitions in VM
+        :return: New disks/partitions in VM
+        """
+        session = None
+        try:
+            session = vm.wait_for_login()
+            if platform.platform().count('ppc64'):
+                time.sleep(10)
+            added_partitions = utils_disk.get_added_parts(session, old_partitions)
+            LOG.debug("Newly added partition(s) is: %s", added_partitions)
+            return added_partitions
+        except Exception as err:
+            test.fail("Error happens when get new disk: %s", str(err))
+        finally:
+            if session:
+                session.close()
+
+    def check_usb_disk_io(partition):
+        """
+        Check if the disk partition in vm can be normally used.
+
+        :param partition: The disk partition in vm to be checked.
+        :return: If the disk can be used, return True.
+        """
+        session = None
+        try:
+            session = vm.wait_for_login()
+            cmd = ("fdisk -l /dev/{0} && mkfs.ext4 -F /dev/{0} && "
+                   "mkdir -p test && mount /dev/{0} test && echo"
+                   " teststring > test/testfile && umount test"
+                   .format(partition))
+            status, output = session.cmd_status_output(cmd)
+            logging.debug("Disk operation in VM:\nexit code:\n%s\noutput:\n%s",
+                          status, output)
+            return status == 0
+        except (remote.LoginError, virt_vm.VMError, aexpect.ShellError) as err:
+            logging.debug("Error happens when check disk io in vm: %s", str(err))
+            return False
+        finally:
+            if session:
+                session.close()
+
+    def get_usable_usb_device():
+        """
+        Get usable usb devices
+
+        """
+        usable_usb = None
+        all_usbs = libvirt_usb.get_usbs_lists_on_host()
+        for usb in all_usbs:
+            if ((usb.split())[5]) in usb_device_label:
+                usable_usb = usb
+                break
+        if usable_usb is None:
+            test.cancel("No usable USB device found.")
+
+    def create_customized_disk(params):
+        """
+        Create one customized disk with related attributes
+
+        :param params: dict wrapped with params
+        """
+        type_name = params.get("type_name")
+        disk_device = params.get("device_type")
+        device_target = params.get("target_dev")
+        device_bus = params.get("target_bus")
+        device_format = params.get("target_format")
+        source_file_path = params.get("virt_disk_device_source")
+        source_dict = {}
+        if source_file_path:
+            if 'block' in type_name:
+                source_dict.update({"dev": source_file_path})
+            else:
+                source_dict.update({"file": source_file_path})
+        startup_policy = params.get("startup_policy_value")
+        if startup_policy:
+            source_dict.update({"startupPolicy": startup_policy})
+        disk_src_dict = {"attrs": source_dict}
+        customized_disk = libvirt_disk.create_primitive_disk_xml(
+            type_name, disk_device,
+            device_target, device_bus,
+            device_format, disk_src_dict, None)
+        LOG.debug("create customized xml: %s", customized_disk)
+        return customized_disk
+
+    usb_device_label = params.get("usb_device_label")
+    if 'ENTER.YOUR.DEV.NAME' in usb_device_label:
+        test.cancel("please specify usable usb device labels in cfg file")
+
+    hotplug = "yes" == params.get("virt_device_hotplug")
+    status_error = "yes" == params.get("status_error")
+    define_error = "yes" == params.get("define_error")
+    pkgs_host = params.get("pkgs_host", "")
+
+    # Get disk partitions info before hot/cold plug virtual disk
+    if vm.is_dead():
+        vm.start()
+    session = vm.wait_for_login()
+    old_partitions = utils_disk.get_parts_list(session)
+    session.close()
+    vm.destroy(gracefully=False)
+
+    # Backup vm xml
+    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml = vmxml_backup.copy()
+
+    try:
+        get_usable_usb_device()
+        # install essential package usbutils in host
+        if not shutil.which('lsusb'):
+            test.error("package {} not installed, please install them before testing".format(pkgs_host))
+
+        device_obj = create_customized_disk(params)
+        if not hotplug:
+            vmxml.add_device(device_obj)
+            vmxml.sync()
+        vm.start()
+        vm.wait_for_login().close()
+    except virt_vm.VMStartError as details:
+        if status_error:
+            LOG.debug("VM failed to start as expected."
+                      "Error: %s", str(details))
+            usb_start_error_message = params.get('usb_start_error_message')
+            if usb_start_error_message and not str(details).count(usb_start_error_message):
+                test.fail('VM start error should contain messages:\n%s' % usb_start_error_message)
+    except xcepts.LibvirtXMLError as xml_error:
+        if not define_error:
+            test.fail("Failed to define VM:\n%s" % xml_error)
+        else:
+            LOG.debug("VM failed to define as expected."
+                      "Error: %s", str(xml_error))
+    else:
+        utils_misc.wait_for(lambda: get_new_disks(old_partitions), 20)
+        new_disks = get_new_disks(old_partitions)
+        if len(new_disks) != 1:
+            test.fail("Attached 1 virtual disk but got %s." % len(new_disks))
+        new_disk = new_disks[0]
+        if platform.platform().count('ppc64'):
+            time.sleep(10)
+        session = vm.wait_for_login()
+        if not libvirt_usb.check_usb_disk_type_in_vm(session, new_disk):
+            test.fail("The newly attached disk is not a usb one.")
+        session.close()
+        if not check_usb_disk_io(new_disk):
+            test.fail("Cannot operate the newly added disk in vm.")
+    finally:
+        if vm.is_alive():
+            vm.destroy(gracefully=False)
+        # Restoring vm
+        vmxml_backup.sync()


chunfuwen:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.positive.coldplug.set_mandatory_startuppolicy.with_usb_device_plugged.start_guest
PASS 2-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.positive.coldplug.set_optional_startuppolicy.start_guest
PASS 3-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.negative.coldplug.set_mandatory_startuppolicy.without_usb_device_plugged.start_guest
PASS 4-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.negative.coldplug.set_requisite_startuppolicy.start_guest

chunfuwen:
it covers :xx-103820,xx-103817,xx-17751,xx-103819

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
```diff
@@ -0,0 +1,48 @@
...
+    device_type = "disk"
+    target_dev = "sdh"
+    target_bus = "usb"
+    status_error = "no"
+    define_error = "no"
+    pkgs_host = "usbutils,usbredir-server"
+    no s390-virtio
+    # something like verdor:product
+    usb_device_label = '0781:5591'
+    # usb_device_label = 'ENTER.YOUR.DEV.NAME'
```
Yingshun, line 17: Remove one of them?

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+    # Backup vm xml
+    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml = vmxml_backup.copy()
+
+    try:
+        get_usable_usb_device()
+        # install essential package usbutils in host
+        for pkg in pkgs_host.split(','):
+            if not utils_package.package_install(pkg):
+                test.fail("package {} installation fail".format(pkg))
```
Yingshun, line 149: We had a tips about host package installation, so I would suggest moving it into ci and checking if the commands can be used as the tips

chunfuwen:
@qiankehan 
This automate startupolicy on usb device, please take a review

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
```diff
@@ -0,0 +1,48 @@
...
+    driver_type = 'raw'
+    device_type = "disk"
+    target_dev = "sdh"
+    target_bus = "usb"
+    status_error = "no"
+    define_error = "no"
+    pkgs_host = "usbutils,usbredir-server"
+    no s390-virtio
+    # something like verdor:product
+    usb_device_label = '0781:5591'
```
qiankehan, line 16: Why to hard-code the VENDOR:PRODUCT id here?  Are you sure the hosts for running the testing scripts always have such a USB device?

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: @Yingshun 
@qiankehan 

1. I assume the tips mean here that it is using install-pkgs in jobs.yaml . Yes, we can enable them in downstream CI. But this doesn't conflict with using utils_package.package_install()here in upstream since utils_package.package_install can pass the logic quickly if packages already installed on host . e.g provisioned already by downstream CI
2.  For visibility in upstream , we may need those logic to inform that those scripts are dependent on some packages on host
3.  Even in our development or test in local environment, it is not always that you environment is provisioned by downstream CI. You may setup your environment from the scratch, those logic here give your insights and helps about what dependences you may encounter.

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
chunfuwen: updated, see first comment

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+
+        device_obj = create_customized_disk(params)
+        if not hotplug:
+            vmxml.add_device(device_obj)
+            vmxml.sync()
+        vm.start()
```
dzhengfy, line 155: If you need to check the vm start result and match the specified error message, it will be easier to use 
```
ret = virsh.start() 
libvirt.check_result(ret, ...)
```
But it depends on you. Current codes are also ok, but a little redundant to me.

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+        session = vm.wait_for_login()
+    except virt_vm.VMStartError as details:
+        if status_error:
+            LOG.debug("VM failed to start as expected."
+                      "Error: %s", str(details))
+        usb_start_error_message = params.get('usb_start_error_message')
+        if usb_start_error_message and not str(details).count(usb_start_error_message):
+            test.fail('VM start error should contain messages:\n%s' % usb_start_error_message)
```
dzhengfy, line 163: For lines from 161 to 163,  I suggest to move them under the scope of line 158 because they should be executed when  status_error  = True.  And when status_error = false,  a test.error() should be executed.

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+    except xcepts.LibvirtXMLError as xml_error:
+        if not define_error:
+            test.fail("Failed to define VM:\n%s" % xml_error)
+        else:
+            LOG.debug("VM failed to define as expected."
+                      "Error: %s", str(xml_error))
+    else:
+        utils_misc.wait_for(lambda: get_new_disks(old_partitions), 20)
```
dzhengfy, line 171: To make the execution more efficient, I think we could also reuse and pass the session in line 156 to `get_new_disks(old_partitions, session)` , then we do not need to create the vm session again and again.

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+
+    1.Prepare a vm with usb type disk with related startup policy
+    2.Attach the virtual disk to the vm
+    3.Start vm
+    4.Check the disk in vm
+    """
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+
+    def get_new_disks(old_partitions):
```
dzhengfy, line 32: Add 'session' parameter to reuse existing vm session.

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+    old_partitions = utils_disk.get_parts_list(session)
+    session.close()
+    vm.destroy(gracefully=False)
+
+    # Backup vm xml
+    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml = vmxml_backup.copy()
+
+    try:
+        get_usable_usb_device()
```
dzhengfy, line 145: If no need to check the result, please remove the return in this function.

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+                      "Error: %s", str(xml_error))
+    else:
+        utils_misc.wait_for(lambda: get_new_disks(old_partitions), 20)
+        new_disks = get_new_disks(old_partitions)
+        if len(new_disks) != 1:
+            test.fail("Attached 1 virtual disk but got %s." % len(new_disks))
+        new_disk = new_disks[0]
+        if platform.platform().count('ppc64'):
+            time.sleep(10)
+        if not libvirt_usb_utils.check_usb_disk_type_in_vm(session, new_disk):
```
dzhengfy, line 178: `libvirt_usb_utils -> libvirt_usb`

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: @dzhengfy 
The difference between utils_package.package_install() and check the command only( suggested here) is minor: utils_package.package_install() just save one manual step :yum install package_name, which is potentially to be executed manually if without utils_package.package_install(). Beside this, I don't see any other difference between what it is implemented here and comment suggested. To move this PR forward, I moderately accept the change. However, I also replied  comments here.

> @chunfuwen , we already reached the agreement in the team that we only check if the packages are available on the host in test scripts instead of installing them because we usually do not modify the test machine's environment. This will avoid our scripts of breaking a 3rd party's test machine. This practice is just upstream oriented.
[chunfuwen] If this practices is just upstream oriented, I believe people likes to see it somewhere on tp-libvrit, e.g p-libvirt_review_comment_summary.rst.  Do you have prior actual experiences of breaking 3rd party's test machine when using package_install() since comment mention script will break 3rd machine?
> 
> 1. utils_package.package_install() explicitly usage will also modify a test machine if NO internal CI installs the package. And its usage will make people confuse about the different strategies.
[chunfuwen]utils_package.package_install() is here used for simple purpose, and is very specific problem-oriented
I am not sure how it related to the comment"And its usage will make people confuse about the different strategies"
How one simple function usage here is relate to high level  concept 'strategies' ?

> 2. Checking the availability of these packages can also make the visibility to the users. So it is not a big problem.
[chunfuwen] I agree it is not a big problem
> 3. If you decide to setup your environment from scratch, without CI, then you have to be responsible for setting up all dependencies by yourself. That should be known matter to us all. If there is dependent missing in your environment, the script will raise test.cancel() so that you definitely know what is depended and install it. That is also a clear and easy way.
[chunfuwen] In today's software defined world, if something can be automated, eg. auto install here ,and try it as many as possible, it is commonly recognized that it is a clear and easy way to move things forward efficiently.
>    So could you help align to this practice?
[chunfuwen] see top replied comment

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+            added_partitions = utils_disk.get_added_parts(session, old_partitions)
+            LOG.debug("Newly added partition(s) is: %s", added_partitions)
+            return added_partitions
+        except Exception as err:
+            test.fail("Error happens when get new disk: %s", str(err))
+        finally:
+            if session:
+                session.close()
+
+    def check_usb_disk_io(partition):
```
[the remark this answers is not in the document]
chunfuwen: @dzhengfy 
I checked linux_disk_check implementation, which involve more actions that this validation is not needed.
For example, it firstly call configure_empty_linux_disk ,which has large number of lines of code
`def configure_empty_linux_disk(session, did, size, start="0M", n_partitions=1,
                               fstype="ext4", labeltype=PARTITION_TABLE_TYPE_MBR,
                               timeout=360):
    """
    Create partition on disk in linux guest, format and mount it.
    Only handle an empty disk and will create equal size partitions onto the disk.
    Note: Make sure '/mnt/' is not mount point before run this function,
          in order to avoid unknown exceptions.

    :param session: session object to guest.
    :param did: disk kname. e.g. sdb
    :param size: partition size. e.g. 2G
    :param start: partition beginning at start. e.g. 0G
    :param n_partitions: the number of partitions on disk
    :param fstype: filesystem type for the disk.
    :param labeltype: label type for the disk.
    :param timeout: Timeout for cmd execution in seconds.
    :return a list: mount point list for all partitions.
    """
    mountpoint = []
    create_partition_table_linux(session, did, labeltype)
    size = utils_numeric.normalize_data_size(size, order_magnitude="M") + "M"
    start = float(utils_numeric.normalize_data_size(start, order_magnitude="M"))
    partition_size = float(size[:-1]) / n_partitions
    extended_size = float(size[:-1]) - partition_size
    if labeltype == PARTITION_TABLE_TYPE_MBR and n_partitions > 1:
        part_type = PARTITION_TYPE_EXTENDED
    else:
        part_type = PARTITION_TYPE_PRIMARY
    for i in range(n_partitions):
        if i == 0:
            new_partition = create_partition_linux(
                    session, did, str(partition_size) + size[-1],
                    str(start) + size[-1], timeout=timeout)
        else:
            if part_type == PARTITION_TYPE_EXTENDED:
                create_partition_linux(session, did, str(extended_size) + size[-1],
                                       str(start) + size[-1], part_type, timeout)
                part_type = PARTITION_TYPE_LOGICAL
                new_partition = create_partition_linux(
                        session, did, str(partition_size) + size[-1],
                        str(start) + size[-1], part_type, timeout)
            else:
                new_partition = create_partition_linux(
                        session, did, str(partition_size) + size[-1],
                        str(start) + size[-1], part_type, timeout)
        start += partition_size
        create_filesyetem_linux(session, new_partition, fstype, timeout)
        mount_dst = "/mnt/" + new_partition
        session.cmd("rm -rf %s; mkdir %s" % (mount_dst, mount_dst))
        if not mount("/dev/%s" % new_partition, mount_dst, fstype=fstype, session=session):
            err_msg = "Failed to mount partition '%s'"
            raise exceptions.TestError(err_msg % new_partition)
        mountpoint.append(mount_dst)
    return mountpoint
`

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: updated by check package availability only

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: @dzhengfy 
After this line, there is vm.wait_for_login() called, which may be more concrete measure to validate whether VM really started and worked

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: I made some change, and move session creation to the place where it is used.
In the clean code philosophy, open/close come in pair,and better they can close the place or positions where they used

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: see replied comment previously

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,186 @@
...
+                session.close()
+
+    def get_usable_usb_device():
+        """
+        Get usable usb devices
+
+        :return: usable usb device list
+        """
+        usable_usb = None
+        all_usbs = libvirt_usb_utils.get_usbs_lists_on_host()
```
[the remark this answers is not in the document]
chunfuwen: updated

chunfuwen:
retest result:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.positive.coldplug.set_mandatory_startuppolicy.with_usb_device_plugged.start_guest
PASS 2-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.positive.coldplug.set_optional_startuppolicy.start_guest
PASS 3-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.negative.coldplug.set_mandatory_startuppolicy.without_usb_device_plugged.start_guest
PASS 4-type_specific.io-github-autotest-libvirt.virtual_disks.usb_startuppolicy_config.negative.coldplug.set_requisite_startuppolicy.start_guest

chunfuwen:
@qiankehan 
@Yingshun 
@dzhengfy 
Comments are addressed, please double check

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,185 @@
...
+    vm.destroy(gracefully=False)
+
+    # Backup vm xml
+    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml = vmxml_backup.copy()
+
+    try:
+        get_usable_usb_device()
+        # install essential package usbutils in host
+        if not shutil.which('lsusb'):
```
dzhengfy, line 145: path.find_command() is widely used for checking if a command is available on the environment.  But it is just FYR. Your implementation is also OK to me.

dzhengfy (commented):
Others LGTM

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
```diff
@@ -0,0 +1,185 @@
...
+            test.fail("package {} not installed, please install them before testing".format(pkgs_host))
```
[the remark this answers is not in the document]
chunfuwen: updated

libvirt/tests/src/virtual_disks/virtual_disks_usb_startuppolicy.py
chunfuwen: it seems that there is no difference here since shutil and path are basic python primitive lib.

dzhengfy (approved):

Yingshun (approved):

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
```diff
@@ -0,0 +1,47 @@
+- virtual_disks.usb_startuppolicy_config:
```
qiankehan, line 1: Is these test is for the XML like:
```xml
  <disk type='block' device='disk'>
    <driver name="qemu" type="raw" />
    <source file='/dev/sdX' startupPolicy='optional'/>
    <target dev='sdx' bus='usb'/>
  </disk>
```
If so, it is not a good idea. The emulated disk target **disk->target@bus** is not associated with startupPolicy at all.
It doesn't matter with the startupPolicy of disk.  
The variants of startupPolicy of disk should be:
- the value of startupPolicy
- the disk source is missing or not
- disk@device affects the result as well
  - when startupPolicy==optional, disk source missing, disk@device == disk: start VM will trigger the disk-change event with the reason DISK_DROPPED_MISSING_ON_START. Then the disk XML will be dropped.
  -  when startupPolicy==optional, disk source missing, disk@device == cdrom: start VM will trigger the disk-change event with the reason DISK_CHANGE_MISSING_ON_START. Then only the disk source is missing or empty

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
```diff
@@ -0,0 +1,47 @@
...
+    type = virtual_disks_usb_startuppolicy
+    take_regular_screendumps = "no"
+    start_vm = "no"
+    target_format = "raw"
+    type_name = "block"
```
qiankehan, line 6: Any reason to prepare block type? If the startupPolicy is for disk, file type is OK. If it is for the hostdev of USB, prepare the host USB device not the block device

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
```diff
@@ -0,0 +1,47 @@
...
+    driver_type = 'raw'
+    device_type = "disk"
+    target_dev = "sdh"
+    target_bus = "usb"
+    status_error = "no"
+    define_error = "no"
+    pkgs_host = "usbutils,usbredir-server"
```
qiankehan, line 13: No need for usbredir-server. This test is not for usb-redir device

qiankehan (commented):
It seems that the startupPolicy of disk has been covered by `libvirt/tests/src/virtual_disks/virtual_disks_optional_startuppolicy.py` and `libvirt/tests/src/virtual_disks/virtual_disks_cdrom_device.py`. If this test is aimed to the hostdev of USB, please move it to the hostdev or host-passthough part. And prepare the host USB device not the disk device. Because the host USB device could be USB disk, mouse, keyboard, and so on

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
chunfuwen: @qiankehan , this will be handled in later modular design activity

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
chunfuwen: @qiankehan ,"block" can be candidate, if change is needed, this will be handled in later modular design activity

libvirt/tests/cfg/virtual_disks/virtual_disks_usb_startuppolicy.cfg
chunfuwen: updated

qiankehan:
This case looks duplicated with previous disk startupPolicy tests `libvirt/tests/src/virtual_disks/virtual_disks_optional_startuppolicy.py` and `libvirt/tests/src/virtual_disks/virtual_disks_cdrom_device.py`. Others LGTM

chunfuwen:
This PR aims to polarion case: xx-103820,xx-103817,xx-17751,xx-103819 before module design. For the issue related to cases steps, it will be handled in later modular design effort.
Let's merge it since two approves this already.

PR #4449, opened 2022-08-24
serial_pty_log: fix log decode failure

dzhengfy:
When booting OVMF guest, there are some un-decoded special characters in the
console log which caused file reading failed. So the solution is to add this
error='ignore' parameter in order to not raise the error like below:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd9 in position 0: unexpected end of data`

The ignoring will not impact some key strings matched in the log.

### DIFF
--- a/libvirt/tests/src/serial/serial_functional.py
+++ b/libvirt/tests/src/serial/serial_functional.py
@@ -122,8 +122,7 @@ class Console(aexpect.ShellSession):
                 address + '.out',
                 os.O_RDONLY | os.O_CREAT | os.O_NONBLOCK)
         elif console_type == 'file':
-            self.read_fd = open(address,
-                                'r')
+            self.read_fd = open(address, 'r', errors='ignore')
 
     def _tcp_thread(self):
         """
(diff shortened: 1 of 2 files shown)


dzhengfy:
Before the fix:
```
[serial.functional.type_file]
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd9 in position 0: unexpected end of data&#10;

[serial.pty.log.file]
TestError: Unexpected error: 'utf-8' codec can't decode byte 0xd9 in position 317: invalid continuation byte&#10;
```
After the fix:
```
 (1/1) type_specific.io-github-autotest-libvirt.serial.pty.log.file: WARN: Test passed but there were warnings during execution. Check the log for details. (71.76 s)
 (1/1) type_specific.io-github-autotest-libvirt.serial.pty.log.logd: WARN: Test passed but there were warnings during execution. Check the log for details. (70.81 s)
 (1/1) type_specific.io-github-autotest-libvirt.serial.functional.type_file: WARN: Test passed but there were warnings during execution. Check the log for details. (74.85 s)
```

chunfuwen (approved):
LGTM

PR #4462, opened 2022-08-31
Remove allow_output_check option

yusuke-okada:
This option is removed in beee20f45cbe63b9a6c70a344c1b13097243151d and not work.

Current multifunction test script specifies allow_output_check argument for avocado.utils.process.run() function.
However, this argument was removed in the following patch. So the script always errors because of the type mismatch.


chunfuwen (approved):
LGTM

chunfuwen:

PR #4504, opened 2022-09-09
Fix channel functional case failure on 390x

chunfuwen:
Fix channel functional case failure

channel.functional.positive_tests.auto_gen_port.pty_type, this case will attach channel device with channel type="spicevmc", which is not supported on s390x, so filter out this case on s390x


chunfuwen:
@smitterl 
please take a review

dzhengfy (approved):
LGTM

PR #4558, opened 2022-09-28
migration: Add case about recover migration

cliping:
VIRT-294709 - [postcopy]VM live migration - recover migration when it is not paused

Signed-off-by: cliping <lcheng@redhat.com>

PR #4584, opened 2022-10-18
migration: Add case about pause migration

cliping:
VIRT-294705 - [postcopy]VM live migration - pause migration->do disruptive operations->recover it

Signed-off-by: cliping <lcheng@redhat.com>

### DIFF
--- a/provider/migration/migration_base.py
+++ b/provider/migration/migration_base.py
@@ -247,12 +247,9 @@ def poweroff_vm(params):
     """
     poweroff_vm_dest = "yes" == params.get("poweroff_vm_dest", "no")
     migration_obj = params.get("migration_obj")
-    test_case = params.get('test_case', '')
 
     if poweroff_vm_dest:
         dest_uri = params.get("virsh_migrate_desturi")
-        if test_case == "poweroff_vm":
-            time.sleep(90)
         backup_uri, migration_obj.vm.connect_uri = migration_obj.vm.connect_uri, dest_uri
         migration_obj.vm.cleanup_serial_console()
         migration_obj.vm.create_serial_console()
@@ -489,6 +486,7 @@ def do_common_check(params):
     migration_options = params.get("migration_options")
     second_bandwidth = params.get("second_bandwidth")
     migration_obj = params.get("migration_obj")
+    vm_name = params.get("main_vm")
 
     if migration_options == "migrateuri":
         libvirt_network.check_established(params)
@@ -496,8 +494,8 @@ def do_common_check(params):
         libvirt_domjobinfo.check_domjobinfo(migration_obj.vm, params)
 
     # check job info when migration is in paused status
-    expected_list = {"Job type": "Unbounded", "Operation": "Outgoing migration"}
-    libvirt_monitor.check_domjobinfo(params, expected_list)
+    expected_dict = {"Job type": "Unbounded", "Operation": "Outgoing migration"}
+    libvirt_monitor.check_domjobinfo(vm_name, expected_dict)
 
     # check domain state with reason
     check_vm_state(params)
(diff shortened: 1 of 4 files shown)


cliping:
Depends on https://github.com/autotest/tp-libvirt/pull/4558.

provider/migration/migration_base.py
```diff
@@ -430,3 +450,40 @@ def kill_libvirt_daemon(params):
...
+        cmd = "pkill -9 qemu-kvm"
+    elif operations == "poweroff":
+        poweroff_vm(params)
+        return
+    logging.debug("do disruptive operations: %s" % cmd)
+    if host == "dest":
+        remote.run_remote_cmd(cmd, params, ignore_status=False)
+    else:
+        process.run(cmd, shell=True, ignore_status=False)
+    time.sleep(sleep_value)
```
Yingshun, line 479: May I ask why we need to sleep for a while 'before' and 'after' doing some operations in L464 and L479, are they both needed?

provider/migration/migration_base.py
```diff
@@ -430,3 +450,40 @@ def kill_libvirt_daemon(params):
...
+
+
+def do_domjobabort(params):
```
Yingshun, line 482: Can you reuse virsh.domjobabort() directly?

provider/migration/migration_base.py
```diff
@@ -229,14 +229,35 @@ def check_event_output(params, test, virsh_session=None, remote_virsh_session=No
...
+
+    :param params: dict, get vm session or migration object
+    """
+    poweroff_vm_dest = "yes" == params.get("poweroff_vm_dest", "no")
+    migration_obj = params.get("migration_obj")
+    test_case = params.get('test_case', '')
+
+    if poweroff_vm_dest:
+        dest_uri = params.get("virsh_migrate_desturi")
+        if test_case == "poweroff_vm":
```
dzhengfy, line 244: Because this file is for shared purpose,  it is not good to make a judgement against a concrete case name here. Could you make it more transparent in order other cases might be able to reuse in future? For example, via a parameter like `need_sleep_time = 90` ?

provider/migration/migration_base.py
```diff
@@ -430,3 +450,40 @@ def kill_libvirt_daemon(params):
...
+
+
+def do_domjobabort(params):
+    vm_name = params.get("migrate_main_vm")
+    sleep_value = params.get("sleep_value", "10")
+
+    virsh_dargs = {'ignore_status': True, 'debug': True}
+    logging.debug("sleep %s", sleep_value)
+    time.sleep(int(sleep_value))
+    virsh.domjobabort(vm_name + " --postcopy", **virsh_dargs)
```
dzhengfy, line 489: Is the reason that we must sleep for few seconds then we have to define this function instead of using virsh.domjobabort directly?

dzhengfy, line 489: For `vm_name + " --postcopy"`, I suggest we enable` --postcopy ` in virsh.domjobabort function as a parameter instead of workaround here

dzhengfy (commented):
Others LGTM

provider/migration/migration_base.py
cliping: Updated, thanks.

cliping: Yes. Add need_sleep_time parameter in acting_during_mig, so we can use virsh.domjobabort now. Thanks.

provider/migration/migration_base.py
cliping: Updated, thanks.

provider/migration/migration_base.py
cliping: Updated, thanks.

dzhengfy (approved):
LGTM

cliping:
Depends on https://github.com/avocado-framework/avocado-vt/pull/3568 and https://github.com/autotest/tp-libvirt/pull/4619.

Yingshun (approved):

PR #4654, opened 2022-11-28
memory_misc: disable test on s390x

smitterl:
memory_misc..size_check requires memory hotplug which is not available on s390x.


chunfuwen (approved):
LGTM

chunfuwen:
one line change ,mark it as minor_fix

chunfuwen:
it is filtering out case, so no test result is needed

PR #4665, opened 2022-12-01
v2v: update the way of enabling sha1 algorithm

xiaodwan:
1. The 'update-crypto-policies --set LEGACY' changes the whole system setting which is not recommended by v2v.
The new way is a new solution of doing same thing.
2. Delete the orphan function set_graphics.

### DIFF
--- a/v2v/tests/src/function_test_xen.py
+++ b/v2v/tests/src/function_test_xen.py
@@ -15,6 +15,7 @@ from virttest import data_dir
 from virttest.libvirt_xml import vm_xml
 from virttest.staging import service
 from virttest.utils_test import libvirt
+from virttest.utils_conn import update_crypto_policy
 
 from provider.v2v_vmcheck_helper import VMChecker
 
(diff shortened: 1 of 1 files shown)


xiaodwan:
@yantian0 Please help to review this PR, thanks.

yantian0 (approved):

v2v/tests/src/function_test_xen.py
```diff
@@ -74,17 +75,31 @@ def log_fail(msg):
...
+        Enable the legacy sha1 algorithm.
         """
-        vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(
-            vm_name, virsh_instance=virsh_instance)
-        graphic = vmxml.xmltreefile.find('devices').find('graphics')
-        for key in param:
-            LOG.debug('Set %s=\'%s\'' % (key, param[key]))
-            graphic.set(key, param[key])
-        vmxml.sync(virsh_instance=virsh_instance)
+        ssh_config = ("Host %s\n"
```
chunfuwen, line 82: it looks it more sense that those config items is put into relevant cfg file

xiaodwan: Thanks for the suggestion. I'll think about a better place for them and modify it in future PR.

PR #4672, opened 2022-12-07
Fix SoftwareManager import

harihare:
avocado-framework > 94 clear init.py file
url：avocado-framework/avocado@1b8360c
Ref : https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2347 patch fixes the import issue.

Signed-off-by: Hariharan T S <hariharan.ts@ibm.com>


chunfuwen:
We may need use try ... except  when import since tp-libvirt also need keep compatibility with lower avocado version, eg. 92lts
```
try :
  from avocado.utils.software_manager import SoftwareManager
except:
   from avocado.utils.software_manager.manager import SoftwareManager
```

smitterl:
> We may need use try ... except when import since tp-libvirt also need keep compatibility with lower avocado version, eg. 92lts
> 
> ```
> try :
>   from avocado.utils.software_manager import SoftwareManager
> except:
>    from avocado.utils.software_manager.manager import SoftwareManager
> ```

IIUC, avocado-vt and tp-libvirt are intended to be aligned with latest master in other projects. If we want to fix usage on a specific avocado version I think we should introduce releases ourselves instead of adding complexity trying to make the import backwards compatible.

dzhengfy:
> > We may need use try ... except when import since tp-libvirt also need keep compatibility with lower avocado version, eg. 92lts
> > ```
> > try :
> >   from avocado.utils.software_manager import SoftwareManager
> > except:
> >    from avocado.utils.software_manager.manager import SoftwareManager
> > ```
> 
> IIUC, avocado-vt and tp-libvirt are intended to be aligned with latest master in other projects. If we want to fix usage on a specific avocado version I think we should introduce releases ourselves instead of adding complexity trying to make the import backwards compatible.

@smitterl ， To make our test stable, we do use avocado lts branch (like lts 92.0) instead of master branch. And we will upgrade to new lts in future.

libvirt/tests/src/cpu/powerpc_hmi.py
```diff
@@ -3,7 +3,7 @@
 import logging as log
 
 from avocado.utils import cpu
-from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.software_manager.manager import SoftwareManager
```
dzhengfy, line 6: Please consider compatibility with old versions, for example some teams might use old LTS. Thanks.

PR #4690, opened 2022-12-28
Revert "numa_memory: Check appropriate error messages for aarch64"

hs0210:
If the numa node setting is out of range, for aarch64, the error message should also be "NUMA node.*unavailable" which should be rasied in the configuration check.

When memory_mode is "strict" and memory_nodeset is out of range, the guest does not fail to start until it writes to cpuset.mems with the error message "Numerical result out of range", which should be a bug.


chunfuwen (approved):
LGTM

chunfuwen:
mark it as minor_fix since only two lines are changed

PR #4694, opened 2023-01-03
cpu/lifecycle: Test that time works

smitterl:
Add test case to confirm that after suspending and resuming a VM the clock still works as expected.

### DIFF
--- a/libvirt/tests/src/cpu/lifecycle_time.py
+++ b/libvirt/tests/src/cpu/lifecycle_time.py
@@ -1,0 +1,48 @@
+from time import sleep
+import logging as log
+
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest import virsh
+
+
+logging = log.getLogger('avocado.' + __name__)
+
+
+def run(test, params, env):
+    """
+    Test that the clock works after some action
+
+    :param test: test object
+    :param params: Dict with the test parameters
+    :param env: Dict with the test environment
+    :return:
+    """
+    vm_name = params.get("main_vm")
+
+    vm = env.get_vm(vm_name)
+    vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
+    vmxml_backup = vmxml.copy()
+
+    sleep_before_resume = int(params.get('sleep_before_resume', 5))
+
+    try:
+        session = vm.wait_for_login()
+        session.close()
+
+        virsh.suspend(vm_name)
+        sleep(sleep_before_resume)
+        virsh.resume(vm_name)
+
+        session = vm.wait_for_login()
+        cmd = 'time sleep 1'
+        out = session.cmd_output(cmd, print_func=logging.debug)
+        session.close()
+
+        lines = out.split('\n')
+        if len(lines) < 2 or 'real\t0m1.00' not in lines[1]:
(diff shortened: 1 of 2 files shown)


smitterl:
```
JOB ID     : a4499c800e6169f5ac24bbc637e66f31b2e8215d
JOB LOG    : /var/lib/avocado/job-results/job-2023-01-03T10.23-a4499c8/job.log
 (1/1) type_specific.io-github-autotest-libvirt.lifecycle.time.suspend_resume: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.lifecycle.time.suspend_resume: PASS (15.03 s)
JOB HTML   : /var/lib/avocado/job-results/job-2023-01-03T10.23-a4499c8/results.html
JOB TIME   : 16.37 s
```

smitterl:
@fbq815  @chunfuwen Please help review.

libvirt/tests/src/cpu/lifecycle_time.py
```diff
@@ -0,0 +1,45 @@
...
+        session = vm.wait_for_login()
+        cmd = 'time sleep 1'
+        out = session.cmd_output(cmd, print_func=logging.debug)
+        lines = out.split('\n')
+        if len(lines) < 2 or 'real\t0m1.00' not in lines[1]:
+            test.fail("VM seems to have slept longer than expected: %s" % out)
+    except Exception as e:
+        test.error("Test error: %s" % e)
+    finally:
+        vmxml_backup.sync()
```
chunfuwen, line 45: we may need destroy VM first, then vmxml_backup.sync()

libvirt/tests/src/cpu/lifecycle_time.py
```diff
@@ -0,0 +1,45 @@
...
+
+        virsh.suspend(vm_name)
+        sleep(sleep_before_resume)
+        virsh.resume(vm_name)
+
+        session = vm.wait_for_login()
+        cmd = 'time sleep 1'
+        out = session.cmd_output(cmd, print_func=logging.debug)
+        lines = out.split('\n')
+        if len(lines) < 2 or 'real\t0m1.00' not in lines[1]:
```
fbq815, line 40: 1. As we discussed with the test case, maybe 'real\t0m1.0' be enough? I'm afraid 1.00 too strict.
2. For libvirt testing, are we sure we could trigger this test case with SE guest? cause it seems under the cpu loop right now

smitterl: > 1. As we discussed with the test case, maybe 'real\t0m1.0' be enough? I'm afraid 1.00 too strict.
In my testing I always achieved an error < 0.01 so I'd like to try this and if we see it fails we can adjust. What do you think?

> 2. For libvirt testing, are we sure we could trigger this test case with SE guest? cause it seems under the cpu loop right now

I beiieve the test case is valid not only for secure guests on s390x, therefore I put it under /cpu because it's related with clock. Does this make sense?

libvirt/tests/src/cpu/lifecycle_time.py
smitterl: Done :)

fbq815 (approved):

chunfuwen (approved):
LGTM

PR #4701, opened 2023-01-09
passthrough/pci: add test for ISM device on s390x

smitterl:
ISM is a special device on s390x that can be passed through with vfio_pci. It's managed by libvirt as node device.

Add a test case that confirms that the device can be attached and is listed inside the guest.

### DIFF
--- a/libvirt/tests/cfg/passthrough/pci/ism_pci_passthrough.cfg
+++ b/libvirt/tests/cfg/passthrough/pci/ism_pci_passthrough.cfg
@@ -1,0 +1,7 @@
+- libvirt_pci_passthrough.ism:
+    only s390-virtio
+    type = ism_pci_passthrough
+    variants:
+        - happy_path:
+            start_vm = no
+            pci_dev = LIBVIRT_PCI_NAME
--- a/libvirt/tests/src/passthrough/pci/ism_pci_passthrough.py
+++ b/libvirt/tests/src/passthrough/pci/ism_pci_passthrough.py
@@ -1,0 +1,40 @@
+import logging as log
+
+from virttest.libvirt_xml.vm_xml import VMXML
+from virttest.libvirt_xml.nodedev_xml import NodedevXML
+
+
+logging = log.getLogger('avocado.' + __name__)
+
+
+def run(test, params, env):
+    """
+    Test for ISM device passthrough to libvirt guest.
+    """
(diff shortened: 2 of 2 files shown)


smitterl:
```
JOB ID     : 91a84b1481c57409537a05e8aa8442b4d312cf45
JOB LOG    : /var/lib/avocado/job-results/job-2023-01-09T10.44-91a84b1/job.log
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_pci_passthrough.ism.happy_path: PASS (17.23 s)
JOB HTML   : /var/lib/avocado/job-results/job-2023-01-09T10.44-91a84b1/results.html
JOB TIME   : 17.61 s
```

smitterl:
@chunfuwen @fbq815 Please help review, thanks!

libvirt/tests/cfg/passthrough/pci/ism_pci_passthrough.cfg
```diff
@@ -0,0 +1,7 @@
+- libvirt_pci_passthrough.ism:
+    only s390-virtio
+    type = ism_pci_passthrough
+    variants:
+        - happy_path:
+            start_vm = no
+            pci_dev = pci_0000_00_00_0
```
chunfuwen, line 7: Is pci_dev_address hard coded or it need to be configured per host ?

smitterl: It is hard coded here and can be configured per host. This way I want to avoid adding logic to select the right PCI device. Using logic to select the right CCW device in the passthrough/ccw tests has led to issues in the past, so I think it's better to set an explicit value and have for example the CI substitute it by the right value.

Alternative we could define a placeholder to make this intention more explicit:

```
        pci_dev = <LIBVIRT_PCI_NAME>
```
and in the py file we keep the default value `pci_0000_00_00_0` so that the expected format is still documented.

What do you think?

chunfuwen: I agree with this idea, so you can proceed it

chunfuwen (commented):
see inline comment

fbq815:
LGTM

chunfuwen (approved):
LGTM

PR #4714, opened 2023-01-18
v2v: fix HOME env issue when running in CI

xiaodwan:


xiaodwan:
The test is PASS in our internal CI environment.

PR #4720, opened 2023-01-20
virtual_network: make interface selection configurable

smitterl:
Currently, the several tests select the first host interface. In order to allow for testing with a specific interface in other test environments allow for value to be configured. Don't assign a value in the cfg in order to keep legacy behavior.


chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

dzhengfy:
@chloerh @yalzhang @Yingshun , FYI. When we design cases or automate cases, it 's better to take this into consideration. Thanks.

PR #4721, opened 2023-01-20
v2v: fix firmware check issue

xiaodwan:
The old firmware check is as a public checkpoint, but in fact, not all VMs have that attribute. And the original test case and bug only report the attribute should be show in xml only for vmware guests.

So I moved the check for all test cases and put it into a specific test case.


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.vmware_os_firmware.esx.esx_70.output_mode.rhev.rhv.it_default: PASS (11.02 s)

PR #4737, opened 2023-02-03
guest_numa.py: cancel test when out of memory for host

meinaLi:
In gating, rhel.guest_numa.possitive_test.hugepage.per_node.2M.topology.numatune_memnode.m_strict.numatune_mem will fail when the host has not enough memory for node1. It may affectthe gating result. So  decided to cancel it in that case。

### DIFF
--- a/libvirt/tests/src/numa/guest_numa.py
+++ b/libvirt/tests/src/numa/guest_numa.py
@@ -321,9 +321,13 @@ def run(test, params, env):
                     node_dict = i.copy()
                     node_dict['num'] = node_val
                     backup_list.append(node_dict)
-                    hp_cl.set_node_num_huge_pages(i['num'],
-                                                  i['nodenum'],
-                                                  i['size'])
+                    try:
+                        hp_cl.set_node_num_huge_pages(i['num'],
+                                                      i['nodenum'],
+                                                      i['size'])
+                    except ValueError as details:
+                        if "please check if the node has enough memory" in str(details):
+                            test.cancel(str(details))
                     node_val_after_set = hp_cl.get_node_num_huge_pages(i['nodenum'],
                                                                        i['size'])
                     if node_val_after_set < int(i['num']):


meinaLi:
```
# avocado run --vt-type libvirt --test-runner=runner --vt-machine-type q35 guest_numa.possitive_test.hugepage.per_node.2M
JOB ID     : f88e547246f5a2322ff227c3f9551c9772d12a96
JOB LOG    : /var/lib/avocado/job-results/job-2023-02-03T01.47-f88e547/job.log
 (01/16) type_specific.io-github-autotest-libvirt.guest_numa.possitive_test.hugepage.per_node.2M.topology.no_numatune_memnode.numatune_mem: PASS (27.86 s)
 (02/16) type_specific.io-github-autotest-libvirt.guest_numa.possitive_test.hugepage.per_node.2M.topology.no_numatune_memnode.no_numatune_mem: PASS (26.34 s)
 (03/16) type_specific.io-github-autotest-libvirt.guest_numa.possitive_test.hugepage.per_node.2M.topology.numatune_memnode.m_strict.numatune_mem: PASS (26.12 s)
......
 (15/16) type_specific.io-github-autotest-libvirt.guest_numa.possitive_test.hugepage.per_node.2M.no_topo.numatune_memnode.m_interleave.numatune_mem: PASS (26.22 s)
 (16/16) type_specific.io-github-autotest-libvirt.guest_numa.possitive_test.hugepage.per_node.2M.no_topo.numatune_memnode.m_interleave.no_numatune_mem: PASS (26.95 s)
RESULTS    : PASS 16 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /var/lib/avocado/job-results/job-2023-02-03T01.47-f88e547/results.html
JOB TIME   : 433.41 s
```

meinaLi:
Why update it in this method?
1) The skiplist.yaml can't support this skip because it doesn't have a bug.
2) We can use the free size of node1 in host as a condition, but can not get specific value. It mainly depends on different hosts.
3) I don't suggest update the function hp_cl.set_node_num_huge_pages in __init__.py of avocado. I think it may affect other tests(not only libvirt) if we directly update it.
So here I directly move this function to the guest_numa.py to cancel it when the auto case encounter this failure.

meinaLi:
@kylazhang @liang-cong-red-hat
This is that numa issue with not enough memory we discussed before. Please help to review it. Thanks.

chunfuwen (approved):
LGTM

libvirt/tests/src/numa/guest_numa.py
```diff
@@ -98,6 +99,24 @@ def run(test, params, env):
...
+        sys_node_path = "/sys/devices/system/node"
+        node_page_path = "%s/node%s" % (sys_node_path, node)
+        node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+        obj = kernel_interface.SysFS(node_page_path)
+        obj.sys_fs_value = num
+        # If node has some used hugepage, result will be larger than expected.
+        if obj.sys_fs_value < int(num):
+            test.cancel("Cannot set %s hugepages on node %s, please check"
+                        " if the node has enough memory" % (num, node))
+
```
chloerh, line 119: There's no need to copy the function here.

libvirt/tests/src/numa/guest_numa.py
```diff
@@ -321,9 +340,9 @@ def _update_qemu_conf():
                     node_dict = i.copy()
                     node_dict['num'] = node_val
                     backup_list.append(node_dict)
-                    hp_cl.set_node_num_huge_pages(i['num'],
-                                                  i['nodenum'],
-                                                  i['size'])
+                    set_node_num_huge_pages(i['num'],
+                                            i['nodenum'],
+                                            i['size'])
```
chloerh, line 345:
```suggestion
                    try:
                        hp_cl.set_node_num_huge_pages(i['num'],
                                                  i['nodenum'],
                                                  i['size'])
                     except ValueError as details:
                         if "please check if the node has enough memory" in str(details):
                             test.cancel(str(details))
                         else:
                             raise
                             
```

chloerh (changes requested):

liang-cong-red-hat (approved):
LGTM

libvirt/tests/src/numa/guest_numa.py
dzhengfy: @meinaLi , I agree with @chloerh because her solution is easier and less codes.

chunfuwen: I is just wondering why caught exception, rather than front judgement whether it satisfy pre-conditions like line 116, is recommended. Is there any industry practices ,which can be shared here?

chunfuwen: In my opinion, in python world, we may try to avoid catching  exception, particularly fatal exception something like ValueError, although we can from technologies view.

chloerh: > I is just wondering why caught exception, rather than front judgement whether it satisfy pre-conditions like line 116, is recommended. Is there any industry practices ,which can be shared here?

The function you mentioned at line 116 was copied from another file.

dzhengfy:
@Yingshun @cliping @nanli1 , I would like to hear your opinion. Thanks.

libvirt/tests/src/numa/guest_numa.py
Yingshun: > @meinaLi , I agree with @chloerh because her solution is easier and less codes.

+1. I think we should avoid copying codes from avocado-vt to tp-libvirt, especially for this case. In hp_cl.set_node_num_huge_pages(), the exception ValueError is a fixed error type if sysFS().sys_fs_value is smaller than the number we set, and from @chloerh 's suggestion, the error message will be checked as well. So it's ok for me.

meinaLi: Thanks for your opinions. This method is valid and clear. I've already updated it.

libvirt/tests/src/numa/guest_numa.py
meinaLi: Remove it now.

chloerh (approved):

Yingshun (approved):

dzhengfy (approved):

PR #4739, opened 2023-02-06
virsh_domdirtyrate_calc: no dirty-ring on s390 and aarch64

dzhengfy:
dirty-ring is not enabled on s390x and aarch64 so far in qemu-kvm.


dzhengfy:
Before fix:
[rhel.virsh.domdirtyrate_calc.positive.dirty_ring_mode]
TestFail: error: internal error: unable to execute QEMU command 'calc-dirty-rate': mode dirty-ring is not enabled, use other method instead.

After fix:
The case does not exist.

chloerh (approved):

PR #4744, opened 2023-02-06
Libvirt RNG, remove test cancel due to bug

vasekhodina:
Previously there was a test cancel step that stopped the test when multiple RNG devices were created. THis was done due to an existing bug. The bug is resolved and now this commit removes the test cancel.

### DIFF
--- a/libvirt/tests/src/libvirt_rng.py
+++ b/libvirt/tests/src/libvirt_rng.py
@@ -417,15 +417,27 @@ def run(test, params, env):
                           " the limitation of configuration")
         if device_num > 1:
             rng_dev = rng_avail.split()
-            if len(rng_dev) != device_num:
-                test.cancel("Multiple virtio-rng devices are not"
-                            " supported on this guest kernel. "
-                            "Bug: https://bugzilla.redhat.com/"
-                            "show_bug.cgi?id=915335")
+            compare_device_numbers(rng_dev)
             session.cmd("echo -n %s > %s" % (rng_dev[1], rng_files[1]))
             # Read the random device
             if session.cmd_status(cmd, timeout=timeout):
                 test.fail("Failed to read the random device")
+
+    def compare_device_numbers(rng_dev):
+        """
+        Compares number of entries in rng_dev list, while doing some cleanup
+        of said entries.
+
+        :param rng_dev: List of names of RNG devices
+        """
+        if vm_xml.VMXML.get_devices("tpm"):
+            rng_dev.remove("tpm-rng-0")
+        if "trng" in rng_dev:
+            rng_dev.remove("trng")
+        if len(rng_dev) != device_num:
+            test.fail("Number of rng devices defined and available does not match.\n"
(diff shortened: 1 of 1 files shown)


yanqzhan:
Looks good enough to resolve the tpm-rng issue.
But from L361-L372, I remember in s390 there's also an extra rng device that's not virtio_rng. (Sebas didn't modify codes from L418 maybe because s390 doesn't test this multiple rng case.) So how about only counting 'virtio_rng' for rng_avail and switch(L421) to another one from them?  @dzhengfy How do you think?

yanqzhan (approved):
Thanks! LGTM now.

libvirt/tests/src/libvirt_rng.py
```diff
@@ -417,16 +417,20 @@ def check_guest(session, expect_fail=False,
...
             if session.cmd_status(cmd, timeout=timeout):
                 test.fail("Failed to read the random device")
 
+    def compare_device_numbers(rng_dev):
+        if vm_xml.VMXML.get_devices("tpm"):
+            rng_dev.remove("tpm-rng-0")
+        if "trng" in rng_dev:
+            rng_dev.remove("trng")
+        if len(rng_dev) != device_num:
+            test.fail("Number of rng devices defined and available does not match")
```
dzhengfy, line 432: I would like to have more detailed information to be included in the test.fail, like rng_dev list info and device_num value.  They are helpful for debugging. Thanks

dzhengfy (commented):
Others LGTM

libvirt/tests/src/libvirt_rng.py
vasekhodina: I've added more debug output.

libvirt/tests/src/libvirt_rng.py
```diff
@@ -417,16 +417,28 @@ def check_guest(session, expect_fail=False,
...
+        :param rng_dev: List of names of RNG devices
+        """
+        if vm_xml.VMXML.get_devices("tpm"):
+            rng_dev.remove("tpm-rng-0")
+        if "trng" in rng_dev:
+            rng_dev.remove("trng")
+        if len(rng_dev) != device_num:
+            test.fail(("Number of rng devices defined and available does not match.\n"
+                       "Rng devices: %s\n"
+                       "Number of devices: %i"), rng_dev, device_num)
```
dzhengfy, line 440: The statement is incorrect. It should be 
```
test.fail("Number of rng devices defined and available does not match.\n"
                       "Rng devices: %s\n"
                       "Number of devices: %i" % (rng_dev, device_num))
```

vasekhodina: Fixed.

dzhengfy (approved):

dzhengfy:
@vasekhodina , there is a fix for this PR. FYI. https://github.com/autotest/tp-libvirt/pull/4767
I am not sure if this pr did pass for your test because there was no result provided here.

PR #4745, opened 2023-02-07
Fix virtual_network reconnect_tap case failures 

chunfuwen:
On some lab, when start ipv6 network, it may fail as below: virsh net-start nattest
error: Failed to start network nattest
error: internal error: Check the host setup: interface enc1 has kernel autoconfigured IPv6 routes and enabling forwarding  without accept_ra set to 2 will cause the kernel to flush them, breaking networking.

Refer to https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virtual_network/iface_network.py#L694, it need enable test_ipv6_address=yes to enforce accept_ra set to 2

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_network.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_network.cfg
@@ -163,7 +163,7 @@
                             net2_dhcp_end_ipv4 = '192.168.150.254'
                         - reconnect_tap:
                             reconnect_tap = "yes"
-                            test_ipv6_address = "no"
+                            test_ipv6_address = "yes"
                             test_ipv4_address = "no"
                             variants:
                                 - default:


chunfuwen:
Command 'virsh list' finished with 0 after 0.026764931s
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_forward.net_nat.reconnect_tap.default

chunfuwen:
previous failed:

[root@s390x-kvm-virtqez2 ~]# virsh net-start nattest
error: Failed to start network nattest
error: internal error: Check the host setup: interface enc1 has kernel autoconfigured IPv6 routes and enabling forwarding  without accept_ra set to 2 will cause the kernel to flush them, breaking networking.

chunfuwen:
@yalzhang 
@smitterl 
please take a look

smitterl (approved):
LGTM Thank you very much!

chunfuwen:
@smitterl  I updated that this is not specific to s390x change. 
After running change on x86, it also works well as is proved as by below x86 test result

Command 'virsh list' finished with 0 after 0.054166333s
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_forward.net_nat.reconnect_tap.default


PASS 1-type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_forward.net_nat.reconnect_tap.bridge_type

smitterl (approved):
LGTM thank you!

yalzhang:
Hi @chunfuwen  As we have discussed in mail sevaral months ago, the network start fail only happens on system with kernel autoconfigured IPv6 routes. Could we fix it like this: 
1). try to start the network with ipv6 setting;
2). if it succeed, then the case pass; 
3). if start fail, echo "2" *only* to the interface which was mentioned in the error message, and try to start the network again; it should succeed then.
Update the settings to all the interfaces is not necessary and dangerous since not all the values are the same. 
And do you think it is needed to restore the kernel parameters after the test?

chunfuwen:
@yalzhang  Do you see some risks that this will be exposed? In my understanding,  the purpose why libvirt add checks is to prevent network security. Since we test tap device filter, so the primary task to ensure that we can launch nattest  without comprising current feature net-filter.  Moreover, https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virtual_network/iface_network.py#L694 , I referred previous code here, so I believe it can follow up the workflow.

yalzhang:
> Do you see some risks that this will be exposed?
No. I don't know any risks that may be caused by this change. I just thought that it may be a overkill to change all the values.  
> Moreover, https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virtual_network/iface_network.py#L694 , I referred to previous code here, so I think we can follow up similar workflow.
There was a libvirt change: https://gitlab.com/libvirt/libvirt/-/blame/master/src/util/virnetdevip.c#L518 
with this change, network can start successfully without accept_ra=2 in some cases.  

> One more thing, I use sysctl net.ipv6.conf.all.accept_ra, and it return 1 or 2. So is it true for different interfaces can be different value ?
Yes, it can be different. On one of my system:ca
# tail -n +1 /proc/sys/net/ipv6/conf/*/accept_ra
==> /proc/sys/net/ipv6/conf/all/accept_ra <==
1

==> /proc/sys/net/ipv6/conf/default/accept_ra <==
1

==> /proc/sys/net/ipv6/conf/eno1/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/eno2/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/eno3/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/eno4/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/ens1f0/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/ens1f1/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/lo/accept_ra <==
1

Since I have no idea about  the risks, and considering the priority of the feature and ipv6, approve.

yalzhang (approved):

chunfuwen:
@dzhengfy  please help take a look.

dzhengfy (approved):

PR #4777, opened 2023-03-05
vcpu_cve: add case for cve status

dzhengfy:
This is to check guest cpu's CVE status.
Case ID: VIRT-294940

### DIFF
--- a/libvirt/tests/src/cpu/vcpu_cve.py
+++ b/libvirt/tests/src/cpu/vcpu_cve.py
@@ -1,0 +1,99 @@
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright Red Hat
+#
+#   SPDX-License-Identifier: GPL-2.0
+
+#   Author: Dan Zheng <dzheng@redhat.com>
+#
+
+"""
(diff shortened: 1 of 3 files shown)


dzhengfy:
` (1/1) type_specific.io-github-autotest-libvirt.vcpu.cve.guest_cpu_cve_status.host_passthrough: WARN: Test passed but there were warnings during execution. Check the log for details. (72.18 s)`

libvirt-9.0.0-7.el9.x86_64

dzhengfy:
@LuyaoHuang, please help review

Yingshun (approved):
LGTM

libvirt/tests/src/cpu/vcpu_cve.py
```diff
@@ -0,0 +1,86 @@
+import re
```
chunfuwen, line 1: what cve means here? is it one cpu sub feature name or is it something like bugzilla level we often mention. So maybe add one description in this python about what this python want to achieve.

LuyaoHuang:
ACK from my point view

chloerh (approved):

libvirt/tests/src/cpu/vcpu_cve.py
dzhengfy: Updated.

chunfuwen (approved):
LGTM

PR #4782, opened 2023-03-08
memory_misc: Disable numa tests on s390x

smitterl:
No NUMA on s390x, disable tests.


chunfuwen (approved):
LGTM

chunfuwen:
disable case on s390x, no need for test result

PR #4786, opened 2023-03-09
Fix for dommemstat error when libvirtd is off

bskjois:
Signed-off-by: Kowshik Jois B S <kowsjois@linux.ibm.com>

Description: libvirtd is restarted after issuing virsh commands from libvirt version 5.6.0 on. Hence need to fix test expectations when libvirtd is off

### DIFF
--- a/libvirt/tests/src/virsh_cmd/monitor/virsh_dommemstat.py
+++ b/libvirt/tests/src/virsh_cmd/monitor/virsh_dommemstat.py
@@ -1,6 +1,10 @@
+import logging as log
 from virttest import virsh
 from virttest import utils_libvirtd
 from virttest import ssh_key
+from virttest import libvirt_version
+
+logging = log.getLogger('avocado.' + __name__)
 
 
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virsh_cmd/monitor/virsh_dommemstat.py
```diff
@@ -69,6 +75,11 @@ def run(test, params, env):
     # check status_error
     if status_error == "yes":
         if status == 0:
+            if libvirtd == "off" and \
+                    libvirt_version.version_compare(5, 6, 0):
+                logging.info("From libvirt version 5.6.0 libvirtd is "
+                             "restarted and command should succeed")
+        else:
```
harihare, line 82: Else is place for wrong if, please check it.

bskjois: ohh yea.. thanks for finding that out. I will change it and push a new commit

PR #4798, opened 2023-03-13
virtual_devices/video_devices: disable unsupported cases on s390x

smitterl:
s390x only supports virtio-gpu.


chunfuwen:
minor fix

PR #4819, opened 2023-03-23
fix blockcopy not find xml

nanli1:
blockcopy case need to get xml path when do blockcopy command


chunfuwen (approved):
LGTM

chunfuwen:
suggest move this result out of pull request description, and append them as independent part

chunfuwen:
very trivial change ,and mark it as minor_fix

PR #4908, opened 2023-05-26
passt:Install socat if not installed

chloerh:


libvirt/tests/src/virtual_network/passt/passt_function.py
```diff
@@ -76,7 +77,7 @@ def run(test, params, env):
     if selinux_status != 'enforcing':
         utils_selinux.set_status('enforcing')
 
-    if not os.path.exists('/usr/bin/socat'):
+    if not utils_package.package_install('socat'):
```
Yingshun, line 80: can you add 'socat' package into install-pkgs in ci? As I know, we recommend checking if the command is available or not in the host and raising an exception if not, so I dont think we need to do anything to the previous code but update ci, what do you think?

chloerh: @Yingshun This was suggested by kvm qe who would like to include passt test to their test scope.

Yingshun: Test cases should focus on the test itself, and the installation of the required packages should be placed in the env preparation stage(eg. in ci). There are several OS and the package name of this command may be different. It's impossible to solve these kinds of problems in every test scripts, so I still think it's better to move package preparation for host to ci.

chloerh: Reasonable, I'll close this one. Thanks

PR #4928, opened 2023-06-05
Fixup:passt:Set shell=True when running check command

chloerh:
Or the command would fail.


cliping (approved):
Please upload test result. Others LGTM.

PR #4985, opened 2023-06-21
backingchain: fix s390x issue by using correct guest disk to check

meinaLi:
In s390x, the disk drive number vdb will be changed but not a stable one. So the hash checking will check the bootable disk but not the tested disk. That will cause different hash value contrast. This PR can be fixed to check test disks.

### DIFF
--- a/libvirt/tests/src/backingchain/blockcommit/blockcommit_all_block_chain.py
+++ b/libvirt/tests/src/backingchain/blockcommit/blockcommit_all_block_chain.py
@@ -59,7 +59,7 @@ def run(test, params, env):
         Do blockcommit and check backingchain result
         """
         session = vm.wait_for_login()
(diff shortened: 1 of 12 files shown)


meinaLi:
The issues are fixed in the RHEL 9 test result of backingchain jobs in s390x and x86_64 arches.

chunfuwen (approved):
lgtm

cliping (approved):
LGMT

dzhengfy (approved):
LGTM

dzhengfy:
@meinaLi Could you provide test result ?

Yingshun (commented):
@meinaLi  Could you please update commit message to explain the change?  BTW, it's better to describe the problem in simple language than give an internal bug id.

Yingshun (approved):

PR #5026, opened 2023-07-11
guest_os_booting_base.py: add firmware_type argument for ovmf guest

meinaLi:
We use libvirt_bios.remove_bootconfig_items_from_vmos(vmxml.os) in guest_os_booting_base.py. But for some ovmf guest which test the common attributes same with seabios guest. There's no need to run this step. So add a condition here.


dzhengfy (approved):
LGTM

PR #5032, opened 2023-07-17
cpu: Remove cases about iothread 'poll-' value validation

Yingshun:
The type of iothread's 'poll*' attributes were changed from unsigned int to unsigned long long and the validation of iothread 'poll-' value were removed from codes, so deleting the related cases.


dzhengfy (approved):
LGTM

PR #5035, opened 2023-07-17
migration: Replace set_vm_disk with modify_vm_device

Yingshun:
Updated to use a simpler function to modify the 1st disk of VM.

**Test results:**
` (1/1) type_specific.io-github-autotest-libvirt.migration.migration_uri.desturi.ipv4.tcp.p2p: PASS (515.96 s)
`

### DIFF
--- a/provider/migration/base_steps.py
+++ b/provider/migration/base_steps.py
@@ -14,6 +14,7 @@ from virttest import utils_iptables
 from virttest import utils_misc
 
 from virttest.utils_libvirt import libvirt_disk
+from virttest.utils_libvirt import libvirt_vmxml
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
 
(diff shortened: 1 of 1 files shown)


Yingshun:
@cliping could you plz help review this pr? Thanks!

cliping (approved):
LGTM

dzhengfy (approved):
LGTM

dzhengfy (commented):
But we need to pay attention to the replacement.  In old set_vm_disk, there is a start vm step at the end. So I am not sure if there are some tp-libvirt cases are depending on this step. @Yingshun

Yingshun:
> But we need to pay attention to the replacement. In old set_vm_disk, there is a start vm step at the end. So I am not sure if there are some tp-libvirt cases are depending on this step. @Yingshun

Yes, that's right. I've run 'function-migration_modular_1' job with this patch and all passed except for 2 cases with ioerror , which have nothing to do with this change.  And in setup_default(), it has 1 parameter(start_vm) to control vm's status in the following lines, so i think it's ok.

PR #5041, opened 2023-07-19
Exportfs command not found error fixing!

Anushree-Mathur:
Some tests require "exportfs" command which comes with the "nfs-utils" package so this patch will check if the nfs-utils package is installed else it will install that package, so the following error won't occur:

Error: Command 'exportfs' could not be found in any of the PATH dirs: ['/bin', '/root/.local/bin', '/root/bin', '/sbin', '/usr/bin', '/usr/libexec', '/usr/local/bin', '/usr/local/sbin', '/usr/sbin', '/usr/share/Modules/bin'] .
Signed-off-by :Anushree Mathur <anushree.mathur@linux.vnet.ibm.com>

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
@@ -14,6 +14,7 @@ from virttest.utils_test import libvirt
 from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml.devices.disk import Disk
 from virttest.staging.service import Factory
+from virttest.utils_package import package_install
 
 
 # Using as lower capital is not the best way to do, but this is just a
@@ -27,6 +28,9 @@ def run(test, params, env):
     1. unspecified error
     2. no space
     """
+    # Install nfs-utils package.
+    if not package_install("nfs-utils"):
+        test.cancel("Need nfs-utils package for exportfs command, can't install it")
 
     if not virsh.has_help_command('domblkerror'):
         test.cancel("This version of libvirt does not support domblkerror "
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_restore.py
@@ -15,7 +15,7 @@ from virttest import libvirt_version
 
 from virttest.libvirt_xml import vm_xml
 from virttest.utils_test import libvirt
-
+from virttest.utils_package import package_install
 
 # Using as lower capital is not the best way to do, but this is just a
 # workaround to avoid changing the entire file.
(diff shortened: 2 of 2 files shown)


libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -27,6 +30,17 @@ def run(test, params, env):
     1. unspecified error
     2. no space
     """
+    # Install virt-top package if missing.
+    software_mgr = manager.SoftwareManager()
+    if not software_mgr.check_installed('nfs-utils'):
+        logging.info('Installing nfs-utils package:')
+        software_mgr.install('nfs-utils')
+
```
bskjois, line 38: Can we check negative scenario as well here? 
What if it is not able to install the package ?

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -27,6 +30,17 @@ def run(test, params, env):
     1. unspecified error
     2. no space
     """
+    # Install virt-top package if missing.
```
smitterl, line 33: There is already function for this: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_package.py#L229

smitterl (changes requested):

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
Anushree-Mathur: @smitterl Thanks for mentioning this. I will modify the patch using this function.

Anushree-Mathur:
@smitterl I have modified the patch!

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
Anushree-Mathur: @bskjois In the modified patch i mentioned if the nfs-utils packages can't be installed it will give message!

bskjois (approved):
LGTM

Anushree-Mathur:
@chunfuwen @dzhengfy : Can you review and help me in merging this PR!

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
chunfuwen: there are some suggestions or common practices here
 we are not handling package install or uinstall in upstream code since it impact underlying system under testing if we do.
an alternative way is to install those packages in downstream CI

libvirt/tests/src/virsh_cmd/domain/virsh_domblkerror.py
```diff
@@ -27,6 +28,9 @@ def run(test, params, env):
     1. unspecified error
     2. no space
     """
+    # Install nfs-utils package.
+    if not package_install("nfs-utils"):
```
Anushree-Mathur, line 32: @chunfuwen I have seen this package installation policy in tp-libvirt code at many places. Is this something new change that have been introduced !

chunfuwen: @Anushree-Mathur previously there is package installation policy at many places in tp-libvirt. In practices, we learned some lessons, e.g package install conflict during calling software install, and suggestions from other upstream projects, so we don't suggest install sw in tp-libvirt code

PR #5085, opened 2023-08-04
Add  scsi device case

chunfuwen:
Add more scsi device case
    
    xx-186123: [SCSI] Hot-plug the same scsi device to two guests with shareable element
    xx-145209: [scsi] Check the qemu-pr-helper process can be restarted when vm issue pr cmds
    xx-98980:  [scsi] [negative] Generate duplicated addresses for scsi devices -- bug 1210587 (fixe

### DIFF
--- a/libvirt/tests/cfg/scsi/scsi_device.cfg
+++ b/libvirt/tests/cfg/scsi/scsi_device.cfg
@@ -28,6 +28,12 @@
                             controller_type = "scsi"
                             controller_model = "virtio-scsi"
                             controller_index = "0"
+                        - shared_by_two_guests:
+                            only hotplug
+                        - qemu_pr_helper:
+                            only coldplug
+                            managed = "yes"
+                            target_device = "vdb"
                 - negative:
                     variants test_scenario:
                         - max_sectors_controller:
@@ -38,6 +44,11 @@
                             max_sectors = "12345678676754"
                             define_error = "yes"
                             error_msg = "Invalid value for attribute 'max_sectors' in element 'driver': '12345678676754'"
+                        - duplicated_addresses_generate:
+                            only coldplug
+                            error_msg = "using disk target name 'sda' conflicts with SCSI host device address"
+                            define_error = "yes"
+                            target_device = "sda"
     variants plug_mode:
         - coldplug:
             coldplug = "yes"
--- a/libvirt/tests/src/scsi/scsi_device.py
+++ b/libvirt/tests/src/scsi/scsi_device.py
@@ -12,6 +12,8 @@
 import logging
 import os
 
+from avocado.utils import process
+
 from virttest import virt_vm
 from virttest import virsh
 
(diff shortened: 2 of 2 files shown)


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.scsi_device_manipulate.coldplug.scsi_hostdev.positive.qemu_pr_helper

libvirt/tests/cfg/scsi/scsi_device.cfg
```diff
@@ -2,6 +2,7 @@
     type = scsi_device
     status_error = "no"
     start_vm = "no"
+    vms = avocado-vt-vm1 vm1
```
dzhengfy, line 5: Ususally we config vm names in our local CI job definition instead of here.

For example, you can change it in shared/cfg/base.cfg or this cfg file
In shared/cfg/base.cfg, there is the definition which we use to configure the vms names
```
# List of virtual machine object names (whitespace separated)
vms = avocado-vt-vm1
```

libvirt/tests/src/scsi/scsi_device.py
```diff
@@ -438,6 +440,138 @@ def check_scsi_controller(test, params, env):
...
+    check scsi shareable attribute
+
+    :param test: one test object instance
+    :param params: dict wrapped with params
+    :param env: environment instance
+    """
+    vm_names = params.get("vms").split()
+    for vm_name in vm_names:
+        vm_output = virsh.dumpxml(vm_name).stdout_text
+        if 'shareable' not in vm_output:
```
dzhengfy, line 475: This is to match 'shareable' word in the whole vm xml. How about using more acurate searching by xpath, like using check_guest_xml_by_xpaths() in utils_libvirt/libvirt_vmxml.py ?

libvirt/tests/cfg/scsi/scsi_device.cfg
chunfuwen: remove vms parameter, and enable it in local CI

libvirt/tests/src/scsi/scsi_device.py
chunfuwen: In this case, shareable is special attribute to disk device. So matching keyword 'shareable' is enough.
Moreover, statement " if 'shareable' not in vm_output" also give reader meaningful expression that shareable attribute is checked in VM output.
So it is not necessary to check xpath here.

dzhengfy (approved):
LGTM

libvirt/tests/src/scsi/scsi_device.py
```diff
@@ -438,6 +440,140 @@ def check_scsi_controller(test, params, env):
...
+    for vm_name in vm_names:
+        vm = env.get_vm(vm_name)
+        if vm.is_dead():
+            vm.start()
+        vm.wait_for_login().close()
+        virsh.attach_device(vm_name, hostdev_xml.xml, flagstr="--live",
+                            ignore_status=False)
+
+
+def check_hostdev_shareable_attr(test, params, env):
```
cliping, line 464: There is no 'env' parameter to this function.

libvirt/tests/src/scsi/scsi_device.py
```diff
@@ -438,6 +440,140 @@ def check_scsi_controller(test, params, env):
...
+    vm_names = params.get("vms").split()
+    for vm_name in vm_names:
+        vmxml_checked = vm_xml.VMXML.new_from_dumpxml(vm_name)
+        hostdev_xml = vmxml_checked.get_devices('hostdev')[0]
+        shareable_attr = hostdev_xml.shareable
+        if shareable_attr is not True:
+            test.fail("VM: % failed to find shareable attribute in output: %s" % (vm_name, str(hostdev_xml)))
+
+
+def test_coldplug_scsi_hostdev_qemu_pr_helper(test, params, env):
```
cliping, line 481: Same as above

cliping (commented):
Others LGTM

libvirt/tests/src/scsi/scsi_device.py
chunfuwen: updated

libvirt/tests/src/scsi/scsi_device.py
chunfuwen: updated

cliping (approved):

PR #5099, opened 2023-08-09
Draft: passt: fix test cases on s390x

smitterl:
1. s390x doesn't have ACPI
2. s390x virtual interface name inside VM is encX
3. Remove unsupported DAC security driver for unprivileged user
   if present
4. Pass 'vm_iface' to 'passt.check_vm_ip' to avoid default value
   with wrong interface naming scheme
5. passt/check_nameserver: remove interface name from ipv6 entry as
   it might be different between host and guest
6. passt/check_default_gw: let us select which entry to choose because
   utils_net.get_default_gateway might return more than one, e.g. with
   multipath routes or when there's more than one interface on the host;
   same for passt/check_connection
7. passt/check_port_listen: only check for partial process user "'passt"
   because on x86_64 there's specializatoin 'passt.avx2' but on s390x it's
   just 'passt'
8. Don't use utils_net.get_host_ip_address because it doesn't let us choose
    which interface to get the ip address from; use utils_net.get_ip_address_by_interface
    instead

### DIFF
--- a/libvirt/tests/cfg/virtual_network/passt/passt_attach_detach.cfg
+++ b/libvirt/tests/cfg/virtual_network/passt/passt_attach_detach.cfg
@@ -2,6 +2,7 @@
     type = passt_attach_detach
     func_supported_since_libvirt_ver = (9, 0, 0)
     host_iface =
+    host_iface_index =
     outside_ip = 'www.redhat.com'
     start_vm = no
     mtu = 65520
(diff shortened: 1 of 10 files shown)


smitterl:
This has been split. 
Ref
https://github.com/autotest/tp-libvirt/pull/5361
https://github.com/avocado-framework/avocado-vt/pull/3818
https://github.com/autotest/tp-libvirt/pull/5363

PR #5135, opened 2023-08-28
add case for seclabel with file console

nanli1:
RHEL-115565: Test seclabel element can not be successfully validated in character devices

**Here are some issues needs to confirm with feature owner** 
1. May duplicate with another polarion case
2. Scenario 2> 3 step ,detail checkpoint is...?
3. Update xml with virsh edit or virsh define


nanli1:
Polarion case duplicate with VIRT-296954 and so close this pr.

PR #5153, opened 2023-09-04
memory: fix several tests on s390x

smitterl:
memory..define_value_unit: disable numa tests on s390x

There's no NUMA on s390x. Disable tests.

memory_backing/lifecycel_for_hugepage: fix tests on s390x

On s390x, supported huge page size for KVM is 1M.
Also, add kvm_module_parameters which allow to set 'hpage=1' to enable hugepage support in CI.

memory_allocation_config: disable numa test on s390x

No NUMA on s390x, disable test.

virtio.virtio_page_per_vq: disable on s390x

virtio-balloon-ccw doesn't support the page-per-vq property. Disable test.

### DIFF
--- a/libvirt/tests/cfg/memory/memory_backing/lifecycle_for_hugepage.cfg
+++ b/libvirt/tests/cfg/memory/memory_backing/lifecycle_for_hugepage.cfg
@@ -10,22 +10,38 @@
     set_pagesize ="1048576"
     set_pagenum = "2"
     free_hugepages_cmd= "cat /sys/kernel/mm/hugepages/hugepages-1048576kB/free_hugepages"
+    s390-virtio:
+        free_hugepages_cmd= "cat /sys/kernel/mm/hugepages/hugepages-1024kB/free_hugepages"
+        set_pagesize = "1024"
+        mount_size = "1024"
+        set_pagenum = "2048"
+        vm_nr_hugepages = 2048
+        kvm_module_parameters =
     variants:
         - memory_hugepage:
             variants:
                 - 4k:
+                    no s390-virtio
                     page_size = "4"
                     page_unit = "KiB"
                     memory_backing_dict = "'mb': {'hugepages': {'pages': [{'unit': '${page_unit}', 'size': '${page_size}'}]}}"
                     vm_attrs = {${memory_backing_dict}, 'memory_unit':'${mem_unit}','memory':${mem_value},'current_mem':${current_mem},'current_mem_unit':'${current_mem_unit}'}
                     HugePages_Free = "1024"
                     free_hugepages = "2"
+                - 1M:
+                    only s390-virtio
+                    memory_backing_dict = "'mb': {'hugepages': {}}"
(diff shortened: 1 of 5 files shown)


smitterl:
```
# avocado run --test-runner=runner --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system memory.backing.lifecycle,memory.allocation.define_value_unit.with_numa.positive_test.bytes,memory.allocation.config.positive_test.with_numa,virtio.virtio_page_per_vq.memballoon.default_start --vt-extra-params "kvm_module_parameters = hpage=1"
No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
No python imaging library installed. PPM image conversion to JPEG disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
JOB ID     : 86d7eaa219fa263db43cfce951dd3c737bf52245
JOB LOG    : /var/lib/avocado/job-results/job-2023-09-04T06.14-86d7eaa/job.log
 (1/3) type_specific.io-github-autotest-libvirt.memory.backing.lifecycle.memory_hugepage.1M: PASS (40.43 s)
 (2/3) type_specific.io-github-autotest-libvirt.memory.backing.lifecycle.memory_hugepage.0: PASS (7.80 s)
 (3/3) type_specific.io-github-autotest-libvirt.memory.backing.lifecycle.memory_hugepage.scarce_mem: PASS (8.36 s)
JOB HTML   : /var/lib/avocado/job-results/job-2023-09-04T06.14-86d7eaa/results.html
JOB TIME   : 57.02 s
```

chunfuwen (approved):
lgtm

libvirt/tests/cfg/memory/memory_backing/lifecycle_for_hugepage.cfg
```diff
@@ -10,22 +10,38 @@
     set_pagesize ="1048576"
     set_pagenum = "2"
     free_hugepages_cmd= "cat /sys/kernel/mm/hugepages/hugepages-1048576kB/free_hugepages"
+    s390-virtio:
+        free_hugepages_cmd= "cat /sys/kernel/mm/hugepages/hugepages-1024kB/free_hugepages"
+        set_pagesize = "1024"
+        mount_size = "1024"
+        set_pagenum = "2048"
+        vm_nr_hugepages = 2048
+        kvm_module_parameters =
```
nanli1, line 19: Just confirm , Is this  `kvm_module_parameters = ` correct or it is same as `kvm_module_parameters = ''`

nanli1 (approved):

libvirt/tests/cfg/memory/memory_backing/lifecycle_for_hugepage.cfg
smitterl: @nanli1 Practically it's the same because both cases cause the kvm_module_parameters not to be set and both lines can be replaced in CI or manually to set parameters. However, technically it's not exactly the same, I believe. I assume `kvm_module_parameters =` sets the value to `None` and `''` sets it to empty string. When doing substitution in CI we use yaml to describe that. Using `''` in yaml strings _might_ conflict with `""`. Therefore I chose here to use `kvm_module_parameters =` just to be on the safe side and I think I will use it in the same way from now on.

PR #5154, opened 2023-09-04
virsh.emulatorpin: add alternative error msg for negative test

smitterl:
On s390x, there's a different but valid error message. Expect the message alternatively on failure.


chunfuwen (approved):
lgtm

PR #5156, opened 2023-09-05
virsh_event: change mount-point to avoid conflict at teardown

smitterl:
The test
 io-error_event.no_timeout.loop
created a small file inside of `tmpdir` and then mounted it upon its parent directory `tmpdir`. While this seems to lead to the desired io error events later, it also seems to lead to conflicts during teardown and the test dies "without reporting status".

Instead, create a subdirectory to test. This makes also sense because tmpdir is used by avocado-vt for vm xmls and other data.


chunfuwen (approved):
lgtm

Yingshun (approved):

PR #5165, opened 2023-09-08
virtio: update virtio pattern to virtio_attributes

meinaLi:
That because virtio pattern is not suitable when searching series of these cases to run.


dzhengfy (approved):
LGTM

PR #5169, opened 2023-09-11
Improve compatibility by filtering serial device operations

Aakanksha0311:
Previously, an error occurred when attempting to hot-plug or unplug serial devices on the arm64 architecture. This is due to the lack of support for these operations on arm64.
To resolve this issue, this commit refactors the configuration file to exclude hot-plugging and unplugging of serial devices specifically for arm64. By doing so, we ensure better compatibility and prevent unknown errors.


Yingshun:
@Aakanksha0311  Can you plz update message in [commit?](https://github.com/autotest/tp-libvirt/pull/5169/commits/bd0faec1ffacb219647d278ed72340e186fd0493)

PR #5184, opened 2023-09-15
npiv: update the check result of negative test

meinaLi:
In negative test, we only need to check the mkfs result. If it was not permitted, that means it's expected for readonly npiv disk. So here don't use the check_vm_disk() function to check detailly.

### DIFF
--- a/libvirt/tests/src/npiv/npiv_virtual_disk.py
+++ b/libvirt/tests/src/npiv/npiv_virtual_disk.py
@@ -213,8 +213,13 @@ def run(test, params, env):
         # Check disk in VM
         output = session.cmd_status_output('mkfs.ext4 -F %s' % new_disk)
         logging.debug("mkfs.ext4 the disk in vm, result: %s", output[1])
-        if not check_vm_disk(session, new_disk, readonly):
-            raise exceptions.TestFail("Failed check the disk in vm.")
+        if readonly == "yes":
+            if "Operation not permitted" in str(output):
+                logging.debug("Can't write readonly disk in vm")
+                return
+            else:
+                test.fail("A readonly virtual disk operated as "
+                          "a non-readonly disk.")
         session.cmd_status_output('umount %s' % new_disk)
         # Detach disk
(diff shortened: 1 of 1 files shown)


meinaLi:
```
# avocado run --vt-type libvirt --test-runner=runner --vt-machine-type q35 npiv.virtual_disk.negative_testsing.hot_attach_detach.readonly
JOB ID     : a1bcbb7631746abd5c114d337e17fa22ff4f0efc
JOB LOG    : /var/lib/avocado/job-results/job-2023-09-14T23.37-a1bcbb7/job.log
 (1/1) type_specific.io-github-autotest-libvirt.npiv.virtual_disk.negative_testsing.hot_attach_detach.readonly: PASS (31.83 s)
JOB HTML   : /var/lib/avocado/job-results/job-2023-09-14T23.37-a1bcbb7/results.html
JOB TIME   : 33.44 s
```

libvirt/tests/src/npiv/npiv_virtual_disk.py
```diff
@@ -213,8 +213,14 @@ def run(test, params, env):
         # Check disk in VM
         output = session.cmd_status_output('mkfs.ext4 -F %s' % new_disk)
         logging.debug("mkfs.ext4 the disk in vm, result: %s", output[1])
-        if not check_vm_disk(session, new_disk, readonly):
-            raise exceptions.TestFail("Failed check the disk in vm.")
+        if readonly == "yes":
```
chunfuwen, line 216: there are confusion here:
1)what if  readonly == "no", sound like we miss this condition
2) Do you mean if readonly == "yes",  ,the program need return whatever something happen ?
 if that is true, it only need use return statement once at end of these code

meinaLi: 1) If readonly=="no", the codes will skip this `if` condition and continue to run the next steps. So I think may be no need to add it.
2) I only keep one return statement in it now. And use test.fail() in `else` statement.

chunfuwen (approved):
lgtm

chloerh (approved):

PR #5210, opened 2023-09-28
graphics: Cleanup env

cliping:
On RHEL 8, after run graphics.functional.positive_tests.spice_only .listen_type_network.network_bridge case, a bridge named linux_bridge will be left. This will cause some cases to fail. So update case.


cliping:
Before:
TestFail: Expect not listening on 2620:0052:0000:49c4:def4:01ff:fee8:062d%42:5900 but true

After:
(1/1) type_specific.io-github-autotest-libvirt.graphics.functional.positive_tests.vnc_only.vnc_listen_ipv4: PASS (12.63 s)

dzhengfy (approved):
LGTM.   Thanks for fixing

PR #5285, opened 2023-11-20
migration: Remove disk size checkpoint

cliping:
Since the image on source host will be compressed and the image on targe host will not be compressed, the disk size of the image will be inconsistent on the source host and target host. So remove disk size checkpoint.


cliping:
Test result:

Before:
TestFail: Check disk size failed: old: 1084479242, new: 2233382993.

After:
 (1/1) type_specific.io-github-autotest-libvirt.migration_with_copy_storage.migration_with_backingchain.with_backing_chain.copy_storage_inc.local_storage.p2p: PASS (371.13 s)

nanli1 (approved):

chloerh (approved):

PR #5299, opened 2023-11-27
memory_balloon: diable new virtio-model tests on s390x

smitterl:
The virtio models transitional and non-transitional are not available on s390x. Disable tests.


chunfuwen (approved):
lgtm

PR #5309, opened 2023-11-29
Filter out pcibridge.positive_test.hotplug.pci_br.no_device on s390x

chunfuwen:
Filter out pcibridge.positive_test.hotplug.pci_br.no_device on s390x

It also filter out on q35 on x86_64


chunfuwen:
@smitterl please double confirm this

smitterl:
> @smitterl please double confirm this

IIUC we don't have devices downstream for this but there are upstream ref. https://virtualpenguins.blogspot.com/2018/02/notes-on-pci-on-s390x.html

Maybe let's skip this only downstream.

chunfuwen:
as per smitterl's suggestion , close this PR

PR #5346, opened 2023-12-14
virsh_dump: add new case when no space left

smitterl:
VMs should keep running if dumping them fails because there's no space.

### DIFF
--- a/libvirt/tests/cfg/guest_kernel_debugging/virsh_dump.cfg
+++ b/libvirt/tests/cfg/guest_kernel_debugging/virsh_dump.cfg
@@ -95,3 +95,6 @@
                     setup_libvirt_polkit = "yes"
                     unprivileged_user = "EXAMPLE"
                     virsh_uri = "qemu:///system"
+                - no_space_left:
+                    dump_options = "--memory-only"
+                    dump_dir = "/var/tmp/too_small"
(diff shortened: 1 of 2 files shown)


smitterl:
@fbq815 please help review, thanks!

fbq815:
LGTM

libvirt/tests/cfg/virsh_cmd/domain/virsh_dump.cfg
```diff
@@ -95,3 +95,6 @@
                     setup_libvirt_polkit = "yes"
                     unprivileged_user = "EXAMPLE"
                     virsh_uri = "qemu:///system"
+                - no_space_left:
```
dzhengfy, line 98: Codes LGTM. 
Could you please move the cfg and python files to libvirt/tests/cfg/guest_kernel_debugging/ and  libvirt/tests/src/guest_kernel_debugging/? This is because we plan to make virsh_cmd folder deprecated in future and put cases into the feature folder accordingly.

smitterl: @dzhengfy OK done.

dzhengfy (approved):
LGTM

PR #5353, opened 2023-12-19
Unify the name of the vdpa cases

yalzhang:
The jobs.yaml run vdpa related case in a separate job with a pattern "vdpa_interface". And these cases are excluded from the network job by the pattern "vdpa_interface" since it needs specific hardware.


yalzhang:
@Yingshun  Please help to review, Thank you!

Yingshun (approved):

PR #5361, opened 2023-12-21
provider//passt: add changes for multiarch

smitterl:
Depends on https://github.com/avocado-framework/avocado-vt/pull/3818

'passt.avx2' is a specialization only for x86_64.
On other architectures the process might just be "passt".

Change the regular expression to match any process containing
"passt".

`check_vm_ip` calculated the guest interface name for x86_64
for other architectures allow for configuration of this value.

Remove zone index when comparing routes as they might differ
between host and guest (e.g. if the interfaces are named differently).

Check gateways by confirming that routes in guest are available on
host instead of confirming they must be exactly the same on both ends.

Pass the target interface name to the check_connection function to
allow for the correct selection of the gateway. When a host
has two interfaces the function used a list and failed.

`check_protocol_connection` sent a command to start listening to
the server VM and the client send a message to quickly. We could
establish a wait condition like checking ports of the machine
but a static sleep seemed like a simpler and still reliable solution.

The port list definition for `check_portforward` needs to contain
the host interface, not the VM interface. Therefore, pass the parameter
and remove the unused vm_iface.

### DIFF
--- a/provider/virtual_network/passt.py
+++ b/provider/virtual_network/passt.py
@@ -1,6 +1,8 @@
 import logging
 import os
+import re
 import random
+import time
 from socket import socket
 
 import aexpect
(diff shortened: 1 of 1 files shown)


smitterl:
@chloerh Please help review
@dzhengfy for visibility

smitterl:
@chloerh please help review

smitterl:
@yalzhang please help review

yalzhang (commented):
LGMT

yalzhang (approved):

chloerh (approved):

PR #5369, opened 2023-12-28
numa: Fix nodeset sequence

hs0210:
set() is unordered, list() cannot sort data, so there is no guarantee that nodes will be spliced in order. Use sort() instead.

For numa_base.convert_to_string_with_dash(), 
"0, 1" will be converted to "0-1" which is an expected result.
"1, 0" will just keep "1, 0".

Before:
```
FAIL 1-type_specific.io-github-autotest-libvirt.guest_numa_node_tuning.host_guest_mixed_memory_binding.memnode_mode_strict.mem_mode_strict.single_node -> TestFail: Expect cpuset.mems to be 1,0, but found 0-1
```

After:
```
PASS 1-type_specific.io-github-autotest-libvirt.guest_numa_node_tuning.host_guest_mixed_memory_binding.memnode_mode_strict.mem_mode_strict.single_node
```


chunfuwen (approved):
lgtm

dzhengfy (approved):
LGTM, thanks for fixing

PR #5375, opened 2024-01-03
Get the vm_iface name from the guest OS

yalzhang:
With acpi feature, we can assume that the vm_iface is "eno1" with the <acpi index="1"/> setting. But it only works on x86 now. Update it to get the vm_iface name from the guest OS instead.


yalzhang:
depends on https://github.com/autotest/tp-libvirt/pull/5361

libvirt/tests/src/virtual_network/passt/passt_attach_detach.py
```diff
@@ -124,7 +125,9 @@ def run(test, params, env):
                 test.fail(f'Logfile of passt "{log_file}" not created')
 
             session = vm.wait_for_serial_login(timeout=60)
-            passt.check_vm_ip(iface_attrs, session, host_iface)
+            if platform.machine() != 'x86_64':
+                vm_iface = utils_net.get_linux_ifname(session, mac)
+            passt.check_vm_ip(iface_attrs, session, host_iface, vm_iface)
```
smitterl, line 130: `get_linux_ifname` also work on x86_64? If so, I think you should remove the `if` condition and just run the same code for all archs.

yalzhang: Updated. Thank you.

chloerh (approved):

yalzhang:
Test on x86, pass

smitterl (approved):
LGTM, thank you!

PR #5382, opened 2024-01-08
passt/function: fix for multi-arch

smitterl:
Depends on https://github.com/autotest/tp-libvirt/pull/5361

1. Configure values for s390x
2. Get IP addresses by host interface to avoid issues when there's more than one functional NIC.
3. Use the host_iface in several functions to make sure the right interface is tested.
4. The test needs to check the ports on the host not in the guest. Therefore, configure the host_iface.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/passt/passt_function.cfg
+++ b/libvirt/tests/cfg/virtual_network/passt/passt_function.cfg
@@ -22,6 +22,9 @@
             iface_attrs = {'model': 'virtio', 'acpi': {'index': '1'}, 'type_name': 'user', 'backend': {'type': 'passt'}, 'source': {'dev': host_iface}}
             vm_iface = eno1
             vm_ping_outside = pass
+            s390-virtio:
+                iface_attrs = {'model': 'virtio', 'type_name': 'user', 'backend': {'type': 'passt'}, 'source': {'dev': host_iface}}
+                vm_iface = enc1
         - ip_portfw:
             ipv6_prefix = 128
             alias = {'name': 'ua-c87b89ff-b769-4abc-921f-30d42d7aec5b'}
@@ -30,6 +33,8 @@
             vm_iface = eno1
             vm_ping_outside = pass
             vm_ping_host_public = pass
+            s390-virtio:
+                vm_iface = enc1
             variants:
                 - ip_addr:
                     portForward_0 = {'attrs': {'proto': 'tcp'}, 'ranges': [{'start': '31339', 'to': '41339'}]}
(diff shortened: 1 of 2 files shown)


smitterl:
Expect checks to pass after dependency has been merged.

smitterl:
@chloerh @yalzhang please help review
@dzhengfy for visibility on arm

libvirt/tests/cfg/virtual_network/passt/passt_function.cfg
```diff
@@ -65,9 +70,11 @@
                     portForward_1 = {'attrs': {'proto': 'udp', 'address': host_ip, 'dev': host_iface}, 'ranges': [{'start': '2025'}]}
                     portForward_2 = {'attrs': {'proto': 'tcp', 'address': host_ip_v6, 'dev': host_iface}, 'ranges': [{'start': '51339'}]}
                     portForwards = [${portForward_0}, ${portForward_1}, ${portForward_2}]
-                    tcp_port_list = [f'{host_ip}%{vm_iface}:31339', f'[{host_ip_v6}]%{vm_iface}:51339']
-                    udp_port_list = [f'{host_ip}%{vm_iface}:2025']
+                    tcp_port_list = [f'{host_ip}%{host_iface}:31339', f'[{host_ip_v6}]%{host_iface}:51339']
+                    udp_port_list = [f'{host_ip}%{host_iface}:2025']
```
yalzhang, line 74: good catch!

libvirt/tests/src/virtual_network/passt/passt_function.py
```diff
@@ -52,17 +52,17 @@ def run(test, params, env):
...
 
-    host_ip = utils_net.get_host_ip_address(ip_ver='ipv4')
-    params['host_ip_v6'] = host_ip_v6 = utils_net.get_host_ip_address(
-        ip_ver='ipv6')
+    host_iface = params.get('host_iface')
+    host_iface = host_iface if host_iface else utils_net.get_net_if(
+        state="UP")[0]
+    host_ip = utils_net.get_ip_address_by_interface(host_iface, ip_ver='ipv4')
+    host_ip_v6 = utils_net.get_ip_address_by_interface(host_iface, ip_ver='ipv6')
+    params['host_ip_v6'] = host_ip_v6
```
yalzhang, line 60: Maybe we can check the interface xml first, if the "dev" is specified, use it directly. If it is null, try to get the host_iface then. And for the host_ip_v6, if the host interface has 2 global ipv6 addresses(like in your test env.),  it will use the more specific one - the one got by dhcp with prefix length as 128. Do we need to add some check here?

yalzhang: > The host_iface value would be used later to set the dev value l. 86.
Oh, sorry, I misunderstood it.
> At this point, I have a test environment like on x86_64 without dhcp. Therefore the check is not needed IMO.
Okay, then it should work properly.

yalzhang (approved):

chloerh (approved):

PR #5383, opened 2024-01-09
virsh_crash: Test to check environment behavior after test

smitterl:
The test crashes virsh. It can be used to check if the core file will be recorded and if the coredump will be reported.


smitterl:
This change was only used to test somethign in our CI. Closing.

PR #5417, opened 2024-01-29
Skip the ipv6 default route check

yalzhang:
There are some systems with multipath ipv6 route which passt can not recognize. Skip the ipv6 default route check on such system.


yalzhang:
depend on https://github.com/avocado-framework/avocado-vt/pull/3836

chloerh (approved):

PR #5427, opened 2024-01-30
virtual_disks/file_descriptor: fix disk discovery

smitterl:
The test identified new disks by comparing the list of disks before and after plug. However, a reboot can mess up the non-persistent disk namings /dev/vdX.

Use instead a function that identifies a new disk inside of a VM as the one disk that has no root mount on itself, its partitions or volumes.


chunfuwen (approved):
lgtm

PR #5436, opened 2024-02-02
virtual_networks: Fix the cutting of ip addr

hs0210:
replace() here will cut "xxx.xxx.122.122" to "xxx.xxx" which is not expected.

Before:
```
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.connectivity_check.mcast_interface.default:  FAIL: omping should fail with remote vm firewalld on (76.35 s)
```

After:
```
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.connectivity_check.mcast_interface.default: PASS (99.06 s
```


chunfuwen (approved):
lgtm

PR #5442, opened 2024-02-05
Check the cpu topology from virsh capabilities is the subset of result from system

liang-cong-red-hat:
Since in rhel8 the cpu cluster_id is not available from virsh capabilities, so change the comparison logic to checking the cpu topology from virsh capabilities is the subset of result from system

result on rhel9:
 (1/1) type_specific.io-github-autotest-libvirt.host_numa.numa_info.default: PASS (8.75 s)

result on rhel8:
 (1/1) type_specific.io-github-autotest-libvirt.host_numa.numa_info.default: PASS (10.94 s)


dzhengfy (approved):
LGTM

PR #5456, opened 2024-02-18
vcpu_misc: fix cpu model for arm

dzhengfy:
- change cpu mode
- no vendor information on arm now
- maxphysaddr is not supported on arm


dzhengfy:
Before the fix:
```
[vcpu_misc.positive_test.vendor_id]
TestFail: error: Failed to start domain 'avocado-vt-vm1'error: unsupported configuration: CPU mode 'host-model' for aarch64 kvm domain on aarch64 host is not supported by hypervisor&#10;
```

chunfuwen (approved):
lgtm

PR #5476, opened 2024-02-27
Fix the error message

yalzhang:
The error message changes for the negative scenario to update the mac address. Update the error message in the cfg file to match it.


yalzhang:
before:
TestFail: Expect should fail with one of ['device not found: no device found at address .* matching MAC address .* and alias .*'], but failed with:error: Failed to update device from /tmp/xml_utils_temp_ga51um9j.xmlerror: Operation not supported: cannot change network interface mac address from 9a:ae:52:d7:16:28 to 52:54:00:66:c3:62&#10;

after:
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.update_device.unsupported_live_update.delete.mac: PASS (50.32 s)

Yingshun (approved):

PR #5495, opened 2024-03-06
libvirt_rng: fix count only for arm

dzhengfy:
- dd count value too small on arm will not make the rate control take effect. Fix dd count only for arm and do not impact for other arches.
- Reorg the rng related files into corresponding directory.


libvirt/tests/cfg/security/rng/libvirt_rng.cfg
```diff
@@ -90,7 +90,8 @@
         - rng_rate:
             test_qemu_cmd = "yes"
             test_guest = "yes"
-            dd_throughput = "bs=512 count=100"
+            aarch64:
```
chunfuwen, line 93: it shows there are 4 files changes, there maybe something wrong when committing

chunfuwen (approved):
lgtm

chunfuwen:
@dzhengfy maybe we have one test result available

xiaodwan (approved):

cliping (approved):
LGTM

PR #5510, opened 2024-03-13
Graphical: fix an intermittent network issue caused by network configuration

meinaLi:
Fix the intermittent error: LibvirtXMLError: Failed to start network virt-test-nat.Detail: error: Failed to start network virt-test-naterror: internal error: Check the host setup: interface eth0 has kernel autoconfigured IPv6 routes and enabling forwarding without accept_ra set to 2 will cause the kernel to flush them, breaking networking.


meinaLi:
Consider closing this PR because this auto case isn't a real use case and most of time it will be skipped by an expected failures. Consequently, there's no need to update it ti fix the trivial intermittent network issue.

PR #5513, opened 2024-03-14
passt.interface_function: get interface name dynamically

yafu-1:


chunfuwen (approved):
lgtm

libvirt/tests/src/virtual_network/passt/passt_function.py
```diff
@@ -97,6 +97,8 @@ def run(test, params, env):
             test.fail(f'Logfile of passt "{log_file}" not created')
 
         session = vm.wait_for_serial_login(timeout=60)
+        vm_iface_info = utils_net.get_linux_iface_info(mac=mac, session=session)
```
dzhengfy, line 100: On line 63,  `vm_iface = params.get('vm_iface', 'eno1') `
It seems above line could be removed. Right?

yafu-1: Existing pr to fix the issue:
https://github.com/autotest/tp-libvirt/pull/5375
So I will close this one. Thanks for reviewing.

PR #5515, opened 2024-03-15
Skip ipv6 default route check with multipath route

yalzhang:
Some systems has ipv6 multipath route which will cause the passt check fail since passt can not recognize the format of it. Skip this check.


chloerh (approved):

yafu-1 (approved):
LGTM

PR #5525, opened 2024-03-18
migartion: Add libvirt version check for domjobabort

cliping:
Add libvirt version check for 'domjobabort --postcopy' command.


cliping:
Before:
TestFail: Expect should fail with one of ['cannot acquire state change lock .*held by monitor=remoteDispatchDomainMigrateFinish3Params'], but failed with:error: command 'domjobabort' doesn't support option --postcopy


After:
 (1/1) type_specific.io-github-autotest-libvirt.migration.abort_precopy_migration.abort_by_domjobabort_on_target.with_postcopy.p2p: CANCEL: This libvirt version doesn't support this function. (94.45 s)

Yingshun (approved):

PR #5527, opened 2024-03-18
Add version check for interface with all zeros mac

yalzhang:
Since the issue is fixed on libvirt-9.9.0, add the version check for interface with all zeros mac.


yalzhang:
before the fix, the case will fail with older libvirt version.
TestFail: Expect mac address on vmxml: fe:, not 00:00:00:00:00:00&#10;
after:
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.elements_and_attributes.mac.specific_addr.all_zeros.pre_set: CANCEL: This libvirt version doesn't support this function.

yalzhang:
@chloerh  Please help to review, Thank you!

cliping (approved):
LGTM

Yingshun (approved):

PR #5534, opened 2024-03-20
virtual_networks: remove spaces before comparing rx_frames

QianJianhua1:
Fixed a bug related to rx_frames comparison.
The issue was that the collected rx_frames contained spaces, while the expected rx_frames did not.

Before:
``` 
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.update_device.coalesce.nat_net.update_iface_coalesce: FAIL: rx-frames of vnet0 should be 64, not  64 (32.15 s)
```

After:
```
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.update_device.coalesce.nat_net.update_iface_coalesce: PASS (36.92 s)
```

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_rerror_policy.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_rerror_policy.py
@@ -57,19 +57,19 @@ def create_customized_nbd_disk(params, nbd_server_host):
     return customized_disk
 
 
-def operate_guest_disk_after_killing_nbdserver(vm, device_target):
+def operate_guest_disk_after_killing_nbdserver(vm):
     """
     Operate guest disk after killing nbd server.
 
     :params vm: VM instance
-    :params device_target: VM device target
     """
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_network/update_device/update_device_coalesce.py
```diff
@@ -25,7 +25,7 @@ def check_coalesce_rx_frames(tap_device, rx_frames, test):
     """
     coal_params = network_base.get_ethtool_coalesce(tap_device)
     LOG.debug(f'coalesce parameters from ethtool: {coal_params}')
-    coal_rx_frames = coal_params.get('rx-frames')
+    coal_rx_frames = coal_params.get('rx-frames').strip()
```
dzhengfy, line 28: I suggest we fix this in network_base.get_ethtool_coalesce() instead of here because that will avoid other referrences doing same fix as you do here in future.

PR #5570, opened 2024-04-14
virtual_disks: Add 2 cases of vhostvdpa_block

Yingshun:
This pr adds:
    VIRT-300428 - Define&Start vm with disk of vhostvdpa backend
    VIRT-300445 - Define/start vm with disk of vhostvdpa backend
        - invalid configuration

Depends on:
- https://github.com/avocado-framework/avocado-vt/pull/3893

**Test results:** The failed case is caused by an issue.
 ```
(1/5) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start.default.none: PASS (46.15 s)
 (2/5) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start.default.directsync: FAIL: Failed to check disk io for vdb! (51.62 s)

 (3/5) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start.default.no_specified_value: PASS (10.23 s)
 (4/5) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start.without_shared_memory.none: PASS (9.54 s)
 (5/5) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start.qcow2_dirver.none: PASS (10.79 s)
```

PR #5576, opened 2024-04-19
virtual_disks: Add a vhostvdpa-blk test

Yingshun:
This PR adds:
    VIRT-300441: Define&Start vm with two disks of vhostvdpa backend

Depends on:
- https://github.com/avocado-framework/avocado-vt/pull/3893
- https://github.com/autotest/tp-libvirt/pull/5570

**Test results:**
```
(1/1) type_specific.io-github-autotest-libvirt.virtual_disk.vhostvdpa.define_start_multi_disks: PASS (56.23 s)
JOB HTML   : /var/lib/avocado/job-results/job-2024-04-18T22.18-67c2b69/results.html
JOB TIME   : 58.10 s
```

### DIFF
--- a/libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
+++ b/libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
@@ -1,0 +1,12 @@
+- virtual_disks.vhostvdpa.define_start_multi_disks:
+    type = define_start_vm_with_multi_vhostvdpa_backend_disks
+    start_vm = no
+    simulator = "yes"
+    func_supported_since_libvirt_ver = (9, 10, 0)
+    only x86_64
+
+    disk_vdpa_attrs = {"source": {"attrs": {"dev": "/dev/vhost-vdpa-0"}}, "type_name": "vhostvdpa"}
+    disk2_vdpa_attrs = {"source": {"attrs": {"dev": "/dev/vhost-vdpa-1"}}, "type_name": "vhostvdpa"}
+    disk_driver = {"driver": {"name": "qemu", "type": "raw", "cache": "none", "io": "threads", "copy_on_read": "on", "discard": "unmap", "detect_zeroes": "on"}}
+    disk_attrs = {"device": "disk", "target": {"dev": "vdb", "bus": "virtio"}, **${disk_vdpa_attrs}, **${disk_driver}}
+    disk2_attrs = {"device": "disk", "target": {"dev": "vdc", "bus": "virtio"}, **${disk2_vdpa_attrs}, **${disk_driver}}
--- a/libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
+++ b/libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
@@ -1,0 +1,45 @@
+from virttest import libvirt_version
+from virttest import utils_vdpa
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_libvirt import libvirt_disk
+from virttest.utils_libvirt import libvirt_vmxml
+
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
+
+
+def run(test, params, env):
+    """
+    Verify that a vm can have more than one vhost-vdpa backend disks,
+    and both the two disks can be successfully written to/read from.
+    """
+    libvirt_version.is_libvirt_feature_supported(params)
+    disk_attrs = eval(params.get("disk_attrs", "{}"))
+    disk2_attrs = eval(params.get("disk2_attrs", "{}"))
+
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    try:
+        test.log.info("TEST_STEP: Define a VM with vhost-vdpa disk.")
+        test_env_obj = utils_vdpa.VDPASimulatorTest(sim_dev_module="vdpa_sim_blk", mgmtdev="vdpasim_blk")
+        test_env_obj.setup(dev_num=2)
+        vm_xml.VMXML.set_memoryBacking_tag(vm_name, access_mode="shared", hpgs=False)
+        libvirt_vmxml.modify_vm_device(vm_xml.VMXML.new_from_dumpxml(vm_name), "disk", disk_attrs, 1)
+        libvirt_vmxml.modify_vm_device(vm_xml.VMXML.new_from_dumpxml(vm_name), "disk", disk2_attrs, 2)
+
+        vm.start()
+        vm_session = vm.wait_for_login()
+
+        test.log.info("TEST_STEP: Check r/w operations on vhost-vdpa disk.")
+        new_disks = libvirt_disk.get_non_root_disk_names(vm_session)
+        vm_session.close()
+        for disk in new_disks:
+            if not libvirt_disk.check_virtual_disk_io(vm, disk[0]):
+                test.fail("Failed to check disk io for %s!" % disk[0])
+
+    finally:
+        bkxml.sync()
+        test_env_obj.cleanup()


dzhengfy (approved):
LGTM

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
```diff
@@ -0,0 +1,12 @@
+- virtual_disks.vhostvdpa.define_start_multi_disks:
```
chunfuwen, line 1: .virtual_disks.vhostvdpa.define_start_multi_disks -->virtual_disks.vhostvdpa.start_multi_disks
since start means you need define  it firstly

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
```diff
@@ -0,0 +1,12 @@
...
+    type = define_start_vm_with_multi_vhostvdpa_backend_disks
```
chunfuwen, line 2: maybe can be simplified as below without losing meaning, since all disks are backend device.
-->start_vm_with_multi_vhostvdpa_disks

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
Yingshun: Thank you for your suggestion. I still want to be as consistent as possible with the polarion case's title and the dir structure, so I prefer to keep the current name.

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
Yingshun: Same as above.

chloerh (approved):

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
smitterl: We could update the Polarion test title instead.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
```diff
@@ -0,0 +1,45 @@
...
+    disk_attrs = eval(params.get("disk_attrs", "{}"))
+    disk2_attrs = eval(params.get("disk2_attrs", "{}"))
+
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    try:
```
smitterl, line 25: Personally, I'd prefer you avoided logs and use smaller functions with descriptive names which I believe contributes to its readability and hence maintainability:

```python
    try:
        define_vm_with_vhost_vdpa()

        vm.start()
        vm_session = vm.wait_for_login()

        check_read_write_for_new_disks()
```

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
```diff
@@ -0,0 +1,45 @@
...
+        libvirt_vmxml.modify_vm_device(vm_xml.VMXML.new_from_dumpxml(vm_name), "disk", disk2_attrs, 2)
+
+        vm.start()
+        vm_session = vm.wait_for_login()
+
+        test.log.info("TEST_STEP: Check r/w operations on vhost-vdpa disk.")
+        new_disks = libvirt_disk.get_non_root_disk_names(vm_session)
+        vm_session.close()
+        for disk in new_disks:
+            if not libvirt_disk.check_virtual_disk_io(vm, disk[0]):
```
smitterl, line 40: Regarding this function, @chunfuwen @Yingshun shouldn't we also call 'sync' to make sure changes are written to the disk? https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_libvirt/libvirt_disk.py#L792 and afterwards read from disk `cat test/testfile`?

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
```diff
@@ -0,0 +1,45 @@
...
+from virttest.utils_libvirt import libvirt_disk
+from virttest.utils_libvirt import libvirt_vmxml
+
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
+
+
+def run(test, params, env):
+    """
+    Verify that a vm can have more than one vhost-vdpa backend disks,
+    and both the two disks can work well.
```
smitterl, line 13: Instead of "work well" maybe be more explicit "can be successfully written to/read from"

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
Yingshun: @fangge1212 What do you think?  Can you update the polarion case title if you agree?

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
Yingshun: Yes, I think we need 'sync' and check the content. I have only a few cases that need to check disk io, but I'm confused if we need to format and remount every time we call the function...

smitterl: If check_virtual_disk_io is run more than once in the same test case it might be enough to format and mount only once. We might want to add an additional parameter "reuse_fs=False" but set it True the second time we use it. OTOH, I can imagine there are not many cases where we need this but I don't know. I mean most tests that I'm aware of reboot/restart the guest, in those cases the mount needs to be executed. Therefore I think it's fine if we leave the format-and-mount part as-is.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
Yingshun: Updated, thx!

libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.cfg
Yingshun: Discussed it with feature owner and understood the reason. She gave the case such a name because she wanted to keep it consistent with the name style of another use case and also wanted to get the meaning or operations performed by reading the name of the case.  I think it's reasonable so still want to keep the current way. If there is no objection, let's move on.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
Yingshun: Thanks for sharing! As you mentioned offline, maybe we can have a new way for logging to avoid writing them into every scripts later.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/define_start_vm_with_multi_vhostvdpa_backend_disks.py
Yingshun: Ok.  I think there will be no change even we decide to call 'sync' or not in this pr. so I think it's better to have another mail thread for libvirt_disk.check_virtual_disk_io() because I think this question has not been solved for almost 2 week. what do you think?

smitterl (approved):
Thank you

chunfuwen (approved):
lgtm

PR #5587, opened 2024-04-24
virtual_disks: Add a vhostvdpa disk nodedev test

Yingshun:
This PR adds:
    VIRT-300446 - Test nodedev related operations against vhostvdpa disk

**Depends on:**
- https://github.com/avocado-framework/avocado-vt/pull/3903
- https://github.com/avocado-framework/avocado-vt/pull/3904

**Test results:**
` (1/1) type_specific.io-github-autotest-libvirt.virtual_disks.vhostvdpa.nodedev: PASS (9.00 s)
`

### DIFF
--- a/libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
+++ b/libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
@@ -1,0 +1,50 @@
+from virttest import libvirt_version
+from virttest import utils_vdpa
+from virttest import virsh
+from virttest.libvirt_xml import nodedev_xml
+
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
+
+
+def run(test, params, env):
+    """
+    Verify nodedev related operations against vhost-vdpa disk.
+    """
+    def check_nodedev_xml(dev_name):
+        """
+        Check nodedev xml for vhost-vdpa device.
+
+        :param dev_name: Device name
+        """
+        test.log.info(f"TEST_STEP: Check xml for node device - {dev_name}.")
+        nodexml = nodedev_xml.NodedevXML.new_from_dumpxml(dev_name)
+        nodedev_attrs = nodexml.fetch_attrs()
+        idx = nodedev_attrs["path"][-1]
+        exp_attrs = {"driver_name": "vhost_vdpa", "cap_type": "vdpa",
+                     "cap": {"chardev": f"/dev/vhost-vdpa-{idx}"}}
+        test.log.debug(f"Exptected: {exp_attrs}")
+        test.log.debug(f"Actual: {nodedev_attrs}")
+        for k, v in exp_attrs.items():
+            if nodedev_attrs.get(k) != v:
+                test.fail("Failed to check node device xml!")
+
+    libvirt_version.is_libvirt_feature_supported(params)
+
+    try:
+        test.log.info("TEST_STEP: Prepare vhost-vdpa disks.")
+        test_env_obj = utils_vdpa.VDPASimulatorTest(
+            sim_dev_module="vdpa_sim_blk", mgmtdev="vdpasim_blk")
+        test_env_obj.setup(dev_num=2)
+
+        test.log.info("TEST_STEP: List node devices with vdpa capability on host.")
+        vdpa_devices = [d for d in test_env_obj.get_vdpa_dev_info()['dev']]
+        devs = virsh.nodedev_list(
+            cap="vdpa", **VIRSH_ARGS).stdout_text.strip().splitlines()
+        if devs != ["vdpa_" + d for d in vdpa_devices]:
+            test.fail("Failed to get vdpa devices!")
+
+        for dev_name in devs:
+            check_nodedev_xml(dev_name)
+
+    finally:
+        test_env_obj.cleanup()
--- a/libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.cfg
+++ b/libvirt/tests/cfg/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.cfg
@@ -1,0 +1,6 @@
+- virtual_disks.vhostvdpa.nodedev:
+    type = nodedev_vhostvdpa_disk
+    start_vm = no
+    simulator = "yes"
+    func_supported_since_libvirt_ver = (9, 10, 0)
+    only x86_64


libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    libvirt_version.is_libvirt_feature_supported(params)
+
+    try:
+        test.log.info("TEST_STEP: Prepare vhost-vdpa disks.")
+        test_env_obj = utils_vdpa.VDPASimulatorTest(
+            sim_dev_module="vdpa_sim_blk", mgmtdev="vdpasim_blk")
+        test_env_obj.setup(dev_num=2)
+
+        test.log.info("TEST_STEP: List node devices with vdpa capability on host.")
```
chunfuwen, line 21: line 22 to line 28 , it looks like it is virsh command operation part, it can be wrapped into one specific method

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
```diff
@@ -0,0 +1,42 @@
...
+        vdpa_devices = [d for d in test_env_obj.get_vdpa_dev_info()['dev']]
+        devs = virsh.nodedev_list(
+            cap="vdpa", **VIRSH_ARGS).stdout_text.strip().splitlines()
+        if devs != ["vdpa_" + d for d in vdpa_devices]:
+            test.fail("Failed to get vdpa devices!")
+
+        for dev_name in devs:
```
chunfuwen, line 28: line 29 to line 40 , looks like it is to validate previous virsh operation, it can be wrapped into independent method

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
Yingshun: Yes, it might be. But this is to check the output of the virsh nodedev-list, I don't see any special need to wrap up virsh.nodedev_list() in my case.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
Yingshun: It's about node device's xml checking. I don't quite understand your meaning of wrapping this part. Is that mean we need a common way to compare the xml? If yes, I remember we had a discussion - there's too many differences in different devices, so it's not easy to have that kind of common function at time. Plz let me know your thought, thx!

nanli1 (approved):

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
smitterl: > line 22 to line 28 , it looks like it is virsh command operation part, it can be wrapped into one specific method

+1 

It improves the readability of the code. Clean code principles recommend to use functions with descriptive names instead of comments (this can help comments get out of sync with code) and functions should strive to maintain the same abstraction level. It can help focus on what's important in each function. The code will be easier to maintain. For run:

```python
try:
     test_env_obj = prepare_vhost_vdpa_disks()
     devs = get_vdpa_host_devices(test_env_obj)
     for dev_name in devs:
          check_nodedev_xml(dev_name)
finally:
    test_env_obj.cleanup()
```

Just by rewriting it in this way I noticed that your code actually doesn't exactly do what the comment promises, or I might be misunderstanding:

```python
def run(test, params, env):
    """
    Verify nodedev related operations against vhost-vdpa disk.
    """
```
IMO, you more accurately:
```python
def run(test, params, env):
    """
    Verify nodedev attributes for vhost-vdpa disk
    """
```
There are no operations against vhost_vdpa.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
smitterl: > line 29 to line 40 , looks like it is to validate previous virsh operation, it can be wrapped into independent method

+1, see comment above

smitterl (changes requested):

smitterl:
Thanks @Yingshun I agree with @chunfuwen that the code could be made more readable. Please, check my comment, do I understand the docstring for the 'run' method correctly?

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
Yingshun: The description of the polarion case is "This case is to verify that nodedev related operations against vhostvdpa disk can succeed.", so I think it will do some operations on the vhost-vdpa disk like virsh nodedev-list and nodedev-dumpxml to make sure all these operations work well, and I think it's not a test only for nodedev attributes, so I'd like to keep the current way. what do you think?

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
Yingshun: updated.

libvirt/tests/src/virtual_disks/vhostvdpa_block_backend_type/nodedev_vhostvdpa_disk.py
smitterl: @Yingshun I'm sorry but I don't agree. I might be missing something. Where are those operations? If there are none, the comment shouldn't say there should be any. Also, please consider that these automated test cases are upstream. Hopefully there will be other people using them and they don't have access to the test management system. As mentioned on another PR, maybe the right thing is to update the documented test cases. If that test case mentions operations but those operations are not run during the test case, then from my POV that's wrong.

smitterl: Clarified misunderstanding offline.

smitterl (approved):
LGTM, thank you!

chunfuwen (approved):
lgtm

PR #5603, opened 2024-05-06
v2v: fix the issue "failed to install virtio drivers on rhel8.10"

vwu-vera:
Add the "skip checking virtio drivers for rhel8.10 issue". Please check #RHEL-30528 #RHEL-17685.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -37,7 +37,8 @@ FEATURE_SUPPORT = {
     'virtio_rng': '2.6.26',
     'cache_none': 'virt-v2v-1.42.0-4',
     'q35': 'virt-v2v-1.43.3-2',
-    'virtio_model': 'virt-v2v-1.45.97-4'}
+    'virtio_model': 'virt-v2v-1.45.97-4',
+    'virtio_skip': 'virt-v2v-2.0.0-1'}
 # bz#1961107
 V2V_ADAPTE_SPICE_REMOVAL_VER = "[virt-v2v-1.45.92,)"
 V2V_VSOCK_SUPPORT_LINUX_VER = "[virt-v2v-2.0.2-1,)"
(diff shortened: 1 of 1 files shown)


vwu-vera:
(1/1) type_specific.io-github-autotest-libvirt.convert_vm_to_libvirt.esx.vm.8_0.windows.win2022.boot_uefi.arch_x86_64.it_vddk: PASS (485.90 s)

checking log:
[stdlog] 2024-05-06 04:57:50,384 v2v_vmcheck_help L0789 INFO | Skip Checking VirtIO drivers: Unsupported if v2v < virt-v2v-2.0.0-1

vwu-vera:
@xiaodwan please help to review. Thanks.:)

provider/v2v_vmcheck_helper.py
```diff
@@ -783,6 +784,11 @@ def check_windows_vm(self):
             self.log_err(err_msg)
 
         # Check Red Hat VirtIO drivers and display adapter
+        if not compare_version(FEATURE_SUPPORT['virtio_skip']):
+            reason = "Unsupported if v2v < %s" % FEATURE_SUPPORT['virtio_skip']
```
xiaodwan, line 788: This also skips the check for 'Red Hat QXL'.  Does the check be skipped, too?

vwu-vera: I noticed in the comment below:
        # bz1997446, bz2012658
        # Those guests are incorrectly inspected to other versions, so
        # the QXL drivers are installed unexpectedly. This only happens
        # in RHEL8 and the issue is not going to be fixed. so add a
        # workaround to pass the checking.

So I think the QXL checking can also be skipped for rhel8.  Right?

xiaodwan: ok

xiaodwan (approved):

PR #5664, opened 2024-06-03
fix virsh domstats block.allocation not changing issue during blockco…

nanli1:
…mmit

   add utils_misc.wait for function

```
avocado run --vt-type libvirt --test-runner=runner --vt-machine-type q35 backingchain.blockcommit.allocation_watermark
 (1/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.file_disk.inactive_layer: PASS (82.96 s)
 (2/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.file_disk.active_layer: PASS (82.68 s)
 (3/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.block_disk.inactive_layer: PASS (79.07 s)
 (4/4) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.block_disk.active_layer: PASS (81.31 s)


```

### DIFF
--- a/libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
+++ b/libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
@@ -1,8 +1,7 @@
-import time
+import re
 
 from avocado.utils import lv_utils
 
-from virttest import data_dir
 from virttest import utils_disk
 from virttest import utils_misc
 from virttest import virsh
@@ -13,6 +12,8 @@ from virttest.utils_test import libvirt
 
 from provider.backingchain import blockcommand_base
 from provider.virtual_disk import disk_base
+
+virsh_debug = {'debug': True, 'ignore_status': False}
 
 
 def run(test, params, env):
@@ -29,7 +30,7 @@ def run(test, params, env):
 
     def setup_test():
         """
-        Set disk
+        Set disk.
         """
         test_obj.new_image_path = disk_obj.add_vm_disk(
             disk_type, disk_dict)
@@ -49,7 +50,6 @@ def run(test, params, env):
                                       snap_num=1, extra=extra_option,
                                       snap_path=path_list[index],
                                       clean_snap_file=False)
-
             alloc_new_dict_1 = get_alloc_result()
             write_file(format_disk, "/mnt/file_%s" % index)
             format_disk = False
@@ -63,42 +63,34 @@ def run(test, params, env):
                                                            test_obj.new_dev)
         test.log.debug("After creating snaps, The xml is:\n%s", vmxml)
 
-        # Manual case requires to sleep 5 seconds and check the domstats again
-        time.sleep(5)
-        test.log.info("TEST_STEP2: Check %s keep changing "
-                      "during blockcommit.", disappear_alloc)
-        cmd = "blockcommit %s %s %s " % (vm_name, target_disk,
-                                         commit_option)
+        test.log.info("TEST_STEP2: Check %s keep changing during blockcommit.", changing_alloc)
+        alloc_value = get_alloc_result().get(changing_index)
+        cmd = "blockcommit %s %s %s " % (vm_name, target_disk, commit_option)
         virsh_session = virsh.VirshSession(virsh_exec=virsh.VIRSH_EXEC,
                                            auto_close=True)
         virsh_session.sendline(cmd)
         test.log.debug("blockcommit cmd: %s", cmd)
 
-        alloc_dict = {}
-        for i in range(0, 2):
-            alloc_new_dict = get_alloc_result()
-            test.log.debug('Current alloc dict is:%s, last alloc dict is :%s',
-                           alloc_new_dict, alloc_dict)
-            if alloc_new_dict.get(disappear_index) == \
-                    alloc_dict.get(disappear_index, ""):
-                test.fail("%s should keep changing during blockcommit %s." % (
-                    disappear_alloc, virsh_session.get_stripped_output()))
-            alloc_dict = alloc_new_dict
-            time.sleep(1)
+        def _alloc_changed():
+            new_alloc = get_alloc_result()
+            test.log.debug("Commit result:%s",  virsh_session.get_stripped_output())
+            test.log.debug("Currently, %s changes from %s to %s" % (
+                    changing_alloc, alloc_value, new_alloc[changing_index]))
+            return new_alloc[changing_index] != alloc_value
+
+        if not utils_misc.wait_for(_alloc_changed, timeout=120, step=0.1):
+            test.fail("%s is not changing during blockcommit in 120s." % changing_alloc)
+        test.log.debug("Checked the %s changing successfully" % changing_alloc)
 
         test.log.info("TEST_STEP3: Check %s should disappear" % disappear_alloc)
-        if utils_misc.wait_for(lambda: commit_success_msg in
-                               virsh_session.get_stripped_output(), 40, step=0.01):
-            alloc_new_dict = get_alloc_result()
-            if disappear_alloc in alloc_new_dict:
-                test.fail("The % should not be in %s" % (disappear_alloc,
-                                                         alloc_new_dict))
-        else:
-            test.log.debug("Current commit result:%s ",
-                           virsh_session.get_stripped_output())
-            test.fail("Blockcommit still not finish in 40s, "
-                      "Please control commit speed to test :%s",
-                      (virsh_session.get_stripped_output()))
+        if not utils_misc.wait_for(
+                lambda: commit_success_msg in virsh_session.get_stripped_output(),
+                30, step=0.01):
+            virsh.blockjob(vm_name, target_disk, options=" --abort", **virsh_debug)
+        alloc_new_dict = get_alloc_result()
+        if disappear_alloc in alloc_new_dict:
+            test.fail("The % should not be in %s" % (disappear_alloc,
+                                                     alloc_new_dict))
 
     def teardown_test():
         """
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
```diff
@@ -63,28 +63,23 @@ def run_test():
...
-            alloc_new_dict = get_alloc_result()
-            test.log.debug('Current alloc dict is:%s, last alloc dict is :%s',
-                           alloc_new_dict, alloc_dict)
-            if alloc_new_dict.get(disappear_index) == \
-                    alloc_dict.get(disappear_index, ""):
-                test.fail("%s should keep changing during blockcommit %s." % (
-                    disappear_alloc, virsh_session.get_stripped_output()))
-            alloc_dict = alloc_new_dict
-            time.sleep(1)
+        def _compare_alloc():
```
smitterl, line 76: maybe call this better `_alloc_changed` to be clearer about the expected comparison and boolean return value

smitterl (changes requested):
I used this PR but the test still fails on my machine

```
JOB ID     : 6a88d361f2f1b8b9625ade6ea922cdbefa6fef6d
JOB LOG    : /var/lib/avocado/job-results/job-2024-06-05T10.17-6a88d36/job.log
 (1/1) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.block_disk.active_layer: FAIL: block.4.allocation is not changing during blockcommit in 15s. (56.96 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /var/lib/avocado/job-results/job-2024-06-05T10.17-6a88d36/results.html
JOB TIME   : 57.25 s
```

libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
nanli1: Thanks Sebas, Not use dict.get() now,

nanli1:
> I used this PR but the test still fails on my machine
> 
> ```
> JOB ID     : 6a88d361f2f1b8b9625ade6ea922cdbefa6fef6d
> JOB LOG    : /var/lib/avocado/job-results/job-2024-06-05T10.17-6a88d36/job.log
>  (1/1) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.allocation_watermark.block_disk.active_layer: FAIL: block.4.allocation is not changing during blockcommit in 15s. (56.96 s)
> RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
> JOB HTML   : /var/lib/avocado/job-results/job-2024-06-05T10.17-6a88d36/results.html
> JOB TIME   : 57.25 s
> ```

Updated this pr, and the file disk scenarios success all the time , block disk scenarios failed all the time now. Seeking meina's help to check if it's a bug.

nanli1:
@smitterl Hi Sebas ,Thanks your comment ,I updated this pr and it has been running on x86 for 10 times and PASS, Please check on 390 ,Thanks.

libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
smitterl: I don't understand. What do you want to say with "Not use dict.get() now"?

What I mean is that the code currently is read in the following way:

```
 if not utils_misc.wait_for(_compare_alloc, timeout=120, step=0.1):
```

So, without further context it can be understood as "couldn't compare allocation".

The code would be more readable IMO if the line said

```
 if not utils_misc.wait_for(_alloc_changed, timeout=120, step=0.1):
```

Which is more easily understood as "if not _alloc_changed" = "if allocation didn't change".

smitterl:
Even with this patch the test fails.

AFAICT,

# The test has been checking `Got alloc dict : {'0': '1745944576', '1': '0', '2': '0', '3': '0', '4': '0'}` 
# the blockcommit finishes `virsh # Block commit: [100.00 %]`
# The test now tries to check again with `changing_index` value `4` which leads to ```
2024-06-12 05:15:02,226 test             L0941 ERROR|   File "/var/log/avocado/data/avocado-vt/virttest/test-providers.d/downloads/io-github-autotest-libvirt/libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py", line 77, in _compare_alloc
    changing_alloc, alloc_value, new_alloc[changing_index]))

2024-06-12 05:15:02,226 test             L0941 ERROR| KeyError: '4'
```

File: libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py [hunk 2]
```diff
@@ -142,8 +133,8 @@ def write_file(format_disk, file_name):
         if format_disk:
             cmd = "mkfs.ext4 /dev/%s;mount /dev/%s /mnt" % (target_disk, target_disk)
             session.cmd_status_output(cmd)
-        utils_disk.dd_data_to_vm_disk(session, file_name, bs='1M',
-                                      count='50')
+        utils_disk.dd_data_to_vm_disk(session, file_name, bs='2M',
+                                      count='100')
```
smitterl, line 137: Given that the size of the written data might need to be increased, please move this to the cfg so we can update it more easily and don't need to touch the code.

smitterl:
> Even with this patch the test fails.
> 
> ```
> backingchain.blockcommit.allocation_watermark.block_disk.active_layer: ERROR: '4' (44.46 s)
> ```
> 
> AFAICT,
> 
> 1. the test has been checking `Got alloc dict : {'0': '1745944576', '1': '0', '2': '0', '3': '0', '4': '0'}`
> 2. the blockcommit finishes `virsh # Block commit: [100.00 %]`
> 3. the test now tries to check again with `changing_index` value `4` which leads to
> 
> ```
> 2024-06-12 05:15:02,226 test             L0941 ERROR|   File "/var/log/avocado/data/avocado-vt/virttest/test-providers.d/downloads/io-github-autotest-libvirt/libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py", line 77, in _compare_alloc
>     changing_alloc, alloc_value, new_alloc[changing_index]))
> 
> 2024-06-12 05:15:02,226 test             L0941 ERROR| KeyError: '4'
> ```

I now got all tests passing in the following way:

1. Write 400MB instead of 200MB: I think the actual size (bs, count) should become configurable, s. comment
2. I had to  manually enable 'virt_use_nfs', that is 'setsebool virt_use_nfs=on'. Isn't the test code doing this automatically? I have seen other tests doing this automatically during test setup with the SELinuxBoolean object in avocado-vt/virttest/test_setup/storage.py, s. https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/cfg/svirt/shared_storage/svirt_nfs_disk.cfg#L6

smitterl (changes requested):
Hi @nanli1 please check my comments. I think we should address two more points:
1. Make sure nfs is allowed for by selinux
2. Make the data size that's written configurable by moving it to the cfg file.

libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
nanli1: Thanks your patience , Sebas , Updated

libvirt/tests/src/backingchain/blockcommit/check_allocation_watermark_during_blockcommit.py
nanli1: Sorry ,Not understand correctly, Updated now :D

nanli1:
> Hi @nanli1 please check my comments. I think we should address two more points:
> 
> 1. Make sure nfs is allowed for by selinux
> 2. Make the data size that's written configurable by moving it to the cfg file.

Updated as your comment, Please review :D

nanli1:
@smitterl Enable nfs by  the below code, Please review again : D
```
     if params.get("virt_use_nfs"):
            selinux_bool = utils_misc.SELinuxBoolean(params)
            selinux_bool.setup_local()
```

smitterl:
> @smitterl Enable nfs by the below code, Please review again : D
> 
> ```
>      if params.get("virt_use_nfs"):
>             selinux_bool = utils_misc.SELinuxBoolean(params)
>             selinux_bool.setup_local()
> ```

不好意思 I might have misunderstood, you are right, this shouldn't be necessary. Just defining the variables

```
    storage_type = nfs
    local_boolean_varible = 'virt_use_nfs'
    setup_local_nfs = "yes"
    nfs_mount_src = "/var/lib/avocado/data/avocado-vt/images"
    nfs_mount_dir = "/var/lib/libvirt/nfs_dir"
    nfs_mount_options = "rw"
    export_ip = "*"
    export_dir = "/var/lib/avocado/data/avocado-vt/images"
    local_boolean_value = "on"
```
 should suffice I assume. I didn't see those in my last review but I think they are there now. They should cause the avocado-vt/virttest/test_setup.py to do the rest.

Please remove these lines again and I'll rerun on my system. Thank you

nanli1:
@smitterl Hi Sebas, Updated, Please help re-review :D

smitterl (approved):
works well now, thank you very much

dzhengfy (approved):
LGTM

PR #5666, opened 2024-06-03
Fix set memory deviation issue

chunfuwen:
In some nested virtualization such as s390x, it needs
more time to settle guest memory


chunfuwen:
virsh.setmem.valid_options.running.half_mem.domid.dom_arg_size_arg.cmd_flag_current.manipulate_action.none.no_action  failed

smitterl:
I could reproduce the issue this on an LPAR, not nested. However, when testing manually, the memory inside the guest is never updated, even after a couple of minutes.

smitterl (changes requested):
This doesn't seem to be working on my environment.

chunfuwen:
close it and it looks like it can not resolve the issue

PR #5681, opened 2024-06-10
snapshot_with_genid: not supported on s390x

smitterl:
&lt;genid&gt; is not supported on s390x. (s390x has no ACPI and is Big-Endian, no BIOS, no UEFI.) Disable the test.


smitterl:
@nanli1 please help review. thanks

nanli1 (approved):

chunfuwen (approved):
lgtm

PR #5702, opened 2024-06-18
migration: Fix stress package issue

cliping:
There is no stress package in the latest repo. So we need to compile the stress package as a workaround.

### DIFF
--- a/libvirt/tests/src/migration/live_migration.py
+++ b/libvirt/tests/src/migration/live_migration.py
@@ -315,6 +315,9 @@ def run(test, params, env):
                                                            test, params)
 
         if stress_package:
+            if libvirt_version.version_compare(10, 4, 0):
+                params.update({"stress_install_from_repo": "no"})
+                params.update({"stress_dependency_packages_list": "['gcc', 'make']"})
             migration_test.run_stress_in_vm(vm, params)
         mode = 'both' if '--postcopy' in postcopy_options else 'precopy'
         if migrate_speed:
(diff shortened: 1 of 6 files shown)


cliping:
Before:
TestError: Installation failed with output: checking for a BSD-compatible install... /usr/bin/install -cchecking whether build environment is sane... yeschecking for a thread-safe mkdir -p... /usr/bin/mkdir -pchecking for gawk... gawkchecking whether make sets $(MAKE)... nochecking for gcc... nochecking for cc... nochecking for cl.exe... noconfigure: error: in `/home/stress-1.0.4':configure: error: no acceptable C compiler found in $PATHSee `config.log' for more details.

After:
 (1/1) type_specific.io-github-autotest-libvirt.migration.migration_uri.network_data_transport.tcp.migration_port_reuse.abort_migration.p2p: PASS (312.90 s)

libvirt/tests/src/migration/live_migration.py
```diff
@@ -315,6 +315,9 @@ def run(test, params, env):
                                                            test, params)
 
         if stress_package:
+            if libvirt_version.version_compare(10, 4, 0):
+                params.update({"stress_install_from_repo": "no"})
+                params.update({"stress_dependency_packages_list": "['gcc', 'make']"})
```
Yingshun, line 320: In the code, I can see it'll try to install the rpm package firtst, if it fails, go to the steps to compile the source tarball. So I think we should keep 'stress_install_from_repo' to yes and update the dependency packages.
The package "make' and "gcc" are basic ones, should we set them as a default value?

https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_test/__init__.py#L2414C9-L2438C1

Yingshun (changes requested):

libvirt/tests/src/migration/live_migration.py
cliping: Since the issue of missing stress package has been solved in ci, so close this PR. Thanks.

PR #5713, opened 2024-06-21
migration: Add case to verify panic device

cliping:
XXXX-301745 - Trigger guest os crash with coredump-restart on_crash event after migartion


cliping:
Test result:

 (1/1) type_specific.io-github-autotest-libvirt.migrate.panic_device: PASS (193.70 s)

libvirt/tests/src/migration/migrate_with_panic_device.py
```diff
@@ -0,0 +1,100 @@
...
+
+        vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+        if not vmxml.xmltreefile.find('devices').findall('panic'):
+            # Set panic device
+            panic_dev = Panic()
+            panic_dev.model = panic_model
+            panic_dev.addr_type = addr_type
+            panic_dev.addr_iobase = addr_iobase
+            vmxml.add_device(panic_dev)
+        vmxml.on_crash = "coredump-restart"
```
Yingshun, line 45: I think it's better to move it to cfg to show that we'll cover "coredump-restart" in this case, what do you think?

libvirt/tests/src/migration/migrate_with_panic_device.py
```diff
@@ -0,0 +1,100 @@
...
+        dump_file_path = params.get("dump_file_path")
+
+        test.log.info("Setup steps.")
+        migration_obj.setup_connection()
+        # Cleanup dump file
+        cmd = "rm -f %s/*" % dump_file_path
+        remote.run_remote_cmd(cmd, params)
+
+        vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+        if not vmxml.xmltreefile.find('devices').findall('panic'):
```
dzhengfy, line 38: Is it simple to reuse libvirt_vmxml.modify_vm_device() here?

dzhengfy (commented):
Others LGTM

libvirt/tests/src/migration/migrate_with_panic_device.py
cliping: Updated, thanks.

libvirt/tests/src/migration/migrate_with_panic_device.py
cliping: Update, thanks.

Yingshun (approved):

libvirt/tests/src/migration/migrate_with_panic_device.py
```diff
@@ -0,0 +1,96 @@
...
+            raise
+
+        remote_vm_session.close()
+        ret = virsh.domstate(vm_name, extra="--reason", uri=dest_uri, debug=True)
+        libvirt.check_result(ret, expected_match=expected_match)
+
+        # Check dump file
+        cmd = "ls %s" % dump_file_path
+        ret = remote.run_remote_cmd(cmd, params, ignore_status=False)
+        if vm_name not in ret.stdout_text.strip():
```
dzhengfy, line 81: This checking is inaccurate.
We need to take the timestamp into the consideration to make sure the coredump file is created during this case.

cliping: The coredump files in the directory will be cleared before migration. If a coredump file exists, it should be generated by this case.

dzhengfy (approved):

PR #5715, opened 2024-06-24
Add case of cgroup vhost

chloerh:
- VIRT-300725 - [cgroup][vhost] Check the vhost* thread's cpu affinity with emulatorpin setting

Test result:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_network.check_vhost_cpu_affinity_with_emulatorpin

### DIFF
--- a/libvirt/tests/cfg/virtual_network/driver/check_vhost_cpu_affinity_with_emulatorpin.cfg
+++ b/libvirt/tests/cfg/virtual_network/driver/check_vhost_cpu_affinity_with_emulatorpin.cfg
@@ -1,0 +1,7 @@
+- virtual_network.check_vhost_cpu_affinity_with_emulatorpin:
+    type = check_vhost_cpu_affinity_with_emulatorpin
+    start_vm = no
+    timeout = 240
+    vcpu_set = 0
+    vm_attrs = {'vcpu': 1, 'cputune': {'vcpupins': [{'vcpu': '0', 'cpuset': '${vcpu_set}'}], 'emulatorpin': epin_set}}
+    iface_attrs = {'model': 'virtio'}
--- a/libvirt/tests/src/virtual_network/driver/check_vhost_cpu_affinity_with_emulatorpin.py
+++ b/libvirt/tests/src/virtual_network/driver/check_vhost_cpu_affinity_with_emulatorpin.py
@@ -1,0 +1,104 @@
+import logging
+import random
+import re
+
+from avocado.core import exceptions
+from avocado.utils import process
+from virttest import cpu
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_libvirt import libvirt_misc
+from virttest.utils_libvirt import libvirt_vmxml
+
+VIRSH_ARGS = {'ignore_status': False, 'debug': True}
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def get_online_cpus():
+    """
+    Get online cpus
+
+    :return: a list of online cpus
+    """
+    cpu_range = cpu.get_cpu_info().get('On-line CPU(s) list')
+    online_cpus = [
+        (lambda x: range(x[0], x[-1] + 1))(list(map(int, r.split('-'))))
+        for r in cpu_range.split(',')
+    ]
+    online_cpus = [cpu for r in online_cpus for cpu in r]
+    LOG.debug(f'Online cpus: {online_cpus}')
+
+    return online_cpus
+
+
+def check_cpu_affinity(vm_name, pid, epin_set):
+    """
+    Check cpu affinity with virsh emulatorpin and taskset
+
+    :param vm_name: vm name
+    :param pid: pid of vhost
+    :param epin_set: pre-set emulatorpin
+    """
+    epin = libvirt_misc.convert_to_dict(
+        virsh.emulatorpin(vm_name, **VIRSH_ARGS).stdout_text,
+        r'\s+(\S+):\s+(\S+)')
+
+    if epin['*'] != epin_set:
+        raise exceptions.TestFail(f'Incorrect cpu affinity (emulatorpin): '
+                                  f'{epin["*"]}, Should be {epin_set}.')
+
+    taskset_output = process.run(
+        f'taskset -cap {pid}', shell=True).stdout_text
+    search_pattern = "pid \d+'s current affinity list:\s(\S+)"
+    search_result = re.search(search_pattern, taskset_output)
+    if not search_result or search_result.group(1) != epin_set:
+        raise exceptions.TestFail(f'Expect current affinity to be {epin_set}, '
+                                  f'not {search_result.group(1)}.')
+
+
+def run(test, params, env):
+    """
+    Check the vhost* thread's cpu affinity with emulatorpin setting
+    """
+    vm_name = params.get('main_vm')
+    vm = env.get_vm(vm_name)
+    vcpu_set = params.get('vcpu_set')
+    iface_attrs = eval(params.get('iface_attrs', '{}'))
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    bkxml = vmxml.copy()
+
+    online_cpus = get_online_cpus()
+    epin_set = str(random.choice(online_cpus))
+    online_cpus.remove(int(epin_set))
+    epin_reset = str(random.choice(online_cpus))
+    vm_attrs = eval(params.get('vm_attrs', '{}'))
+
+    try:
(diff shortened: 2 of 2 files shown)


yalzhang (approved):
It can cover the bug succesfully. Run with old kernel, fail, while run with updated kernel, pass.

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
```diff
@@ -0,0 +1,8 @@
+- virtual_network.check_vhost_cpu_affinity_with_emulatorpin:
```
dzhengfy, line 1: Currently there are many files or folders under virtual_network. 
I think we'd better make an organized structure. 
@chloerh @yalzhang How do you think to put this new file?

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
```diff
@@ -0,0 +1,8 @@
...
+    type = check_vhost_cpu_affinity_with_emulatorpin
+    start_vm = no
+    timeout = 240
+    vcpu_set = 0
+    epin_set = 1
```
dzhengfy, line 6: This hardcoded cpu id 1 is at a little risky. Sometimes the cpu 1 might be offlined or non usable.  I suggest we dynamically choose another online cpu id.

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if epin['*'] != epin_set:
+            test.fail(f'Incorrect cpu affinity (emulatorpin): {epin["*"]},'
+                      f'Should be {epin_set}.')
+
+        pid = process.run(
+            "ps -eL|awk '/vhost/{print $2}'", shell=True).stdout_text.strip()
+
+        taskset_output = process.run(
+            f'taskset -cap {pid}', shell=True).stdout_text
```
dzhengfy, line 58: It is better to firstly check the availability of `taskset` and install it if no.

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
```diff
@@ -0,0 +1,82 @@
...
+
+def run(test, params, env):
+    """
+    Check the vhost* thread's cpu affinity with emulatorpin setting
+    """
+    vm_name = params.get('main_vm')
+    vm = env.get_vm(vm_name)
+    vcpu_set = params.get('vcpu_set')
+    epin_set = params.get('epin_set')
+    epin_reset = random.randint(int(epin_set), 10)
```
dzhengfy, line 25: This is at risk. Please choose from online cpus. @chloerh 
Do you need to consider different numa nodes？ @yalzhang

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
```diff
@@ -0,0 +1,82 @@
...
+            f'taskset -cap {pid}', shell=True).stdout_text
+        search_pattern = "pid \d+'s current affinity list:\s(\S+)"
+        search_result = re.search(search_pattern, taskset_output)
+        if not search_result or search_result.group(1) != epin_set:
+            test.fail(f'Expect current affinity to be {epin_set}, '
+                      f'not {search_result.group(1)}.')
+
+        # Update the emulator pin to other cpu,
+        # and check the vhost thread's cpu affinity
+        virsh.emulatorpin(vm_name, epin_reset, **VIRSH_ARGS)
```
dzhengfy, line 67: Line 67 - 80 seems many duplcation with line 42-58. I think your codes are able to be more beatiful if removing those duplication.

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
chloerh: @yalzhang Could you help confirm? Thanks

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
chloerh: Sure, I'm ok either way.

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
chloerh: ok

chloerh: Actually this is on host, I just need to check its existence?

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
chloerh: Sure, seems the polarion case needs an update too.

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
yalzhang: Yes, we can choose from the online cpus. The test point of this case is that the vhost thread is pinned to the same cpu as the emulator.

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
yalzhang: Maybe we can create a folder called "driver" under the "virtual network" folder. There should be several cases about dirver later like multiqueue, we can put them together here. What do you think?

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
yalzhang: No need. This command is provided by util-linux-core, which is needed by systemd. So I think it must exists on the host.
# yum remove util-linux-core
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use "rhc" or "subscription-manager" to register.

Error: 
 Problem: The operation would result in removing the following protected packages: systemd
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
yalzhang: No, no need to consider the numa nodes. The check point is that the vhost thread will always pinned to the same cpu as the emulator. So in this step, we just update the affinity of th emulator to check it the vhost thread affinity will change accordingly.

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
chloerh: ok

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
chloerh: Done

libvirt/tests/src/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.py
chloerh: Done

chloerh:
All resolved except the one of putting file to another folder, will update after confirmation.
Test result:
` (1/1) type_specific.io-github-autotest-libvirt.virtual_network.check_vhost_cpu_affinity_with_emulatorpin: PASS (7.57 s)`

libvirt/tests/cfg/virtual_network/check_vhost_cpu_affinity_with_emulatorpin.cfg
chloerh: Done

dzhengfy (approved):
LGTM

PR #5722, opened 2024-06-25
Add network case of qos

chloerh:
- VIRT-296327 - [Qos] Check the actual network throughput per bandwidth settings in network/interface

### DIFF
--- a/libvirt/tests/src/virtual_network/qos/check_actual_network_throughput.py
+++ b/libvirt/tests/src/virtual_network/qos/check_actual_network_throughput.py
@@ -1,0 +1,132 @@
+import logging
+
+from avocado.utils import process
+from virttest import utils_misc
+from virttest import utils_net
+from virttest import utils_package
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.network_xml import NetworkXML
+from virttest.staging import service
+from virttest.utils_libvirt import libvirt_network
+from virttest.utils_libvirt import libvirt_vmxml
+from virttest.utils_test import libvirt
+
+from provider.virtual_network import network_base
+
+VIRSH_ARGS = {'ignore_status': False, 'debug': True}
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def run(test, params, env):
+    """
+    Check the actual network throughput
+    per bandwidth settings in network/interface
+
+    """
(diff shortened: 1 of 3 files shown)


libvirt/tests/src/virtual_network/qos/check_actual_network_throughput.py
```diff
@@ -0,0 +1,132 @@
...
+        if not utils_net.check_class_rules(*tc_args):
+            test.fail('Class rule check failed. Please check log.')
+        LOG.debug('Class rule check: PASS')
+
+        if not utils_net.check_filter_rules(tc_args[0], filter_bw):
+            test.fail('Filter rule check failed. Please check log.')
+        LOG.debug('Filter rule check: PASS')
+
+        if not utils_package.package_install('netperf', session=vm_sess):
+            test.error('Failed to install netperf on VM.')
```
Yingshun, line 92: If it fails, can we compile via virttest.utils_netperf.NetperfPackage().install()?

provider/virtual_network/network_base.py
```diff
@@ -336,3 +337,29 @@ def check_iface_attrs(iface, key, expect_val):
...
+    :param cli_runner: runner of netperf client
+    :param ip_addr: ip address
+    :param bw: bandwidth setting
+    :param th_type: inbound or outbound
+    """
+
+    serv_runner('netserver')
+    netperf_out = cli_runner(f'netperf -H {ip_addr}')
+    LOG.debug(netperf_out)
+    serv_runner('pkill netserver')
```
Yingshun, line 356: reuse utils_netperf.NetperfServer() and utils_netperf.NetperfClient()?

libvirt/tests/src/virtual_network/qos/check_actual_network_throughput.py
chloerh: Seems the `install()` function needs to create a session itself, not supporting passing a session as param for now, maybe after the function being updated?

provider/virtual_network/network_base.py
chloerh: I've checked this too, still needs to support vm_session.

yalzhang (commented):
local run pass

yalzhang (approved):

Yingshun (approved):

PR #5726, opened 2024-06-26
Virstual_disk: fix shutdown event of vhostvdpa disk

meinaLi:
Fix the error of "EventNotFoundError: Not found event disconnected.*(.*)*Shutdown Finished.*(.*)*Stopped after 10 seconds". We can only check the last two events because maybe the disconnected event is not in output.


meinaLi:
Close this PR because the failure is caused by qemu-guest-agent package which is not installed in guest by CI.

PR #5738, opened 2024-06-27
network_base: add check_firewall_backend function

yanqzhan:
Add a function to judge nftables or iptables to use for test.

### DIFF
--- a/provider/virtual_network/network_base.py
+++ b/provider/virtual_network/network_base.py
@@ -4,10 +4,11 @@ import shutil
 
 import aexpect
 from avocado.core import exceptions
-from avocado.utils import process
+from avocado.utils import process, distro
 from virttest import remote
 from virttest import utils_net
 from virttest import virsh
+from virttest import libvirt_version
 from virttest.libvirt_xml import network_xml
 from virttest.libvirt_xml import vm_xml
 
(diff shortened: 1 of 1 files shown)


yanqzhan:
Tested in https://github.com/autotest/tp-libvirt/pull/5739.

provider/virtual_network/network_base.py
```diff
@@ -336,3 +336,23 @@ def check_iface_attrs(iface, key, expect_val):
...
+    host_distro = distro.detect().name
+    host_version = distro.detect().version
+    from_rhel10 = (host_distro == 'rhel' and int(host_version) >= 10)
+    if from_rhel10:
+        firewall_backend = "nftables"
+        with open('/etc/libvirt/network.conf') as file:
+            for line in file.readlines():
+                if re.search('^firewall_backend\\s*=\\s*.iptables', line):
+                    firewall_backend = "iptables"
+                    break
```
Yingshun, line 356: Can you plz reuse utils_config in avocado-vt? If there is no proper class, can you add a one for libvirt network config?

provider/virtual_network/network_base.py
```diff
@@ -336,3 +336,23 @@ def check_iface_attrs(iface, key, expect_val):
...
+
+def check_firewall_backend():
+    """
+    Check host firewall_backend to determine to use nft or iptables cmd
+
+    """
+    firewall_backend = "iptables"
+    host_distro = distro.detect().name
+    host_version = distro.detect().version
+    from_rhel10 = (host_distro == 'rhel' and int(host_version) >= 10)
```
Yingshun, line 349: I wonder if there is another way to check the firewall backend or not. can we check what the default firewalld on host is?

chloerh: +1

yanqzhan: I'm not sure, after discussion with some colleagues, the methods maybe:
1.check comments in /etc/libvirt/network.conf：
#   If firewall_backend isn't configured, libvirt will choose the
#   first available backend from the following list:
#
#     [iptables, nftables]
2. check *.conf.in, which's the source file of .conf file. But I didn't find network.conf.in, and they're in source code.
3. check logs while executing corresponding actions.

provider/virtual_network/network_base.py
yanqzhan: I can do it later. Thx.

provider/virtual_network/network_base.py
```diff
@@ -336,3 +337,29 @@ def check_iface_attrs(iface, key, expect_val):
...
+        with open('/etc/libvirt/network.conf') as file:
+            for line in file.readlines():
+                if re.search('^firewall_backend\\s*=\\s*.iptables', line):
+                    firewall_backend = "iptables"
+                    break
+    elif libvirt_version.version_compare(10, 4, 0):
+        with open('/etc/libvirt/network.conf') as file:
+            for line in file.readlines():
+                if re.search('^firewall_backend\\s*=\\s*.nftables', line):
+                    LOG.debug(f'my values: {firewall_backend}')
```
dzhengfy, line 361: It seems this debugging message is not useful enough.
IF you would like to keep it, please avoid oral words , like my,  and give a meaningful sentence.

provider/virtual_network/network_base.py
dzhengfy: just execute like
```
`grep  ^#firewall_backend\\s*=\\s*.nftables /etc/libvirt/network.conf`
#firewall_backend = "nftables"
```

Could it meet your needs?

yanqzhan:
After discussion, reviewer thinks this is not a good method. And from another discussion with feature owner and auto owner, we decide not to check rules anymore but only cares about funciton in other cases. So close it.

PR #5774, opened 2024-07-11
Update the file path to "/"

yalzhang:
In this case, we need to create a temporary file to test the file transfer function. The original /tmp directroy is not enough some times since we need to creat 1G file. So update the file path to be under "/".

### DIFF
--- a/libvirt/tests/cfg/virtual_network/passt/passt_transfer_file.cfg
+++ b/libvirt/tests/cfg/virtual_network/passt/passt_transfer_file.cfg
@@ -35,7 +35,7 @@
             bs = 1M
         - 1G:
             bs = 100M
-    test_file = /tmp/${file_size}_file
+    test_file = /${file_size}_file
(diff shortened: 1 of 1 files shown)


yalzhang:
before the fix, some case will fail with:
ShellCmdError: Shell command failed: 'dd if=/dev/urandom bs=100M count=10 &gt; /tmp/1G_file'    (status: 1,    output: &quot;dd: error writing 'standard output': No space left on device\n9+0 records in\n8+0 records out\n925503488 bytes (926 MB, 883 MiB) copied, 2.82279 s, 328 MB/s\n&quot;)&#10;

after the fix, it pass
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.passt.transfer_file.ipv6.vm_to_host.1G.ip_portfw.root_user: PASS (125.75 s)

chloerh (approved):

PR #5780, opened 2024-07-15
Add passt udp transfer file

yalzhang:
Add more scenarios to test passt interface transfer file via udp.


chunfuwen:
need more results here

PR #5782, opened 2024-07-15
libvirt_hooks.py: Fix string matching

Slancaster1:
### The Issue
Since a recent libvirt commit, the hook log for the hook "release end" provides a reason for the release.
```
/etc/libvirt/hooks/qemu guest_name release end <shutoff-reason>
```
The test currently looks for a logline that does not provide a reason for release. This appears like this:
```
/etc/libvirt/hooks/qemu guest_name release end -
```
This results in the test failing if a reason is provided. E.G. the test will fail if this log line is given:
```
/etc/libvirt/hooks/qemu guest_name release end shutdown
```
### The Fix
Alter test such that it only looks for "release end shutdown"

### Evidence
Console output showing this test passing
```
(.libvirt-ci-venv-ci-runtest-IDDz4J) [root@ampere-mtsnow-altramax-48 tp-libvirt]# avocado run --vt-type libvirt --vt-machine-type arm64-mmio --test-runner=runner libvirt_hooks.vm.start_stop
No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
No python imaging library installed. PPM image conversion to JPEG disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
JOB ID     : f844c41492c07d9c4bab462b3f0d60d5b2486945
JOB LOG    : /var/log/avocado/job-results/job-2024-07-22T09.52-f844c41/job.log
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_hooks.vm.start_stop: PASS (62.05 s)
JOB HTML   : /var/log/avocado/job-results/job-2024-07-22T09.52-f844c41/results.html
JOB TIME   : 62.44 s
```

### DIFF
--- a/libvirt/tests/src/libvirt_hooks.py
+++ b/libvirt/tests/src/libvirt_hooks.py
@@ -85,6 +85,7 @@ def run(test, params, env):
         """
         Do start/stop operation and check the results.
         """
+        release_reason = params.get("release_reason", "-")
         logging.info("Try to test start/stop hooks...")
         hook_para = "%s %s" % (hook_file, vm_name)
         prepare_hook_file(hook_script %
@@ -102,7 +103,7 @@ def run(test, params, env):
             vm.destroy()
             hook_str = hook_para + " stopped end -"
             assert check_hooks(hook_str)
-            hook_str = hook_para + " release end -"
+            hook_str = hook_para + " release end " + release_reason
             assert check_hooks(hook_str)
         except AssertionError:
             utils_misc.log_last_traceback()
--- a/libvirt/tests/cfg/libvirt_hooks.cfg
+++ b/libvirt/tests/cfg/libvirt_hooks.cfg
@@ -24,6 +24,7 @@
             variants:
                 - start_stop:
                     test_start_stop = "yes"
+                    release_reason = "shutdown"
                 - libvirtd_t:
                     test_libvirtd = "yes"
                 - save_restore:


Slancaster1:
@dzhengfy Could you add yourself as a reviewer to this PR?

libvirt/tests/src/libvirt_hooks.py
```diff
@@ -102,7 +102,7 @@ def start_stop_hook():
             vm.destroy()
             hook_str = hook_para + " stopped end -"
             assert check_hooks(hook_str)
-            hook_str = hook_para + " release end -"
+            hook_str = hook_para + " release end"
```
chunfuwen, line 105: Maybe we can use wild card to avoid frequent change. e.g hook_str = hook_para + " release end.*" since test case need care about keyword only

Slancaster1: I'm unsure how a wildcard would be functionally different from this way. Omitting the '-' and using python's 'in' operator in check_hooks works the same a regex search with a wildcard would.

Both ways seem to avoid frequent change, and only care about the keywords "release end", regardless of what follows.

Furthermore, altering check_hooks to accept regexs is more invasive to the test. And I'm unsure if it would effect other test cases in this script that use the check_hooks function.

However, if you still feel a wildcard is better. I can put together this change, and push it.

dzhengfy: From my POV, both are ok.

dzhengfy, line 105: But I am thinking we need to add one more checkpoint here.
```
commit c772f1982d854eebe2131e80c4e95d19e573cc8c
Author: Swapnil Ingle <swapnil.ingle@nutanix.com>
Date:   Tue Jun 18 16:39:58 2024 +0000

    Pass shutoff reason to release hook
```
Based on above commit, the reason is expected. 
And in our script, we need vm.destroy() which uses  gracefully `shutdown_command` in the guest , so the expected reason should be "shutdown" instead of other reason. So I suggest we add the checking for the reason value. And it is good to put the expected reason in the cfg.

Slancaster1: I have made these changes

dzhengfy (approved):
LGTM

cliping (approved):

chunfuwen (approved):
lgtm

PR #5797, opened 2024-07-22
libvirt_ap_passthrough/autostart: wait for device come up

smitterl:
Previously, the device was available immediately after login. Now, it can take another second. Expect a maximum of 5seconds for it to come up.


chunfuwen (approved):
lgtm

Yingshun (approved):

PR #5807, opened 2024-07-24
v2v: fix cpu-topology case on rhel10

vwu-vera:
Please refer bz#1541908. 

* Output to RHV would create the required metadata (<rasd:num_of_sockets> etc) for topology.  But the CPU vendor
and model cannot be passed to RHV.

*Use virt-v2v to convert guest to libvirt, and check cpu info in guest xml after conversion,v2v could parse guest's cpu topology in guest xml correctly


vwu-vera:

xiaodwan (approved):

PR #5813, opened 2024-07-29
requirements-travis: fix dependencies

smitterl:
The systems can't be set up anymore for python versions
since 3.8. Update depdendencies.

There will be a later pylint version producing new errors.
Ignore these as we didn't check them before.

Update pip to latest to avoid other dependency issues.

Allow for pylint to exit with code 32 if there are no files to
check. Otherwise the build will fail.
For that, the exit-on-error must be temporarily switched off (set +e)
in the run commands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,12 +29,19 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: pip install --upgrade pip
       - name: Install dependencies
         run: |
          pip install sphinx${SPHINX_VERSION:+==$SPHINX_VERSION}
          pip install -r requirements-travis.txt
       - name: Run inspekt
-        run: inspekt checkall --disable-style E501,E265,W601,W605,E402,E722,E741 --no-license-check
+        run: inspekt checkall --disable-style E501,E265,W601,W605,E402,E722,E741 --disable-lint W,R,C,E1002,E1101,E1103,E1120,F0401,I0011,E0601,E0602,E0606 --no-license-check
       - run: echo "This job's status is ${{ job.status }}."
       - name: Run spellchecker
-        run: pylint --errors-only --disable=all --enable=spelling --spelling-dict=en_US --spelling-private-dict-file=spell.ignore *
+        run: |
+         set +e
+         pylint --errors-only --disable=all --enable=spelling --spelling-dict=en_US --spelling-private-dict-file=spell.ignore *
+         exitcode=$?
+         set -e
+         test $exitcode == 32 -o $exitcode == 0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
          pip install sphinx${SPHINX_VERSION:+==$SPHINX_VERSION}
          pip install -r requirements-travis.txt
       - name: Run inspekt
-        run: inspekt checkall --disable-style E501,E265,W601,W605,E402,E722,E741 --no-license-check
+        run: inspekt checkall --disable-style E501,E265,W601,W605,E402,E722,E741 --disable-lint W,R,C,E1002,E1101,E1103,E1120,F0401,I0011,E0606,E0601 --no-license-check
```
chunfuwen, line 37: May I ask what is  the reason behind adding this additional disable options:--disable-lint W,R,C,E1002,E1101,E1103,E1120,F0401,I0011,E0606,E0601 ?

.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
          pip install sphinx${SPHINX_VERSION:+==$SPHINX_VERSION}
```
chunfuwen, line 34: I noticed there is warning to upgrade pip, how about add :pip install --upgrade pip before this line

smitterl: I don't know. The warning is not new. In this PR I would want to only fix the current issue with 3.8 and 3.9 builds due to dependency conflicts. How about we discuss the pip install --upgrade separately, what do you think?

.github/workflows/ci.yml
smitterl: The newer pylint version does additional checks like E0606,E0601, so disable them. The others were disabled already per default.

.github/workflows/ci.yml
chunfuwen: agreed

smitterl: I've run into some more issues. I'm trying now with this, ignoring certain new pylint checks to see which tox, virtualenv and pylint versions this could work with. I hope it will pass and if so, I'll fix the component versions.

smitterl:
@chloerh @chunfuwen 
I tested the logic that allows for '32' by temporarily changing the commit to allow for '31'. As expected it would fail, so switched in back to '32' and it succeeds.

.github/workflows/ci.yml
smitterl: In the end I added this to the PR. It seems to have fixed some 'missing member' issue.

chloerh (approved):
Thanks for fixing this issue!

smitterl:
Not sure if we still need this, the more recent workflow passed: https://github.com/autotest/tp-libvirt/actions/runs/10145504991/job/28051343401?pr=5818

chunfuwen (approved):
lgtm

PR #5823, opened 2024-07-31
svirt/selinux: don't update boot disk

smitterl:
The test usually updates the attributes of an existing device. However, in case of disks this means that the boot disk is substituted. s390x behaves here differently from x86_64, for example. x86_64 would have a running VM even if it doesn't boot (the test creates an empty image to boot from) whereas on s390x the VM doesn't start. Therefore, in case of x86_64 the file is relabeled but on s390x it isn't.

Instead of updating the device, attach it.


chloerh (approved):

cliping (approved):

chunfuwen (approved):
lgtm

PR #5826, opened 2024-08-01
virtual_network:Use get_default_gateway to get host iface name

chloerh:
The original way is to get the 1st iface that is UP which is not reliable since an interface might be UP even if it doesn't have an ip address.


chloerh:
Test result:


All Tests
Class | Duration | Fail | (diff) | Skip | (diff) | Pass | (diff) | Total | (diff)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
virtual_network | 5 hr 14 min | 3 | -17 | 7 |   | 369 | 17 | 379 |

yanqzhan (approved):
LGTM.

Yingshun (approved):

PR #5829, opened 2024-08-05
vm_start_destroy_repeatedly: Refinement

Slancaster1:
Ran test 'vm_start_destroy_repeatedly' multiple times, and fixed some test-script related errors

1) Increased test_timeout to allow time for test
2) Properly handled wait timeouts for startup, and resume so that we wait for correct states


Slancaster1:
@dzhengfy Could you add yourself as a reviewer?

libvirt/tests/src/vm_start_destroy_repeatedly.py
```diff
@@ -51,10 +50,16 @@ def run(test, params, env):
...
 
     for i in range(num_cycles):
         logging.info("Starting vm '%s' -- attempt #%d", vm_name, i+1)
-        power_cycle_vm(test, vm, vm_name, login_timeout, startup_wait, resume_wait)
+
+        try:
+            power_cycle_vm(test, vm, vm_name, login_timeout, startup_wait, resume_wait)
+        except Exception as err:
+            logging.info(str(err))
+            test.fail("Power cycle threw exception")
```
dzhengfy, line 63: Usually we directly throw the exception here. Like
` test.fail("Power cycle threw exception with %s" % str(err))`

libvirt/tests/src/vm_start_destroy_repeatedly.py
```diff
@@ -51,10 +50,16 @@ def run(test, params, env):
...
+    resume_wait = float(params.get("resume_wait", 240))          # Controls wait for virsh.resume()
 
     for i in range(num_cycles):
         logging.info("Starting vm '%s' -- attempt #%d", vm_name, i+1)
-        power_cycle_vm(test, vm, vm_name, login_timeout, startup_wait, resume_wait)
+
+        try:
+            power_cycle_vm(test, vm, vm_name, login_timeout, startup_wait, resume_wait)
+        except Exception as err:
+            logging.info(str(err))
```
dzhengfy, line 62: This is not necessary.

Slancaster1: Removed it

dzhengfy (approved):

dzhengfy:
This pr was able to reproduce the target bug.

PR #5739, opened 2024-06-27
iface_network: use nft for new product

yanqzhan:
Replace an ip6tables usage with nft for new product which use the default firewall_backend='nftables'.

PR #5838, opened 2024-08-12
network: revert ip6 masquerade check by nft

yanqzhan:
The depending check_firewall_backend can not be approved. But new cases exist and not check rules anymore. Just skip but not fix.


chunfuwen:
this aims to revert previous PR:https://github.com/autotest/tp-libvirt/pull/5739, which fail s390x gating testing, I will merge it as urgent request

PR #5847, opened 2024-08-17
v2v:fix v2v-options failed case -adding version control

vwu-vera:
Confirm with developer, the patch is included upstream in virt-v2v >= 2.4.  Any version
above this doesn't need the separate patch.


vwu-vera:

xiaodwan (approved):

PR #5882, opened 2024-09-06
virtio_page_per_vq: enhance input event file

dzhengfy:
Regarding /dev/input/event* files, there might be some different situation.
For example,
- No event file exsit at all
- Only /dev/input/event0 exists
- Only /dev/input/event1 exists
- Both /dev/input/event0 and /dev/input/event1 exists
- and more ...

So this patch is intending to support above situations instead of hardcoding with event1.

### DIFF
--- a/libvirt/tests/cfg/virtio/virtio_page_per_vq.cfg
+++ b/libvirt/tests/cfg/virtio/virtio_page_per_vq.cfg
@@ -41,7 +41,7 @@
                 - mouse:
                 - tablet:
                 - passthrough:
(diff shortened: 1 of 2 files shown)


yafu-1 (approved):
LGTM

chloerh (approved):

chloerh:
Oh, conflicting

cliping (approved):

dzhengfy:
Fix conflicts, so merge it now.

PR #5916, opened 2024-09-25
svirt/selinux/per_device: return if expected error is found

chloerh:
`status_error` expected to be `yes` on invalid configuration, that is, either during xml setup

a) through `vmxml.sync` which undefines and tries to define the VM, or b) through attach through `virsh.attach_device`.

Both will raise LibvirtXMLError on failure. So, if as expected this error is raised for coldplug, we can stop the test execution by returning and considering the test PASS.


chloerh:
@smitterl Thank you for fixing the issue! 

Test result:
PASS 1-type_specific.io-github-autotest-libvirt.svirt.selinux.seclabel.per_device.relabel_yes.without_label.serial.cold_plug

Yingshun (approved):

PR #5959, opened 2024-10-24
Fix nbd disk failure due to the issue that partitions not shown up with

chunfuwen:
Fix nbd disk failure due to the issue that partitions not shown up with limited time

Use utils_misc.wait_for() to avoid hard time waiting


libvirt/tests/src/virtual_disks/virtual_disks_nbd.py
```diff
@@ -275,7 +275,7 @@ def check_in_vm(old_parts):
         if check_partitions and not status_error:
             logging.debug("wait seconds for starting in checking vm part")
             time.sleep(2)
```
cliping, line 277: When we use utils_misc.wait_for(), maybe we can delete 'time.sleep(2)'.

PR #5994, opened 2024-11-14
Prepare certain state before managedsave

yalzhang:
For the case "managedsave.status_error_no.name_option.paused_status" we need to prepare the VM to be in paused state before managedsave.


yalzhang:
@chloerh  Please help to review, Thank you!

chloerh (approved):

PR #6001, opened 2024-11-19
virtual_network: disable test cases with mixed boot order elements

smitterl:
On s390x, only per-device <boot/> is supported. Disable test expecting conflict with //os/boot


chunfuwen (approved):
lgtm

PR #6016, opened 2024-11-26
virsh-snapshot: Update logic to verify snapshot creation

misanjumn:
### virsh-snapshot: Update logic to verify snapshot creation

- check_snaplist function is used to verify the created snapshot by checking it's presence in the guest xml.
- But since there might be multiple disk name values, drivers, source files, snapshot types, etc, the verification is failing leading to false negative test result.
- This patch will allow us to check the disk values by checking if it is present in the list of xml parameter values.

Additionally, this patch also fixes the Syntax Warning: Invalid Escape Sequence, line: 76

Signed-off-by: Misbah Anjum N [misanjum@linux.vnet.ibm.com](mailto:misanjum@linux.vnet.ibm.com)


harihare (approved):
LGTM

chunfuwen (approved):
lgtm

PR #6029, opened 2024-11-28
Fix virtio_win_dir issue and change cases target

mxie91:


mxie91:
@xiaodwan pls review, thanks

xiaodwan (approved):

PR #6032, opened 2024-11-29
virsh_detach_device_alias: remove test.skip

dzhengfy:
test.skip is not used any more.


dzhengfy:
before the fix:
```
virsh.detach_device_alias.config.input.passthrough
'VirtTest' object has no attribute 'skip'
```

cliping (approved):

chloerh (approved):

PR #6064, opened 2024-12-12
iface_network/dhcp_range: expect second guest not to get IP

smitterl:
In 1e88841514 `utils_net.get_guest_ip_addr` was replaced by `network_base.get_vm_ip`. However, the first function returned `None` when IP wasn't available whereas the second function raises a test error. Update the code to handle the test error in the same way it would have handle a return value `None`.

In this specific case, when dhcp_range is tested, the range is very small so the second VM is expected not to get an ip.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_network.py
+++ b/libvirt/tests/src/virtual_network/iface_network.py
@@ -6,6 +6,7 @@ import logging as log
 import platform
 import shutil
 
+from avocado.core.exceptions import TestError
 from avocado.utils import process
 from avocado.utils import stacktrace
 
@@ -1107,10 +1108,14 @@ TIMEOUT 3"""
                         vms_mac = vms.get_virsh_mac_address()
                         # restart guest network to get ip addr
                         utils_net.restart_guest_network(sess, vms_mac)
-                        vms_ip = network_base.get_vm_ip(sess, vms_mac)
-                        if not vms_ip and dhcp_range:
-                            test.fail("Guest has invalid ip address")
-                        elif vms_ip and not dhcp_range:
+                        vms_ip = None
+                        try:
+                            vms_ip = network_base.get_vm_ip(sess, vms_mac)
+                        except TestError:
+                            logging.debug("Didn't get guest IP.")
+                            if dhcp_range:
+                                test.fail("Guest has invalid ip address")
+                        if vms_ip and not dhcp_range:
                             # Get IP address on guest should return Null
                             # if it exceeds the dhcp range
                             test.fail("Guest has ip address: %s" % vms_ip)


smitterl:
test cases with `multiple_guests > 0`, only these execute the loop over `vms_list`
```
JOB ID     : d00742007e146e1105bd47e61ad010baea920716
JOB LOG    : /var/log/avocado/job-results/job-2024-12-12T12.27-d007420/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_macvtap.net_bridge.multi_guests: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_network.net_macvtap.net_bridge.multi_guests: PASS (89.14 s)
JOB HTML   : /var/log/avocado/job-results/job-2024-12-12T12.27-d007420/results.html
JOB TIME   : 90.69 s
JOB ID     : 68bb20618e9a4a7bd7477a62df9694223a26ba4e
JOB LOG    : /var/log/avocado/job-results/job-2024-12-12T12.24-68bb206/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_network.dnsmasq_test.net_dhcp_range: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_network.dnsmasq_test.net_dhcp_range: PASS (122.79 s)
JOB HTML   : /var/log/avocado/job-results/job-2024-12-12T12.24-68bb206/results.html
JOB TIME   : 124.55 s
```

chloerh:
@smitterl maybe the function get_vm_ip() could be updated to return None instead of raising error, to avoid different behavior. I think I could just add an arg and make it work. What do you think?

smitterl:
> @smitterl maybe the function get_vm_ip() could be updated to return None instead of raising error, to avoid different behavior. I think I could just add an arg and make it work. What do you think?

Do you mean I should add an arg like "raise_on_error" that per default would be `True` to keep the current behavior but enable behavior of `get_guest_ip_addr` via `raise_on_error=False`? I think that would work.

chloerh:
> > @smitterl maybe the function get_vm_ip() could be updated to return None instead of raising error, to avoid different behavior. I think I could just add an arg and make it work. What do you think?
> 
> Do you mean I should add an arg like "raise_on_error" that per default would be `True` to keep the current behavior but enable behavior of `get_guest_ip_addr` via `raise_on_error=False`? I think that would work.

Yes. I could update it too. 🤝

chloerh:
Updated function get_vm_ip:
https://github.com/autotest/tp-libvirt/pull/6157

smitterl:
> Updated function get_vm_ip:
> 
> * [get_vm_ip:Return None if needed #6157](https://github.com/autotest/tp-libvirt/pull/6157)

smitterl:
Seems this request has become obsolete #6157

PR #6071, opened 2024-12-17
migrate_with_virtual_devices: fix rng check method

dzhengfy:
As there might be other built-in rng devices, we only need to check
if attached virtio rng is in rng_available.


dzhengfy:
Before the fix:
migrate.virtual_devices.rng.backend_builtin.precopy
Error in phase POST_MIGRATE_BACK:&#10;Device rng is not removed from vm&#10;

After the fix:
 (1/1) type_specific.io-github-autotest-libvirt.migrate.virtual_devices.rng.backend_builtin.precopy: PASS (426.23 s)

libvirt-10.10.0

Yingshun (approved):

PR #6078, opened 2024-12-19
Remove the workaround since qemu bug fixed

yalzhang:
There was qemu bugs about virtio multiqueue, and now the bugs are fixed. Remove the workaround.


nanli1:
hi yalan, any results

Yingshun (approved):

PR #6088, opened 2024-12-24
update_device/iface: disable test on s390x

smitterl:
On s390x, there's only one model. Trying to unset the type will be ignored.


chunfuwen (approved):
lgtm

PR #6089, opened 2024-12-24
Clean the env before vm clone

yalzhang:
There could be scenario that existing VMs with the same name which block the following operations. Remove these VMs to clean the env.

### DIFF
--- a/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
+++ b/libvirt/tests/src/virsh_cmd/domain/virsh_managedsave.py
@@ -538,5 +538,5 @@ def run(test, params, env):
         if multi_guests:
             for i in range(int(multi_guests)):
                 virsh.remove_domain("%s_%s" % (vm_name, i),
-                                    "--remove-all-storage",
+                                    "--remove-all-storage --nvram",
                                     debug=True)


yalzhang:
before the test, there could be error:
rhel.managedsave.libvirt_guests.positive_test.no_parallel_shutdown.shutdown_on_shutdown.none_on_boot.with_transient_vm.persistent_only_true (from avocado-vt)
Error message:
Command 'virt-clone --original 'avocado-vt-vm1' --name 'avocado-vt-vm2' --auto-clone' failed.&#10;stdout: b''&#10;stderr: b"ERROR    Disk /var/lib/avocado/data/avocado-vt/images/jeos-27-x86_64-clone.qcow2 is already in use by other guests ['avocado-vt-vm1_0']. (Use --check path_in_use=off or --check all=off to override)\n"&#10;additional_info: None

after the fix, run the same with with existing VM 'avocado-vt-vm2'
# virsh list --all
 Id   Name             State
---------------------------------
 -    avocado-vt-vm1   shut off
 -    avocado-vt-vm2   shut off
(1/1) type_specific.io-github-autotest-libvirt.managedsave.libvirt_guests.positive_test.parallel_shutdown.shutdown_on_shutdown.ignore_on_boot.without_transient_vm: PASS (167.62 s)

yalzhang:
@nanli1  Hi Please help to review, Thank you!

libvirt/tests/src/daemon/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
```diff
@@ -308,8 +308,14 @@ def transfer_to_transient(per_guest_name):
         utils_path.find_command("virt-clone")
     except utils_path.CmdNotFoundError:
         test.cancel("No virt-clone command found.")
-
     # Clone additional vms: avocado-vt-vm2, avocado-vt-vm3.....
+    # Before clone, make sure there is no existing VMs
+    all_vm_list = virsh.dom_list('--name --all').stdout.strip().splitlines()
+    logging.debug("All vm on the host: %s", all_vm_list)
+    for vm_i in all_vm_list:
+        if vm_i.startswith("avocado-vt-vm") and vm_i != "avocado-vt-vm1":
```
nanli1, line 316: we use    guest_name = ("%s" % main_vm_name[:-1])+("%s" % str(i+2)) to as cloned name , right? if so,  could we         `if vm_i.startswith(main_vm_name) and vm_i != "main_vm_name":` here ?

yalzhang: Updated and tested pass

libvirt/tests/src/daemon/conf_file/sysconfig_libvirt_guests/libvirt_guests.py
```diff
@@ -310,6 +310,13 @@ def transfer_to_transient(per_guest_name):
         test.cancel("No virt-clone command found.")
 
     # Clone additional vms: avocado-vt-vm2, avocado-vt-vm3.....
+    # Before clone, make sure there is no existing VMs
```
Yingshun, line 313: What about adding some random strings in the vm to be created. 
for eg. in L321: 
guest_name = ("%s" % main_vm_name[:-1])+("%s" % str(i+2))  

what about using  utils_misc.generate_random_string(4) instead of str(i+2)?

Yingshun: Your deletion is very specific. If there happens to be a virtual machine with this name(main_vm_name[:-1])+2...) used and it's also vms in additional_vm settings for other tests in the environment, it will become a problem.

yalzhang: It's not 100% fail. I think we need to figure out which case can left these legacy VMs, and add cleanup code on that case. Let me analyse it further. Thank you~

yalzhang: Hi @Yingshun  I finally figure out the root cause of these legacy VMs. Fix it in the virsh_managedsave.py file. Please help to check, Thank you!

PR #6090, opened 2024-12-24
Update expected result for ovs bridge throughput

yalzhang:
For ovs bridge, if we set differnet value for peak and average, the average will map to the min rate, and peak to the max rate, which is different with linux bridge. For linux bridge, the average will map to max rate. To make the behavior consistent, set peak and average to be the same. When we set different value for average and peak on ovs bridge, libvirt will report warning on the libvirtd log.


Yingshun:
@yalzhang  can you plz upload test results?

PR #6115, opened 2025-01-07
Add new features support and enable more utils

chunfuwen:
Add new features support and enable more utils
1. Remove virt-manager package install since it is never necessary
2. Enable local storage as default since it is default behaviour after
   bootc image builder feature updates
3. Enable -use-librepo when building out anaconda-iso since this is
   default feature for anaconda-iso
4. Add util: podman_pull to support podman download image before run
   bootc image builder
5. Add files and directories customization when creating json files
   since it is new feature for bootc image builder
6. Enable vhd on rhel_9.6 and rhel_10.0 since this feature are
   newly supported on those two releases
7. Add check_squashfs when anaconda-iso since it is new feature
   requirement when building out anaconda-iso

### DIFF
--- a/provider/bootc_image_builder/bootc_image_build_utils.py
+++ b/provider/bootc_image_builder/bootc_image_build_utils.py
@@ -34,7 +34,7 @@ def install_bib_packages():
     install necessary bootc image builder necessary packages
 
     """
-    package_list = ["podman", "skopeo", "virt-install", "curl", "virt-manager"]
+    package_list = ["podman", "skopeo", "virt-install", "curl"]
     for pkg in package_list:
         try:
             path.find_command(pkg)
@@ -71,8 +71,7 @@ def podman_command_build(bib_image_url, disk_image_type, image_ref, config=None,
         else:
             cmd += " -v %s:/config.json  " % config
 
-    if local_container:
-        cmd += " -v /var/lib/containers/storage:/var/lib/containers/storage "
+    cmd += " -v /var/lib/containers/storage:/var/lib/containers/storage "
 
     if options is not None:
         cmd += " %s " % options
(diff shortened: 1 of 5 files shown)


provider/bootc_image_builder/bootc_image_build_utils.py
```diff
@@ -169,6 +168,17 @@ def podman_push(podman_username, podman_password, registry, container_url):
...
+def podman_pull(container_url):
+    """
+    Use podman pull registry image
+
+    :param container_url: image url
+    :return: CmdResult object
+    """
+    pull_image_cmd = f"sudo podman pull {container_url}"
+    process.run(pull_image_cmd, timeout=600, verbose=True, ignore_status=False, shell=True)
+
```
smitterl, line 180: The commit comment is "Enable vhd on more releases", but this and other lines look like you also changed more logic in the test code. Can you please update your commit message with more information about the changes?

smitterl (changes requested):

provider/bootc_image_builder/bootc_image_build_utils.py
chunfuwen: updated

smitterl: I'm not sure if I'm missing something. AFAICT you are now listing in one line code you touched but I'm missing the reasoning why it was necessary. It doesn't seem to be related to the vhd change so I'm my concern is that during debugging in the future we might not be able to understand why something has changed and if it's a valid change.

smitterl (changes requested):
I'm not sure if I'm missing something. AFAICT you are now listing in one line code you touched but I'm missing the reasoning why it was necessary. It doesn't seem to be related to the vhd change so I'm my concern is that during debugging in the future we might not be able to understand why something has changed and if it's a valid change.
I believe this is a general requirement for commit messages. Please help clarify.

chunfuwen:
@smitterl  I added more details about what changes this PR has, please help check again.

smitterl (approved):
LGTM thank you!

dzhengfy (approved):
LGTM

PR #6134, opened 2025-01-15
remot_with_unix: fix socket reload

smitterl:
Reload libvirtd sockets when using unix sockets for connection was always done, also for modular daemons.

Use libvirtd only if system is not configured with modular daemons.


smitterl:
Set draft because test is currently failing not sure if bug.

smitterl:
After rereviewing test case traditional mode is intended.

PR #6138, opened 2025-01-16
Add boottype for windows uefi guests

mxie91:


mxie91:
@xiaodwan pls help review, thanks

PR #6157, opened 2025-01-23
get_vm_ip:Return None if needed

chloerh:
We use get_vm_ip to replace utils_net.get_guest_ip_addr in some test code to avoid potential risks, but it changes the original behavior which is to return None when cannot get valid ip address. Add an argument to choose between returning None or raising error to fix this problem.

Test result:
```
 (1/1) type_specific.local.virtual_network.iface_network.net_macvtap.net_bridge.multi_guests: STARTED
 (1/1) type_specific.local.virtual_network.iface_network.net_macvtap.net_bridge.multi_guests: PASS (98.11 s)
```

### DIFF
--- a/provider/virtual_network/network_base.py
+++ b/provider/virtual_network/network_base.py
@@ -19,13 +19,15 @@ VIRSH_ARGS = {'ignore_status': False, 'debug': True}
 LOG = logging.getLogger('avocado.' + __name__)
 
 
-def get_vm_ip(session, mac, ip_ver="ipv4", timeout=5):
+def get_vm_ip(session, mac, ip_ver="ipv4", timeout=5, ignore_error=False):
     """
     Get vm ip address
 
     :param session: vm session
     :param mac: mac address of vm
     :param ip_ver: ip version, defaults to "ipv4"
+    :param ignore_error: True to return None, False to raise exception,
+           defaults to False
     :return: ip address of given mac
     """
     def _get_vm_ip():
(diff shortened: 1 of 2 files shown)


nanli1 (approved):

provider/virtual_network/network_base.py
```diff
@@ -39,16 +41,16 @@ def _get_vm_ip():
...
-            LOG.warn(f'No ip addr of given mac: {mac}')
+            LOG.warning(f'No ip addr of given mac: {mac}')
             return
         elif len(target_addr) > 1:
-            LOG.warn(f'Multiple ip addr: {target_addr}')
+            LOG.warning(f'Multiple ip addr: {target_addr}')
 
         return target_addr[0]['local']
 
     vm_ip = utils_misc.wait_for(_get_vm_ip, timeout, ignore_errors=True)
```
Yingshun, line 51: May I ask if you need to replace 'ignore_errors=True' with 'ignore_errors=ignore_error'?

Yingshun (commented):
Others LGTM

provider/virtual_network/network_base.py
chloerh: After checking the history and refreshing my memory, I found that the 'wait_for' was supposed to extend waiting time when trying to get ip address. Therefore ignoring errors when not getting ip during that time. 
The newly added arg 'ignore_error' is supposed to control the behavior of the function when not getting ip address after timeout.

Yingshun (approved):

PR #6158, opened 2025-01-23
virsh_net_update:Remove negative cases of update dns-txt dns-host

chloerh:
Updating dns-txt and dns-host of network is supported now


yanqzhan (approved):
background: “after 10.6, libvirt support to modify the dns-txt and dns-host record, still do not support live update the dns-srv record.”
So approved.

Yingshun (approved):

PR #6178, opened 2025-02-10
migration_iommu_device: using migrate_vm_back option

rh-jugraham:
Use migrate_vm_back option in avocado-vt/share/cfg/base.cfg to determine whether to migrate the vm back to the source host.

### DIFF
--- a/libvirt/tests/src/sriov/vIOMMU/migration_iommu_device.py
+++ b/libvirt/tests/src/sriov/vIOMMU/migration_iommu_device.py
@@ -80,8 +80,10 @@ def run(test, params, env):
         migration_obj.verify_default()
         check_iommu_xml(vm, params)
 
-        test.log.info("TEST_STEP: Migrate back the VM to the source host.")
-        migration_obj.run_migration_back()
-        migration_obj.migration_test.ping_vm(vm, params)
+        migrate_vm_back = "yes" == params.get("migrate_vm_back", "yes")
+        if migrate_vm_back:
+            test.log.info("TEST_STEP: Migrate back the VM to the source host.")
+            migration_obj.run_migration_back()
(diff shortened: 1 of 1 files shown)


rh-jugraham:
@dzhengfy Added migrate_vm_back option that, if enabled, skips backward migration for this test case.

libvirt/tests/src/sriov/vIOMMU/migration_iommu_device.py
```diff
@@ -80,8 +80,13 @@ def setup_test():
...
-        test.log.info("TEST_STEP: Migrate back the VM to the source host.")
-        migration_obj.run_migration_back()
+        migrate_vm_back = "yes" == params.get("migrate_vm_back", "yes")
+        if not migrate_vm_back:
+            return
+        else:
+            test.log.info("TEST_STEP: Migrate back the VM to the source host.")
+            migration_obj.run_migration_back()
+
         migration_obj.migration_test.ping_vm(vm, params)
```
dzhengfy, line 90: Is this ping_vm still needed? I am not sure, @Yingshun please help check.

dzhengfy: @rh-jugraham , After discussion with @Yingshun, we found below codes were already included in migration_obj.run_migration_back(). So the root cause is below `ping_vm` should not be executed.
`migrate_vm_back = "yes" == params.get("migrate_vm_back", "yes")`
So could you help update the code to only run ping_vm when migrate_vm_back is True? Thanks

rh-jugraham: Updated - only running run_migration_back() and ping_vm() if migrate_vm_back is True.

dzhengfy (approved):
LGTM

PR #6179, opened 2025-02-11
virt_admin: pause after daemon restart

smitterl:
We've seen virtqemud trying to access some unconfined service in these tests, making them fail only in our CI.

We have not been able to identify which PID it is trying to read.

Given that both tests do a restart before the access violation and in a previous test round executing some bash command right after didn't reproduce the violation, let's try to wait 10sec to see if it makes a difference.


smitterl:
Not sure about the final solution. We decided to have another check to learn more. Closing for now.

PR #3923, opened 2021-12-09
virttools: add new test type and virt-install test

smitterl:
Depends on avocado-vt PR https://github.com/avocado-framework/avocado-vt/pull/3270

1. /virttools:
    Add new test type for tools from virt-manager repository; by
    using avocado-vt we get access to many useful test functions
    ready to use for qemu/libvirt testing. The test type assumes
    the same default setup as for tp-libvirt/libvirt and builds
    on top of that. During testing, switching from a bootstrapped
    libvirt environment to a bootstrapped virttools environment
    worked and the test case passed.
2. hostdev_mdev:
    Add a new test case.
    a. 'run': This virt-install test case is currently ccw device
              specific but other types on other platforms can be
              implemented by creating a subclass for the
              MdevHandler for the desired mdev_type.
    b. 'clean_up/sleep': The introduction of 1sec sleep is
              necessary to make sure setting the device offline will
              work. Using the usual 'wait_for' here seemed to
              overcomplicate the code.
    c. 'virt_install_with_hostdev': The function hardcodes several
              vm installation parameters. This is to keep the code
              simple at this point but can be changed in the future.
3. spell.ignore:
    Add command and device names used in ccw.py and hostdev_mdev.py.
4. ccw.py:
    Add a new function to set devices offline to reset them to their
    previous state. Test cases with real ccw devices usually assume
    the test environment is configured in a way that the first unused
    ccw device will be used for the test case, possibly in a
    destructive way. Some functions have been added return value 'True'
    on success to allow for usage in 'wait_for' constructions.
5. README.md:
    Add readme file to give a bit more context and intention for the
    new test type.

Signed-off-by: Sebastian Mitterle <smitterl@redhat.com>

PR #6195, opened 2025-02-19
in_place_upgrade_guest: automates the case that verifies in place upgrade on a guest

rh-jugraham:
Case ID: VIRT-293757

Depends on: https://github.com/avocado-framework/avocado-vt/pull/4070

Automates the case that verifies that an in place upgrade can be successfully completed on a guest. This case has already been automated by /tp-qemu/qemu/tests/in_place_upgrade.py - this specific test in tp-libvirt exists such that an in place guest upgrade test can be accomplished without depending on 'kar'.

This test specifically supports upgrading from rhel X-1 -> rhel X on a rhel X host.

Tests passing:
```
(.libvirt-ci-venv-ci-runtest-ZbmaDy) [root@ampere-mtsnow-altra-09 ~]# avocado run --vt-type libvirt --vt-omit-data-loss --vt-machine-type arm64-mmio in_place_upgrade_guest --vt-connect-uri qemu:///system
JOB ID     : 3947dcc69ad9e030e06a3828ae35e8470aa508dc
JOB LOG    : /var/log/avocado/job-results/job-2025-02-24T15.09-3947dcc/job.log
 (1/1) type_specific.io-github-autotest-libvirt.in_place_upgrade_guest.rhel9_to_rhel10.0: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.in_place_upgrade_guest.rhel9_to_rhel10.0: PASS (1233.92 s)
JOB HTML   : /var/log/avocado/job-results/job-2025-02-24T15.09-3947dcc/results.html
JOB TIME   : 1235.61 s
```

### DIFF
--- a/libvirt/tests/cfg/in_place_upgrade_guest.cfg
+++ b/libvirt/tests/cfg/in_place_upgrade_guest.cfg
@@ -1,0 +1,11 @@
+- in_place_upgrade_guest:
+    type = in_place_upgrade_guest
+    backup_image_before_testing = "yes"
+    restore_image_after_testing = "yes"
+    release_check_cmd = "cat /etc/redhat-release"
+    kernel_check_cmd = "uname -r"
+    pagesize_check_cmd = "getconf PAGESIZE"
+    compose_url = "EXAMPLE.COMPOSE.URL"
+    upgrade_repos_path = "/etc/leapp/files/leapp_upgrade_repositories.repo"
+    leapp_preupgrade_cmd = "leapp preupgrade --debug --no-rhsm"
+    leapp_upgrade_cmd = "leapp upgrade --debug --no-rhsm"
--- a/libvirt/tests/src/in_place_upgrade_guest.py
+++ b/libvirt/tests/src/in_place_upgrade_guest.py
@@ -1,0 +1,210 @@
+# pylint: disable=spelling
+# disable pylint spell checker to allow for leapp and preupgrade
+import re
+import os
+import textwrap
+
+from avocado.utils import process
+
+from virttest import utils_package
+from virttest import utils_test
+
+
+def run(test, params, env):
+    """
+    This case verifies that an in place upgrade on a guest succeeds.
+    This test only supports upgrading a RHEL X-1 guest to RHEL X on a RHEL X host.
+    This test ensures that, when relevant, the page size does not change after upgrade.
+
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params env: environment object
+    """
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+
+    try:
+        if not vm.is_alive():
+            vm.start()
+        session = vm.wait_for_login()
+
+        source_release, target_release = get_release_info(test, params, session)
+        check_support(test, source_release, target_release)
+
+        pagesize_check_cmd = params.get("pagesize_check_cmd")
+        target_page_size = session.cmd_output(pagesize_check_cmd).strip()
+
+        utils_package.package_install(["leapp-upgrade*"], session=session, timeout=360)
+
+        prepare_repos_in_guest(test, params, vm, target_release)
+
+        run_leapp_cmd(test, params, session, step="preupgrade")
+        run_leapp_cmd(test, params, session, step="upgrade")
+        session = vm.reboot(session=session, timeout=900)
+
+        verify_upgrade_succeeded(test, params, vm, session, target_release, target_page_size)
+
+        session.close()
+
+    finally:
+        if vm.is_alive():
+            vm.destroy(gracefully=False)
+
+
+def get_release_info(test, params, session):
+    """
+    Get the source release and target release versions
+
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params session: the vm session
+    :return: tuple, (source  version, target  version)
+    """
+    release_check_cmd = params.get("release_check_cmd")
+    test.log.debug("Getting source release (same as release on vm)")
+    source_release = extract_release_version(test, session.cmd_output(release_check_cmd))
+    test.log.debug("Getting target release (same as release on host)")
+    target_release = extract_release_version(test, process.run(release_check_cmd, shell=True).stdout_text)
+    return (source_release, target_release)
+
+
+def extract_release_version(test, release_output):
+    """
+    Extract the release version from the full release output
+
+    :params test: test object
+    :params release_output: output from running release_check_cmd
+    """
+    test.log.debug("Full release output: %s" % release_output)
+    return re.findall("\d+\.\d+", release_output)[0]
+
+
+def check_support(test, source_release, target_release):
+    """
+    Check whether the test supports upgrading from source release to target release
+
+    :params test: test object
+    :params source_release: source release version (upgrading from version)
+    :params target_release: target release version (upgrading to version)
+    """
+    source_major_release = int(source_release.split(".")[0])
+    target_major_release = int(target_release.split(".")[0])
+
+    if (source_major_release + 1 != target_major_release):
+        test.cancel("Can not upgrade guest from rhel %s to rhel %s - "
+                    "This test only supports upgrading a RHEL X-1 guest to RHEL X on a RHEL X host" %
+                    (source_release, target_release))
+    else:
+        test.log.info("Upgrading guest from rhel %s to rhel %s on a rhel %s host" %
+                      (source_release, target_release, target_release))
+
+
+def prepare_repos_in_guest(test, params, vm, target_release):
+    """
+    Prepare repos in the guest by including the upgrade repos paths
+
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params session: the vm
+    :params target_release: target release version
+    """
+    compose_url = params.get("compose_url")
+    upgrade_repos_path = params.get("upgrade_repos_path")
+    vm_arch_name = params.get("vm_arch_name")
+
+    target_major_release = target_release.split(".")[0]
+
+    upgrade_repos_content = textwrap.dedent(f"""
+        [APPSTREAM]
+        name=APPSTREAM
+        baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/AppStream/{vm_arch_name}/os/
+        enabled=1
+        gpgcheck=0
+        [BASEOS]
+        name=BASEOS
+        baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/BaseOS/{vm_arch_name}/os/
+        enabled=1
+        gpgcheck=0
+    """)
+
+    test.log.debug("Temporarily creating leapp_upgrade_repositories file on the host: %s" % upgrade_repos_content)
+    tmp_upgrade_repos_path = os.path.join(test.debugdir, "tmp_leapp_upgrade_repositories.repo")
+    with open(tmp_upgrade_repos_path, 'w+') as f:
+        f.write(upgrade_repos_content)
+
+    test.log.info("Copying leapp_upgrade_repositories file onto the guest")
+    vm.copy_files_to(host_path=tmp_upgrade_repos_path, guest_path=upgrade_repos_path)
+
+
+def run_leapp_cmd(test, params, session, step):
+    """
+    Run the leapp (pre)upgrade cmd
+
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params session: the vm session
+    :params step: "preupgrade" or "upgrade"
+    """
+    leapp_cmd = params.get("leapp_preupgrade_cmd")
+    if step == "upgrade":
+        leapp_cmd = params.get("leapp_upgrade_cmd")
+
+    status = session.cmd_status(leapp_cmd, timeout=900)
+    if status == 0:
+        test.log.info("Leapp %s executed successfully" % step)
+    else:
+        test.fail("Leapp %s failed, see logs for more info" % step)
+
+
+def verify_upgrade_succeeded(test, params, vm, session, target_release, target_page_size):
+    """
+    Verify whether the upgrade succeeded by checking
+    the release version, the kernel version, and the page size
+
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params vm: the vm
+    :params session: the vm session
+    :params target_release: target release version (guest version after upgrade)
+    :params target_page_size: target page size (result of pagesize_check_cmd)
+    """
+    vm_arch_name = params.get("vm_arch_name")
+    release_check_cmd = params.get("release_check_cmd")
+    kernel_check_cmd = params.get("kernel_check_cmd")
+    pagesize_check_cmd = params.get("pagesize_check_cmd")
+
+    test.log.debug("Getting guest release after upgrade")
+    post_release_output = extract_release_version(test, session.cmd_output(release_check_cmd))
+    if target_release not in post_release_output:
+        test.fail("The guest after upgrade should be at rhel %s, but is at %s" %
+                  (target_release, post_release_output))
+    else:
+        test.log.info("The guest after upgrade is at rhel %s, as expected" % post_release_output)
+
+    target_major_release = target_release.split(".")[0]
+    test.log.debug("Getting guest kernel version after upgrade")
+    post_kernel_output = session.cmd_output(kernel_check_cmd).strip()
+    if (".el%s" % target_major_release) not in post_kernel_output:
+        test.fail("The guest after upgrade should be at kernel version .el%s, but is at %s" %
+                  (target_major_release, post_kernel_output))
+    else:
+        test.log.info("The guest after upgrade is at kernel version .el%s (%s), as expected" %
+                      (target_major_release, post_kernel_output))
+
+    post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+    if vm_arch_name == "aarch64" and target_page_size != post_page_size:
+        # Note: the following is a workaround that should be removed after the leapp tool provides 64k support on aarch64
+        test.log.info("Guest page size before upgrade (%s) does not match page size after upgrade (%s)" %
+                      (target_page_size, post_page_size))
+        test.log.info("Attempting temporary workaround to update the guest default kernel")
+        kernel_pattern = ".*el%s.*%sk" % (target_release.replace(".", "_"), int(target_page_size) // 1024)
+        kernel_version = utils_test.get_available_kernel_paths(session, kernel_pattern)[0]
+        utils_test.update_vm_default_kernel(vm, kernel_version, reboot=True, guest_arch_name=vm_arch_name, timeout=900)
+
+        session = vm.wait_for_login()
+        post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+        if target_page_size != post_page_size:
+            test.fail("Guest page size before upgrade (%s) does not match page size after upgrade and workaround (%s)" %
+                      (target_page_size, post_page_size))
+        else:
+            test.log.info("Guest page size after upgrade and workaround is %s, as expected" % post_page_size)


rh-jugraham:
@dzhengfy @Yingshun Could you help review this PR? 

A few notes:
1. I added a 64k kernel workaround for now that can be removed when/if there is a change
2. I was considering adding a new test type as this is not technically a libvirt specific test; however, I wasn't sure if it made sense to create a new test type for 1 test
3. I wasn't sure how to best add in version specific information - I added it in the cfg under the rhel9_to_rhel10.0 variant (similar to the tp-qemu ipu test) but unsure if there is an alternative

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,16 @@
...
+    upgrade_repos_path = "/etc/leapp/files/leapp_upgrade_repositories.repo"
+    leapp_preupgrade_cmd = "leapp preupgrade --debug --no-rhsm"
+    leapp_upgrade_cmd = "leapp upgrade --debug --no-rhsm"
+    list_kernel_cmd = "grubby --info=ALL | grep ^kernel"
+    default_kernel_cmd = "grub2-set-default"
+    variants:
+        - rhel9_to_rhel10.0:
+            pre_major_release = "9"
+            target_major_release = "10"
+            target_release = "RHEL-10.0"
```
iccaszhulili, line 16: Hi, Julia

As I heard from dzheng, that we just need to test upgrade on RHEL10 host, so , we I think we may just need to check the host major and minor version, then we can get the target_release. And I think what we need to input is the pre_minor_release, since the major release is always RHEL.(X-1), we do not support upgrade from RHEL.(X-2) to RHEL.X in one shot.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,16 @@
+- in_place_upgrade_guest:
+    type = in_place_upgrade_guest
+    release_check_cmd = "cat /etc/redhat-release"
+    kernel_check_cmd = "uname -r"
+    pagesize_check_cmd = "getconf PAGESIZE"
```
iccaszhulili, line 5: Hi, Julia 

Do we need to do the following steps
```
# yum -y install kernel-64k 
# k=$(echo /boot/vmlinuz*64k) 
# grubby --set-default=$k --update-kernel=$k --args="crashkernel=2G-:640M" 
# reboot
```
before doing upgrade. Is there an existing 64k image for testing?

libvirt/tests/cfg/in_place_upgrade_guest.cfg
rh-jugraham: @iccaszhulili 

I'm not sure if I am understanding your comment correctly, so let me know if I misunderstood. What I realized is that we know the release before upgrade (RHEL-X-1.y) and we know the target major release (RHEL-X) but we don't know the target minor release. So I tried modifying the code such that everything is calculated (no need to hard code what doesn't need to be) except for the target minor release, which needs to be given with, for example, the variant `to_RHELX.0` for target minor release RHEL-X.0 (such as migrating rhel-9.6 --> rhel10.0).

Updated tests passing:
```
(.libvirt-ci-venv-ci-runtest-ZbmaDy) [root@ampere-mtsnow-altra-09 ~]# avocado run --vt-type libvirt --vt-omit-data-loss --vt-machine-type arm64-mmio in_place_upgrade_guest --vt-connect-uri qemu:///system
JOB ID     : d1ac967df701dbdc57a74e246ade58a97ac0a4aa
JOB LOG    : /var/log/avocado/job-results/job-2025-02-25T16.45-d1ac967/job.log
 (1/1) type_specific.io-github-autotest-libvirt.in_place_upgrade_guest.to_rhelX.0: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.in_place_upgrade_guest.to_rhelX.0: PASS (994.40 s)
JOB HTML   : /var/log/avocado/job-results/job-2025-02-25T16.45-d1ac967/results.html
JOB TIME   : 996.08 s
```

iccaszhulili: Hi, Julia 

> @iccaszhulili
> 
> I'm not sure if I am understanding your comment correctly, so let me know if I misunderstood. What I realized is that we know the release before upgrade (RHEL-X-1.y) and we know the target major release (RHEL-X) but we don't know the target minor release. 
I think we can get the minor release from the host, for example, in RHEL10.1 release testing cycle, we need to test upgrade from RHEL9.7 to RHEL10.1. During testing, we will provision a host with RHEL10.1 OS, then we can know that we need upgrade the guest to RHEL10.1, the guest OS version will be aligned with the host OS version.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,14 @@
...
+    compose_url = "http://download.eng.bos.redhat.com"
+    upgrade_repos_path = "/etc/leapp/files/leapp_upgrade_repositories.repo"
+    leapp_preupgrade_cmd = "leapp preupgrade --debug --no-rhsm"
+    leapp_upgrade_cmd = "leapp upgrade --debug --no-rhsm"
+    list_kernel_cmd = "grubby --info=ALL | grep ^kernel"
+    default_kernel_cmd = "grub2-set-default"
+    variants:
+        - to_rhelX.0:
```
iccaszhulili, line 13: Hi, Julia

I am Okay with it. Just to mention that, we may need to do some settings in CI. We need to ask CI to provision a guest with latest RHEL(X-1) OS. We need to make sure CI knows the minor release of RHEL(X-1).

iccaszhulili (approved):

libvirt/tests/cfg/in_place_upgrade_guest.cfg
dzhengfy: We have various guest in different versions in US/PEK2 storage server. So I think we just fetch it as we want.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,134 @@
...
+                test.fail("Page size before upgrade (%s) is not the same after upgrade (%s)" % (target_page_size, post_page_size))
+
+        session.close()
+
+    finally:
+        # Recover VM
+        if vm.is_alive():
+            vm.destroy(gracefully=False)
+        test.log.info("Restoring vm %s...", vm.name)
+        vmxml_backup.sync()
```
dzhengfy, line 134: If I did not miss anything, I did not see any changes to vm xml. So we can remove this line.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,134 @@
...
+    release_check_cmd = params.get("release_check_cmd")
+    kernel_check_cmd = params.get("kernel_check_cmd")
+    pagesize_check_cmd = params.get("pagesize_check_cmd")
+    compose_url = params.get("compose_url")
+    upgrade_repos_path = params.get("upgrade_repos_path")
+    target_minor_release = params.get("target_minor_release")
+
+    # 1) Prepare a running guest
+    vm = env.get_vm(vm_name)
+    vmxml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
```
dzhengfy, line 46: If I did not miss anything, I did not see any changes to vm xml. So we can remove this line.

dzhengfy (commented):
Others LGTM

libvirt/tests/cfg/in_place_upgrade_guest.cfg
rh-jugraham: @dzhengfy This does make more sense to me (we have a rhelX.z machine and we are testing upgrading from rhelX-1.y --> X.z). However, how do we make sure the CI knows the minor release of rhelX-1.y - how do we know which guest is the latest to fetch? And what is the best way to fetch it?

libvirt/tests/cfg/in_place_upgrade_guest.cfg
rh-jugraham: This makes more sense, thanks!

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,14 @@
+- in_place_upgrade_guest:
+    type = in_place_upgrade_guest
+    release_check_cmd = "cat /etc/redhat-release"
+    kernel_check_cmd = "uname -r"
+    pagesize_check_cmd = "getconf PAGESIZE"
+    compose_url = "http://download.eng.bos.redhat.com"
```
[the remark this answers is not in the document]
rh-jugraham: @dzhengfy Done - created a PR in ci-shared-datas. One question: is it an issue that the test has `baseurl={compose_url}/rhel-{target_major_release}/nightly...`? I'm not sure if the full path like this belongs in the code but I had struggled defining it completely in the cfg (especially getting vm_arch_name).

libvirt/tests/cfg/in_place_upgrade_guest.cfg
rh-jugraham: Update: understood - the test will assume that it already has a guest at version X-1.y on a host with version X.z.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,11 @@
+- in_place_upgrade_guest:
+    type = in_place_upgrade_guest
+    release_check_cmd = "sed 's/[^0-9.]*//g' /etc/redhat-release"
```
smitterl, line 3: It's generally preferrable to handle strings in Python code to make debugging easier. It might be unlikely in this specific case but the pattern or the place for the searched-for information might change, in that case the release check would return an empty string but the logs wouldn't record what the output was.
So, I'd prefer to see you read `/etc/redhat-release` as string and use python module `re`.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,138 @@
...
+
+from virttest import utils_package
+
+
+def run_leapp_cmd(test, params, session, step):
+    leapp_cmd = params.get("leapp_preupgrade_cmd")
+    if step == "upgrade":
+        leapp_cmd = params.get("leapp_upgrade_cmd")
+
+    leapp_status, leapp_output = session.cmd_status_output(leapp_cmd, timeout=900)
```
smitterl, line 15: This line is longer than the usual 80 characters. The context (function name, command variable) already indicate it's related to leapp, so `status, output` should suffice.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,138 @@
...
+    if leapp_status == 0:
+        test.log.info("Leapp %s executed successfully" % step)
+    else:
+        test.log.debug("Leapp %s output: %s" % (step, leapp_output))
```
smitterl, line 19: The leapp output with --debug can be really long.
Alternatively, it could make more sense to copy the leapp pre/upgrade log into the test folder, origin `/var/log/leapp/leapp-(pre)upgrade.log` target.

The test might handle it similar to the `LibvirtdDebugLog` in https://github.com/avocado-framework/avocado-vt/blob/master/virttest/env_process.py#L1012
But even if not, the debug log is also stored in the test log directory so it could at least help understand how the code can construct the path to the test log directory where to copy the file to.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,138 @@
...
+            baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/AppStream/{vm_arch_name}/os/
+            enabled=0
+            gpgcheck=0
+            [BASEOS]
+            name=BASEOS
+            baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/BaseOS/{vm_arch_name}/os/
+            enabled=0
+            gpgcheck=0
+        """)
+        session.cmd("cat <<EOF > %s %s\nEOF" % (upgrade_repos_path, upgrade_repos_content))
```
smitterl, line 87: As mentioned above, it's usually better to use the python library instead of bash in this case:

with(upgrade_repos_path...) as f:
     f.write(f"""
...
gpgcheck=0
""")
```

File: libvirt/tests/src/in_place_upgrade_guest.py [hunk 11]
```diff
@@ -0,0 +1,138 @@
...
+
+    # 1) Prepare a running guest
+    vm = env.get_vm(vm_name)
+
+    try:
+        if not vm.is_alive():
+            vm.start()
+        session = vm.wait_for_login()
+
+        # 2) Calculate target release and pagesize
```
smitterl, line 53: We often try to apply clean code principles. One demands putting logic into smaller units (functions) and maintain the same abstraction level.

The test code could hence be written as:

```
from_release, to_release = get_release_info(release_check_cmd)
check_support(from_release, to_release)
utils_package.package_install(...)
configure_leapp_tool(compose_url, to_release)
pre_ugprade_info.append(get_guest_info(session, guest_cmd)) # << this can be generalized for other conditions
if pre_upgrade_succeeds():
    run_upgrade_and_reboot()
check_upgrade_post_conditions(pre_upgrade_info, vm, params)

...
def check_upgrade_post_conditions(pre_upgrade_info, vm, params):
...
    session = vm.wait_for_login()
    check_kernel_version(session)
    post_upgrade_info.append(guest_info(...))
    confirm_are_equal(pre_upgrade_info, post_upgrade_info)

```

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,138 @@
...
+
+            # and check again
+            post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+            if target_page_size != post_page_size:
+                test.fail("Page size before upgrade (%s) is not the same after upgrade (%s)" % (target_page_size, post_page_size))
+
+        session.close()
+
+    finally:
+        # Destroy VM
```
smitterl, line 136: Is this comment necessary?

smitterl (changes requested):
Thanks for this. I think this can be improved to be more generally useful, esp. on more architectures.

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: I do know that the vm logs do include the the full output of the (pre)upgrade cmd so technically nothing needs to be done. The reason why I had originally decided to add it into specifically the debug log was because I know that, for example, when running tests with jenkins, you often only get access to the debug log. However, given that this test will likely not be run in that context, having this output in the debug log is unnecessary as it is very long.

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: @smitterl How would I do so on the vm since `with open() as f` tries to open the file on the host - I haven't been able to find an alternative for the guest.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,11 @@
...
+    kernel_check_cmd = "uname -r"
+    pagesize_check_cmd = "getconf PAGESIZE"
+    compose_url = "EXAMPLE.COMPOSE.URL"
+    upgrade_repos_path = "/etc/leapp/files/leapp_upgrade_repositories.repo"
+    leapp_preupgrade_cmd = "leapp preupgrade --debug --no-rhsm"
+    leapp_upgrade_cmd = "leapp upgrade --debug --no-rhsm"
+    list_kernel_cmd = "grubby --info=ALL | grep ^kernel"
+    default_kernel_cmd = "grub2-set-default"
```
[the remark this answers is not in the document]
rh-jugraham: Good call - I created a [PR](https://github.com/avocado-framework/avocado-vt/pull/4070) for that!

rh-jugraham:
> Thanks for this. I think this can be improved to be more generally useful, esp. on more architectures.

@smitterl Thank you for the detailed feedback! I believe I have addressed all comments or added a question if confused.

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: Updated! Now using `with open() as f` to create the file on the host then copying it over to the guest.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
```diff
@@ -0,0 +1,10 @@
+- in_place_upgrade_guest:
+    type = in_place_upgrade_guest
+    release_check_cmd = "cat /etc/redhat-release"
+    kernel_check_cmd = "uname -r"
+    pagesize_check_cmd = "getconf PAGESIZE"
```
iccaszhulili, line 5: @yicui Is this command better than `uname -r`?

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+
+    finally:
+        if vm.is_alive():
+            vm.destroy(gracefully=False)
+
+
+def get_release_info(params, session):
+    """
+    Get the pre release and target release versions
+    """
```
dzhengfy, line 57: Please add docstring like  `:return: tuple, (source  version, target  version)` as well as the parameters

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    release_check_cmd = params.get("release_check_cmd")
+    pre_release = extract_release_version(session.cmd_output(release_check_cmd))
+    target_release = extract_release_version(process.run(release_check_cmd, shell=True).stdout_text)
+    return (pre_release, target_release)
+
+
+def extract_release_version(release_output):
+    """
+    Extract the release version from the full release output
+    """
```
dzhengfy, line 67: Add dostring for paremeter

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    return re.findall("\d+\.\d+", release_output)[0]
+
+
+def check_support(test, pre_release, target_release):
+    """
+    Check whether the test supports upgrading from pre_release to target_release
+    """
```
dzhengfy, line 74: Complete the docstring, thanks

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def get_release_info(params, session):
+    """
+    Get the pre release and target release versions
+    """
+    release_check_cmd = params.get("release_check_cmd")
+    pre_release = extract_release_version(session.cmd_output(release_check_cmd))
+    target_release = extract_release_version(process.run(release_check_cmd, shell=True).stdout_text)
+    return (pre_release, target_release)
```
dzhengfy, line 61: Could you add more debug infomation for your returned values?

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+        gpgcheck=0
+    """)
+
+    tmp_dir = data_dir.get_tmp_dir()
+    tmp_upgrade_repos_path = os.path.join(tmp_dir, "tmp_leapp_upgrade_repositories.repo")
+    with open(tmp_upgrade_repos_path, 'w+') as f:
+        f.write(upgrade_repos_content)
+
+    vm.copy_files_to(host_path=tmp_upgrade_repos_path, guest_path=upgrade_repos_path)
+
```
dzhengfy, line 116: Could you add more debug infomation to describe what this function does?

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    os.remove(tmp_upgrade_repos_path)
+
+
+def run_leapp_cmd(test, params, session, step):
+    """
+    Run the leapp (pre)upgrade cmd
+    """
```
dzhengfy, line 123: Complete docstring, please

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+        test.log.info("Leapp %s executed successfully" % step)
+    else:
+        test.fail("Leapp %s failed, see logs for more info" % step)
+
+
+def verify_upgrade_succeeded(test, params, vm, session, target_release, target_page_size):
+    """
+    Verify whether the upgrade succeeded by checking
+    the release version, the kernel version, and the page size
+    """
```
dzhengfy, line 139: Complete docstring, please

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    vm_arch_name = params.get("vm_arch_name")
+    release_check_cmd = params.get("release_check_cmd")
+    kernel_check_cmd = params.get("kernel_check_cmd")
+    pagesize_check_cmd = params.get("pagesize_check_cmd")
+
+    post_release_output = extract_release_version(session.cmd_output(release_check_cmd))
+    if target_release not in post_release_output:
+        test.fail("The guest after upgrade should be at rhel %s, but is at %s" %
+                  (target_release, post_release_output))
+
```
dzhengfy, line 149: I suggest we add `else` clause to display debug information like what result is as expected.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    target_major_release = target_release.partition(".")[0]
+    post_kernel_output = session.cmd_output(kernel_check_cmd)
+    if (".el%s" % target_major_release) not in post_kernel_output:
+        test.fail("The guest after upgrade should be at kernel version .el%s, but is at %s" %
+                  (target_major_release, post_kernel_output))
```
dzhengfy, line 154: ditto

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+
+    post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+    if target_page_size != post_page_size:
+        kernel_version = get_kernel_version(test, params, session, target_release, target_page_size)
```
dzhengfy, line 158: Please add some INFO like this is a workaround so far and should be removed after leapp tool support 64k on aarch64. Thanks

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+        utils_test.update_default_kernel(vm, kernel_version, timeout=900, guest_arch_name=vm_arch_name)
+
+        session = vm.wait_for_login()
+        post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+        if target_page_size != post_page_size:
+            test.fail("Page size before upgrade (%s) is not the same after upgrade (%s)" %
+                      (target_page_size, post_page_size))
+
```
dzhengfy, line 166: `Else` clause

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+
+def get_kernel_version(test, params, session, target_release, target_page_size):
+    """
+    Get the kernel version given the release version and page size
+    """
```
dzhengfy, line 171: Docstring, please

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    list_kernel_cmd = params.get("list_kernel_cmd")
+
+    kernels = session.cmd_output(list_kernel_cmd).splitlines()
+    release = ".el%s" % (target_release.partition(".")[0])
+    page_size = "+%sk" % (int(target_page_size) // 1024)
+    try:
+        kernel = [kernel for kernel in kernels if release in kernel and page_size in kernel][0]
+        kernel_version = kernel.split("/")[-1].strip('\"')
+        return kernel_version
```
dzhengfy, line 180: Before return, please add more debug info.

libvirt/tests/cfg/in_place_upgrade_guest.cfg
iccaszhulili: @dzhengfy `getconf PAGESIZ` is better than `uname -r`?

dzhengfy: @iccaszhulili I think both are ok.` uname -r ` is mainly for checking the kernel version, not pagesize here.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,183 @@
...
+    upgrade_repos_path = params.get("upgrade_repos_path")
+    vm_arch_name = params.get("vm_arch_name")
+
+    target_major_release = target_release.partition(".")[0]
+
+    upgrade_repos_content = textwrap.dedent(f"""
+        [APPSTREAM]
+        name=APPSTREAM
+        baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/AppStream/{vm_arch_name}/os/
+        enabled=0
```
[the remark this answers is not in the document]
rh-jugraham: @dzhengfy Yes, I believe so? But could you clarify this question?

dzhengfy (commented):
I am fine with other codes. Just waiting for the clarification about enable repo or not

libvirt/tests/src/in_place_upgrade_guest.py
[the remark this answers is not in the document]
rh-jugraham: I see, makes sense!

dzhengfy (approved):

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+        session = vm.wait_for_login()
+        post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+        if target_page_size != post_page_size:
+            test.fail("Guest page size before upgrade (%s) does not match page size after upgrade and workaround (%s)" %
+                      (target_page_size, post_page_size))
+        else:
+            test.log.info("Guest page size after upgrade and workaround is %s, as expected" % post_page_size)
+
+
+def get_kernel_version(test, params, session, target_release, target_page_size):
```
smitterl, line 215: IMO, this function should live in https://github.com/avocado-framework/avocado-vt/pull/4070 or where ever the `update_vm_default_kernel` function lives because it can help others who use that function. They will likely have to re-implement the logic presented here, or similar logic, to achieve that.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+                      (target_major_release, post_kernel_output))
+
+    post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+    if target_page_size != post_page_size:
+        # Note: the following is a workaround that should be removed after the leapp tool provides 64k support on aarch64
+        test.log.info("Guest page size before upgrade (%s) does not match page size after upgrade (%s)" %
+                      (target_page_size, post_page_size))
+        test.log.info("Attempting temporary workaround to update the guest default kernel")
+        kernel_version = get_kernel_version(test, params, session, target_release, target_page_size)
+        utils_test.update_vm_default_kernel(vm, kernel_version, timeout=900, guest_arch_name=vm_arch_name)
```
smitterl, line 204: If you trigger the reboot here through the default parameter, IMO you should still set it explicitly 'reboot=True' because the next time this code is reviewed the reviewer might not be familiar with the `update_vm_default_kernel` function and its defaults. However, here it's crucial.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+        run_leapp_cmd(test, params, session, step="upgrade")
+        session = vm.reboot(session=session, timeout=900)
+
+        verify_upgrade_succeeded(test, params, vm, session, target_release, target_page_size)
+
+        session.close()
+
+    finally:
+        if vm.is_alive():
+            vm.destroy(gracefully=False)
```
smitterl, line 51: Usually, we expect tests to restore the VM as it was before. This is not possible here and is only achieved by avocado-vt's `backup_image` and `restore_image` configuration values (s. https://github.com/avocado-framework/avocado-vt/blob/master/virttest/shared/cfg/base.cfg#L582)

You can decide to back up and restore the original image file in your test case or at least add a WARNING at the start of the `run` docstring. If the backup/restore functionality is not configured (AFAICT it isn't s. `base.cfg` above) then the test can't be rerun, the test data has to be set up again.

You could also check for the configuration and raise a "TestError" to inform the person who wants to execute this test to please consider these configuration parameters. Unfortunately, the functionality is implemented in `virttest/env_process.py` so you can't force the image to be backed up.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+
+from virttest import utils_package
+from virttest import utils_test
+from virttest import data_dir
+
+
+def run(test, params, env):
+    """
+    This case verifies that an in place upgrade on a guest succeeds
+    This test only supports upgrading a RHEL X-1 guest to RHEL X on a RHEL X host
```
smitterl, line 17: You might want to add that it also confirms that the page size isn't changed. In a later PR somebody could expand this to 'test upgrade invariants' creating a dictionary before and after the upgrade and confirming they have the same values, if not reporting them.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+def check_support(test, source_release, target_release):
+    """
+    Check whether the test supports upgrading from source release to target release
+
+    :params test: test object
+    :params source_release: source release version (upgrading from version)
+    :params target_release: target release version (upgrading to version)
+    """
+    source_major_release = int(source_release.partition(".")[0])
+    target_major_release = int(target_release.partition(".")[0])
```
smitterl, line 91: Is there a reason why you are using `partition` instead of `split`? TBH, I see this function for the first time. Usually, in the framework we use `split`. Out of convention, I'd prefer you used it, too, but I wouldn't insist on it.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+    :params test: test object
+    :params params: wrapped dict with all parameters
+    :params session: the vm
+    :params target_release: target release version
+    """
+    compose_url = params.get("compose_url")
+    upgrade_repos_path = params.get("upgrade_repos_path")
+    vm_arch_name = params.get("vm_arch_name")
+
+    target_major_release = target_release.partition(".")[0]
```
smitterl, line 115: s. above

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+        gpgcheck=0
+        [BASEOS]
+        name=BASEOS
+        baseurl={compose_url}/rhel-{target_major_release}/nightly/RHEL-{target_major_release}/latest-RHEL-{target_release}/compose/BaseOS/{vm_arch_name}/os/
+        enabled=1
+        gpgcheck=0
+    """)
+
+    test.log.debug("Temporarily creating leapp_upgrade_repositories file on the host: %s" % upgrade_repos_content)
+    tmp_dir = data_dir.get_tmp_dir()
```
smitterl, line 131: If you use `test.debugdir` instead the file will be stored in the test result folder (e.g. `job-results/latest/test-results/<test-name>/libvirt_upgrade_repositories.repo`) which is the first place where I would check for the file. The `get_tmp_dir` often points to a path below `/var/tmp` which is test-run specific but not as easy to identify as the test run folder.

libvirt/tests/src/in_place_upgrade_guest.py
```diff
@@ -0,0 +1,240 @@
...
+    if (".el%s" % target_major_release) not in post_kernel_output:
+        test.fail("The guest after upgrade should be at kernel version .el%s, but is at %s" %
+                  (target_major_release, post_kernel_output))
+    else:
+        test.log.info("The guest after upgrade is at kernel version .el%s (%s), as expected" %
+                      (target_major_release, post_kernel_output))
+
+    post_page_size = session.cmd_output(pagesize_check_cmd).strip()
+    if target_page_size != post_page_size:
+        # Note: the following is a workaround that should be removed after the leapp tool provides 64k support on aarch64
```
smitterl, line 199: In order for this to be valid for other archs you should add an arch switch:

```
import platform
...

if platform.machine() == "aarch64":
...

```

smitterl (changes requested):

smitterl:
Considering that the SUT here is not libvirt but Leapp in KVM guests, you could also consider creating a new test type and moving these files there. E.g. "tp-libvirt/ipu/tests/(cfg|src)/..."
How to do this: https://github.com/autotest/tp-libvirt/pull/3923

Maybe it would be too much effort to just run a single test scenario. I don't know. In that case, maybe the test type could be "misc" and could also host 

https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/security/virt_what_cvm.py

I am not sure about this. Just mentioning it for consideration.

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: I honestly don't remember why I chose partition - it may have made more sense in an older version of this code? So split works well.

rh-jugraham:
> Considering that the SUT here is not libvirt but Leapp in KVM guests, you could also consider creating a new test type and moving these files there. E.g. "tp-libvirt/ipu/tests/(cfg|src)/..." How to do this: #3923
> 
> Maybe it would be too much effort to just run a single test scenario. I don't know. In that case, maybe the test type could be "misc" and could also host
> 
> https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/security/virt_what_cvm.py
> 
> I am not sure about this. Just mentioning it for consideration.

I wasn't sure as, as you mention, it is just a single test and it may be more effort than it is worth. Given that not many test types currently exist, having one that is just for one test feels out of place to me? I do like the idea of adding other tests to this test type (to both justify it better as well as clean up the repo more) but given that doing so 1) is out of the scope of this specific test and 2) may have some unintended side effects with tests being moved around, my initial reaction is to perhaps keep this test as it is but after this is merged, consider moving this test along with others into a new test type.

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: I used vm_arch_name (which is just from params) but good call, it is easy to forget about other architectures

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: Could do so. 

Do you know how I could get more information about kernel naming conventions/differences between architectures? Since I am used to aarch64, and this code currently only runs for aarch64, I don't know how many assumptions I am making that would not translate well onto other architectures (besides of course page size being an arm specific thing).

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: So I am not sure if this is what you are referring to, but when running this test, there is the following line in the debug log after the test has completed:

`avocado.virttest.storage storage          L1189 DEBUG| Copying /var/lib/avocado/data/avocado-vt/images/jeos-27-aarch64.qcow2.backup -> /var/lib/avocado/data/avocado-vt/images/jeos-27-aarch64.qcow2`

There are also the following options already set by default, as I can see in the first few lines of the debug.log, which I am assuming are responsible for the above code:
```
backup_image_before_testing = yes
restore_image_after_testing = yes
```

So I believe that the VM is being restored (though not obvious from my implementation as this relies on defaults)? I can rerun this test multiple times and everything seems fine.

rh-jugraham:
@smitterl Thanks for the thorough review! I added a few questions to the comments that I was uncertain of

libvirt/tests/src/in_place_upgrade_guest.py
smitterl: Yes, this is what I'm referring to bug apparently there's also some confusion on both sides because there seem to be several parameters with similar names. Sorry in advance for the long reply.

`backup_image`
-----------------------

`backup_image` is set to `no` per default in `base.cfg` our CI doesn't modify this.
On review, I noticed this parameter is apparently not used everywhere
```
# Enable backup_image = yes only in some specific tests, such as
#    unattended_install. In all other tests, it should be no, so keep the global
#    as is.
```
and actually I have not found any occurrence of `backup_image'` or `backup_image"` in projects avocado, avocado-vt, tp-qemu or tp-libvirt - maybe it's actually obsolete and should be removed?

`backup_image_before_test`
-----------------------------------------
Available in config section `vt.setup` section of file `avocado_vt/conf.d/vt.conf`
I can't see the default value. But on one hand, we can assume it's `False` because the sample configuration implies to comment out the `True` option:
```
[vt.setup]
# Backup image before testing (if not already backed up)
#backup_image_before_test = True
# Restore image after testing (if backup present)
#restore_image_after_test = True
# Keep guest running between tests (faster, but unsafe)
#keep_guest_running = False
```
Another indication is that in our CI project files you'll find that it's set explicitly to `True`, although likely not effective (because it's not commented out).

`backup_image_before_testing`
---------------------------------------------
For our CI aarch64 job configurations I see that the `avocado_vt/options.py` is updated to set `backup_image_before_testing` to `no`.

So at this point, I'm really very confused about how your environment got those values. Digging further I see that

`avocado_vt/options.py` per default sets `yes` for this parameter IF `vt.config` section wasn't defined but `vt.setup.backup_image_before_test` is `True`

```
    def _process_restore_image(self):
        if not get_opt(self.config, "vt.config"):
            if get_opt(self.config, "vt.setup.backup_image_before_test"):
                self.cartesian_parser.assign("backup_image_before_testing", "yes")
            if get_opt(self.config, "vt.setup.restore_image_after_test"):
                self.cartesian_parser.assign("restore_image_after_testing", "yes")
```

but also that many tests apparently set this value explicitly.

Conclusion
----------------

You are right, env_process uses `backup_image_before_testing` to determine if to back up (same for the restore variants). So, you can likely leave things as they are for your current setup. Given however the above picture, I'd personally prefer your test confirms that `backup_image_before_testing` really is `yes` because the upgrade is too big of a change for me.

libvirt/tests/src/in_place_upgrade_guest.py
smitterl: I don't know the kernel naming conventions but I'm not sure if it's absolutely necessary. By using the proposed function from 

your code could have a switch:

```
if params.get("use_64k") == "yes":
    kernel_pattern += ".*64k"

kernel_version = utils_test.get_available_kernel_path(kernel_pattern)
update_vm_default_kernel(vm, kenrel_version, reboot=True, guest_arch_name=machine.platform())
```

smitterl:
Hi @juteman thanks for your patience, please check my replies

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: I understand now. I was assuming that the helper function would get the kernel path given the os release version, but I realized that I can simply get the kernel path(s) given any regex pattern

libvirt/tests/src/in_place_upgrade_guest.py
rh-jugraham: Oh, I understand now. I did not realize there were so many params with such similar names - maybe sometime in the future, cleaning up the params would be a helpful task. Since my test does rely on backup_image_before_testing being yes and I don't know exactly where it is being set, I will also set it (and restore_image_after_testing) explicitly.

rh-jugraham:
@smitterl All updated!! Thanks again!

PR #6230, opened 2025-03-12
Fix path problem for nbdkit cve-starttls case

mxie91:


mxie91:
@xiaodwan pls help review, thanks

PR #6240, opened 2025-03-20
Remove dhclient from tests

chloerh:
dhclient is deprecated on RHEL 10 and will have to be removed from our
code base. Use avocado-vt functions instead of invoking it directly so
this can be handled from avocado-vt when needed.

### DIFF
--- a/libvirt/tests/src/virtual_network/update_device/update_iface_source.py
+++ b/libvirt/tests/src/virtual_network/update_device/update_iface_source.py
@@ -76,7 +76,7 @@ def run(test, params, env):
         vm_iface = utils_net.get_linux_ifname(session, mac)
         LOG.debug(f'Interface inside vm: {vm_iface}')
 
-        LOG.info('Run dhclient to refresh network.')
+        LOG.info('Run command to refresh network.')
         utils_net.restart_guest_network(session, mac)
 
         vm_iface_info = utils_net.get_linux_iface_info(
--- a/libvirt/tests/src/virtual_network/virtual_network_multivms.py
+++ b/libvirt/tests/src/virtual_network/virtual_network_multivms.py
@@ -23,26 +23,6 @@ from provider.virtual_network import network_base
 logging = log.getLogger('avocado.' + __name__)
 
 
-def create_bridge(br_name, iface_name):
-    """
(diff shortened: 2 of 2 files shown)


smitterl (changes requested):
Please add a commit message and PR description indicating why the reference to dhclient has been removed and you chose to use the utils_net function.

Rest LGTM, thanks

chloerh:
> Sorry, I hope I'm not too nitpicky but I'm not sure if I understand your commit message. Are you saying that the dhclient will not be available anymore for testing (in any of our available repositories)? What do you plan to substitute the `dhclient` command with? If there's already a commit in avocado-vt or some could maybe you can just point to that.

@smitterl 
Actually, I had been working on replacing dhclient in avocado-vt, but it's blocked. And I'm not working on it now.
We're going to replace dhclient eventually, and I replaced code of creating bridge with function of avocado-vt to keep consistency

smitterl:
> > Sorry, I hope I'm not too nitpicky but I'm not sure if I understand your commit message. Are you saying that the dhclient will not be available anymore for testing (in any of our available repositories)? What do you plan to substitute the `dhclient` command with? If there's already a commit in avocado-vt or some could maybe you can just point to that.
> 
> @smitterl Actually, I had been working on replacing dhclient in avocado-vt, but it's blocked. And I'm not working on it now. We're going to replace dhclient eventually, and I replaced code of creating bridge with function of avocado-vt to keep consistency

Still too nitpicky maybe but I think you should explain that in your commit message. Something like:
```
dhclient is deprecated on RHEL 10 and will have to be removed from our code base. Use avocado-vt functions instead of invoking it directly so this can be handled from avocado-vt when needed.
```

chloerh:
@smitterl updated, thank you for the suggestion 👍

smitterl (approved):

chunfuwen:
Can we have some basic test result available?

chloerh:
> Can we have some basic test result available?

 (1/1) virtual_network.multivms.macTableManager.linux_br.set_libvirt: STARTED
 (1/1) virtual_network.multivms.macTableManager.linux_br.set_libvirt: PASS (157.31 s)

@chunfuwen

PR #6242, opened 2025-03-21
Update connectivity check of bridge type interface

chloerh:
Remove nwfilter from test

### DIFF
--- a/libvirt/tests/src/virtual_network/connectivity/connectivity_check_bridge_interface.py
+++ b/libvirt/tests/src/virtual_network/connectivity/connectivity_check_bridge_interface.py
@@ -2,9 +2,9 @@ import logging
 import re
 
 from avocado.utils import process
-
 from virttest import utils_misc
 from virttest import utils_net
+from virttest import libvirt_version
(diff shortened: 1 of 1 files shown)


smitterl (changes requested):
Please indicate in your commit message and PR description the reason why the nwfilter is removed.
Also, the pipeline seems to need some fix.

Rest LGTM, thanks.

chloerh:
@smitterl Updated with bug id.

chloerh:
> Thanks, I hope I'm not too nitpicky but could you add to your commit message that it's removed because "the nwfilter driver makes exensive use of ebtables in addition to iptables, but both of those have been deprecated, and in RHEL10 they've started removing parts of those packages with the intent of eventually removing them completely." and that therefore, the nwfilter check should not be executed for now because it's currently broken on recent RHEL versions?
> 
> This way in the future we don't have to read the bug report to understand this.

@smitterl  Actually, I'm not quite familiar with the change. I'll invite @yanqzhan to help

smitterl (approved):
Thank you!

PR #6260, opened 2025-04-03
virtio_page_per_vq: fix var reference

dzhengfy:


dzhengfy:
Before the fix:
virtio_attributes.virtio_page_per_vq.input.passthrough.default_start
Error: local variable 'disk_image_path' referenced before assignment

This was due to the following issue in the test environment and  disk_image_path did not get correct initialization.
```
 Running 'ls /dev/input/event*'
2025-03-07 11:02:42,204 avocado.utils.process process          L0473 DEBUG| [stderr] ls: cannot access '/dev/input/event*'
2025-03-07 11:02:42,204 avocado.utils.process process          L0473 DEBUG| [stderr] : No such file or directory
```

Yingshun (approved):

PR #6271, opened 2025-04-08
update msg for virsh hypervisor_cpu_compare with virsh capa

nanli1:
hypervisor_cpu_compare with virsh capabilities result are not
recommanded in Bug xxxx-79460 

virsh.hypervisor_cpu_compare.cpu_xml.f_capa_xml
virsh.hypervisor_cpu_compare.capa_xml


smitterl (commented):
I also don't understand your cmmit message

>  hypervisor_cpu_compare with virsh capabilities result are not
recommanded

Please update. IIUC, you update the test cases for when the host cpu is used for comparison, that there's an error message indicating in those cases the comparison might lead to incorrect results?

PR #6272, opened 2025-04-09
filesystem_device: add data collection on failing virtiofs usage

smitterl:
Sometimes the mount command or other commmands would time out. In this case, log some information about the processes and retry. If the retry worked we still raise a test error because we expect it on the first try and want to inspect the logged information.

### DIFF
--- a/libvirt/tests/src/virtiofs/virtiofs.py
+++ b/libvirt/tests/src/virtiofs/virtiofs.py
@@ -73,6 +73,34 @@ def run(test, params, env):
         1.Mount dir in guest;
         2.Write a file in guest;
         3.Check the md5sum value are the same in guests and host;
+
+        :param vm_names: list of the VM names in the test
+        :param fs_devs: list of device XML instances
+        """
+        try:
+            _shared_data(vm_names, fs_devs)
+        except ShellTimeoutError as e:
+            logging.error(f"Issue using virtiofs {e}")
+            logging.error("Collect more information and retry")
+            for cmd in [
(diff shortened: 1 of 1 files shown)


libvirt/tests/src/virtual_device/filesystem_device.py
```diff
@@ -68,6 +69,34 @@ def shared_data(vm_names, fs_devs):
...
+        except ShellTimeoutError as e:
+            logging.error(f"Issue using virtiofs {e}")
+            logging.error("Collect more information and retry")
+            for cmd in [
+                "ps aux|grep virtiofsd",
+                "ps aux|grep qemu-kvm",
+                "ls /var/tmp|grep mount",
+                "cat /var/log/libvirt/qemu/*virtiofsd.log"
+            ]:
+                logging.error(utils_misc.cmd_status_output(cmd)[1])
```
iccaszhulili, line 87: Hi, Sebas

If we logging error, this case will still pass if retry mount is successful, right? Correct me if I am wrong. If it is the case, can we get the notification in jira issue?

smitterl: No, the test will still fail but we will have more information, see further below
```
            logging.error("Seems retrying succeeded")
            raise e
```

iccaszhulili:
LGTM

iccaszhulili (approved):

chunfuwen (approved):
lgtm

PR #6273, opened 2025-04-10
Scsi_device: automate 1 new case for rawio attribute

meinaLi:
Automate case:
VIRT-302169 - [virtual disks] [SCSI]Start vm with block lun disk(scsi-block) and rawio attribute


smitterl (approved):
LGTM, thank you

chunfuwen (approved):
lgtm

PR #6283, opened 2025-04-18
guest_kernel_debugging: Add case about pstore device

cliping:
XXX-303982 - [pstore][acpi-erst] start a guest with pstore device

### DIFF
--- a/libvirt/tests/src/guest_kernel_debugging/pstore.py
+++ b/libvirt/tests/src/guest_kernel_debugging/pstore.py
@@ -1,0 +1,65 @@
+from aexpect import ShellProcessTerminatedError
+from aexpect import ShellTimeoutError
+
+from avocado.utils import process
+
+from virttest import libvirt_version
+from virttest.libvirt_xml import vm_xml
+from virttest.utils_libvirt import libvirt_vmxml
+
+
+def run(test, params, env):
+    """
+    1. Start a guest with pstore device
+    2. Check qemu-cmdline
+    3. Login to guest and change kernel parameters
+    4. Trigger a crash
+    5. Check the log in /var/lib/systemd/pstore
+    """
+    vm_name = params.get("main_vm", "avocado-vt-vm1")
+    vm = env.get_vm(vm_name)
+    pstore_dict = eval(params.get("pstore_dict", "{}"))
+    qemu_cmdline_str = params.get("qemu_cmdline_str")
+    pstore_path = params.get("pstore_path")
+
+    libvirt_version.is_libvirt_feature_supported(params)
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    backup_xml = vmxml.copy()
+
+    try:
+        libvirt_vmxml.modify_vm_device(vmxml, 'pstore', pstore_dict)
+        vm.start()
+
+        # Check qemu-cmdline
+        ret = process.run(f"ps axu | grep {vm_name} | grep erst", shell=True, verbose=True).stdout_text.strip()
+        if qemu_cmdline_str not in ret:
+            test.fail(f"Not found '{qemu_cmdline_str}' in qemu cmdline.")
+
+        vm_session = vm.wait_for_login()
+        # Cleanup pstore path
+        vm_session.cmd(f"rm -rf {pstore_path}/*")
+        # Set kernel parameters
+        vm_session.cmd("echo Y > /sys/module/printk/parameters/always_kmsg_dump")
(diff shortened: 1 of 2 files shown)


cliping:
Test result:
 (1/1) type_specific.io-github-autotest-libvirt.guest_kernel_debugging.pstore: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.guest_kernel_debugging.pstore: PASS (77.43 s)

cliping:
Depend on: https://github.com/avocado-framework/avocado-vt/pull/4108

libvirt/tests/src/guest_kernel_debugging/pstore.py
```diff
@@ -0,0 +1,65 @@
...
+        # Set kernel parameters
+        vm_session.cmd("echo Y > /sys/module/printk/parameters/always_kmsg_dump")
+        vm_session.cmd("echo Y > /sys/module/kernel/parameters/crash_kexec_post_notifiers")
+
+        try:
+            # Trigger a crash
+            vm_session.cmd("echo c > /proc/sysrq-trigger", timeout=3)
+        except (ShellTimeoutError, ShellProcessTerminatedError):
+            pass
+        vm_session.close()
```
smitterl, line 51: If any other error occurs the sessoin won't be closed. You can handle this usually by defining `vm_session = None` at the top and in the `finally clause`
```
if vm_session:
    vm_session.close()
```

libvirt/tests/src/guest_kernel_debugging/pstore.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Check log
+        vm.destroy()
```
smitterl, line 54: It seems you expect the VM to crash, why would you destroy it? If this is not necessary please remove it.

libvirt/tests/src/guest_kernel_debugging/pstore.py
```diff
@@ -0,0 +1,65 @@
...
+        vm.start()
+        new_session = vm.wait_for_login()
+        ret = new_session.cmd_output(f"ll {pstore_path}")
+        if "dmesg-erst" not in ret:
+            test.fail("Not found dmesg file in {pstore_path}")
+        new_session.close()
```
smitterl, line 60: My previous comment about `vm_session` applies to `new_session`, too. Actually, instead of dealing with two differently named sessions I think you could just reuse the variable name after closing the first above, this way you only have to check one session parameter in the `finally` clause.

smitterl (changes requested):
Please check my comments, thank you

libvirt/tests/src/guest_kernel_debugging/pstore.py
```diff
@@ -0,0 +1,65 @@
...
+
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    backup_xml = vmxml.copy()
+
+    try:
+        libvirt_vmxml.modify_vm_device(vmxml, 'pstore', pstore_dict)
+        vm.start()
+
+        # Check qemu-cmdline
+        ret = process.run(f"ps axu | grep {vm_name} | grep erst", shell=True, verbose=True).stdout_text.strip()
```
dzhengfy, line 35: We usually use `utils_test/libvirt.py:4150:def check_qemu_cmd_line()` to check qemu command line.

libvirt/tests/src/guest_kernel_debugging/pstore.py
dzhengfy: "echo c > /proc/sysrq-trigger" will trigger a vm crash. So we have to destroy the vm here, I think @smitterl

libvirt/tests/src/guest_kernel_debugging/pstore.py
```diff
@@ -0,0 +1,65 @@
...
+        except (ShellTimeoutError, ShellProcessTerminatedError):
+            pass
+        vm_session.close()
+
+        # Check log
+        vm.destroy()
+        vm.start()
+        new_session = vm.wait_for_login()
+        ret = new_session.cmd_output(f"ll {pstore_path}")
+        if "dmesg-erst" not in ret:
```
dzhengfy, line 58: For better debugging, I suggest we also print out the pstore file list

libvirt/tests/src/guest_kernel_debugging/pstore.py
dzhengfy, line 51: I suggest from 39 to 51 should be put into `try: except: finally: `block to ensure any unexpected errors in vm_session.cmd and close could be caught and threw out if you want to use seperate vm sessions. Otherwise. let's handle all execeptions in the last finally clause.

PR #6297, opened 2025-04-27
numa_config_with_auto_placement:fix incorrect iothreadid

dzhengfy:
This case has no iothreadid setting, so previous fix is incorrect. The iothread id is ordered from 1. So this patch is to use new way to count the iothread id.


dzhengfy:
Before fix:
 (1/1) type_specific.io-github-autotest-libvirt.numa_config_with_auto_placement.default: FAIL: Unexpected failure during the test: Error in Getter for property "iothreadids", element tag "iothreadids" not found on parent at xpath "/" in XML\n<domain xmlns:ns0="http://libosinfo.org/xmlns/libvirt/domain/1.0" type="kvm" id="2">\n  <name>avocado-vt-vm1</... (80.17 s)

After fix:
 (1/1) type_specific.io-github-autotest-libvirt.numa_config_with_auto_placement.default: PASS (69.62 s)

PR #6314, opened 2025-05-15
libvirtd: ensure older libvirt version is correctly found

rh-jugraham:
Addressed edge cases that caused the older libvirt version to not be found correctly.

**Issues**

1. `\s+(\d+.*\d+)\s+\w+` will also match "Last metadata expiration check: 1:11:34 ago on Thu 15 May 2025 03:16:53 PM EDT." if included in the `dnf list libvirt --showduplicates --available` output. The new regex - `\s+(\d+[^:\s]*\.[^:\s]*\d+)\s+` - will now ensure that between the digits, there may be any characters except a white space character or a `:`. To ensure that simple numbers are not also matched (ex: 2025), also ensured that at least 1 `.` is included (as found in versions).

2. checking if `one_version < installed_version` works in most cases, but not all. For example, if the `installed_version` is "10.10.0-10" but the `one_version` is either "10.10.0-8" or "10.9.0-10", then "10.10.0-10" will be incorrectly considered less than, since 1 is less than 8 or 9. Therefore, LooseVersion is used instead. edit: see comments, `rpmdev-vercmp` used instead.

**Passing Test**

```
(.libvirt-ci-venv-ci-runtest-VOkL3B) [root@nvidia-jetson-xavier-nx-02 ~]# avocado run --vt-type libvirt --vt-omit-data-loss --vt-machine-type arm64-mmio libvirtd.libvirtd.upgrade_with_legacy_mode --vt-connect-uri qemu:///system
JOB ID     : 462e4e5c73cb78fe1247531ce5b045d883a5306a
JOB LOG    : /var/log/avocado/job-results/job-2025-05-15T17.38-462e4e5/job.log
 (1/1) type_specific.io-github-autotest-libvirt.libvirtd.libvirtd.upgrade_with_legacy_mode: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.libvirtd.libvirtd.upgrade_with_legacy_mode: PASS (243.78 s)
JOB HTML   : /var/log/avocado/job-results/job-2025-05-15T17.38-462e4e5/results.html
JOB TIME   : 247.27 s
```

### DIFF
--- a/libvirt/tests/src/daemon/libvirtd/libvirtd.py
+++ b/libvirt/tests/src/daemon/libvirtd/libvirtd.py
@@ -4,6 +4,7 @@ import logging as log
 import os.path
 import re
 import threading
+import shutil
 
 import queue
 
@@ -96,11 +97,16 @@ def get_installable_old_libvirt(installed_version, test):
                           shell=True,
                           ignore_status=False,
                           verbose=True)
-    versions = re.findall(r"\s+(\d+.*\d+)\s+\w+", cmd_ret.stdout_text.strip())
+    versions = re.findall(r"\s+(\d+[^:\s]*\.[^:\s]*\d+)\s+", cmd_ret.stdout_text.strip())
     versions = versions[::-1]
     old_version = None
+    if not shutil.which('rpmdev-vercmp'):
+        test.cancel("rpmdevtools package is not installed on the host")
     for one_version in versions:
-        if one_version < installed_version:
+        cmd = 'rpmdev-vercmp %s %s' % (one_version, installed_version)
+        cmd_ret = process.run(cmd, shell=True, ignore_status=True)
+        # Exit status is 0 if the EVR's are equal, 11 if EVR1 is newer, and 12 if EVR2
+        if cmd_ret.exit_status == 12:
             old_version = one_version
             break
     if not old_version:


rh-jugraham:
@dzhengfy Could you review this PR? Thanks!

libvirt/tests/src/daemon/libvirtd/libvirtd.py
```diff
@@ -4,9 +4,10 @@
 import os.path
 import re
 import threading
-
 import queue
 
+from distutils.version import LooseVersion  # pylint: disable=E0611
```
dzhengfy, line 9: I saw distutils will be removed from python 3.12. There is a risk to use it. Could you try other library?

rh-jugraham: Thanks for the information - I need to look into another library. I used this one since 1) it is already used by provider/v2v_vmcheck_helper.py and 2) the updated packaging.version.Version unfortunately does not allow '-' in versions. So need to do some more research for a good alternative.

rh-jugraham: @dzhengfy updated!

libvirt/tests/src/daemon/libvirtd/libvirtd.py
```diff
@@ -96,11 +96,15 @@ def get_installable_old_libvirt(installed_version, test):
                           shell=True,
                           ignore_status=False,
                           verbose=True)
-    versions = re.findall(r"\s+(\d+.*\d+)\s+\w+", cmd_ret.stdout_text.strip())
+    versions = re.findall(r"\s+(\d+[^:\s]*\.[^:\s]*\d+)\s+", cmd_ret.stdout_text.strip())
     versions = versions[::-1]
     old_version = None
+    utils_package.package_install("rpmdevtools")
```
dzhengfy, line 102: We do not install any new package on the host for the test to use. For image mode, it also can not install successfully.  Instead, we will update the job definition to install the package by CI in advance. 
Here, we just check if the package is available on the host.

dzhengfy (commented):
Others LGTM

rh-jugraham:
@dzhengfy Since the MR to install rpmdevtools has been merged, I believe this PR is ready for a final review?

PR #6328, opened 2025-05-22
passt: Replace the test file dir '/' with '/var/tmp'

qiankehan:
For RHEL image mode, the directory '/' is read-only. It will cause the case failures when trying to write a file on it. Fix it with the writable directory '/var/tmp'.


qiankehan:
@dzhengfy It is a tiny fix. Please help review it

chloerh (approved):

PR #6330, opened 2025-05-25
passt: support multiple default routes

dzhengfy:
In some environment, there are more than one default routes in ip route output, for example on ARM host.
This patch is to support this situation instead of only selecting the first default route which is not accurate.

For example:
"nexthops":[{"gateway":"fe80::4a5a:d00:6631:2920","dev":"eno1","weight":1,"flags":[]}, {"gateway":"fe80::4a5a:d00:6631:b20","dev":"eno1","weight":1,"flags":[]}

### DIFF
--- a/libvirt/tests/cfg/virtual_network/passt/passt_attach_detach.cfg
+++ b/libvirt/tests/cfg/virtual_network/passt/passt_attach_detach.cfg
@@ -5,6 +5,7 @@
     outside_ip = 'www.redhat.com'
     start_vm = no
     mtu = 65520
+    multiple_nexthops = no
     variants user_type:
         - root_user:
             test_user = root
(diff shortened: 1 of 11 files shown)


dzhengfy:
Depend on
- https://github.com/avocado-framework/avocado-vt/pull/4131

Yingshun (commented):
May I ask why there are 2 default gateway? You have modified many cases here, so is this a problem caused by normal behavior? and what behavior caused it? after having 2 default gateway, is the vm's connectivity still working?

libvirt/tests/cfg/virtual_network/passt/passt_attach_detach.cfg
```diff
@@ -5,6 +5,7 @@
     outside_ip = 'www.redhat.com'
     start_vm = no
     mtu = 65520
+    multiple_nexthops= no
```
Yingshun, line 8: what about adding a space before '='?

Yingshun (approved):

chloerh (approved):

PR #6336, opened 2025-05-28
Scsi_device: add new queues test to scsi controller

meinaLi:
Automate new test scnearios in case:
RHEL-137060 - Attach/detach SCSI controller with driver configured should be supported


smitterl (approved):
LGTM Thanks

PR #6340, opened 2025-05-29
nbdkit: fix the value which is supported after 1.40

vwu-vera:
the value:10s is supported after 1.40. So remove this value for 10.1.

DELAY FORMAT
In nbdkit ≥ 1.40, [nbdkit_parse_delay(3)](https://libguestfs.org/nbdkit_parse_delay.3.html) is used to parse the delay parameters, so a wide variety of formats are accepted, such as:

delay-read=1.5
delay-write=100ms
delay-zero=1.2μs
delay-trim=1000ns
In nbdkit ≤ 1.38, only a whole number of seconds, or a whole number of milliseconds (if suffixed by ms) were allowed.


vwu-vera:

PR #6347, opened 2025-06-03
nbdkit: fix the failed case with the version issue

vwu-vera:
the package nbdkit on RHEL10 + has more than one rpm dependencies and it caused the rpmbuild failed.

Case on rhel9 can be passed successfully. 

Adding version control to skip this source pkg check on rhel10+.


vwu-vera:
@xiaodwan  please help to review and merge. Thanks.:)

PR #6355, opened 2025-06-05
migration: Fix decode failure

cliping:
Fix following issue:
    'utf-8' codec can't decode byte 0xff in position 1465040: invalid start byte

The ignoring will not impact some key strings matched in the log.


chunfuwen (approved):
lgtm

PR #6379, opened 2025-06-17
Snapshot revert: fix VM is dead reason: Domain avocado-vt-vm1 is inactive detail: 'shut off'

BulaYoungR:


dzhengfy:
@BulaYoungR 
According to libvirt(11.3.0) man page:
```
Normally, reverting to a snapshot leaves the domain in the state it was at the time the snapshot was  created,  ex‐
       cept that a disk snapshot with no vm state leaves the domain in an inactive state.  Passing either the --running or
       --paused  flag  will perform additional state changes (such as booting an inactive domain, or pausing a running do‐
       main).
```
So I suggest we check if the snapshot is with disk-only, then we do the snapshot revert  with --running option instead of starting the vm explicitely. What do you think of?

libvirt/tests/src/snapshot/revert_disk_external_snap.py
```diff
@@ -30,7 +30,13 @@ def revert_snap_and_check_files(params, vm, test, snap_name, expected_files):
     :param snap_name: snapshot name.
     :param expected_files: expected file list.
     """
-    virsh.snapshot_revert(vm.name, snap_name, **virsh_dargs)
+    snap_type = params.get("snap_type")
+    revert_cmd_str = snap_name
+
+    if snap_type == "disk_only" and libvirt_version.version_compare(10, 10, 0):
+        revert_cmd_str += " --running"
```
dzhengfy, line 37: `def snapshot_revert(name, snapshot, options="", **dargs):`  in virsh.py
You can directly set "--running" for `options`

BulaYoungR: I changed it in my new commit

PR #6382, opened 2025-06-18
passt_function: Allow multiple nexthops

qiankehan:
The cases will failed with error "FAIL: Host default ipv6 gateway not consistent with vm" for the host with multiple nexthops in default route like: default proto ra metric 100 pref medium
	nexthop via fe80::4a5a:d01:f631:3320 dev eno1 weight 1
	nexthop via fe80::4a5a:d01:f631:2420 dev eno1 weight 1

Allow multiple nexthops to fix the failures.


dzhengfy:
@qiankehan please also provide another case test result.

PR #6391, opened 2025-06-20
Fix the avc inproper AVC check

yalzhang:
This test script checks for any AVC denials during test execution. However, it doesn't clear the audit log history before running, which causes the test to always fail. The script has been updated to check timestamps and only report a failure if AVC denials occur within the test start and end time window.


cliping (approved):
LGTM

chloerh (approved):

PR #6396, opened 2025-06-23
v2v:fix the failed no-shutdown case

vwu-vera:
The error message is changed after 9.7/10.1. So modify the msg_content accordingly.


vwu-vera:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.negative_test.not_shutdown.running.default.kvm.output_mode.local.it_default: PASS (50.32 s)

PR #6399, opened 2025-06-24
Start vm with --bypass-cache after managedsave

yalzhang:
Automate case VIRT-297730. There are 2 scenarios: start vm with '--bypass-cache' option, or set 'auto_start_bypass_cache = 1' in qemu.conf and check the function of it.

### DIFF
--- a/libvirt/tests/cfg/save_and_restore/start_with_bypass_cache.cfg
+++ b/libvirt/tests/cfg/save_and_restore/start_with_bypass_cache.cfg
@@ -1,0 +1,10 @@
+- save_and_restore.start_with_bypass_cache:
+    type = start_with_bypass_cache
+    check_cmd = "while(true); do [ -e {} ] && cat /proc/$(lsof -w {}|awk '/libvirt_i/{{print $2}}')/fdinfo/0 ;done"
+    save_path = "/var/lib/libvirt/qemu/save/"
+    variants scenario:
+        - by_opt:
+            start_option = --bypass-cache
+        - by_qemu_conf:
+            set_qemu_conf = yes
+            auto_file_path = '/run/libvirt/qemu/autostarted'
--- a/libvirt/tests/src/save_and_restore/start_with_bypass_cache.py
+++ b/libvirt/tests/src/save_and_restore/start_with_bypass_cache.py
@@ -1,0 +1,113 @@
+import logging
+import os
+import re
+
+from avocado.utils import process
+from avocado.utils import software_manager
+from virttest import utils_config
+from virttest import utils_libvirtd
+from virttest import virsh
+from virttest.libvirt_xml import vm_xml
+
+from provider.save import save_base
+
+LOG = logging.getLogger('avocado.test.' + __name__)
+VIRSH_ARGS = {'debug': True, 'ignore_status': False}
+
+
+def setup_bypass_cache(check_cmd, test):
+    """
+    Setup for restore --bypass-cache option
+
+    :param check_cmd: command to check bypass cache
+    :param test: test instance
+    :return: subprocess instance to run check command
+    """
+    sm = software_manager.manager.SoftwareManager()
+    if not sm.check_installed('lsof'):
+        test.error("Need to install lsof on host")
+    sp = process.SubProcess(check_cmd, shell=True)
+    sp.start()
+
+    return sp
+
+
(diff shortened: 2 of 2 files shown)


libvirt/tests/cfg/save_and_restore/start_with_bypass_cache.cfg
```diff
@@ -0,0 +1,9 @@
+- save_and_restore.start_with_bypass_cache:
+    type = start_with_bypass_cache
+    check_cmd = "while(true);do cat /proc/$(lsof -w /var/lib/libvirt/qemu/save/avocado-vt-vm1.save|awk '/libvirt_i/{print $2}')/fdinfo/0 ;done"
```
chunfuwen, line 3: avocado-vt-vm1.save is hard string. Vm name may be not this one.
Moreover, it looks like this is infinite loop, do you need to consider tertiniate it in some conditions?

yalzhang: Updated, Thank you!  For the loop, updated to be with " do [ -e {} ] &&", with next step to start the vm, the file will be deleted to close the loop.

libvirt/tests/src/save_and_restore/start_with_bypass_cache.py
```diff
@@ -0,0 +1,121 @@
...
+        # Setup and start checking bypass cache
+        sp = setup_bypass_cache(check_cmd, test)
+        LOG.info("TEST STEP 2: Start restore process")
+
+        if set_qemu_conf:
+            LOG.info('Start libvirtd to trigger VM startup')
+            libvirtd.start()
+        else:
+            # Run restore with "--bypass-cache" option
+            virsh.start(vm_name, "--bypass-cache", debug=True, ignore_status=False)
```
nanli1, line 107: hi yalan, I find you have defined VIRSH_ARGS = {'debug': True, 'ignore_status': False}
 ,could VIRSH_ARGS  be used to replaced `debug=True, ignore_status=False ` here

yalzhang: Updated, Thank you!

nanli1 (approved):

libvirt/tests/src/save_and_restore/start_with_bypass_cache.py
```diff
@@ -0,0 +1,113 @@
...
+        LOG.info("TEST STEP 4: Ensure VM is running after restore")
+        if not vm.is_alive():
+            test.fail(f"VM should be running after restore, instead of {vm.state()}")
+        # Compare VM info with the state before managedsave
+        LOG.info("TEST STEP 5: Verify VM time and ping status on VM")
+        save_base.post_save_check(vm, pid_ping, upsince)
+
+    finally:
+        bkxml.sync()
+        qemu_conf.restore()
```
yalzhang, line 113: gemini-cli-reviewer:
Overall, the test case is well-structured and covers the two scenarios described in the PR. The use of a `finally` block for cleanup is good. I have a few suggestions to improve the robustness and efficiency of the test.

1.  **`check_cmd` fragility and inefficiency**: The `check_cmd` in `start_with_bypass_cache.cfg` uses a `while(true)` loop without a `sleep`, which can cause high CPU usage. It's also a bit fragile as it assumes `lsof` returns a single, valid PID and hardcodes file descriptor `0`. It would be more robust to parse the file descriptor from `lsof` output (4th column) and add error handling for the `lsof` command. Consider adding a `sleep` to the loop.

2.  **Potential for hanging**: In `check_bypass_cache`, `sp.get_stdout().decode()` will block indefinitely because the subprocess runs an infinite loop. This could cause the test to hang. It would be safer to use a read with a timeout, like `sp.read_stdout(timeout=60)`, and then terminate the subprocess. This ensures the test will eventually continue or fail even if no output is produced.

3.  **Clarity**: The `check_cmd` is quite complex. It might be clearer to implement the logic for finding the process and checking the file descriptor flags within the Python code itself, perhaps in the `check_bypass_cache` function. This would make the test easier to read, debug, and maintain. For example, you could use the `psutil` library to find the process and inspect its open files.

yalzhang (commented):
I have reviewed the pull request and left some comments. Overall, the changes look good, but there are a few areas that could be improved for robustness and efficiency.

PR #6427, opened 2025-07-06
tpm_device: check vtpm support in domcapabilities

yanqzhan:
virsh domcapabilities should be able to expose vtpm info, such as
version, model, backend type, should be same as qemu-kvm or swtpm
supported. There's a bug RHEL-102057 in which the return code of
'qemu-kvm -tpmdev help' is 1, so collect stderr instead of stdout.

### DIFF
--- a/libvirt/tests/src/virtual_device/tpm_device.py
+++ b/libvirt/tests/src/virtual_device/tpm_device.py
@@ -17,6 +17,7 @@ from virttest import utils_misc
 from virttest import utils_libguestfs
 from virttest import utils_libvirtd
 
+from virttest.libvirt_xml import domcapability_xml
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_device/tpm_device.py
```diff
@@ -34,6 +35,39 @@
 logging = log.getLogger('avocado.' + __name__)
 
 
+def get_supported_values():
+    """
+    Get supported vtpm attr values from qemu or swtpm.
+    :return: expected vtpm attr values
+    """
+    out_m = process.run('/usr/libexec/qemu-kvm -device help').stdout_text.strip()
```
chunfuwen, line 43: /usr/libexec/qemu-kvm may diff on different arch. Can be get it dynamically?

PR #6448, opened 2025-07-14
base_steps: reload firewall on source host during cleanup

rh-jugraham:
Reload firewall on source host during `cleanup_connection()` if migrate_desturi_port is set. Although this is technically unnecessary, as any test that modifies the firewall on the source host should clean it up within the test, in edge cases where the test fails early and cleanup isn't finished, this should ensure that the firewall is reset on both the target and the source.


rh-jugraham:
@dzhengfy ready for review. Let me know if you have any suggestions regarding this issue.

chunfuwen:
please upload test result as well

PR #6452, opened 2025-07-15
in_place_upgrade_guest: add LEAPP_DEVEL_SKIP_CHECK_OS_RELEASE variable

rh-jugraham:
From RHEL9.7, we need to keep the original upgrade-paths until devel-freeze, because it could happen the package could be released in advance when it's required, and such a released package must not officially support IPU form 9.7 to 10.1.


rh-jugraham:
@dzhengfy @iccaszhulili ready for review. While this could be added as --vt-extra-params within /virt/CUT/Sanity/leapp_guest_upgrade, I believe it makes more sense to add it within the test code itself here as this is where the command is actually being run. But I am open to alternatives.

iccaszhulili (approved):

iccaszhulili:
> @dzhengfy @iccaszhulili ready for review. While this could be added as --vt-extra-params within /virt/CUT/Sanity/leapp_guest_upgrade, I believe it makes more sense to add it within the test code itself here as this is where the command is actually being run. But I am open to alternatives.

Hi, Julia, I also think this is acceptable to add code here. LGTM

PR #6455, opened 2025-07-17
Virtual_disk: automate new case for the nested datastore element

meinaLi:
Automate new case:
VIRT-304570 - [virtual disks][dataStore] Edit and start guest with disk with nested <dataStore> and <backingStore>

### DIFF
--- a/libvirt/tests/src/virtual_disks/virtual_disks_datastore.py
+++ b/libvirt/tests/src/virtual_disks/virtual_disks_datastore.py
@@ -1,5 +1,7 @@
 import os
 
+from avocado.utils import process
+
 from virttest import data_dir
 from virttest import libvirt_version
 from virttest import utils_disk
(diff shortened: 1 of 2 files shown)


libvirt/tests/src/virtual_disks/virtual_disks_datastore.py
```diff
@@ -66,22 +71,58 @@ def check_result():
...
     vmxml_backup = vmxml.copy()
     disk_obj = disk_base.DiskBase(test, vm, params)
+    file_list = []
 
     try:
         test.log.info("TEST_STEP: Prepare the guest.")
         data_file = prepare_disk()
+        file_list.append(data_file)
         data_file_option = params.get("data_file_option") % data_file
+        if with_nested:
```
chunfuwen, line 90: maybe wrap line 91 to line 110 into one standalone method

meinaLi: Use a new function as mentioned.

chunfuwen (approved):
lgtm

PR #6470, opened 2025-07-30
aarch64_cpu_sve: fix sve check image mode

dzhengfy:
We need to check different path for kernel config in image mode


dzhengfy:
Before the fix:
```
rhel.aarch64_cpu_sve.boot_test.enable_sve (from avocado-vt)

Skip Message
Guest kernel doesn't enable CONFIG_ARM64_SVE
```
After the fix:
`[boot_test.enable_sve]	2 min 7 sec	Passed`

chloerh (approved):

PR #6471, opened 2025-07-30
Correct version format of two v2v cases

mxie91:


mxie91:
@xiaodwan please help review, thanks

Before fix:
function_test_esx.positive_test.windows.uefi.win2019.esx_70.libvirt.it_vddk.vpx_uri  failed with error 'Invalid version: '1.45.1-1.el9''
function_test_esx.positive_test.without_file_architecture.esx_70.libvirt.it_vddk.vpx_uri  failed with error 'Invalid version: '1.56.0-1.el10''

After fix:
 (1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.without_file_architecture.esx_70.libvirt.it_vddk.vpx_uri: PASS (564.80 s)
 (1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.windows.uefi.win2019.esx_70.libvirt.it_vddk.vpx_uri: PASS (497.11 s)

v2v/tests/src/function_test_esx.py
```diff
@@ -701,12 +701,6 @@ def vm_check(status_error):
                       (len(error_list), error_list))
 
     try:
-        # Added version_required condition due to different versions of v2v for RHEL9 and RHEL10
-        # RHEL9 V2V version - "[virt-v2v-2.7.1-12,)"
-        # RHEL10 V2V version - "[virt-v2v-2.8.1-1,)"
-        if "print_blkhash" in checkpoint and "el9" in process.run("uname -r", shell=True).stdout_text:
```
xiaodwan, line 707: The code here checks the required v2v version on both rhel9 and rhel10. 
Which error do you see?  
also cc. @ganeshhubale

mxie91: I have discussed this with Ganesh before, there is no need to differentiate between RHEL9 and RHEL10 versions, because RHEL10 version is usually higher than RHEL9 version, it is enough to set the minimum supported version for the case

xiaodwan: This is not true. For example, virt-v2v-2.8.0-1.el10 is greater than virt-v2v-2.7.1-12, but it's not supported the new feature.

xiaodwan: The v2v is not supported on rhel10.0 which is virt-v2v-2.7.1-4. so the change can work on every rhel release.

xiaodwan (approved):

PR #6479, opened 2025-08-04
Add the selinux_change_from parameter to all external variants

iccaszhulili:


dzhengfy (commented):
Could you give a simple explaination for the parameter?

PR #6506, opened 2025-08-13
sriov_attach_detach_device_vfio_variant_driver: enhance test

dzhengfy:
The ping test is better to have. But we do not want to skip the test for detaching device when ping test fails. On some hosts, there is no NIC with cable for the vm to use, so ping test will fail as expected.

### DIFF
--- a/libvirt/tests/src/sriov/plug_unplug/sriov_attach_detach_device_vfio_variant_driver.py
+++ b/libvirt/tests/src/sriov/plug_unplug/sriov_attach_detach_device_vfio_variant_driver.py
@@ -9,6 +9,8 @@ from virttest.libvirt_xml import vm_xml
 from virttest.utils_libvirt import libvirt_vfio
 from virttest.utils_libvirt import libvirt_vmxml
 
+from provider.sriov import sriov_vfio
+
 
 def run(test, params, env):
     """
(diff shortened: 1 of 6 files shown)


libvirt/tests/src/sriov/plug_unplug/sriov_attach_detach_device_vfio_variant_driver.py
```diff
@@ -43,7 +43,8 @@ def parse_iface_dict(pf_pci):
     test_pf = params.get("test_pf", "ens3f0np0")
     pf_pci = utils_sriov.get_pf_pci(test_pf=test_pf)
     iface_number = int(params.get("iface_number", "1"))
-    ping_dest = params.get("ping_dest", "www.redhat.com")
+    ping_dest = params.get("ping_dest")
+    is_ping_fail = False
```
Yingshun, line 47: Instead of setting a hardcode value, I think it's better to detect the network connection in the host if you want to make ping check flexible. Otherwise we need to change it manually. And also, even we don't have this change, we can also manually comment out ping check part. what do you think?

dzhengfy:
Depend on 
- https://github.com/autotest/tp-libvirt/pull/6507

libvirt/tests/src/sriov/plug_unplug/sriov_attach_detach_device_vfio_variant_driver.py
dzhengfy: Updated in https://github.com/autotest/tp-libvirt/pull/6507

PR #6510, opened 2025-08-14
Virtual_disk: update guest disk mountpoint to /mnt

meinaLi:
This PR is to fix the error "mkdir: cannot create directory /test: Read-only file system" in image mode.


chunfuwen (approved):
LGTM

PR #6523, opened 2025-08-20
Add one negative:no_start_qemu_guest_agent case into domhostname

chunfuwen:
This change adds a negative test case to the `domhostname` test. The new case, `no_start_qemu_guest_agent`, verifies that the `domhostname` command fails as expected when the QEMU guest agent is not running in the guest.

The test logic in `domhostname.py` is updated to handle this case by optionally preventing the guest agent from starting and then checking for the appropriate error message. The test configuration in `domhostname.cfg` is updated to include this new negative test variant.

### DIFF
--- a/libvirt/tests/cfg/guest_agent/domhostname.cfg
+++ b/libvirt/tests/cfg/guest_agent/domhostname.cfg
@@ -8,3 +8,8 @@
             domain_ref = "id"
         - domain_UUID:
             domain_ref = "uuid"
+        - negative:
+            status_error = "yes"
+            variants:
+                - no_start_qemu_guest_agent:
+                    no_start_qemu_ga = "yes"
--- a/libvirt/tests/src/guest_agent/domhostname.py
+++ b/libvirt/tests/src/guest_agent/domhostname.py
@@ -11,21 +11,32 @@ def run(test, params, env):
     vm_name = params.get("main_vm")
     vm = env.get_vm(vm_name)
     expr_hostname = params.get("expr_hostname", "myhostname")
+    status_error = params.get("status_error", "no") == "yes"
+    start_qemu_ga = params.get("no_start_qemu_ga", "no") == "no"
 
     vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     backup_xml = vmxml.copy()
 
     try:
-        vm.prepare_guest_agent()
+        vm.prepare_guest_agent(start=start_qemu_ga)
 
-        test.log.info(f"TEST_STEP: Set VM's hostname to {expr_hostname}.")
-        vm_session = vm.wait_for_login()
-        vm_session.cmd("hostname %s" % expr_hostname)
+        if not status_error:
+            test.log.info(f"TEST_STEP: Set VM's hostname to {expr_hostname}.")
+            vm_session = vm.wait_for_login()
+            vm_session.cmd("hostname %s" % expr_hostname)
+
         domain_ref = vm.get_id() if params.get("domain_ref") == 'id' \
-            else eval('vm.%s' % params.get("domain_ref"))
+            else eval('vm.%s' % params.get("domain_ref", "name"))
 
         test.log.info(f"TEST_STEP: Get VM's hostname via guest agent.")
-        res = virsh.domhostname(domain_ref, debug=True)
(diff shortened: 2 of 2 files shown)


libvirt/tests/cfg/guest_agent/domhostname.cfg
```diff
@@ -8,3 +8,8 @@
             domain_ref = "id"
         - domain_UUID:
             domain_ref = "uuid"
+        - negative:
+            status_error = "yes"
+            variants:
```
chunfuwen, line 13: The structure of this new `negative` block seems a bit off. Based on the typical structure of these configuration files, `negative` should likely be a top-level key at the same level as `variants`, not nested within it. Also, the `variants` key under `negative` seems unintentional.

I would suggest structuring it like this for clarity and consistency:

```yaml
- guest_agent.domhostname:
    type = domhostname

    variants:
        - domain_name:
            domain_ref = "name"
        - domain_ID:
            domain_ref = "id"
        - domain_UUID:
            domain_ref = "uuid"
    negative:
        - no_start_qemu_guest_agent:
            no_start_qemu_ga = "yes"
            status_error = "yes"
            domain_ref = "uuid" # Or name, as appropriate
```

libvirt/tests/src/guest_agent/domhostname.py
```diff
@@ -11,21 +11,32 @@ def run(test, params, env):
...
     vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     backup_xml = vmxml.copy()
 
     try:
-        vm.prepare_guest_agent()
+        vm.prepare_guest_agent(start=start_qemu_ga)
+
+        if not status_error:
+            test.log.info(f"TEST_STEP: Set VM's hostname to {expr_hostname}.")
+            vm_session = vm.wait_for_login()
```
chunfuwen, line 25: Using `eval()` can be a security risk as it can execute arbitrary code. While the input here is controlled by the config file, it's a best practice to avoid `eval()` when a safer alternative exists.

You can achieve the same result more safely by using `getattr()`:

```python
domain_ref_str = params.get("domain_ref", "name")
domain_ref = vm.get_id() if domain_ref_str == 'id' else getattr(vm, domain_ref_str)
```

chunfuwen (commented):
remove

chunfuwen (commented):
gemini-cli-reviewer: This pull request adds a useful negative test case for the `domhostname` command, which is a great addition.

I've left a few comments with suggestions for improvement:
- The structure of the `.cfg` file seems incorrect.
- The use of `eval()` in the Python script should be replaced with `getattr()` for better security and maintainability.
- A minor suggestion to improve the clarity of boolean parameter handling.

Overall, this is a good change. Once the feedback is addressed, this will be ready to merge.

libvirt/tests/src/guest_agent/domhostname.py
```diff
@@ -11,21 +11,32 @@ def run(test, params, env):
...
+    status_error = params.get("status_error", "no") == "yes"
+    start_qemu_ga = params.get("no_start_qemu_ga", "no") == "no"
 
     vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
     backup_xml = vmxml.copy()
 
     try:
-        vm.prepare_guest_agent()
+        vm.prepare_guest_agent(start=start_qemu_ga)
+
```
mxie91, line 22: gemini-cli-reviewer: For better security and maintainability, consider using `getattr()` instead of `eval()`.

```python
getattr(vm, params.get("domain_ref", "name"))
```

This avoids the potential risks associated with `eval()` and is generally considered a best practice in Python.

mxie91 (commented):
gemini-cli-reviewer: The changes for the new negative test case are correct and the test logic is properly updated. I've added one comment suggesting a minor improvement for maintainability. Overall, this is a good addition to the test suite.

chunfuwen (commented):
Thanks for your contribution. There are a couple of things that need to be addressed before this PR can be merged:

1. The structure of `domhostname.cfg` is incorrect. The `negative` block should be a top-level key.
2. The use of `eval()` in `domhostname.py` is a security risk and should be replaced with `getattr()`.

Please address these issues and I'll be happy to review the PR again.

PR #6529, opened 2025-08-29
iface_option: Fix int() argument must be a string, a bytes-like object or a number, not NoneType

BulaYoungR:
Add fallback parser for ethtool -l and guard against int(None) If utils_net.get_channel_info() returns empty, parse Combined from ethtool -l <iface> via regex (no header dependency). Use first Combined as max, second as current (fallback to first if only one). Explicitly error if Combined can’t be parsed.


dzhengfy:
@BulaYoungR please double check this problem still exists. if not, let's close this PR.

PR #6531, opened 2025-09-02
virtual_network: add case for pktgen with burst mode

nanli1:
This introduces test cases for pktgen burst mode functionality
in virtual network environments using virtio-net-pci devices. The tests
validate packet generation performance and reliability under different
burst configurations.

This change is to ensure comprehensive coverage of
network performance testing scenarios, particularly for high-throughput
use cases that require burst packet transmission capabilities.

Reference: XXXX-300599

```
 (1/1) Host_RHEL.m10.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.1.x86_64.io-github-autotest-libvirt.virtual_network.qemu_test.pktgen_burst_mode_test.q35: PASS (425.71 s)


```

## Summary by CodeRabbit

* **Tests**
  * Added a pktgen burst-mode performance test for virtual networks with configurable threads, packet size, burst and timeout; excludes Windows and targets virtio_net. Supports VM lifecycle control, firewall handling, OVMF with 8-queue NIC, adaptive package steps by distro/version, connectivity checks, crash verification, and structured result recording.

* **New Features**
  * Added end-to-end pktgen benchmarking utilities: configuration, command generation, test execution with MPPS calculation, timeouts, installation helpers, formatted results, and category-based test runners.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
+++ b/libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
@@ -1,0 +1,24 @@
+- virtual_network.qemu_test.pktgen_burst_mode_test:
+    type = pktgen_burst_mode_test
+    no Windows
+    only virtio_net
+    no Host_RHEL.m5, Host_RHEL.m6
+    start_vm = no
+    vm_ping_host = "pass"
+    # tx: Guest transmission performance to external network.
+    pkg_dir = "tx"
+    pktgen_script = "pktgen_sample03_burst_single_flow"
+    pktgen_threads = 1
+    pkt_size = 64
+    burst = 3
+    pktgen_test_timeout = 30
+    record_list = pkt_size run_threads burst mpps
+    guest_ver_cmd = "uname -r"
+    kvm_ver_chk_cmd = "rpm -qa qemu-kvm-rhev qemu-kvm"
+    test_vm = no
+    flush_firewall_rules_cmd = "iptables -F ; nft flush ruleset"
+    queues_nic1 = "8"
+    create_vm_libvirt = "yes"
+    kill_vm_libvirt = "yes"
+    ovmf:
+        kill_vm_libvirt_options = " --nvram"
--- a/libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
+++ b/libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
@@ -1,0 +1,101 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright Redhat
+#
+#   SPDX-License-Identifier: GPL-2.0
+#   Author: Nannan Li<nanli@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+import os
+
+from avocado.utils import distro
+from avocado.utils import process
+
+from virttest import utils_misc
+from virttest import utils_net
+from virttest.libvirt_xml import vm_xml
+
+from provider.virtual_network import pktgen_utils
+from provider.virtual_network import network_base
+
+
+def write_to_result_file(test, params):
+    """
+    Write some qemu, kvm version info and write them into pktgen test result.
+
+    :params, params
+    :return result_file, pktgen test result file.
+    """
+    kvm_ver_chk_cmd = params.get("kvm_ver_chk_cmd")
+
+    result_path = utils_misc.get_path(test.resultsdir, "pktgen_perf.RHS")
+    result_file = open(result_path, "w")
+    kvm_ver = process.system_output(kvm_ver_chk_cmd, shell=True).decode()
+    host_ver = os.uname().release
+    result_file.write("### kvm-userspace-ver : %s\n" % kvm_ver)
+    result_file.write("### kvm_version : %s\n" % host_ver)
+
+    return result_file
+
+
+def run(test, params, env):
+    """
+    Pktgen tests with burst mode.
+    """
+    def run_test():
+        """
+        Run pktgen in burst mode for vm.
+        """
+        test.log.debug("TEST_STEP 1: Prepare qemu version into pktgen test result")
+        result_file = write_to_result_file(test, params)
+        if flush_firewall_rules_cmd:
+            process.system(flush_firewall_rules_cmd, shell=True)
+
+        _distro = distro.detect()
+        if _distro.name == 'rhel' and int(_distro.version) > 7:
+            is_64k_page_size = "64k" in host_ver
+            if is_64k_page_size:
+                pktgen_utils.install_package(host_ver, pagesize="64k")
+            else:
+                pktgen_utils.install_package(host_ver)
+
+        test.log.debug("TEST_STEP 2: Test with guest and host connectivity")
+        vm.start()
+        test.log.debug("Test with Guest xml:%s", vm_xml.VMXML.new_from_dumpxml(vm_name))
+        session = vm.wait_for_serial_login(restart_network=True)
+        network_base.ping_check(params, ips, session, force_ipv4=True)
+
+        test.log.debug("TEST_STEP 3: Install package and Run pktgen in burst mode.")
+        guest_ver = session.cmd_output(guest_ver_cmd)
+        result_file.write("### guest-kernel-ver :%s" % guest_ver)
+        if _distro.name == 'rhel' and int(_distro.version) > 7:
+            guest_is_64k_page_size = "64k" in guest_ver
+            if guest_is_64k_page_size:
+                pktgen_utils.install_package(
+                    guest_ver.strip(),
+                    pagesize="64k",
+                    vm=vm,
+                    session_serial=session)
+            else:
+                pktgen_utils.install_package(
+                    guest_ver.strip(), vm=vm, session_serial=session)
+
+        pktgen_utils.run_tests_for_category(
+            params, result_file, test_vm, vm, session)
+
+        test.log.debug("TEST_STEP: Check the guest dmesg without error messages.")
+        vm.verify_dmesg()
+        vm.verify_kernel_crash()
+
+        result_file.close()
+
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+    flush_firewall_rules_cmd = params.get("flush_firewall_rules_cmd")
+    guest_ver_cmd = params.get("guest_ver_cmd")
+    test_vm = params.get_boolean("test_vm")
+    host_ver = os.uname().release
+    ips = {'host_ip': utils_net.get_host_ip_address(params)}
+
+    run_test()
--- a/provider/virtual_network/pktgen_utils.py
+++ b/provider/virtual_network/pktgen_utils.py
@@ -1,0 +1,235 @@
+import aexpect
+import logging
+import os
+
+from avocado.utils import process
+
+from virttest import data_dir
+from virttest import utils_net
+from virttest import utils_misc
+from virttest import utils_package
+
+
+LOG_JOB = logging.getLogger("avocado.test")
+
+
+class PktgenConfig:
+    def __init__(self):
+        self.interface = None
+        self.dsc = None
+        self.runner = None
+        self.dest_path = "/tmp/"
+
+    def configure_pktgen(
+        self,
+        pkt_cate,
+        vm=None,
+        session_serial=None,
+    ):
+        """
+        Configure pktgen test environment for different packet categories.
+
+        :param pkt_cate: Packet category (tx, rx, loopback)
+        :param vm: VM instance
+        :param session_serial: Serial session for guest command execution
+        :return: Configured PktgenConfig instance
+        """
+        source_path = os.path.join(data_dir.get_shared_dir(), "scripts/pktgen_perf")
+
+        if pkt_cate == "tx":
+            LOG_JOB.info("test guest tx pps performance")
+            vm.copy_files_to(source_path, self.dest_path)
+            guest_mac = vm.get_mac_address(0)
+            self.interface = utils_net.get_linux_ifname(session_serial, guest_mac)
+            host_iface = utils_net.get_default_gateway(iface_name=True, force_dhcp=False, json=True)
+            dsc_dev = utils_net.Interface(host_iface)
+            self.dsc = dsc_dev.get_mac()
+            self.runner = session_serial.cmd
+        return self
+
+    def generate_pktgen_cmd(
+        self,
+        script,
+        pkt_cate,
+        interface,
+        dsc,
+        threads,
+        size,
+        burst,
+        session_serial=None,
+    ):
+        """
+        Generate pktgen command based on test parameters.
+
+        :param script: Script name to execute
+        :param pkt_cate: Packet category (tx, rx, loopback)
+        :param interface: Network interface
+        :param dsc: Destination MAC address or IP
+        :param threads: Number of threads
+        :param size: Packet size
+        :param burst: Burst size
+        :param session_serial: Serial session for guest command execution
+        :return: Generated command string
+        """
+        cmd = "%s -i %s -m %s -n 0 -t %s -s %s -b %s -c 0" % (
+            "%spktgen_perf/%s.sh" % (self.dest_path, script),
+            interface,
+            dsc,
+            threads,
+            size,
+            burst,
+        )
+
+        if (
+            session_serial
+            and self.runner.__name__ == session_serial.cmd.__name__
+        ):
+            cmd = f"{cmd} &"
+
+        return cmd
+
+
+def run_test(script, cmd, runner, interface, timeout):
+    """
+    Run pktgen  script on remote and gather packet numbers/time and
+    calculate mpps.
+    :param script: pktgen script name.
+    :param cmd: The command to execute the pktgen script
+    :param runner: The command runner function
+    :param interface: The network interface used by pktgen.
+    :param timeout: The maximum time allowed for the test to run
+    :return: The calculated MPPS (Million Packets Per Second)
+    """
+
+    packets = "cat /sys/class/net/%s/statistics/tx_packets" % interface
+    LOG_JOB.info("Start pktgen test by cmd '%s'", cmd)
+    try:
+        packet_b = runner(packets)
+        packet_a = None
+        runner(cmd, timeout)
+    except aexpect.ShellTimeoutError:
+        # when pktgen script is running on guest, the pktgen process
+        # need to be killed.
+        kill_cmd = (
+            "kill -9 `ps -ef | grep %s --color | grep -v grep | "
+            "awk '{print $2}'`" % script
+        )
+        runner(kill_cmd)
+        packet_a = runner(packets)
+    except process.CmdError:
+        # when pktgen script is running on host, the pktgen process
+        # will be quit when timeout triggers, so no need to kill it.
+        packet_a = runner(packets)
+
+    return "{:.2f}".format((int(packet_a) - int(packet_b)) / timeout / 10 ** 6)
+
+
+def install_package(ver, pagesize=None, vm=None, session_serial=None):
+    """
+    Check module pktgen, install kernel-modules-internal package.
+
+    :param ver: Kernel version string
+    :param pagesize: Page size specification for kernel package selection
+    :param vm: VM instance for guest installation
+    :param session_serial: Serial session for guest command execution
+    """
+    result = process.run("which brew", ignore_status=True, shell=True)
+    if result.exit_status != 0:
+        utils_package.package_install("brewkoji")
+
+    if pagesize:
+        kernel_ver = "kernel-%s-modules-internal-%s" % (pagesize, ver.split("+")[0])
+    else:
+        kernel_ver = "kernel-modules-internal-%s" % ver
+    cmd_download = "cd /tmp && brew download-build %s --rpm" % kernel_ver
+    cmd_install = "cd /tmp && rpm -ivh  %s.rpm --force --nodeps" % kernel_ver
+
+    utils_misc.cmd_status_output(cmd_download, shell=True)
+    cmd_clean = "rm -rf /tmp/%s.rpm" % kernel_ver
+    if session_serial:
+        local_path = "/tmp/%s.rpm" % kernel_ver
+        remote_path = "/tmp/"
+        vm.copy_files_to(local_path, remote_path)
+    utils_misc.cmd_status_output(cmd_install, session=session_serial)
+    utils_misc.cmd_status_output(cmd_clean, session=session_serial)
+
+
+def format_result(result):
+    """Format result with fixed width (12) and decimals (2)"""
+    if isinstance(result, str):
+        return "%12s" % result
+    elif isinstance(result, int):
+        return "%12d" % result
+    elif isinstance(result, float):
+        return "%12.2f" % result
+    else:
+        raise TypeError(f"unexpected result type: {type(result).__name__}")
+
+
+def run_tests_for_category(
+    params,
+    result_file,
+    test_vm=None,
+    vm=None,
+    session_serial=None,
+):
+    """
+    Run Pktgen tests for a specific category.
+
+    :param params: Dictionary with the test parameters
+    :param result_file: File to write the test results
+    :param test_vm: Flag indicating whether the test is running on a VM
+    :param vm: VM instance
+    :param session_serial: Session serial for VM
+    """
+
+    timeout = params.get_numeric("pktgen_test_timeout", "240")
+    category = params.get("pkg_dir")
+    record_list = params.get_list("record_list")
+    pktgen_script = params.get("pktgen_script")
+    # Get single values for threads and burst instead of looping
+    threads = params.get("pktgen_threads", "")
+    burst = params.get("burst", "")
+
+    record_line = ""
+    for record in record_list:
+        record_line += "%s|" % format_result(record)
+
+    pktgen_config = PktgenConfig()
+
+    for script in pktgen_script.split():
+        for pkt_cate in category.split():
+            result_file.write("Script:%s " % script)
+            result_file.write("Category:%s\n" % pkt_cate)
+            result_file.write("%s\n" % record_line.rstrip("|"))
+
+            # Use single values directly since they're not lists
+            size = params.get("pkt_size", "")
+            if pkt_cate != "loopback":
+                pktgen_config = pktgen_config.configure_pktgen(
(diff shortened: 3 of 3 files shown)


nanli1:
@yanglei-rh This case already included multiple queues as your requirement, Please check. Thanks

chloerh:
There are conflicts

chloerh (approved):

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
```diff
@@ -0,0 +1,22 @@
...
+    pktgen_script = "pktgen_sample03_burst_single_flow"
+    pktgen_threads = 1
+    pkt_size = 64
+    burst = 3
+    pktgen_test_timeout = 30
+    record_list = "pkt_size run_threads burst mpps"
+    guest_ver_cmd = "uname -r"
+    kvm_ver_chk_cmd = "rpm -qa qemu-kvm-rhev && rpm -qa qemu-kvm"
+    variants:
+        - ethernet_multi_queues:
```
yanglei-rh, line 14: Hi @nanli1 Please also consider the ovs bridge.

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
```diff
@@ -0,0 +1,152 @@
...
+        Prepare the test environment with required guest type and interface.
+
+        """
+        utils_net.create_linux_bridge_tmux(bridge_name, host_iface)
+        if params.get_boolean("pre_tap") :
+            network_base.create_tap(tap_name, bridge_name, 'root', flag=tap_flag)
+
+        test.log.info("TEST_SETUP: guest and interface.")
+        network_base.prepare_single_vm(params, vm_name, iface_list)
+
```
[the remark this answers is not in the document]
smitterl: @nanli1 ^^ AFAICT coderabbitai is correct, the third argument in your network_base.py of this commit is `disk_path` - please fix.

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
```diff
@@ -0,0 +1,147 @@
...
+    def run_test(vm_list):
+        """
+        Run pktgen in burst mode for vm.
+        """
+        result_file = write_in_result_file(test, params)
+
+        if disable_iptables_rules_cmd:
+            process.system(disable_iptables_rules_cmd, shell=True)
+
+        pktgen_runner = pktgen_utils.PktgenRunner()
```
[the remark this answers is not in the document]
smitterl: What will haveppen if `utils_net.delete_linux...` raised an error? It seems the iptables rules are not recovered and the system will be in bad state. If so, please make sure this won't happen.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
+import aexpect
```
smitterl: If this file was created with a code assistant please read https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/ai_policy.html and make sure to tag the tool in your commit message - https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/ai_policy.html
If not, please add your own copyright notice on top like you did in your test script file.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+    host_version = int(host_version_list[0])
+    repo_files = {
+        9: "rcm-tools-rhel-9-baseos.repo",
+        10: "rcm-tools-rhel-10-baseos.repo"
+    }
+
+    repo_link = repo_files.get(host_version)
+    if not repo_link:
+        raise ValueError(f"Unsupported RHEL version: {host_version}")
+    download_url = f'http://download.eng.bos.redhat.com/rel-eng/internal/{repo_link}'
```
smitterl: I'm not sure if this URL is still up and generally our commits must not contain internal URLs, please make this configurable.

Installing brew on the system shouldn't be done in this way by the test case. Instead, you should make sure that CI at least adds the needed repositories and you can use the `utils_package` module to install the package if not present. Don't you also have to set up the koji configuration (internal) to make this work? All of this is better taken care of in CI and should not be part of this upstream test file.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+
+def format_result(result, base, fbase):
+    """
+    Format the result to a fixed length string.
+
+    :param result: result need to convert
+    :param base: the length of converted string
+    :param fbase: the decimal digit for float
+    """
+    if isinstance(result, six.string_types):
```
smitterl: Why do you need `six` for this? if you just want to see if it's a string do `isinstance(result, str)`.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        session_serial=None,
+        interface=None,
+    ):
+        """
+        Configure pktgen test environment for different packet categories.
+
+        :param params: Dictionary with test parameters
+        :param script: Script name to execute
+        :param pkt_cate: Packet category (tx, rx, loopback)
+        :param test_vm: Flag indicating whether test is running on VM
```
smitterl: I find this parameter very confusing. Either call it `test_in_vm` or I believe you can remove it altogether because `test_in_vm == (session_serial is not None)` holds, am I correct?

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+            host_iface = utils_net.get_default_gateway(iface_name=True, force_dhcp=False, json=True)
+            dsc_dev = utils_net.Interface(host_iface)
+            self.dsc = dsc_dev.get_mac()
+            self.runner = session_serial.cmd
+        elif pkt_cate == "rx":
+            LOG_JOB.info("test guest rx pps performance")
+            process.run("cp -r %s %s" % (local_path, remote_path))
+            host_bridge = params.get("netdst", "switch")
+            utils_net.Interface(host_bridge)
+            if script == "pktgen_perf":
```
smitterl: In your test configuration you only seem to need  `pktgen_script = "pktgen_sample03_burst_single_flow"` - is there a miss in your test config or can you omit this code?

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+            # host_iface = utils_net.get_host_iface_name_for_vm(vm)
+            host_iface = utils_net.get_default_gateway(iface_name=True, force_dhcp=False, json=True)
+            dsc_dev = utils_net.Interface(host_iface)
+            self.dsc = dsc_dev.get_mac()
+            self.runner = session_serial.cmd
+        elif pkt_cate == "rx":
+            LOG_JOB.info("test guest rx pps performance")
+            process.run("cp -r %s %s" % (local_path, remote_path))
+            host_bridge = params.get("netdst", "switch")
+            utils_net.Interface(host_bridge)
```
smitterl: This line doesn't do anything, remove it.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+from virttest import data_dir
+from virttest import utils_net
+from virttest import utils_misc
+
+
+LOG_JOB = logging.getLogger("avocado.test")
+
+
+class PktgenConfig:
+    def __init__(self, interface=None, dsc=None, runner=None):
```
smitterl: IMO, it would be a better design to have:
1. A class `PktgenConfigParams` that holds the parameters that are passed here and/or to `configure_pktgen`
2. I don't see a reason why `configure_pktgen` shouldn't be already part of the constructor (`__init__`) except for the part where you copy files. But even that could be part of the constructor.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        :param threads: Number of threads
+        :param size: Packet size
+        :param burst: Burst size
+        :param session_serial: Serial session for guest command execution
+        :return: Generated command string
+        """
+        script_path = "/tmp/pktgen_perf/%s.sh" % script
+        if script in "pktgen_perf":
+            dsc_option = "-m" if pkt_cate == "tx" else "-d"
+            cmd = "%s -i %s %s %s -t %s -s %s" % (
```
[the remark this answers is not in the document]
smitterl: ^^coderabbit is right @nanli1 or maybe you want to say `pktgen_perf in scrip`? In any case, this seems not to be necessary.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+    base = params.get("format_base", "12")
+    fbase = params.get("format_fbase", "2")
+
+    # get record_list
+    record_line = ""
+    for record in record_list.split():
+        record_line += "%s|" % format_result(record, base, fbase)
+
+    pktgen_config = PktgenConfig()
+    pktgen_runner = PktgenRunner()
```
smitterl: Why isn't it `PktgenConfig(runner=PktgenRunner())`?

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+    if vp_vdpa:
+        pktgen_config.vp_vdpa_bind(session_serial)
+
+    for script in pktgen_script.split():
+        for pkt_cate in category.split():
+            result_file.write("Script:%s " % script)
+            result_file.write("Category:%s\n" % pkt_cate)
+            result_file.write("%s\n" % record_line.rstrip("|"))
+
+            for size in params.get("pkt_size", "").split():
```
smitterl: The complexity of these nested if else and for is too high. This needs to be simplified so that a human can review the logic more easily.

smitterl (changes requested):
Hi @nanli1 Please check my comments. There were some things that actually look like bugs, some things that can be improved but also two major areas where I'd recommend:

1. Remove the package installation logic and delegate that responsibility to CI
2. Redesign classes `PktgenConfig` and `PktgenRunner` to make code readable and reduce duplication.

Please ping me on slack to address any question you might have.

provider/virtual_network/pktgen_utils.py
nanli1: hi Sebas, thanks your help about reviewing. That's very generous.

Actually, this file this was not generated from AI, and it'is a migrated from qemu case and we almost just copy pktgen_utils.py from qemu repo, because virtual network team choose to believe tp-qemu ever reviewed 
for pktgen_utils.py,    so I invite wenli  @wenli(the source author) to give some reply for your comment for this file 

CC @yanglei-rh

quanwenli: The original test workflow was as follows: first, install the required packages on the host using brew, and then copy them to the vm via scp. This approach was used because the performance tests run in a private network environment, where the virtual machines cannot directly install packages from the external network. Therefore, we pre-install the necessary packages on the host and transfer them to the vm for use in the internal network environment.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        packets = "cat /sys/class/net/%s/statistics/tx_packets" % interface
+        LOG_JOB.info("Start pktgen test by cmd '%s'", cmd)
+        try:
+            packet_b = runner(packets)
+            packet_a = None
+            runner(cmd, timeout)
+        except aexpect.ShellTimeoutError:
+            # when pktgen script is running on guest, the pktgen process
+            # need to be killed.
+            kill_cmd = (
```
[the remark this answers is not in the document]
quanwenli: During the performance tests, I tried using killall pktgen, but it did not reliably terminate the pktgen processes in our test environment.

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
```diff
@@ -0,0 +1,24 @@
...
+    start_vm = no
+    category = "tx"
+    pktgen_script = "pktgen_sample03_burst_single_flow"
+    pktgen_threads = 1
+    pkt_size = 64
+    burst = 3
+    pktgen_test_timeout = 30
+    record_list = "pkt_size run_threads burst mpps"
+    guest_ver_cmd = "uname -r"
+    kvm_ver_chk_cmd = "rpm -qa qemu-kvm-rhev && rpm -qa qemu-kvm"
```
qiankehan: Use simpler cmd `rpm -qa qemu-kvm-rhev qemu-kvm` and it takes less time:
```
(base) bash-5.3$ time echo $(rpm -qa qemu-kvm-rhev && rpm -qa qemu-kvm)
qemu-kvm-10.1.2-1.fc43.x86_64

real    0m4.215s
user    0m3.605s
sys     0m0.447s
(base) bash-5.3$ time echo $(rpm -qa qemu-kvm-rhev qemu-kvm)
qemu-kvm-10.1.2-1.fc43.x86_64

real    0m2.142s
user    0m1.812s
sys     0m0.242s
```

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+    params.get("pktgen_threads")
+    burst = params.get("burst")
+    record_list = params.get("record_list")
+    pktgen_script = params.get("pktgen_script")
+    base = params.get("format_base", "12")
+    fbase = params.get("format_fbase", "2")
+
+    # get record_list
+    record_line = ""
+    for record in record_list.split():
```
qiankehan: Set record_list as list format in cfg file and get the list by `params.get_list()`. No need to convert str to list

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
```diff
@@ -0,0 +1,24 @@
+- virtual_network.qemu_test.pktgen_burst_mode_test:
+    type = pktgen_burst_mode_test
+    start_vm = no
+    category = "tx"
```
qiankehan: What category is here? The name could be more detailed

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
```diff
@@ -0,0 +1,24 @@
...
+    pktgen_threads = 1
+    pkt_size = 64
+    burst = 3
+    pktgen_test_timeout = 30
+    record_list = "pkt_size run_threads burst mpps"
+    guest_ver_cmd = "uname -r"
+    kvm_ver_chk_cmd = "rpm -qa qemu-kvm-rhev && rpm -qa qemu-kvm"
+    queues = 8
+    test_vm = no
+    disable_iptables_rules_cmd = "iptables -F ; nft flush ruleset"
```
qiankehan: The cmd is to flush firewall rules. Use the prefix `flush_firewall`

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
```diff
@@ -0,0 +1,151 @@
...
+        if pktgen_runner.is_version_lt_rhel(process.getoutput("uname -r"), 7):
+            if host_ver.count("64k"):
```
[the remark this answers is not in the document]
qiankehan: Without context it is hard to know what the magic string "64k" means. Please define a constant or param to replace it. Fix it for line 70, 96, 99 as well.

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
```diff
@@ -0,0 +1,151 @@
...
+    test_vm = params.get_boolean("test_vm")
+    host_ver = os.uname()[2]
+
+    rand_id = utils_misc.generate_random_string(3)
+    bridge_name = 'br_' + rand_id
+    tap_name = 'tap_' + rand_id
+    host_iface = utils_net.get_default_gateway(iface_name=True, force_dhcp=True, json=True)
+    host_ip = utils_net.get_ip_address_by_interface(host_iface, ip_ver='ipv4')
+
+    iface_list = eval(params.get('iface_list', '{}'))
```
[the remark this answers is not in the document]
qiankehan: What about `params.get_list()`? But I am not sure if it could covert the items in the list to dict type

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
qiankehan: The cmd is to flush fireall.  
s/disable_iptables/flush_firewall/

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
qiankehan: What category is here? The name could be more detailed

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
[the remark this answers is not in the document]
qiankehan: Hard to understand the magic string "64k" without context. Please define and use it via a constant or param.   
Fix it at lines 70,96,99.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        except process.CmdError:
+            # when pktgen script is running on host, the pktgen process
+            # will be quit when timeout triggers, so no need to kill it.
+            packet_a = runner(packets)
+        count = int(packet_a) - int(packet_b)
+        pps_results = count / timeout
+
+        # convert pps to mpps
+        power = 10**6
+        mpps_results = float(pps_results) / float(power)
```
qiankehan: No need to covert float

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        raise ValueError(f"Unsupported architecture: {ARCH}")
+
+    # Determine repository link based on RHEL version
+    host_version_list = get_host_version(process.getoutput("uname -r"))
+    if not host_version_list:
+        raise ValueError("Unable to determine RHEL version from uname output")
+    host_version = int(host_version_list[0])
+    repo_files = {
+        9: "rcm-tools-rhel-9-baseos.repo",
+        10: "rcm-tools-rhel-10-baseos.repo"
```
qiankehan: Use str type for index.

qiankehan (changes requested):

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
qiankehan: I was intended to say the word `category` is a little confusing. For me, I prefer the `pkg_dir` (packet direction)

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,318 @@
...
+        "x86_64": "/dev/ttyS0",
+        "s390x": "/dev/ttysclp0"
+    }
+
+    console_dev = console_devices.get(ARCH)
+    if not console_dev:
+        raise AssertionError(f"Unsupported architecture: {ARCH}")
+
+    # Determine repository link based on RHEL version
+    host_version_list = get_major_version(process.getoutput("uname -r"))
```
qiankehan: Use `distro.major_version()` instead. Remove the function `get_major_version`

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,318 @@
...
+    """
+    Extract RHEL version from uname string.
+
+    :param uname_str: Output from uname command containing kernel version info
+    :return: List of RHEL version numbers found in the string
+    """
+    return re.findall("el(\\d+)", uname_str)
+
+
+def install_brewkoji():
```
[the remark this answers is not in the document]
nanli1: This is because there is a scenario is to test vm with inner network(ip 192.* or 10.*) , vm could not install by yum, Also we have decide which CI (libvirt ci or qemu ci)would we use  to run qemu cases,  what do you think

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+        result_file.close()
+
+    vm_name = params.get("main_vm")
+    vm = env.get_vm(vm_name)
+    flush_firewall_rules_cmd = params.get("flush_firewall_rules_cmd")
+    guest_ver_cmd = params.get("guest_ver_cmd")
+    test_vm = params.get_boolean("test_vm")
+    host_ver = os.uname().release
+    ips = {'host_ip': utils_net.get_host_ip_address(params)}
```
quanwenli: Tested in perf setup with the following parameters:
    nics = 'nic1'
    netdst = private
    physical_nic = shell:ip link | sed -n '/3c:fd:fe:1a:84:22/{g;1!p;};h' | awk -F': ' '{print $2}' | head -1

Issue observed:
    get_host_ip_address(params) fails because 'netdst'='private' is not a real interface name.

For this case, physical_nic should be used to get the real host interface and IP.

libvirt/tests/cfg/virtual_network/qemu/pktgen_burst_mode_test.cfg
```diff
@@ -0,0 +1,24 @@
+- virtual_network.qemu_test.pktgen_burst_mode_test:
+    type = pktgen_burst_mode_test
+    no Windows
+    only virtio_net
+    no Host_RHEL.m5, Host_RHEL.m6
```
smitterl: AFAIK tp-libvirt doesn't use these parameters, only tp-qemu/kar does. Can you please confirm if they are relevant, if not, please remove them.

libvirt/tests/src/virtual_network/qemu/pktgen_burst_mode_test.py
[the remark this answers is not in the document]
smitterl: @quanwenli the question here specifically from me was why use `count`, it seems you just want to confirm if it's a 64K page size. Using `count` this way is not common in python, so I would use something like the following
a) in order to avoid misunderstanding
b) to cover also @qiankehan 's concerns
c) don't repeat 'detect' call for performance
```python
_distro = distro.detect()
if _distro.name == 'rhel' and int(_distro.version) > 7:
    is_64k_page_size = "64k" in host_ver
    if is_64k_page_size:
...
```

provider/virtual_network/pktgen_utils.py
smitterl: The constructor is never used with 3 parameters, only with empty parameter list.
A long parameter list is not clean code and very difficult to review.

provider/virtual_network/pktgen_utils.py
nanli1: Thanks all !!  I removed the function

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,476 @@
...
+        if pkt_cate == "tx":
+            LOG_JOB.info("test guest tx pps performance")
+            vm.copy_files_to(local_path, remote_path)
+            guest_mac = vm.get_mac_address(0)
+            self.interface = utils_net.get_linux_ifname(session_serial, guest_mac)
+            # host_iface = utils_net.get_host_iface_name_for_vm(vm)
+            host_iface = utils_net.get_default_gateway(iface_name=True, force_dhcp=False, json=True)
+            dsc_dev = utils_net.Interface(host_iface)
+            self.dsc = dsc_dev.get_mac()
+            self.runner = session_serial.cmd
```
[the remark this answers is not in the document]
nanli1: hi sebas,That's a good method. But now I would only use self.runner twice, I would like to keep it as it is ,is that ok?

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+    for script in pktgen_script.split():
+        for pkt_cate in category.split():
+            result_file.write("Script:%s " % script)
+            result_file.write("Category:%s\n" % pkt_cate)
+            result_file.write("%s\n" % record_line.rstrip("|"))
+
+            for size in params.get("pkt_size", "").split():
+                for threads in params.get("pktgen_threads", "").split():
+                    for burst in params.get("burst", "").split():
```
qiankehan: The param `pktgen_threads` and `burst` are not in the list format. So the for loops are not need

qiankehan (approved):
One more comment for the unnecessary loops. The other LGTM

provider/virtual_network/pktgen_utils.py
nanli1: hi Sebas, I deleted all brew related code.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    cmd_clean = "rm -rf /tmp/%s.rpm" % kernel_ver
+    if session_serial:
+        local_path = "/tmp/%s.rpm" % kernel_ver
+        remote_path = "/tmp/"
+        vm.copy_files_to(local_path, remote_path)
+    utils_misc.cmd_status_output(cmd_install, session=session_serial)
+    utils_misc.cmd_status_output(cmd_clean, session=session_serial)
+
+
+def format_result(result, base, fbase):
```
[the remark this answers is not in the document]
nanli1: Absolutely, Updated

smitterl (changes requested):

quanwenli:
https://github.com/autotest/tp-libvirt/pull/6531#discussion_r2642657469

Here, category = "tx"/"rx" is actually used to indicate packet direction. The name category is a bit generic and may be slightly confusing at first glance, but it’s consistent with the existing perf usage. It's up to @nanli1  if she wants, she can rename it, but this naming is commonly used in perf.

quanwenli:
https://github.com/autotest/tp-libvirt/pull/6531#discussion_r2642689315

I understand your point. limiting the test to a single environment could cause it to fail elsewhere. tp-libvirt now supports create_vm_libvirt = yes / kill_vm_libvirt = yes, which allows using pre-prepared images from other sources, rather than relying on the images previously used in tp-libvirt. This keeps the tests flexible.

provider/virtual_network/pktgen_utils.py
```diff
@@ -0,0 +1,230 @@
...
+    :param ver: Kernel version string
+    :param pagesize: Page size specification for kernel package selection
+    :param vm: VM instance for guest installation
+    :param session_serial: Serial session for guest command execution
+    """
+    if pagesize:
+        kernel_ver = "kernel-%s-modules-internal-%s" % (pagesize, ver.split("+")[0])
+    else:
+        kernel_ver = "kernel-modules-internal-%s" % ver
+    cmd_download = "cd /tmp && brew download-build %s --rpm" % kernel_ver
```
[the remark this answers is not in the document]
smitterl: Got it. I think using brew is not desirable, the kernel-modules-internal package is in BUILDROOT. That's how I've been handling it. But I guess it's okay.

yanglei-rh (approved):
LGTM

PR #6532, opened 2025-09-03
Inc_backup: fix checkpoint already existes when no readonly disk

meinaLi:
For some test with readonly disk, we need to create chekcpoint first. But now the condition affects other test cases, so we need to add another condition with_readonly to judge.


chunfuwen (approved):
lgtm

PR #6534, opened 2025-09-04
Delete the pid file after starting the daemon in the foreground

yafu-1:


chunfuwen:
any test result is available?

chloerh (approved):

PR #6555, opened 2025-09-10
migration_with_external_tpm: adapt to image mode

yanqzhan:
'chcon' on /usr/bin/swtpm_setup or /usr/bin/swtpm is not allowed in image mode, but higher rhel versions do not have prevention on directly excuting the two bin files. So use the new tool utils_sys.is_image_mode to identify image or package mode, then only do 'chcon' on package modes especillay benefits the lower rhel product versions to avoid errors.

## Summary by CodeRabbit

* **Tests**
  * Added environment-aware handling for external vTPM during migration, detecting image-based runs.
  * When not in image mode, launch swtpm via systemd and apply SELinux context adjustments for reliable startup.
  * When in image mode, retain existing nohup-based swtpm startup.
  * Improved cleanup: remove local CA artifacts and socket files, apply restorecon where needed, and ensure connections close after cleanup.
  * Increased robustness for both local and remote execution paths.

### DIFF
--- a/libvirt/tests/src/migration/migration_with_vtpm/migration_with_external_tpm.py
+++ b/libvirt/tests/src/migration/migration_with_vtpm/migration_with_external_tpm.py
@@ -6,6 +6,7 @@ from avocado.utils import process
 from virttest import remote
 from virttest import utils_package
 from virttest import utils_misc
+from virttest import utils_sys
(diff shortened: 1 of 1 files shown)


cliping (approved):
LGTM

PR #6569, opened 2025-09-17
[InProgress] Auto case: Migrate VM with nvme disk

ganeshhubale:


v2v/tests/cfg/function_test_esx.cfg
```diff
@@ -287,6 +287,23 @@
...
                             msg_content_yes = "ovirt:id='36'"
                             os_version = win11
+                - windows_vm_with_nvme_disk:
+                    only esx_80
+                    only dest_libvirt
+                    version_required = "[libvirt-11.5.0-1,)"
+                    os_type = 'windows'
+                    os_version = 'win2022'
+                    boottype = 2
+                    main_vm = 'Auto-esx8.0-win2022-nvme-disk-thin-provision'
```
xiaodwan, line 297: It's better to use variant in ci-share-data for the main_vm. The guest name may change in future.

xiaodwan (approved):

PR #6670, opened 2025-11-21
Cancel the testcase as maximum mode is not supported for ppc!

Anushree-Mathur:
Added a condition to cancel a testcase if machine is ppc64 or ppc64le and cpu_mode is maximum as this mode is not supported for power architecture.

## Summary by CodeRabbit

* **Tests**
  * Tests now automatically cancel on Power (PPC) systems when CPU mode is set to "maximum", stopping execution early (before CPU XML creation) to avoid running unsupported configurations on that hardware.

### DIFF
--- a/libvirt/tests/src/cpu/vcpu_misc.py
+++ b/libvirt/tests/src/cpu/vcpu_misc.py
@@ -1,5 +1,6 @@
 import logging as log
 import os
+import platform
 import re
 
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


Anushree-Mathur:
Here is the command and the output to showcase that this mode is not supported for ppc!

```
virsh domcapabilities --arch ppc64 --machine pseries | grep -A2 maximum
    <mode name='maximum' supported='no'/>
    <mode name='host-model' supported='yes'>
      <model fallback='allow'>POWER11</model>
```

harihare (commented):
Effects only PPC, patch LGTM. Please review and merge the patch.

PR #6674, opened 2025-11-23
Fixing maxvcpu mismatch with topology issue, making it dynamic!

Anushree-Mathur:
Maxvcpu is hardcoded for the specific testcase in virsh_emulatorpin.py. When smp value, vcpu_cores, vcpu_threads, vcpu_sockets is specified at the time of running the testcase then the hardcoded value will fail. Adding the code to get this value dynamically based on the smp value provided in the test suite config file.
1 example is that hardcoded value is 7 in virsh_emulatorpin.cfg but in the testsuite config smp value given is 32 not topology will be affected in this case my check will help.

## Summary by CodeRabbit

* **Tests**
  * Enhanced vCPU configuration handling to properly compute maximum vCPUs from available CPU topology information when processing emulator pin attributes.


harihare (commented):
LGTM, as two parameters topology and max_vcpus are not required. we can computed max_cpus when topology is provided. Please may I request to review and merge the patch.

PR #6724, opened 2025-12-10
Fix case - cow_on_read_path

ganeshhubale:
## Summary by CodeRabbit

* **Tests**
  * Added a test variant that exercises nbdkit behavior as an unprivileged user: test harness now creates the unprivileged account, provisions temporary sudo/SSH access to run the scenario non-root, and performs cleanup afterward to improve coverage of privilege-separated paths.


crobinso:
IMO this is really hairy for not a ton of value. Setting up sudoers etc is pretty intense, though I see it is done in other parts of tp-libvirt. But IMO we should not pursue this so I am closing it

PR #6801, opened 2026-02-11
iface_update: Support new libvirt error message for boot order conflicts

BulaYoungR:
Fix negative test failure caused by libvirt error message format change in newer versions.

The test checks that libvirt correctly rejects setting boot order on a network interface when os/boot is already configured. However, libvirt changed the error message format between versions.

Changed file: libvirt/tests/src/virtual_network/iface_update.py
Changed lines: 283-302

Old libvirt error message:
    "per-device boot elements cannot be used together with os/boot elements"

New libvirt error message:
    "boot order 1 is already used by another device"

The test was checking only for the old message format, causing it to fail with newer libvirt versions even though the actual error was correct.

Fix: Modified error checking logic to accept both error message formats:
- First check the expected message from test config (expect_err_msg)
- If not matched, try alternative known patterns for boot order conflicts
- This ensures compatibility with both old and new libvirt versions

The solution uses fallback patterns instead of updating test config files, making it work across different libvirt versions without needing separate test configurations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for libvirt boot-order conflicts, recognizing both legacy and current error formats.

* **Refactor**
  * Reworked interface update logic to use a generalized, dictionary-driven approach for broader and more maintainable attribute coverage.

* **Other**
  * Added explicit network XML state logging around interface updates for better traceability.

### DIFF
--- a/libvirt/tests/src/virtual_network/iface_update.py
+++ b/libvirt/tests/src/virtual_network/iface_update.py
@@ -1,11 +1,9 @@
-import os
 import re
 import logging as log
 import time
 
 from xml.etree import ElementTree
 from avocado.utils import process
-from avocado.utils.software_manager.backends import rpm
 
 from virttest import virsh
 from virttest import utils_net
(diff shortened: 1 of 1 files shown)


harvey0100:
Closing this PR due to current team constraints. This is part of a broader effort to triage all in-flight work across our upstream repos. If this work is still needed, please feel free to reopen and it will be picked up. Apologies for any inconvenience.

PR #6802, opened 2026-02-11
iface_attach_detach: Support new libvirt error message for boot order conflicts

BulaYoungR:
Fix test failure caused by libvirt error message format change in newer versions when validating boot order conflicts.

libvirt/tests/src/virtual_network/iface_attach_detach.py
Lines changed: 46-61 (added 15 lines)

The test validates that libvirt correctly rejects attaching a network interface with boot order when os/boot is already configured on another device. However, libvirt changed the error message format between versions, causing the test to fail even though the error itself was correct.

Old libvirt error message:
    "per-device boot elements cannot be used together with os/boot elements"

New libvirt error message:
    "boot order 1 is already used by another device"

Problem:
The test expected only the old error message format from the error_msg parameter, causing it to fail when libvirt returned the new message format. The actual error was correct - libvirt properly rejected the invalid configuration - but the message text differed.

Solution:
Modified error checking logic in test_boot_order() function to accept both error message formats:

- If status_error is expected and error_msg is provided:
  * Convert error_msg to list if it's a string
  * Extend the list with known alternative patterns for boot order conflicts
  * Pass the extended list to libvirt.check_result()
  * This allows the test to pass if any pattern matches

- If status_error is expected but error_msg is empty:
  * Call libvirt.check_result() with original parameters

The solution adds fallback patterns locally in the test rather than modifying the shared libvirt.check_result() function in utils_test/libvirt.py, ensuring this specific test works across libvirt versions without affecting other tests that may depend on check_result() behavior.

Alternative patterns cover both old and new libvirt error message formats:
- "boot order .* is already used by another device"
- "per-device boot elements cannot be used together with os/boot elements"

## Summary by CodeRabbit

* **Tests**
  * Broadened boot-order error matching so virtual-network tests accept both legacy and new libvirt messages, improving compatibility across libvirt versions.
* **Chores**
  * Cleaned up and standardized test code formatting and parameter handling for readability and more consistent test behavior.

### DIFF
--- a/libvirt/tests/cfg/virtual_network/iface_attach_detach.cfg
+++ b/libvirt/tests/cfg/virtual_network/iface_attach_detach.cfg
@@ -13,13 +13,15 @@
             variants scenario:
                 - with_os_boot:
                     no s390-virtio
-                    error_msg = 'per-device boot elements cannot be used together with os/boot elements'
+                    # old/new libvirt variants
+                    error_msg = 'per-device boot elements cannot be used together with os/boot elements|boot order .* is already used by another device'
                 - with_boot_disk:
-                    error_msg = 'boot order 1 is already used by another device'
+                    # old/new libvirt variants (number may vary, keep regex)
+                    error_msg = 'boot order .* is already used by another device|per-device boot elements cannot be used together with os/boot elements'
                 - unsupported_value:
                     variants boot_order:
                         - -1:
-                            error_msg = 'Expected .*integer value'
+                            error_msg = 'Expected integer value|Expected non-negative integer value'
                         - ss:
-                            error_msg = 'Expected .*integer value'
+                            error_msg = 'Expected integer value|Expected non-negative integer value'
             iface_attrs = {'boot': '${boot_order}'}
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virtual_network/iface_attach_detach.cfg
```diff
@@ -13,13 +13,15 @@
...
+                    error_msg = 'per-device boot elements cannot be used together with os/boot elements|boot order .* is already used by another device'
                 - with_boot_disk:
-                    error_msg = 'boot order 1 is already used by another device'
+                    # old/new libvirt variants (number may vary, keep regex)
+                    error_msg = 'boot order .* is already used by another device|per-device boot elements cannot be used together with os/boot elements'
                 - unsupported_value:
                     variants boot_order:
                         - -1:
-                            error_msg = 'Expected .*integer value'
+                            error_msg = 'Expected integer value|Expected non-negative integer value'
```
dzhengfy: Does not old value match "Expected non-negative integer value"? Do you need this change indeed?

libvirt/tests/cfg/virtual_network/iface_attach_detach.cfg
```diff
@@ -13,13 +13,15 @@
...
                         - ss:
-                            error_msg = 'Expected .*integer value'
+                            error_msg = 'Expected integer value|Expected non-negative integer value'
```
dzhengfy: same as above

harvey0100:
Closing this PR due to current team constraints. This is part of a broader effort to triage all in-flight work across our upstream repos. If this work is still needed, please feel free to reopen and it will be picked up. Apologies for any inconvenience.

PR #6875, opened 2026-05-31
Skip CPU driver check for win2016 (known issue RHEL-17685)

millevy:
After v2v conversion, win2016 guests show a Device Manager error (code 45) for the CPU — the Intel/AMD Processor driver does not appear in the driver list. This is a win2016-specific issue (confirmed not reproducible on win10, win11, win2019, win2022) and is cosmetic — the guest boots and runs normally.

The current code treats a missing CPU driver as a test failure, but for win2016 this is a known, unfixable VMware/Windows interaction (RHEL-17685, closed as minor). Skipping the check for win2016 avoids a false test failure.

The fix is self-healing: if the underlying issue is ever resolved and the CPU driver appears in win2016's driver list, the normal regex check will match first and this elif branch will not be reached.

Ref: RHEL-17685, LNXVAST-2542

## Summary by CodeRabbit

* **Bug Fixes**
  * CPU driver validation now accommodates a known issue on Windows 2016 without reporting false failures.

### DIFF
--- a/provider/v2v_vmcheck_helper.py
+++ b/provider/v2v_vmcheck_helper.py
@@ -843,6 +843,8 @@ class VMChecker(object):
         if re.search(r"(Intel|AMD) Processor", win_dirvers):
             if re.search(r"(Intel|AMD) Processor", win_dirvers).group(0) in cpu_drivers:
                 LOG.info("CPU driver '%s' is found" % re.search(r"(Intel|AMD) Processor", win_dirvers).group(0))
+        elif self.os_version == 'win2016':
+            LOG.info("CPU driver not found for win2016, known issue (RHEL-17685), skipping check")
         else:
             err_msg = "CPU driver is not found"
             self.log_err(err_msg)


millevy:
> 🧹 Nitpick comments (1)
> > provider/v2v_vmcheck_helper.py (1)> `843-845`: _⚡ Quick win_
> > > **Consider caching the regex result to avoid redundant searches.**
> > > The regex `re.search(r"(Intel|AMD) Processor", win_dirvers)` is executed three times on lines 843, 844, and 845. Additionally, the inner condition on line 844 is redundant since if the regex matches, `group(0)` will always be in `cpu_drivers`.
> > > 
> > > ♻️ Proposed refactor
> > > ```diff
> > > -        if re.search(r"(Intel|AMD) Processor", win_dirvers):
> > > -            if re.search(r"(Intel|AMD) Processor", win_dirvers).group(0) in cpu_drivers:
> > > -                LOG.info("CPU driver '%s' is found" % re.search(r"(Intel|AMD) Processor", win_dirvers).group(0))
> > > +        cpu_driver_match = re.search(r"(Intel|AMD) Processor", win_dirvers)
> > > +        if cpu_driver_match:
> > > +            LOG.info("CPU driver '%s' is found" % cpu_driver_match.group(0))
> > >          elif self.os_version == 'win2016':
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@provider/v2v_vmcheck_helper.py` around lines 843 - 845, Cache the regex
> > > result instead of calling re.search three times: call re.search(r"(Intel|AMD)
> > > Processor", win_dirvers) once, store it in a variable (e.g., match), check if
> > > match is truthy, then test match.group(0) membership in cpu_drivers and log once
> > > using match.group(0); remove the redundant nested re.search and the unnecessary
> > > second conditional so only the cached match is used for both the membership
> > > check and the LOG.info call.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@provider/v2v_vmcheck_helper.py`:
> - Around line 843-845: Cache the regex result instead of calling re.search three
> times: call re.search(r"(Intel|AMD) Processor", win_dirvers) once, store it in a
> variable (e.g., match), check if match is truthy, then test match.group(0)
> membership in cpu_drivers and log once using match.group(0); remove the
> redundant nested re.search and the unnecessary second conditional so only the
> cached match is used for both the membership check and the LOG.info call.
> ```
> 
> ℹ️ Review info

Good catch, but this is pre-existing code — keeping this PR scoped to the win2016 fix only.

millevy:
@crobinso could you review this when you get a chance?

crobinso:
LGTM but I can not merge. We are using a downstream fork until maintenance is figured out for these public repos

millevy:
Moved the change to downstream repo , until maintenance is figured out for these public repos.

PR #6878, opened 2026-06-03
libvirt_mem:Add test for memory attach-detach with different option

sneh-3:
supported options are --config, --live, --current, --persistent, --live --config

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive memory attach/detach test scenarios covering multiple operation modes.
  * Introduced enhanced verification mechanisms for memory configuration changes.

* **Chores**
  * Updated memory test configurations to reflect new baseline scenarios.
  * Expanded test coverage for memory attach/detach operations across different configuration modes.

### DIFF
--- a/libvirt/tests/cfg/libvirt_mem.cfg
+++ b/libvirt/tests/cfg/libvirt_mem.cfg
@@ -17,10 +17,11 @@
         wait_before_save_secs = 10
     variants:
         - positive_test:
-            max_mem_rt = 3584000
-            max_mem = 2048000
-            current_mem = 2048000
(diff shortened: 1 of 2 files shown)


hholoubk:
First of all please extend the PR description with ... what this do... why it is necessary... 
(even it looks like obvious, e.g. there is inside the change of memory size (numbers) hardcoded ... and not clear why those numbers are changed. 

Second .. please check existing code .. mainly avocado-vt functions and existing memory providers in tp-libvirt. 

... I will add detailed info directly

hholoubk:
There are some additional comments from AI that I believe are correct but will require much more detailed check and I am not so involved into the memory tests.

PR #6880, opened 2026-06-05
tests/iface_options: add pp64 compatibility for command line test

aniket-sahu-ibmx:
The command line used for the test explicitly states "qemu-kvm", which is not used for ppc64le systems. Hence, this patch adds qemu-system-ppc64 to the command line after checking host architecture.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QEMU process detection in tests to correctly identify running processes across different processor architectures, including ppc64le systems.


aniket-sahu-ibmx:
@harihare please check this as well

PR #6884, opened 2026-06-10
Add fallback logic for interface detection when ifconfig is unavailable!

Anushree-Mathur:
The test fails with error "Device None does not exist" because the underlying utils_misc.get_interface_from_pci_id()
function relies on the deprecated ifconfig command, which is not installed by default.

When ifconfig is unavailable, the function returns None, causing the test to attempt configuring a network interface named "None", which fails.

This patch adds a fallback mechanism that:
1. Detects when get_interface_from_pci_id() returns None
2. Uses the modern 'ip' command to list network interfaces
3. Reads PCI addresses from /sys/class/net/{iface}/device/uevent
4. Matches PCI addresses to find the correct interface name

The fallback only activates when the primary method fails, ensuring no impact on systems where ifconfig is available.

## Summary by CodeRabbit

* **Tests**
  * Improved NIC interface detection for PCI passthrough device validation with enhanced fallback mechanism.
  * Strengthened error handling to gracefully skip devices when interface identification fails, with additional logging for troubleshooting.

### DIFF
--- a/libvirt/tests/src/passthrough/pci/libvirt_pci_passthrough.py
+++ b/libvirt/tests/src/passthrough/pci/libvirt_pci_passthrough.py
@@ -207,6 +207,36 @@ def run(test, params, env):
             # ping to server from each function
             for val in bus_info:
                 nic_name = str(utils_misc.get_interface_from_pci_id(val, session))
+
+                # If get_interface_from_pci_id returns None, use uevent file
+                if nic_name == "None" or not nic_name:
+                    logging.warning("get_interface_from_pci_id returned None for %s, trying uevent method", val)
+                    try:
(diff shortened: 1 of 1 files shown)


Anushree-Mathur:
Following error is being  fixed with this PR :
 
 ```
 (2/3) type_specific.io-github-autotest-libvirt.libvirt_pci_passthrough.default.NIC.Normal_passthrough: STARTED
 (2/3) type_specific.io-github-autotest-libvirt.libvirt_pci_passthrough.default.NIC.Normal_passthrough: ERROR: Shell command failed: 'ip addr flush dev None'    (status: 1,    output: 'Device "None" does not exist.\n') (96.40 s)

```

PR #6885, opened 2026-06-10
Convert netmask to CIDR notation for ip command!

Anushree-Mathur:
The 'ip addr add' command requires CIDR notation (e.g., /24) but the test was using dotted decimal netmask format (e.g., 255.255.255.0), causing the command to hang and timeout.

Add netmask_to_cidr() helper function that converts dotted decimal netmask to CIDR notation by counting the number of 1 bits in the binary representation.
Example: 255.255.255.0 → 11111111.11111111.11111111.00000000 → /24

## Summary by CodeRabbit

* **Bug Fixes**
  * Network interface configuration now uses proper CIDR notation for IP addresses.
  * Netmask inputs are strictly validated; invalid netmasks are reported as errors instead of being silently accepted.


aniket-sahu-ibmx:
Instead of using a helper function to convert decimal netmask to CIDR, perhaps you can just change the value of the netmask to equivalent CIDR notation value in the config, since that's where the test is getting the values from?

PR #6886, opened 2026-06-15
libvirt_mem: Add vhost configuration for SLES

sneh-3:
1. Enhanced memory hotplug testing with conditional vhost parameter configuration for SLES systems.
2. Added automatic kernel module management and configuration verification for improved test reliability on SLES environments.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for memory-related test setup on SLES systems, reducing failures when virtual machines are already running.
  * Added safer checks so the test skips configuration changes when the system state can’t be verified, helping avoid disruptive changes during execution.
  * Strengthened validation of runtime settings before continuing with memory hotplug and domain configuration steps.

### DIFF
--- a/libvirt/tests/src/libvirt_mem.py
+++ b/libvirt/tests/src/libvirt_mem.py
@@ -116,6 +116,60 @@ def run(test, params, env):
         config.restore()
         utils_libvirtd.libvirtd_restart()
 
+    def setup_vhost_max_mem_regions():
+        """
+        Configure vhost max_mem_regions for SLES systems.
+        SLES has lower default values compared to other distros.
+        This is required for memory hotplug tests with max_slots.
+        Must be called when no VM is using vhost modules.
+        This function applies runtime configuration only.
+        For persistent configuration across reboots, manually create:
+        /etc/modprobe.d/vhost.conf with content: options vhost max_mem_regions=512
+        """
+        try:
+            # Check if running on SLES
+            distro_check = process.run("cat /etc/os-release | grep -i sles",
+                                      shell=True, ignore_status=True)
+            if distro_check.exit_status == 0:
+                logging.info("SLES detected, configuring vhost max_mem_regions=512")
+                vm_check = process.run("virsh list --state-running --name",
+                                       shell=True, ignore_status=True)
+                if vm_check.exit_status != 0:
+                    logging.warning("Unable to determine running VMs; skipping vhost configuration")
+                    return
+                running_vms = [name for name in vm_check.stdout_text.splitlines() if name.strip()]
+                if running_vms:
+                    logging.warning("Running VMs detected, skipping vhost configuration for safety:")
+                    logging.warning("%s", "\n".join(running_vms))
+                    return
+
+                logging.info("No running VMs detected, proceeding with vhost configuration")
+                process.run("modprobe -r vhost_net vhost_scsi vhost_vsock vhost",
+                           shell=True, ignore_status=True)
+                result = process.run("modprobe vhost max_mem_regions=512",
+                                    shell=True, ignore_status=True)
+
+                if result.exit_status == 0:
+                    logging.info("Successfully configured vhost max_mem_regions=512 at runtime")
+                    verify_cmd = "cat /sys/module/vhost/parameters/max_mem_regions"
+                    verify_result = process.run(verify_cmd, shell=True, ignore_status=True)
+                    if verify_result.exit_status == 0:
+                        current_value = verify_result.stdout_text.strip()
+                        logging.info("Verified vhost max_mem_regions is set to: %s", current_value)
+                        if current_value != "512":
+                            test.cancel("vhost max_mem_regions verification failed: expected 512, got %s"
+                                        % current_value)
+                    else:
+                        logging.debug("Could not verify vhost max_mem_regions setting")
+                else:
+                    test.cancel("Failed to configure vhost at runtime: %s" % result.stderr_text)
+            else:
+                logging.debug("Non-SLES system detected, skipping vhost configuration")
+
+        except Exception as e:
+            test.cancel("Error setting up vhost max_mem_regions: %s" % str(e))
+
+
     def check_qemu_cmd(max_mem_rt, tg_size):
         """
         Check qemu command line options.
@@ -528,6 +582,8 @@ def run(test, params, env):
         # Destroy domain first
         if vm.is_alive():
             vm.destroy(gracefully=False)
+        # Configure vhost max_mem_regions for SLES (VM is now stopped)
+        setup_vhost_max_mem_regions()
         modify_domain_xml()
         numa_info = utils_misc.NumaInfo()
         logging.debug(numa_info.get_all_node_meminfo())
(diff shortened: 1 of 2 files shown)


sneh-3:
before fix:
(114/413) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.memory.hot.plug.max_slots.without_reboot:  FAIL: error: Failed to attach device from /var/tmp/xml_utils_temp_lmk15107.xml\nerror: internal error: unable to execute QEMU command 'device_add': a used vhost backend has not enough free memory slots left\n (34.58 s)

after fix:
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.memory.hot.plug.max_slots.without_reboot: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.memory.hot.plug.max_slots.without_reboot: PASS (177.27 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 |

libvirt/tests/src/libvirt_mem.py
```diff
@@ -116,6 +116,60 @@ def restore_hugepages(page_size=4):
...
+        SLES has lower default values compared to other distros.
+        This is required for memory hotplug tests with max_slots.
+        Must be called when no VM is using vhost modules.
+        This function applies runtime configuration only.
+        For persistent configuration across reboots, manually create:
+        /etc/modprobe.d/vhost.conf with content: options vhost max_mem_regions=512
+        """
+        try:
+            # Check if running on SLES
+            distro_check = process.run("cat /etc/os-release | grep -i sles",
```
hholoubk: Use avocado method to check the distro 

already used in virsh_dump.py, libvirt_pci_passthrough_hotplug.py, etc.:

from avocado.utils import distro
detected = distro.detect()
if detected.name.lower() in ("suse", "sles"):

BUT
... there is better way how to do it via configuration.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -116,6 +116,60 @@ def restore_hugepages(page_size=4):
...
+                                      shell=True, ignore_status=True)
+            if distro_check.exit_status == 0:
```
hholoubk: It is wrong approach mixing os detection into the value setting. Please move the whole detection outside (into the test run) 

so the row will be 

```
if detected.name.lower() in ("suse", "sles"):
     setup_vhost_max_mem_regions()

```

libvirt/tests/src/libvirt_mem.py
```diff
@@ -116,6 +116,60 @@ def restore_hugepages(page_size=4):
...
+                    return
+
+                logging.info("No running VMs detected, proceeding with vhost configuration")
+                process.run("modprobe -r vhost_net vhost_scsi vhost_vsock vhost",
+                           shell=True, ignore_status=True)
+                result = process.run("modprobe vhost max_mem_regions=512",
+                                    shell=True, ignore_status=True)
+
+                if result.exit_status == 0:
+                    logging.info("Successfully configured vhost max_mem_regions=512 at runtime")
```
hholoubk: To hardcode the value is not good approach. 
this value should be configurable in the cfg. which allow us to distinguish values passed for various host os

e.g. 

```
- max_slots:
    max_mem_slots = 256
    attach_times = 255
    Host_SuSE:
        set_vhost_max_mem_regions = "yes"
        vhost_max_mem_regions = 512
```
(also probably - attach_gt_max: variants)

introducing also the flag if the value should be set ... otherwise the set will go in every test .. regardless it is needed or not. 

the check will be then in the test run part:
```

if params.get("set_vhost_max_mem_regions") == "yes":
    max_mem_regions_backup = setup_vhost_max_mem_regions()
```

libvirt/tests/src/libvirt_mem.py
```diff
@@ -116,6 +116,60 @@ def restore_hugepages(page_size=4):
...
+                running_vms = [name for name in vm_check.stdout_text.splitlines() if name.strip()]
+                if running_vms:
+                    logging.warning("Running VMs detected, skipping vhost configuration for safety:")
+                    logging.warning("%s", "\n".join(running_vms))
+                    return
+
+                logging.info("No running VMs detected, proceeding with vhost configuration")
+                process.run("modprobe -r vhost_net vhost_scsi vhost_vsock vhost",
+                           shell=True, ignore_status=True)
+                result = process.run("modprobe vhost max_mem_regions=512",
```
hholoubk: setting this value regardless what is actually set on the host is not good approach ... 

Also setting the value without storing it and recover to previous value to return the  host in same configuration as before the test is against best practices. 

1) store actual value
2) check if it is lower than what you want 

```
current = process.run(cmd, shell=True).stdout_text.strip()
needed = int(params.get("vhost_max_mem_regions", 512))
if current < needed:
    # do the setting

return current
```


restore later in finally of run test.

libvirt/tests/src/libvirt_mem.py
```diff
@@ -116,6 +116,60 @@ def restore_hugepages(page_size=4):
...
+        try:
+            # Check if running on SLES
+            distro_check = process.run("cat /etc/os-release | grep -i sles",
+                                      shell=True, ignore_status=True)
+            if distro_check.exit_status == 0:
+                logging.info("SLES detected, configuring vhost max_mem_regions=512")
+                vm_check = process.run("virsh list --state-running --name",
+                                       shell=True, ignore_status=True)
+                if vm_check.exit_status != 0:
+                    logging.warning("Unable to determine running VMs; skipping vhost configuration")
```
hholoubk: I understood that the test need this setting so if we are not able to set it ... it should not run at all. 

so here should be the test.cancel instead of run and probably fail later (similar as is in case, that the value after setting it is not as expected)

hholoubk (changes requested):
There are several things that can be better, some of them kind of severe, some just unoptimized. 

The main issue is, that test is changing host configuration and not recovering the environement after. 

Commited changes will run regardless they are needed. 

Suggested changes:

1) Use the avocado cfg file to change the configuration of host only in cases that really need it. 
2) Do not hardcode set value, introduce the flag if it should be set in the test and the value, that should be set.
3) move the check (if the value should be done) outside the setter method
4) backup the value and recover it at the end of the test
5) reconsider if the test shouldn't be canceled in case that there is any running VM.

libvirt/tests/src/libvirt_mem.py
sneh-3: Fixed in v2 based on all review comments:

1. Distro detection removed from setup_vhost_max_mem_regions() — now
   handled via Host_SuSE cfg block with set_vhost_max_mem_regions flag.

2. Value 512 no longer hardcoded — configurable via vhost_max_mem_regions
   param in cfg, read via params.get() in Python.

3. Silent return replaced with test.cancel() for both cases:
   virsh check failure and running VMs detected.

4. Original value is backed up, function only sets if current < needed,
   and original is restored in the finally block after the test.

5. Function is only called when set_vhost_max_mem_regions == "yes",
   which is activated only under Host_SuSE in the max_slots variant.

sneh-3:
--logs--
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.memory.hot.plug.max_slots.without_reboot: STARTED
 (1/1) type_specific.io-github-autotest-libvirt.libvirt_mem.positive_test.memory.hot.plug.max_slots.without_reboot: PASS (152.88 s)

PR #6902, opened 2026-06-17
v2v: set boottype = 3 for vmx_default latest8 efi vm

surygupta-rh:
The vmx_default latest8 VM is efi-based and needs boottype = 3 to convert correctly. 
Without this, v2v defaults to bios boot and the converted VM fails to boot.
Also fixes the VM name placeholder mappings.

## Summary by CodeRabbit

* **Tests**
  * Updated test configuration to support ESX 8.0 compatibility testing with RHEL8 VM support

* **Chores**
  * Modified NFS source configuration for VMX metadata retrieval


crobinso:
@surygupta-rh maint structure still isn't fixed here yet. please send to the internal mirror

PR #6906, opened 2026-06-26
tests/virsh_cmd: add tests for virsh await

aniket-sahu-ibmx:
Added tests for checking the virsh await functionality. Currently only two await states are available and the test covers both of them.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end coverage for `virsh await` with support for VM state waits and guest-agent availability checks.
  * Supports multiple VM reference formats and timeout handling for both success and failure scenarios.

* **Bug Fixes**
  * Added validation for missing, invalid, and unsupported wait conditions.
  * Covers timeout behavior, ACL/polkit access checks, and cleanup after test execution.

### DIFF
--- a/libvirt/tests/cfg/virsh_cmd/domain/virsh_await.cfg
+++ b/libvirt/tests/cfg/virsh_cmd/domain/virsh_await.cfg
@@ -1,0 +1,171 @@
+- virsh.await:
+    type = virsh_await
+    start_vm = "yes"
+    kill_vm = "yes"
+    kill_vm_before_test = "yes"
+    await_vm_ref = "name"
+    await_timeout = 30
+    libvirtd = "on"
+    variants:
+        - domain_inactive:
+            await_condition = "domain-inactive"
+            variants:
+                - normal_test:
+                    status_error = "no"
+                    await_expect_exit = 0
+                    variants:
+                        - shutdown_then_await:
+                            await_pre_state = "running"
+                            await_trigger = "shutdown"
+                            await_timeout = 60
+
+                        - already_shutoff:
+                            no lxc
+                            await_pre_state = "shutoff"
+                            await_trigger = "none"
+                            await_timeout = 5
+
+                        - destroy_then_await:
+                            no lxc
+                            await_pre_state = "running"
+                            await_trigger = "destroy"
+                            await_timeout = 60
+
+                        - ref_by_uuid:
+                            await_pre_state = "shutoff"
+                            await_vm_ref = "uuid"
+                            await_trigger = "none"
+                            await_timeout = 5
+
+                        - ref_by_id:
+                            no lxc
+                            await_pre_state = "running"
+                            await_vm_ref = "id"
+                            await_trigger = "shutdown"
+                            await_timeout = 60
+
+                        - ref_by_hex_id:
+                            no lxc
+                            await_pre_state = "running"
+                            await_vm_ref = "hex_id"
+                            await_trigger = "shutdown"
+                            await_timeout = 60
+
+                - error_test:
+                    status_error = "yes"
+                    variants:
+                        - timeout_expire:
+                            await_pre_state = "running"
+                            await_trigger = "none"
+                            await_timeout = 2
+                            await_expect_exit = 2
+
+                        - invalid_domain:
+                            await_pre_state = "running"
+                            await_vm_ref = "invalid_domain"
+                            await_trigger = "none"
+                            await_timeout = 5
+                            await_expect_exit = 1
+
+                        - empty_domain:
+                            await_pre_state = "running"
+                            await_vm_ref = "empty"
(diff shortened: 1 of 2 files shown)


libvirt/tests/cfg/virsh_cmd/domain/virsh_await.cfg
```diff
@@ -0,0 +1,171 @@
...
+    kill_vm = "yes"
+    kill_vm_before_test = "yes"
+    await_vm_ref = "name"
+    await_timeout = 30
+    libvirtd = "on"
+    variants:
+        - domain_inactive:
+            await_condition = "domain-inactive"
+            variants:
+                - normal_test:
```
hholoubk, line 13: For new tests, please use positive_test / negative_test (please check in [cfg_file_definition_convention.rst](https://github.com/autotest/tp-libvirt/blob/master/cfg_file_definition_convention.rst)). I know normal_test / error_test still exists in older virsh_cmd cases, but it is not the current convention.

libvirt/tests/src/virsh_cmd/domain/virsh_await.py
```diff
@@ -0,0 +1,244 @@
...
+            test.cancel("API acl test not supported in current libvirt version.")
+
+    # virsh await first appeared in libvirt 10.0; cancel gracefully on older builds.
+    if not libvirt_version.version_compare(10, 0, 0):
+        test.cancel("virsh await requires libvirt >= 10.0.0")
+
+    xml_backup = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+
+    try:
+        # ------------------------------------------------------------------ #
```
hholoubk, line 116: Please for better readability and maintainability we would like to avoid too long methods. I know that older test look exactly like this, but with new test we should follow worldwide accepted best practices.

I would suggest to remove those comments that are splitting the code .. and instead of it create methods that are focused just for this part. 

e.g. lines 119-134 will be one method 
lines 139-159 second. 

In the try block you will have just those methods names.

libvirt/tests/src/virsh_cmd/domain/virsh_await.py
```diff
@@ -0,0 +1,244 @@
...
+            # in the base image.
+            vm.prepare_guest_agent(channel=False, start=False)
+
+        # ------------------------------------------------------------------ #
+        # 2. Resolve vm_ref
+        # ------------------------------------------------------------------ #
+        domid = vm.get_id() if vm.is_alive() else ""
+        domuuid = vm.get_uuid()
+
+        if await_vm_ref == "name":
```
hholoubk, line 142: Instead of the long if-elif chain it would be better (better understandable) to use the dictionary. suggestion from AI:

def resolve_vm_ref(vm, vm_name, await_vm_ref):
    domid = vm.get_id() if vm.is_alive() else ""
    refs = {
        "name": vm_name,
        "id": domid,
        "uuid": vm.get_uuid(),
        "hex_id": hex(int(domid)) if domid else None, # but this will fail in case wrong domid
        "invalid_domain": "totally_nonexistent_domain_99999",
        "empty": "",
    }
    if await_vm_ref not in refs:
        test.error(f"Unknown await_vm_ref: {await_vm_ref}")
    if await_vm_ref == "hex_id" and not domid:
        test.error("await_vm_ref=hex_id requires a running domain")
    return refs[await_vm_ref]

I didn't check it, if working and I have still some doubts about ... what if domid="" (vm not alive) the hex(int("")) will fail.

libvirt/tests/src/virsh_cmd/domain/virsh_await.py
```diff
@@ -0,0 +1,244 @@
...
+                unprivileged_user=unprivileged_user,
+                uri=uri,
+            )
+
+        logging.debug("virsh await result: %s", result)
+
+        # ------------------------------------------------------------------ #
+        # 5. Validate result
+        # ------------------------------------------------------------------ #
+        if status_error:
```
hholoubk, line 197: please consider libvirt.check_exit_status() usage

libvirt/tests/src/virsh_cmd/domain/virsh_await.py
```diff
@@ -0,0 +1,244 @@
...
+                        "but guest agent is not responding."
+                    )
+
+    finally:
+        # ------------------------------------------------------------------ #
+        # 6. Restore environment
+        # ------------------------------------------------------------------ #
+        if vm.is_alive():
+            vm.destroy()
+        xml_backup.sync()
```
hholoubk, line 244: please check. I believe xml_backup.sync() already contain the code to  destroye  domain if live. as we do not work after with the vm object. the simple xml_backup should be enough.

hholoubk (commented):
Hi.
Please check my comments below .. feel free to ask me (or any AI)

PR #6912, opened 2026-06-30
memory_hotplug_virshdump: add sequential attach/dump 

sneh-3:
- Validate memory hotplug followed by virsh dump per iteration
- Ensure consistent VM state and avoid race conditions during runtime
- Improve error handling, logging, and cleanup to prevent spurious failures

Signed-off-by: Sneh Shikha Yadav <syadav@linux.ibm.com>

### DIFF
--- a/libvirt/tests/src/memory/memory_hotplug_virshdump.py
+++ b/libvirt/tests/src/memory/memory_hotplug_virshdump.py
@@ -1,0 +1,289 @@
+import os
+import time
+import logging
+import tempfile
+from avocado.utils import process
+from virttest import virsh
+from virttest import utils_hotplug
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.vm_xml import VMXML, VMCPUXML
+
+
+def run(test, params, env):
+    """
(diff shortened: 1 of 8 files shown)


harihare:
As per initial comment, I assume it is only memory releated patch. However I see code related to disk attach and detach. Please correct me if I am wrong.

libvirt/tests/src/memory/memory_hotplug_virshdump.py
```diff
@@ -0,0 +1,289 @@
...
+                break
+
+            dump_file = os.path.join(
+                dump_dir, "dump_%d_%s" % (i + 1, time.strftime("%Y%m%d_%H%M%S"))
+            )
+            dump_ok = False
+            try:
+                logging.info("  [virsh dump] %s", dump_file)
+                dump_result = virsh.dump(vm_name, dump_file, dump_options, debug=True)
+
```
harihare, line 136: as per code for every iteration of hotplug dump is taken. is that you are planning. As dump is a distructive tests it will take good amount of time for multiple iterations. I would recommend to run virsh dump once or twice in parallel to hotplug.

PR #6919, opened 2026-09-02
Attempt to avoid situation of incompatible CPUs between hypervisors

pholica:
Co-authored by ClaudeCode


hholoubk:
Good catch ... and good solution, we definitely needed this one. 

It will be valuable to have this as specific method (part of the base_steps) ... so we will have in the test script calling only one function. (and add one or two rows into the test script)

I would suggest to add new method into the provider/migration/base_steps.py something like:

```
def ensure_cpu_compatibility(vm, params):
        # Ensure the guest CPU is compatible with the destination host before
        # migration. When the source and destination CPUs differ, compute a
        # migratable baseline and apply it; the guest must be off to rewrite
        # its CPU, so it is (re)started below. No-op when the CPUs already
        # match and on aarch64.
        if not check_cpu_for_mig(params):
            if vm.is_alive():
                vm.destroy()
            sync_cpu_for_mig(params)

```

I know that, the sync_cpu_for_mig was not part of your commit, but what will happen if the CPUs are not compatible?

PR #6920, opened 2026-09-07
libvirt: timer_management: Fix logic for checking VM time offsets

pipo:
Previously the test fetched host time and then tried getting guest time, possibly multiple times. This is prone to errors as the VM can take longer to respond.

A much saner approach is to first get time from the guest and then from the host as it's much simpler to fetch time from the host and thus much quicker.

To achieve this the host-time support is removed from 'get_time' into new function 'get_host_time' and the callers are refactored to use it correctly after getting VM time. Debug statements are also rewritten to show all info on a single line.


libvirt/tests/src/timer_management.py
```diff
@@ -120,45 +120,60 @@ def get_time(test, vm=None, time_type=None, windows=False):
...
+    timestr = timestr.replace(" ", "").strip()
+    struct_time = time.strptime(timestr, "%Y/%m/%d/%H/%M/%S")
+
+    try:
+        return int(time.mktime(struct_time))
+    except TypeError as e:
+        test.error("Translate time to seconds error: %s" % e)
+
+
+def get_host_time(test, time_type=None, windows=False):
```
hholoubk, line 163: Two things 

The windows parameter is not used and the test is not covering the windows host, so we don't need this parameter. 

the method is called only with the time_type = "utc" therefore also this parameter is not needed. 

I don't think the test itself will be enlarged with other usage of the host time (other time_type)

therefore it would make sense to rename to get_host_utc_time .. and pass only the test parameter.

libvirt/tests/src/timer_management.py
```diff
@@ -111,7 +111,7 @@ def vm_clock_source(test, vm, target, value=''):
     return result
 
 
-def get_time(test, vm=None, time_type=None, windows=False):
+def get_time(test, vm, time_type=None, windows=False):
```
hholoubk, line 114: please rename the method as get_guest_time so it will be selfexplaining.

hholoubk (changes requested):
Hi @pipo

... thank you for improving the test, even there are still many issues and wrong things in it I understand, you are not responsible for them. 

I like the split of the get_time method into two... but I have some suggestions that could be made. 

Thank you.

